PR #38, opened 2025-07-02
Simplify user registration for comments

christosporios:
Streamline the user registration flow for leaving comments by making it inline and ensuring name collection upfront.

The previous registration flow for commenting was confusing: users had to enter an email, click a link, land on a profile page showing "no admin rights," enter their name, and then manually navigate back to the consultation page. This PR simplifies the process to an inline form, direct email verification return to the consultation page, and upfront name collection, significantly improving the user experience and enforcing the name requirement.

### DIFF
--- a/auth-flow-improvements.md
+++ b/auth-flow-improvements.md
@@ -1,0 +1,152 @@
+# Auth Flow Improvements for Consultation Comments
+
+## Problem Analysis
+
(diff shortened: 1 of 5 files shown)


christosporios:
This is stale; we need to rethink this when we do our next public consultation. Closing.

PR #59, opened 2025-07-15
City creator: instantly import city councils from the web

christosporios:
This is a quick off-roadmap feature to reduce the data entry workload for adding new cities.

- It implements a new Import City form for cities that are `pending`, where city council members, parties and their relationships can be manually entered quickly and on the same screen.
- An "import with AI" button uses claude with web search to import an entire city council with one click, and corrections can be made in the form afterwards before saving.
- A "reset city" button is available on development for cities that are not `pending`: it makes the city `pending` again and deletes all meetings, members, parties etc.

Screenshot 2025-07-15 at 22 09 31


christosporios:
merging this 🤡

PR #61, opened 2025-07-17
Transcript page: fix scrolling bug

### LINKED ISSUE #55: Scrolling bug on the transcript page
When scrolling on the transcript page of a meeting e.g. [here](https://opencouncil.gr/chania/jul9_2025/transcript), there's a an issue that's hard to describe except with a video: https://www.loom.com/share/207c0aca28f1471fb3cedc6415046e26?from_recorder=1&focus_title=1

christosporios:
Fixes #55.

**Fix transcript scrolling bug causing scroll loops**

**Problem:**
When scrolling up through the transcript, users experienced jarring auto-scroll interruptions that created a scrolling loop, making it difficult to navigate upward through the content.

**Root Causes**:
* Broken virtualization logic with conflicting intersection observers
* Dramatic height changes when segments switched between mock and real rendering
* Multiple observers fighting each other during scroll events

**Solution**:

Implemented single, efficient intersection observer for proper virtualization. Fixed layout shifts by ensuring mock rendering maintains exact same height as real content. Added skeleton loading states with consistent dimensions. Optimized performance with debounced updates and efficient state management.

**Result**:
Smooth, natural scrolling behavior throughout the transcript without any jumps or interruptions.

### DIFF
--- a/src/components/meetings/transcript/Transcript.tsx
+++ b/src/components/meetings/transcript/Transcript.tsx
@@ -1,7 +1,7 @@
 "use client";
 import { SpeakerSegment as SpeakerSegmentType } from "@prisma/client";
 import SpeakerSegment from "./SpeakerSegment";
-import { useEffect, useRef, useMemo } from 'react';
+import { useEffect, useRef, useMemo, useState, useCallback } from 'react';
 import { useVideo } from "../VideoProvider";
 import { Utterance } from "@prisma/client";
 import { Transcript as TranscriptType } from "@/lib/db/transcript";
@@ -16,6 +16,7 @@ export default function Transcript() {
     const { options } = useTranscriptOptions();
     const { setCurrentScrollInterval, currentTime } = useVideo();
     const containerRef = useRef<HTMLDivElement>(null);
+    const [visibleSegments, setVisibleSegments] = useState<Set<number>>(new Set());
 
     // Join segments if not in edit mode
     const displayedSegments = useMemo(() => {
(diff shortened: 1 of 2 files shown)


src/components/meetings/transcript/Transcript.tsx
```diff
@@ -57,36 +58,65 @@ export default function Transcript() {
...
+                .filter(Boolean); // Filter out any undefined segments
+
+            if (visibleSegmentsList.length > 0) {
+                const firstVisible = visibleSegmentsList[0];
+                const lastVisible = visibleSegmentsList[visibleSegmentsList.length - 1];
                 debouncedSetCurrentScrollInterval([firstVisible.startTimestamp, lastVisible.endTimestamp]);
             }
-        };
+        }
+    }, [visibleSegments, displayedSegments, debouncedSetCurrentScrollInterval]);
```
kouloumos, line 103: This is a great improvement to the virtualization logic. The single `IntersectionObserver` is much more performant.

I was looking at the `handleIntersection` callback and had a thought on a small refinement we could make. The current implementation is correct, but we can make it slightly more efficient by avoiding the creation of the intermediate `updates` array.

Since the number of visible segments is always small, the main performance benefit comes from reducing memory allocations in a high-frequency event like scrolling.

Here's an alternative approach that lazily creates the new `Set` only when a change is detected:

```typescript
const handleIntersection = useCallback((entries: IntersectionObserverEntry[]) => {
    // 1. Start with a null Set. We won't create a copy unless we absolutely have to.
    let newVisibleSegments: Set<number> | null = null;

    for (const entry of entries) {
        const segmentIndex = parseInt(entry.target.id.split('-')[2]);
        const isCurrentlyVisible = visibleSegments.has(segmentIndex);

        if (entry.isIntersecting && !isCurrentlyVisible) {
            // 2. A segment became visible. If this is the first change,
            // create the copy of the Set now.
            if (!newVisibleSegments) {
                newVisibleSegments = new Set(visibleSegments);
            }
            newVisibleSegments.add(segmentIndex);
        } else if (!entry.isIntersecting && isCurrentlyVisible) {
            // 3. A segment became hidden. Same logic: if this is the first
            // change, create the copy now.
            if (!newVisibleSegments) {
                newVisibleSegments = new Set(visibleSegments);
            }
            newVisibleSegments.delete(segmentIndex);
        }
    }

    // 4. If newVisibleSegments is not null, it means changes occurred.
    // We can then update the state and perform the side effect.
    if (newVisibleSegments) {
        setVisibleSegments(newVisibleSegments);

        const visibleSegmentsList = Array.from(newVisibleSegments)
            .sort((a, b) => a - b)
            .map(index => displayedSegments[index])
            .filter(Boolean);

        if (visibleSegmentsList.length > 0) {
            const firstVisible = visibleSegmentsList[0];
            const lastVisible = visibleSegmentsList[visibleSegmentsList.length - 1];
            debouncedSetCurrentScrollInterval([firstVisible.startTimestamp, lastVisible.endTimestamp]);
        }
    }
}, [visibleSegments, displayedSegments, debouncedSetCurrentScrollInterval]);
```

This version achieves the same result but avoids the extra array allocation and the second loop, which helps reduce pressure on the garbage collector during scrolling.

What do you think of this? No strong opinion either way, as the current code works perfectly, but thought it was worth mentioning.

kouloumos (approved):
Looks great! I was able to verify the initial bug and tested the solution locally. I left a comment inline with a potential improvement.

src/components/meetings/transcript/Transcript.tsx
christosporios: I will omit this for simplicity and merge the PR.

Issue #64, opened 2025-07-18
Within a meeting, we should show roles that were active when the meeting happened


christosporios:
Because roles have time intervals, when displaying e.g. the transcript for a meeting in the past, we should consider roles that were active when the meeting happened. Currently, we don't do this consistently. Yet, the `getPartyFromRoles` method supports an optional `Date` parameter to do this -- it's just not passed correctly everywhere.

christosporios:
this may actually be complete by now -- we need to verify/check that it works

Issue #60, opened 2025-07-17
Transcripts: inconsistent party coloring


christosporios:
See the screenshot below from the transcript page of a meeting: the vertical line appears gray in one speaker segment, but colored in the other. Both line should be in the party color. Additionally, exporting the transcript to .docx has similar issues: party names are shown only on some speaker segments.

This may be related to the deprecated party column on the user table (which has been replaced with the role relation).

<img width="143" height="393" alt="Image" src="https://github.com/user-attachments/assets/2fe7933c-114b-4d2e-903f-eeb80d52734a" />

Issue #43, opened 2025-07-07
Remove deprecated code


christosporios:
Remove deprecated code
- Role related fields
- pg vector & embedding, voyage ai
- waitlist
- party field for People

PR #65, opened 2025-07-18
Remove deprecated party property on Person

christosporios:
A while ago, we replaced our `party` property/column on `Persons` with our `Role` relation, which is time-constrained (has a start date and end date) and can hold extra information about the role a person has on a party. This PR removes the part property entirely, and replaced all remaining usages with looking up the roles relation.

**Highlighting one existing issue that remains after this PR:** Because roles have time intervals, when displaying e.g. the transcript for a meeting in the past, we should consider roles that were active when the meeting happened. Currently, we don't do this consistently. I have opened a issue #64 for this. 

**Also highlighting one issue that this PR introduces**: we change the way we query the database to occasionally include more information, that is often duplicate (e.g. a full party object will be fetched as part of many roles). We should eventually improve this.

**This is a risky change -- thus we should deploy it on staging first and manually test it well before pushing to prod.** Our production data was already migrated to the new roles system, so removing the column should have no data loss.

This is part of #43, and completely solves #60.


src/components/meetings/docx/CouncilMeetingDocx.tsx
```diff
@@ -148,7 +150,7 @@ export const renderDocx = async ({ meeting, transcript, people, parties, speaker
     city: City,
     meeting: CouncilMeeting,
     transcript: Transcript,
-    people: Person[],
+    people: (PersonWithRelations | any)[],
```
kouloumos, line 153: `any` shouldn't be an option. Removing it here will also remove the need for declaring it on the line above 
```suggestion
    people: (PersonWithRelations)[],
```

src/lib/db/speakerSegments.ts
```diff
@@ -509,51 +525,51 @@ export async function getLatestSegmentsForParty(
...
+            const meetingDate = new Date(segment.meeting.dateTime);
+
+            // Check for active role at meeting time
+            const hasActiveRole = person.roles.some(role => {
+                const startDate = role.startDate ? new Date(role.startDate) : null;
+                const endDate = role.endDate ? new Date(role.endDate) : null;
+
+                return (!startDate || startDate <= meetingDate) &&
+                    (!endDate || endDate >= meetingDate);
+            });
```
kouloumos, line 556: Please use the existing `isRoleActive`  method from `src/lib/utils.ts`

src/lib/utils.ts
```diff
@@ -299,6 +299,50 @@ export function filterInactiveRoles<T extends { startDate: Date | null, endDate:
...
+ */
+export function getPartyFromRoles<T extends {
+  startDate: Date | null,
+  endDate: Date | null,
+  party?: { id: string; name: string; colorHex: string;[key: string]: any } | null
+}>(roles: T[], date?: Date): T['party'] | null {
+  const checkDate = date || new Date();
+
+  // Filter roles that are active at the specified date
+  const activeRoles = roles.filter(role => isRoleActiveAt(role, checkDate));
```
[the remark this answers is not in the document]
christosporios: These check for the current date -- here we check for a custom date. We need `filterActiveRolesAt`, but I think it's fine to do it in-line.

PR #71, opened 2025-07-24
Price increase (~50%) and make human corrected transcripts the default 

christosporios:
- Increasing platform pricing about 50% across most tiers
- Human-corrected transcripts are now the default in the pricing page
- Various small improvements in the /about page


kouloumos (approved):
giphy

PR #77, opened 2025-08-15
Remove deprecated role properties

christosporios:
A while ago, we replaced `role`, `role_en` and `isAdministrativeRole` on the Person table with a separate role relationship. This PR removes these properties from the database, and their usage from the code. Finishes #43, and probably the last bit for #64.

Also removes PDF exports.


kouloumos (commented):
Looks good to me! Needs a small change, `BulkExportActions` includes an import from a method that doesn't exist anymore -- `exportMeetingToPDF` resulting to build fail.

Also, the PR includes commits that already exists on the production branch, maybe because you branched out from the production branch?

kouloumos:
> Looks good to me! Needs a small change, `BulkExportActions` includes an import from a method that doesn't exist anymore -- `exportMeetingToPDF` resulting to build fail.
> 

I've fixed this alongside the removal of the pdf related library in b2f1fb713af3fd9f5a925297efda158ff6ad1cca

But, it seems that there is another required change in `src/components/meetings/docx/CouncilMeetingDocx.tsx` which uses `person.role`. You need to adjust it to the new schema.

PR #185, opened 2026-02-02
Remove stale initdb copy from db Dockerfile

gplusdotgr:
## Summary
- remove the COPY of docker-entrypoint-initdb.d that referenced a deleted entrypoint directory
- drop the placeholder initdb folder since it is no longer needed

## Problem
The db Dockerfile still tried to copy ./docker-entrypoint-initdb.d into the image even though we previously removed our custom entrypoint directory. Builds failed with:
```
failed to compute cache key: failed to calculate checksum ... "/docker-entrypoint-initdb.d": not found
```

## Fix
Delete the stale COPY step and remove the now-unused directory. The image now builds using the upstream PostGIS defaults.

## Testing
- docker build ./db -f db/Dockerfile.db --build-arg POSTGIS_IMAGE=imresamu/postgis:16-3.5
- ./run.sh (uses POSTGIS_IMAGE auto-detection)

### DIFF
--- a/db/Dockerfile.db
+++ b/db/Dockerfile.db
@@ -2,6 +2,3 @@ ARG POSTGIS_IMAGE=postgis/postgis:16-3.5
 
 # Use the PostGIS image as the base
 FROM ${POSTGIS_IMAGE}
-
-# Copy initialization scripts
-COPY ./docker-entrypoint-initdb.d/ /docker-entrypoint-initdb.d/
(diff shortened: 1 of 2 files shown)


gplusdotgr:
Hey @christosporios @kouloumos , I tried to start the app with `run.sh` and I run into the above.

kouloumos (commented):
Thanks for taking the time to contribute @gplusdotgr!

It would be better to remove these lines from `Dockerfile.db`.
https://github.com/schemalabz/opencouncil/blob/b0a991d54b37ff6464ccbbab7315c30f8df87503/db/Dockerfile.db#L6-L7

We previously had an entry point, and it looks like it wasn’t fully removed when we deleted it.

gplusdotgr:
even better :-)

PR #188, opened 2026-02-03
Fix performance issues

christosporios:
- Add missing database indexes on Utterance.discussionSubjectId and discussionStatus                                    
  - Skip loading topicLabels relation when statistics don't group by topic                                                
  - Replace array.find() with Map lookups across meeting page components                                         

---

> [!NOTE]
> **Low Risk**
> Primarily performance-focused query/include changes and in-memory lookup refactors; main risk is subtle data-shape differences when `topicLabels` are intentionally not loaded.
> 
> **Overview**
> Improves meeting-page performance by **adding DB indexes** for `Utterance.discussionSubjectId` and `Utterance.discussionStatus`, and wiring them into Prisma via new `@@index` entries.
> 
> Reduces CPU/DB work in hot paths: `getStatisticsFor` now conditionally loads `topicLabels` only when grouping by topic, `getMeetingData` dedupes `speakerTags` in O(n), and several meeting UI contexts/components replace repeated `array.find()` scans with memoized `Map` lookups (e.g., `getSpeakerSegmentById`, `getHighlight`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fc5c13c8e6b18b4b2ba34ee240bc31110c7a677d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/prisma/migrations/20260203145120_add_utterance_indices/migration.sql
+++ b/prisma/migrations/20260203145120_add_utterance_indices/migration.sql
@@ -1,0 +1,5 @@
+-- CreateIndex
+CREATE INDEX "Utterance_discussionSubjectId_idx" ON "Utterance"("discussionSubjectId");
+
+-- CreateIndex
+CREATE INDEX "Utterance_discussionStatus_idx" ON "Utterance"("discussionStatus");
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -303,8 +303,11 @@ model Utterance {
   words Word[]
   highlightedUtterances HighlightedUtterance[]
   podcastPartAudioUtterances PodcastPartAudioUtterance[]
-  @@index([speakerSegmentId, startTimestamp])
   utteranceEdits UtteranceEdit[]
+
+  @@index([speakerSegmentId, startTimestamp])
+  @@index([discussionSubjectId])
+  @@index([discussionStatus])
 }
 
 enum DiscussionStatus {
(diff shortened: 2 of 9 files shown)


kouloumos (commented):
The migration adding indexes on `Utterance.discussionSubjectId` and `Utterance.discussionStatus` connects to something I raised in https://github.com/schemalabz/opencouncil/pull/165#discussion_r2724157291

These indexes exist because we're filtering on the Utterance table directly:

```typescript
// statistics.ts
const utterances = await prisma.utterance.findMany({
    where: {
        discussionSubjectId: subjectId,
        discussionStatus: 'SUBJECT_DISCUSSION'
    },
    ...
});
```

The same pattern appears in `subject.ts` and `voting-utterances/route.ts`. Without indexes, this is a full scan of the Utterance table.

The Utterance table currently has 500,000+ rows. That means each of these two new indexes has 500k+ entries, and both grow with every transcribed meeting. Every utterance insert means two more index entries to maintain.

With the range model I proposed, the filtering moves to a tiny table:

```sql
SELECT u.* FROM Utterance u
JOIN DiscussionRange dr ON u.discussionRangeId = dr.id
WHERE dr.subjectId = X AND dr.status = Y
```

DiscussionRange would have maybe a few hundred rows total (5-10 ranges per subject, 10-20 subjects per meeting). An index there is negligible. You'd still need an FK index on `Utterance.discussionRangeId`, but that's one index instead of two - and it's a standard FK index rather than filtering indexes on columns where the same value is duplicated across hundreds of utterances per range.

I'm fine merging this as-is since it solves the immediate problem, but these indexes are the cost of the per-utterance denormalization. Worth revisiting the range design before we add another 500k utterances.

PR #201, opened 2026-02-19
Fix: secure fetchRegulationData to prevent SSRF

kernelstub:
Hey,

I updated the consultation JSON fetch to add an allowlist for trusted URLs because before, anyone could make the server request any external link, which was a security risk. Now it only fetches JSON from approved domains, keeping our server safe and making sure it uses valid, trusted data.

### DIFF
--- a/src/app/api/og/route.tsx
+++ b/src/app/api/og/route.tsx
@@ -11,6 +11,30 @@ import { Container, MeetingMetaRow, OgHeader, OpenCouncilWatermark, SubjectPills
 // Import the native subject OG image generator for reuse
 import SubjectOgImage from '@/app/[locale]/(city)/[cityId]/(meetings)/[meetingId]/subjects/[subjectId]/opengraph-image';
 
+const ALLOWED_HOSTS = [
+  "opencouncil.gr",
+  "api.opencouncil.gr",
+  "cdn.opencouncil.gr"
+];
+
+function isPrivateIP(hostname: string) {
+  return (
+    hostname === "localhost" ||
+    hostname.startsWith("127.") ||
+    hostname.startsWith("10.") ||
+    hostname.startsWith("192.168.") ||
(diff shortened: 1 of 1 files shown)


kouloumos (commented):
Thanks for looking into this, but I don't think this PR addresses a real vulnerability.

`jsonUrl` is a field on the `Consultation` model that's set directly in the database by us -- there's no API endpoint or form that lets users create consultations or control this value. An attacker would need write access to our database, at which point SSRF is the least of our problems.

Beyond the threat model, a few practical issues:

There are four copies of `fetchRegulationData` in the codebase (`src/lib/db/consultations.ts:199`, `src/app/[locale]/(city)/[cityId]/consultation/[id]/page.tsx:14`, `src/app/[locale]/(city)/[cityId]/consultation/[id]/comments/page.tsx:16`, and the one in `src/app/api/og/route.tsx`). This PR only patches the OG route copy. If this were a real concern, the other three -- which serve actual page content, not just preview images -- would matter more.

The `isPrivateIP` check and the protocol check are redundant with the allowlist. If only `opencouncil.gr`, `api.opencouncil.gr`, and `cdn.opencouncil.gr` are allowed, no private IP or non-https URL can ever pass validation anyway. The allowlist alone would be the entire fix.

If there's something I'm missing about the attack surface, let me know and we can revisit. Otherwise I think we can close this one.

That said -- we definitely need security-minded people looking at this codebase, and I appreciate you taking the time. The four duplicated `fetchRegulationData` copies are a real problem worth fixing, and consolidating them into a single shared function would be a welcome contribution if you're up for it.

kernelstub:
Hey,

It's better to be safe than sorry you never know if you'll add support for consultation later, so having stricter rules now helps avoid problems down the line. I seen it many times from experience  that Companies add something forget about it for 4-7 months add something on top of it and that backbone has vulnerabilities. As said better safe than sorry later.

PR #218, opened 2026-02-25
feat: transcript controls and bulk deletion fixes

### LINKED ISSUE #187: Editing: Be able to delete more than one utterance at a time
When deleting utterances, we are only able to delete them one by one. 
There are times (see photo - σκουπίδια) that we need to delete more than one at a time. Is there a way to do that?

Image

### LINKED ISSUE #189: Editing:  the last subjects of the meeting are not visible when I'm editing a transcription
Image

I wanted to click on the last two subjects to see what will be discussed, but they are hidden behind the play button, and the recording.

-->this does not happen in the transcription page for the general user

angelospk:
Closes #187  
Closes #189

### Summary
This PR improves transcript editing ergonomics in two ways:
1. Adds support for deleting multiple selected utterances in one action.
2. Fixes transcript controls positioning so timeline subjects are no longer hidden behind the floating player controls in editing layout.

### Changes Made
- Added multi-delete support in editing flow:
  - Extended editing context with `deleteSelectedUtterances()`.
  - Wired keyboard shortcut support for deleting current selection (`Delete` / `Backspace`).
  - Updated utterance context menu to show:
    - `Delete utterance` for single selection
    - `Delete selected (N)` for multi-selection.
  - Added localized labels/messages (EN/EL) for new delete actions and toast feedback.
- Added batch-like local state update for deletions:
  - Introduced local deletion application utility (`applyUtteranceDeletions`) to remove multiple utterances and recalculate segment bounds consistently.
  - Updated meeting data context to expose `deleteUtterances(utteranceIds: string[])`.
- Fixed transcript controls overlap in desktop editing layout:
  - Controls now compute left offset based on sidebar state (expanded/collapsed), preventing overlap with transcript subject list.
  - Added a small layout utility for reusable offset calculation.

### Testing
- Manual validation performed for:
  - Multi-selection + delete via context menu
  - Multi-selection + delete via keyboard shortcut
  - Single utterance delete behavior
  - Timeline controls positioning with sidebar expanded/collapsed
- Automated tests were intentionally skipped in this run to reduce machine load.

### Related Issues
- Closes #187
- Closes #189

<h3>Greptile Summary</h3>

This PR adds bulk utterance deletion and adjusts transcript-control placement.
- Adds a transactional bulk-deletion API and optimistic transcript updates with rollback.
- Adds selection-aware deletion dialogs, context-menu actions, keyboard shortcuts, localization, and regression tests.
- Offsets desktop transcript controls according to the expanded or collapsed sidebar state.

The PR is not yet safe to merge because right-clicking outside an active multi-selection still silently discards that selection before deletion.

The prior selection-loss issue remains at the current context-menu guard: an unselected right-click target enters the temporary-selection path even when other utterances are already selected, replacing the group with the clicked utterance.

**Files Needing Attention:** src/components/meetings/transcript/UtteranceContextMenu.tsx

| Filename | Overview |
|----------|----------|
| src/components/meetings/transcript/UtteranceContextMenu.tsx | Adds deferred, snapshot-based context-menu deletion, but the existing multi-selection is still replaced when another utterance is right-clicked. |
| src/components/meetings/EditingContext.tsx | Adds a frozen deletion target, confirmation dialog, localized feedback, and guards selection shortcuts while the dialog is open. |
| src/contexts/KeyboardShortcutsContext.tsx | Adds deletion shortcuts while preserving native behavior inside editable controls and dialogs. |
| src/lib/db/utterance.ts | Adds authorized transactional bulk deletion and preserves authorization errors for the API’s forbidden response. |
| src/components/meetings/CouncilMeetingDataContext.tsx | Adds optimistic single and bulk deletion through the shared endpoint, including failure rollback. |
| src/components/meetings/TranscriptControls.tsx | Computes the desktop player offset from the current sidebar state while leaving mobile positioning unset. |

<sub>Reviews (18): Last reviewed commit: ["feat(meetings): bulk-delete confirmation..."](https://github.com/schemalabz/opencouncil/commit/ba46cf8ff8b49b3eec71eb351c41f2877eb7febe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=23181525)</sub>

### DIFF
--- a/src/components/meetings/EditingContext.tsx
+++ b/src/components/meetings/EditingContext.tsx
@@ -5,13 +5,25 @@ import { ACTIONS, useKeyboardShortcut } from '@/contexts/KeyboardShortcutsContex
 import { useToast } from '@/hooks/use-toast';
 import { useTranslations } from 'next-intl';
 import { calculateUtteranceRange } from '@/lib/selection-utils';
+import {
+    Dialog,
+    DialogContent,
+    DialogDescription,
+    DialogFooter,
+    DialogHeader,
+    DialogTitle,
+} from "@/components/ui/dialog";
+import { Button } from "@/components/ui/button";
 
 interface EditingContextType {
     selectedUtteranceIds: Set<string>;
     lastClickedUtteranceId: string | null;
     toggleSelection: (id: string, modifiers: { shift: boolean, ctrl: boolean }) => void;
+    deselectUtterance: (id: string) => void;
     clearSelection: () => void;
     extractSelectedSegment: () => Promise<void>;
+    confirmDeleteSelected: (explicitIds?: string[]) => void;
+    deleteSelectedUtterances: () => Promise<void>;
     isProcessing: boolean;
 }
 
@@ -21,9 +33,15 @@ export function EditingProvider({ children }: { children: ReactNode }) {
     const [selectedUtteranceIds, setSelectedUtteranceIds] = useState<Set<string>>(new Set());
     const [lastClickedUtteranceId, setLastClickedUtteranceId] = useState<string | null>(null);
     const [isProcessing, setIsProcessing] = useState(false);
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    // Snapshot of ids the open confirm dialog will act on. Decoupled from
+    // `selectedUtteranceIds` so unrelated selection changes between
+    // "open dialog" and "confirm" (e.g. UtteranceContextMenu's close handler
+    // clearing its temp-selection) can't blank the dialog's target list.
+    const [pendingDeleteIds, setPendingDeleteIds] = useState<string[]>([]);
 
     const { transcript, getSpeakerSegmentById } = useCouncilMeetingData();
-    const { extractSpeakerSegment } = useCouncilMeetingActions();
+    const { extractSpeakerSegment, deleteUtterances } = useCouncilMeetingActions();
     const { toast } = useToast();
     const t = useTranslations('editing.toasts');
 
@@ -39,6 +57,15 @@ export function EditingProvider({ children }: { children: ReactNode }) {
     const clearSelection = useCallback(() => {
         setSelectedUtteranceIds(new Set());
         setLastClickedUtteranceId(null);
+    }, []);
+
+    const deselectUtterance = useCallback((id: string) => {
+        setSelectedUtteranceIds(prev => {
+            if (!prev.has(id)) return prev;
+            const next = new Set(prev);
+            next.delete(id);
+            return next;
+        });
     }, []);
 
     const toggleSelection = useCallback((id: string, modifiers: { shift: boolean, ctrl: boolean }) => {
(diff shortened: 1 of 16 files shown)


kouloumos (commented):
The bulk deletion logic is solid -- the transactional `deleteUtterances` in the DB layer, the `applyUtteranceDeletions` utility for local state, and the keyboard shortcut addition are all welcome. The layout fix for transcript controls is clean too.

Two bugs found during testing though:

**The context menu delete is broken for single utterances.** Right-click → "Delete utterance" → the confirmation dialog shows "0 utterances" and the delete button does nothing.

Image

**After dismissing the dialog (or after a successful multi-delete), the screen freezes** and a refresh is needed.

On the backend, `deleteUtterances` is a parallel reimplementation of `deleteUtterance` -- same auth check, same delete-then-recalculate-timestamps pattern. The bulk version is strictly better (transactional, handles missing utterances gracefully), so it would be cleaner to have `deleteUtterance` delegate to `deleteUtterances` rather than maintaining two code paths.

I left inline comments on specifics, but a general observation: there are several spots where the new code doesn't follow existing codebase conventions -- naming patterns, file organization, import style. Things like the `InDb` aliasing (rest of the file imports DB functions without aliases), the `request*` prefix (used elsewhere for task queue calls, not dialogs), a new `src/lib/transcript/` directory for a single utility, etc. Nothing major individually, but it adds up.

Curious -- what's your dev workflow / tooling setup for this? Are you using an agentic coding tool, and if so which one? Not asking to judge, genuinely asking so we can figure out the best way to surface these conventions to whatever tools people use. We have a `CLAUDE.md` with project conventions and we're happy to improve and help make that information accessible in whatever format works for your setup.

angelospk:
Update after the rebase onto main and the latest review pass — all suggestions from the reviews have been incorporated. The branch is now 3 clean commits:

- `feat(db): transactional bulk utterance deletion + API route` — `deleteUtterances()` runs deletion + segment-boundary recalculation in a single Prisma transaction with per-city auth upfront. `DELETE /api/utterances` accepts `{ ids: string[] }`, caps at 500 ids to bound the transaction and avoid an unconstrained deletion vector, and returns distinct 401/403/400/500 so the client can react. `applyUtteranceDeletions()` is a pure helper shared by the DB and client paths.

- `feat(meetings): wire bulk delete into transcript context with optimistic update` — client action calls the new endpoint and applies the deletion locally via `applyUtteranceDeletions`, with a ref-captured pre-deletion snapshot for StrictMode-safe rollback. The single-utterance path delegates to the bulk action so there's only one delete path. The inline trash button on empty utterances calls a new `deselectUtterance(id)` instead of `clearSelection()`, so deleting one utterance no longer wipes the rest of an active multi-selection.

- `feat(meetings): bulk-delete confirmation dialog, keyboard shortcut, menu items` — `EditingContext` exposes `confirmDeleteSelected` / `deleteSelectedUtterances`. No global Enter listener: Radix traps focus inside the dialog and the focused button handles Enter/Escape natively (Cancel is auto-focused, so an accidental Enter dismisses rather than deletes). Backspace/Delete shortcut is registered without breaking browser back-nav elsewhere. While the dialog is open, both `CLEAR_SELECTION` (Escape) and `DELETE_SELECTION` (Backspace/Delete) shortcuts are disabled, so Escape only dismisses the dialog instead of also wiping the selection. The global `KeyboardShortcutsContext` keydown handler also bails out for any target inside a `[role="dialog"]` / `[role="alertdialog"]`, so Enter/Space on dialog buttons can't be swallowed by `EDIT_NEXT_UTTERANCE` / `PLAY_PAUSE`. The shared `UtteranceContextMenu` gets a Delete entry (we keep main's single-shared-menu architecture from the perf refactor rather than reintroducing per-utterance menus). Locale strings added in en/el.

Tests cover both contexts: 5 in `EditingContext.test.tsx` (dialog count under temp-selection clear, freeze regression, `deselectUtterance` preserves the rest of the selection, shortcuts disabled while the dialog is open, multi-select without explicit ids) and 3 in `KeyboardShortcutsContext.test.tsx` (shortcut fires outside dialogs, ignored inside `[role="dialog"]`, ignored inside `[role="alertdialog"]`). tsc clean.

kouloumos:
Hey! Main has moved forward and your branch was out of sync. I've rebased your 3 commits on top of the current main and force-pushed the result.

The original branch state before the rebase is preserved at [`backup/pr-218-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-218-original), so nothing is lost.

I'll continue reviewing shortly!

src/components/meetings/transcript/UtteranceContextMenu.tsx
```diff
@@ -289,6 +289,33 @@ export function UtteranceContextMenu({ children }: { children: React.ReactNode }
...
+                                        ? Array.from(selectedUtteranceIds)
+                                        : [target!.id];
+                                    // Defer dialog opening so the Radix DropdownMenu's
+                                    // DismissableLayer fully unmounts and restores
+                                    // body.style.pointerEvents first. Mounting the
+                                    // Dialog while pointer-events:'none' is still on
+                                    // <body> causes Radix to cache 'none' as the
+                                    // original and re-apply it on close, freezing
+                                    // the page.
+                                    setTimeout(() => confirmDeleteSelected(ids), 0);
```
kouloumos: Your diagnosis in the comment above is exactly right, and I measured that this works: after dismissing, `body` pointer-events returns to `auto` and clicks reach the page again.

One suggestion about the shape of it. This outruns the race; it does not remove it. The menu at line 227 is still modal, so it still puts `pointer-events: none` on `body`, and correctness depends on the `setTimeout(0)` landing after Radix unmounts the layer.

This repository already solves the same problem the other way, in four places:

- `src/components/offer-letter/documents-dropdown.tsx:76-78` — with a comment describing this exact bug
- `src/components/consultations/ConsultationMap.tsx:834`
- `src/components/consultations/DetailPanel.tsx:471`
- `src/components/meetings/decisions/MeetingDecisionsPage.tsx:122`

`modal={false}` on the `DropdownMenu` at line 227 stops the menu locking the body at all, so there is no lock left to leak and no ordering to depend on. The trade is that the menu no longer blocks outside pointer events while open, which is the accepted behaviour at those four sites.

Your version works today. I would still prefer the established one. Your call — say if you disagree and I will take it as it is.

src/components/meetings/EditingContext.tsx
```diff
@@ -21,9 +33,15 @@ export function EditingProvider({ children }: { children: ReactNode }) {
     const [selectedUtteranceIds, setSelectedUtteranceIds] = useState<Set<string>>(new Set());
     const [lastClickedUtteranceId, setLastClickedUtteranceId] = useState<string | null>(null);
     const [isProcessing, setIsProcessing] = useState(false);
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    // Snapshot of ids the open confirm dialog will act on. Decoupled from
+    // `selectedUtteranceIds` so unrelated selection changes between
+    // "open dialog" and "confirm" (e.g. UtteranceContextMenu's close handler
+    // clearing its temp-selection) can't blank the dialog's target list.
+    const [pendingDeleteIds, setPendingDeleteIds] = useState<string[]>([]);
```
kouloumos: This is the part I want to point at, because it is a better fix than the bug asked for.

The symptom was that the context menu's close handler cleared the temporary selection before the dialog read it. The narrow fix is to reorder those two things. Holding a separate snapshot instead means **no** selection change during the dialog's lifetime can blank its target list, including ones nobody has written yet.

That is the difference between fixing the failure and removing the class of failure.

PR #220, opened 2026-02-25
docs(api): expand swagger spec with role-based filtering

### LINKED ISSUE #45: Improve API documentation
### Motivation
To ensure our Open Data API is comprehensively documented, easily discoverable, and maintainable by creating a complete Swagger specification with role-based documentation and clear contribution guidelines.

### Current State
- **Technology**: We use a `swagger.yaml` file to define our API, which is rendered as interactive documentation in Next.js.
- **Configuration**: The interactive API documentation is linked in the site footer, but the `swagger.yaml` file is incomplete—key endpoints like `/api/admin`, `/api/consultations`, and `/api/auth` are missing.
- **Security Concern**: Admin routes are mixed with public routes without proper access control or documentation separation.
- **Issue**: Users and contributors cannot get a full picture of the API's capabilities and there are no clear guidelines on how to document new endpoints as they are added.

### Proposed Outline
1. **Conduct a Full API Audit**: Systematically compare the endpoints in `src/app/api/` with the `swagger.yaml` file to identify all undocumented routes, categorizing them by access level (public, admin, internal).

2. **Implement Role-Based Documentation**: 
   - Create separate Swagger specifications or implement dynamic filtering based on user authentication
   - Design the `/docs` endpoint to show different API documentation based on user privileges
   - Consider serving different documentation URLs (e.g., `/docs/admin` for admin endpoints)

3. **Complete the Swagger Specification**: Update `swagger.yaml` with all missing paths, operations, parameters, and schemas, ensuring proper tagging and security definitions for different access levels.

4. **Implement Dynamic Documentation Serving**: 
   - Modify the Next.js `/docs` route to check user authentication/authorization
   - Filter or conditionally render API endpoints based on user roles
   - Ensure admin routes are only visible to authenticated admin users

5. **Improve Discoverability**: Add a dedicated section to the main `/docs` page that introduces the Open Data API.

6. **Create an API Contribution Guide**: Create a new guide at `docs/guides/api-documentation-guide.md`. This document should outline:
   - The process for adding or updating API documentation when a new endpoint is created
   - Best practices for categorizing endpoints by access level and properly tagging them
   - Security considerations for documenting admin vs public endpoints
   - Instructions for implementing role-based documentation filtering
   - Guidelines for writing clear descriptions for paths, parameters, and schemas
   - Instructions for LLMs on how to approach and maintain the API documentation

### Technical Considerations
- Implement server-side filtering of OpenAPI specs based on user context
- Consider using Next.js middleware or API route guards to control documentation access
- Ensure proper caching strategies for user-specific documentation
- Test all permission combinations to verify correct endpoint visibility

kdngiorgos:
The existing `/docs` page and `GET /api` endpoint served the full OpenAPI spec to everyone — exposing admin/superadmin endpoint names and schemas to unauthenticated users. This PR adds role-based filtering so each viewer only sees the operations they can actually call, and lays the groundwork for keeping the spec in sync with the codebase via code-first generation.

Part of #45

### What this PR does

**Role-based documentation filtering**
- Four-tier access hierarchy: `public < user < admin < superadmin`
- Both the `/docs` page (server component) and `GET /api` route filter paths using the shared `filterSpecByAccessLevel()` utility in `src/lib/utils/openapi.ts`
- An `AccessBanner` on `/docs` tells each viewer how many paths are visible and prompts unauthenticated users to sign in

**Code-first OpenAPI infrastructure**
- Adds `@asteasolutions/zod-to-openapi` to derive specs from existing Zod validation schemas
- Shared `OpenAPIRegistry` in `src/lib/openapi/registry.ts` (security schemes, reusable error schemas)
- `npm run generate-openapi` script writes `swagger.yaml` / `openapi.json` from registered routes
- Proof-of-concept registrations for 4 routes: `GET/POST /api/cities`, `POST /api/search`, `GET/POST /api/cities/{cityId}/meetings`, `GET /api/statistics`

**Shared utility**
- Extracted `getUserAccessLevel(user)` into `src/lib/utils/openapi.ts` to eliminate duplication between the docs page and API route handler

### Security notes
- Unrecognized `x-access-level` values (e.g. typos) are treated as **maximally restrictive** — the path is hidden rather than accidentally exposed
- Session cookie name corrected to `authjs.session-token` (Auth.js v5)
- Defensive handling for optional `administers` field with nullish coalescing

### Files changed
| File | Purpose |
|------|---------|
| `src/lib/openapi/registry.ts` | OpenAPI registry with security schemes and error schemas |
| `src/lib/openapi/index.ts` | Aggregates all route registrations |
| `src/lib/openapi/routes/{cities,meetings,search,statistics}.ts` | Route registrations |
| `scripts/generate-openapi.ts` | Generation script |
| `src/lib/utils/openapi.ts` | `getUserAccessLevel`, `filterSpecByAccessLevel` |
| `src/app/api/route.ts` | Async spec loading + filtering |
| `src/app/[locale]/(other)/docs/page.tsx` | Docs page with AccessBanner |
| `package.json` | Added `@asteasolutions/zod-to-openapi`, `generate-openapi` script |

<h3>Greptile Summary</h3>

This PR replaces the static Swagger file with an in-process, role-filtered OpenAPI registry.
- Adds code-first registrations and shared Zod-backed request and response schemas.
- Filters operations and unreachable component schemas according to viewer access level.
- Serves the generated specification through both GET /api and the localized documentation page.
- Adds focused schema, filtering, and documented-operation coverage tests.

The PR should not merge until meeting authorization failures stop returning generic server errors and the documented authorization responses match the actual permission boundary.

The PUT meeting handler still catches the plain authorization error from withUserAuthorizedToEdit as a generic exception and returns 500, while the generated contract advertises 401; city and meeting admin operations also continue to describe insufficient permission as authentication failure.

**Files Needing Attention:** src/app/api/cities/[cityId]/meetings/[meetingId]/route.ts, src/lib/auth.ts, src/lib/openapi/routes/meetings.ts, src/lib/openapi/routes/cities.ts

| Filename | Overview |
|----------|----------|
| src/lib/utils/openapi.ts | Adds access-level derivation, fail-closed operation filtering, transitive schema-reference traversal, and component pruning. |
| src/lib/openapi/registry.ts | Defines the shared OpenAPI registry, session security scheme, common schemas, and document generator. |
| src/lib/openapi/routes/meetings.ts | Registers meeting operations using shared request schemas and operation-specific response contracts. |
| src/lib/zod-schemas/meeting.ts | Centralizes meeting validation while preserving optional generated IDs and agenda-processing input. |
| src/app/api/route.ts | Replaces static YAML loading with per-viewer filtering of the generated specification. |
| src/app/[locale]/(other)/docs/page.tsx | Renders the generated specification filtered to the current viewer’s access level. |
| src/lib/openapi/__tests__/coverage.test.ts | Guards the intended set of operations against accidental omission during code-first generation. |

<sub>Reviews (20): Last reviewed commit: ["fix(api-docs): restore dropped utterance..."](https://github.com/schemalabz/opencouncil/commit/4a0072bf38d16a151b0c477e24703b4fad4043b2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=24255244)</sub>


kouloumos (commented):
Thanks for the thorough work here @kdngiorgos -- the swagger rewrite is well-structured (access levels match the actual auth calls, all `$ref`s resolve, security declarations are consistent), and the filtering infrastructure in `openapi.ts` is a nice addition.

That said, I want to suggest a different direction for the spec itself. A hand-maintained 3000-line YAML file will inevitably drift from the actual routes -- and the 316-line contributor guide is essentially documentation for fighting that drift. We've been through this before with our task processing backend ([`opencouncil-tasks`](https://github.com/schemalabz/opencouncil-tasks)), where we solved it by deriving the OpenAPI spec directly from TypeScript types. The setup there uses `typescript-json-schema` to walk the AST and auto-discover request/response interfaces, so the docs stay in sync with the code by definition.

For the main app (Next.js), the equivalent approach would be [`@asteasolutions/zod-to-openapi`](https://github.com/asteasolutions/zod-to-openapi). We recently started centralizing Zod validation schemas in `src/lib/zod-schemas/` (see [`city.ts`](https://github.com/schemalabz/opencouncil/blob/main/src/lib/zod-schemas/city.ts) for the pattern -- shared between frontend forms and backend routes). The idea is to build on that: tag those same Zod schemas with `.openapi('SchemaName')` and register routes against them, then generate the spec at build time. The validation schema and the API documentation live in the same file -- change the schema, the docs update.

Concretely, what I'd love to see instead of the YAML rewrite:

1. **Set up `zod-to-openapi` infrastructure** -- a shared `OpenAPIRegistry` in `src/lib/openapi/registry.ts`, a generation script in `scripts/generate-openapi.ts`, and the plumbing to serve the generated spec.
2. **Migrate a handful of routes as proof of concept** — pick 3-4 routes that already have Zod schemas (e.g. `search`, `cities`, `meetings`) and register them. This establishes the pattern for the rest of the codebase to follow incrementally.
3. **Keep the filtering and access banner** -- `src/lib/utils/openapi.ts` with `filterSpecByAccessLevel` and the `AccessBanner` component are useful regardless of how the spec is generated. The `x-access-level` extension and `sessionAuth` scheme can be part of the registry setup.

One concrete thing to fix either way: the user-level determination logic (the `if isSuperAdmin / else if administers.length` block) is duplicated verbatim in `docs/page.tsx` and `api/route.ts`. That should be extracted into a shared function, e.g. `getUserAccessLevel(user)` in `openapi.ts`.

For reference, you can look at how `opencouncil-tasks` handles this in [`src/lib/schemaGenerator.ts`](https://github.com/schemalabz/opencouncil-tasks/blob/main/src/lib/schemaGenerator.ts) (type discovery + JSON Schema generation) and [`src/lib/swaggerGenerator.ts`](https://github.com/schemalabz/opencouncil-tasks/blob/main/src/lib/swaggerGenerator.ts) (path generation from a task registry). The Next.js version would be simpler since `zod-to-openapi` handles the schema-to-OpenAPI conversion for us.

Happy to walk through the setup in more detail if you want to go this route.

src/lib/openapi/routes/cities.ts
```diff
@@ -0,0 +1,113 @@
...
+    officialSupport: z.boolean(),
+    status: z.enum(['active', 'pending', 'inactive']),
+    authorityType: z.enum(['municipality', 'region', 'other']),
+    supportsNotifications: z.boolean(),
+    consultationsEnabled: z.boolean(),
+    meetingCount: z.number().int().optional(),
+    personCount: z.number().int().optional(),
+}).openapi('City');
+
+registry.register('City', CitySchema);
```
kouloumos, line 54: Two things here:

1. `extendZodWithOpenApi(z)` is already called in `registry.ts` (which you import on line 3), so the call on line 5 is redundant. Same in `meetings.ts`, `search.ts`, and `statistics.ts` — you can remove all four.

2. `CitySchema` is defined from scratch, but `src/lib/zod-schemas/city.ts` already has the canonical city validation schemas used by both the frontend forms and the API route handler (`src/app/api/cities/route.ts` imports `createCityFormDataSchema` from there). The existing schema handles things like Prisma enum validation (`z.nativeEnum(AuthorityType)`), `stringToBoolean` transforms for FormData, and fields that are missing here (`peopleOrdering`, `highlightCreationPermission`, `diavgeiaUid`).

The idea behind the zod-to-openapi approach is to tag the *existing* schemas with `.openapi()` so they serve double duty — validation and documentation from the same source. Defining separate schemas for OpenAPI defeats that purpose. Can you import and extend the existing schemas instead?

src/lib/openapi/routes/meetings.ts
```diff
@@ -0,0 +1,116 @@
...
+    administrativeBodyId: z.string().nullable(),
+}).openapi('Meeting');
+
+const CreateMeetingSchema = z.object({
+    name: z.string().min(2).openapi({ description: 'Meeting name in Greek' }),
+    name_en: z.string().min(2).openapi({ description: 'Meeting name in English' }),
+    date: z.string().openapi({ description: 'Meeting date/time (ISO 8601)', example: '2024-09-15T18:00:00Z' }),
+    youtubeUrl: z.string().url().optional().or(z.literal('')).openapi({ description: 'YouTube video URL' }),
+    agendaUrl: z.string().url().optional().or(z.literal('')).openapi({ description: 'Agenda document URL' }),
+meetingId: z.string().min(1).openapi({ description: 'Client-supplied unique meeting ID (e.g., UUID). The server returns 409 if the ID already exists.' }),
```
kouloumos: This line lost its indentation — starts at column 0 while everything else is at 4 spaces. Looks like it came from applying the greptile suggestion via the GitHub UI.

```
meetingId: z.string().min(1).openapi({ ... }),
```

should be:

```
    meetingId: z.string().min(1).openapi({ ... }),
```

src/lib/openapi/routes/meetings.ts
```diff
@@ -0,0 +1,116 @@
...
+        },
+        400: {
+            description: 'Invalid meeting data',
+            content: {
+                'application/json': { schema: ValidationErrorSchema },
+            },
+        },
+    },
+    'x-access-level': 'admin',
+});
```
kouloumos, line 127: A few things don't match the actual handler at `src/app/api/cities/[cityId]/meetings/route.ts`:

- The handler returns **201** (line 124: `NextResponse.json(meeting, { status: 201 })`), but the spec documents `200`. Should be `201`.
- The `meetingId` field description says "The server returns 409 if the ID already exists", but there's no `409` in the responses block.
- Auth failures: the handler uses `withUserAuthorizedToEdit()` which throws a generic `Error("Not authorized")` — this gets caught by the catch-all and returned as `500`, not the `401` documented here. (This is arguably a bug in the handler, not the spec, but the spec should describe what actually happens today or note the discrepancy.)
- `MeetingSchema` is missing fields the handler actually returns: `createdAt`, `updatedAt`, and the full nested `administrativeBody` object (not just the ID).

Also, same as cities — the handler already has a `meetingSchema` (line 10 of the route file) that could be reused here.

src/lib/openapi/routes/search.ts
```diff
@@ -0,0 +1,96 @@
...
+            content: {
+                'application/json': {
+                    schema: z.object({
+                        error: z.object({
+                            code: z.string(),
+                            message: z.string(),
+                            details: z.unknown(),
+                        }),
+                    }),
+                },
```
kouloumos: This uses a custom inline schema for the 400 error:
```typescript
schema: z.object({
    error: z.object({
        code: z.string(),
        message: z.string(),
        details: z.unknown(),
    }),
})
```

That actually matches what `src/app/api/search/route.ts` returns (lines 77-84) — so it's accurate. But it's inconsistent with the other route files which use the shared `ValidationErrorSchema` from the registry. It also produces an anonymous schema in the generated spec.

Two options: either register this as a named schema (e.g., `SearchErrorSchema`) in the registry so other routes with the same pattern can reuse it, or align the actual search handler to use the same error format as the other handlers and then use `ValidationErrorSchema` here.

src/lib/openapi/routes/search.ts
```diff
@@ -0,0 +1,96 @@
...
+            },
+        },
+        500: {
+            description: 'Search engine error',
+            content: {
+                'application/json': { schema: ErrorResponseSchema },
+            },
+        },
```
kouloumos, line 75: The spec uses `ErrorResponseSchema` (`{ error: string }`) for the 500 response, but the actual handler returns the same structured format: `{ error: { code: 'SEARCH_ERROR', message: '...', details: '...' } }` (lines 87-96 of the route). The 500 schema doesn't match.

src/lib/openapi/routes/statistics.ts
```diff
@@ -0,0 +1,64 @@
...
+        count: z.number().int(),
+        duration: z.number(),
+    })).optional(),
+    party: z.array(z.object({
+        id: z.string(),
+        name: z.string(),
+        count: z.number().int(),
+        duration: z.number(),
+    })).optional(),
+}).openapi('StatisticsResponse');
```
kouloumos: This schema doesn't match the actual `Statistics` type from `src/lib/statistics.ts` in several ways:

1. **Field names are wrong**: the spec uses singular (`topic`, `person`, `party`) but the handler returns plural (`topics`, `people`, `parties`) — see the `Statistics` interface at `src/lib/statistics.ts:12-17`.

2. **Property names are wrong**: the spec says `duration` but the actual type uses `speakingSeconds`. There's also a top-level `speakingSeconds` field in `Statistics` that isn't documented at all.

3. **Data structure is wrong**: the spec documents `{ id, name, count, duration }` but the actual `Stat<T>` type (line 7-11 of statistics.ts) is `{ item: T, speakingSeconds: number, count: number }` where `item` is the full entity object (e.g., the entire `Topic`, `Party`, or `PersonWithRelations`).

4. **Undocumented auth behavior**: the handler at `src/app/api/statistics/route.ts:18` conditionally includes unreleased meetings based on `isUserAuthorizedToEdit({ cityId })`. This affects what data is returned but isn't mentioned in the spec.

This route needs to be rebuilt against the actual `Statistics` type.

src/app/[locale]/(other)/docs/page.tsx
```diff
@@ -1,19 +1,92 @@
...
+import { filterSpecByAccessLevel, getUserAccessLevel, type AccessLevel, type OpenApiSpec } from '@/lib/utils/openapi';
 
-async function getSpec() {
-  const yamlString = fs.readFileSync('./swagger.yaml', 'utf8');
-  const spec = yaml.load(yamlString);
-  return spec as Record<string, any>;
+async function getSpec(): Promise<OpenApiSpec> {
+  const yamlString = await fs.readFile('./swagger.yaml', 'utf8');
+  return yaml.load(yamlString) as OpenApiSpec;
 }
```
Comment (kouloumos) [inline src/app/[locale]/(other)/docs/page.tsx:16]: `getSpec()` has no error handling — if `swagger.yaml` is missing or contains invalid YAML, this will crash the entire `/docs` page with an unhandled error. The `/api` route handler (`src/app/api/route.ts`) correctly wraps the same operation in try/catch and returns a 500. Should be consistent.

package.json
```diff
@@ -27,13 +27,15 @@
...
-    "transform-regulation": "tsx scripts/transform-regulation-coordinates.ts"
+    "transform-regulation": "tsx scripts/transform-regulation-coordinates.ts",
+    "generate-openapi": "tsx scripts/generate-openapi.ts"
   },
   "prisma": {
     "seed": "tsx prisma/seed.ts"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.56.0",
+    "@asteasolutions/zod-to-openapi": "^7.3.4",
```
kouloumos: This is listed under `dependencies` but should be `devDependencies`. The library is only consumed by `scripts/generate-openapi.ts` and the `src/lib/openapi/` module — nothing at runtime (the API route and docs page read from the generated `swagger.yaml`, they don't import from `src/lib/openapi/`).

kouloumos (commented):
The infrastructure pieces here are solid -- the registry setup in `registry.ts`, the generation script, `filterSpecByAccessLevel()` with its typo-safe defaults, the `AccessBanner`, and extracting `getUserAccessLevel()` into a shared utility. That's exactly what I was hoping for from the zod-to-openapi pivot.

The main concern is **spec accuracy** — left inline comments with the details, but the pattern across all four route files is the same: the schemas were written from scratch rather than derived from the code they're documenting. The statistics route has wrong field names, wrong property names, and a wrong data structure; the meetings route documents a 200 that's actually a 201, is missing response fields, etc. These are things you'd catch immediately by hitting the actual endpoints and comparing the JSON response to your schema — and I'd strongly recommend making that part of your workflow for this kind of work. For example: start the dev server, `curl localhost:3000/api/statistics?cityId=chania`, and compare the response to `StatisticsResponseSchema`. You'll see the field names don't match.

More broadly, the whole motivation for moving to code-first was to keep docs in sync with the code. The way to do that is to **import and extend the existing Zod schemas** — the ones the handlers already use for validation — rather than writing new ones for OpenAPI. For example, `src/lib/zod-schemas/city.ts` already defines `createCityFormDataSchema` which the `POST /api/cities` handler uses. The search handler in `src/app/api/search/route.ts` already has a `searchRequestSchema` that's identical to what you wrote independently. When you define separate schemas for the spec, the docs can (and do) drift from the code on day one, which is exactly what we were trying to avoid.

The same "run what you build" principle applies to the filtering: `npm run generate-openapi` was never run, so `swagger.yaml` is unchanged from `main` and has no `x-access-level` annotations. That means `filterSpecByAccessLevel()` defaults everything to `public`, the AccessBanner always shows 0 hidden paths, and the security goal isn't achieved yet. Running the script and loading `/docs` as a public vs. authenticated user would have surfaced this. You'll need to decide: run the script and commit the output (which means the spec drops to only the 4 POC routes while we migrate the rest), or keep the old spec until more routes are migrated and document that filtering isn't active yet. Either way it should be explicit.

One more process note: the two "Update" commits (`478c9ee`, `3c914be`) are greptile bot suggestions applied via the GitHub UI. One of them introduced an indentation bug in `meetings.ts` (line 28 — `meetingId` at column 0). Bot suggestions are a starting point, not a guaranteed fix — always review the diff before committing, especially with whitespace-sensitive code. In general, when applying fixes from review feedback, it's better to make changes locally where you can run the linter and verify things still work, rather than editing through the GitHub web UI.

Left inline comments on the specifics — mostly about schema mismatches against the actual handlers and a few code quality things.

kouloumos:
After testing locally and reading through the route files against the actual handlers, I found a number of issues that need to be addressed before we can merge. I pushed two commits to your branch @kdngiorgos  to fix some of them and to set a reference pattern for the rest.

**1b82772 — code quality fixes:**
- Removed the redundant `extendZodWithOpenApi(z)` calls from all four route files — it's already called once in `registry.ts`, that's enough (it's a global side-effect on Zod's prototype).
- Fixed `meetingId` indentation in `meetings.ts`.
- Added error handling to `getSpec()` in `docs/page.tsx` (the other spec loader in `api/route.ts` already had it).
- Moved `@asteasolutions/zod-to-openapi` to `devDependencies` — it's only used by the generate script, not at runtime.
- Switched from `OpenApiGeneratorV31` to `OpenApiGeneratorV3` (3.0 output). When I tested the docs page locally, expanding any endpoint got stuck in a loading state. The cause: the 3.1 generator produces `type: ['string', 'null']` for nullable fields, which `swagger-ui-react` can't render properly. The 3.0 generator produces `nullable: true` instead, which works. This is worth keeping in mind — always test the docs page locally after making changes.

**8043cba — cities route rewrite as exemplar:**
I rewrote `cities.ts` to demonstrate the pattern. Two key ideas:
1. **Reuse existing validation schemas for requests.** The request schema now imports `createCityFormDataSchema` from `src/lib/zod-schemas/city.ts` — the same schema the actual handler uses for validation. `zod-to-openapi` handles `.transform()` correctly (documents the input type, which is the wire format). The only override is `logoImage` because `z.instanceof(File)` doesn't serialize to OpenAPI.
2. **Reuse shared field definitions for responses.** I exported `baseCityFields` from `zod-schemas/city.ts` and spread it into the response schema. The 12 core model fields (names, enums, booleans) are now defined once and shared across frontend forms, backend validation, and OpenAPI docs. The response schema just adds DB-generated fields (`id`, `createdAt`, `updatedAt`, etc.).

The general approach: look at the actual route handler, check what Zod schema it uses for validation (can you import it?), what it returns (does the response schema match?), and what status codes it uses.

## The swagger.yaml problem

This is important: **don't run `npm run generate-openapi` yet.** The existing hand-written `swagger.yaml` documents 10 API paths. The route registrations in `src/lib/openapi/routes/` currently only cover 4 of them (cities, meetings, search, statistics). Regenerating would silently drop these 6 paths:

- `/api/cities/all`
- `/api/cities/{cityId}`
- `/api/cities/{cityId}/meetings/{meetingId}`
- `/api/cities/{cityId}/parties` and `/{partyId}`
- `/api/cities/{cityId}/people` and `/{personId}`

So right now, the new infrastructure generates a *less complete* spec than what we already have. We can only safely regenerate once all the old paths have route registrations. Until then, the hand-written spec stays.

## What needs to happen to merge this

Here's what I'd suggest as a path forward. Let me know if you have questions or want to approach any of this differently.

### 1. Fix the search and meetings routes

**Search** is the easiest win. The handler (`src/app/api/search/route.ts:6-26`) defines a `searchRequestSchema` that's nearly identical to the one in the OpenAPI file — extract it to `src/lib/zod-schemas/search.ts` and import from both places. Also check the response schema against what `search()` actually returns.

**Meetings** has a few issues:
- The handler returns `201` on create (`route.ts:124`), but the OpenAPI spec says `200`.
- The handler's `meetingSchema` has the real validation rules — consider extracting to `src/lib/zod-schemas/meeting.ts`.
- Check the response schema against what `createCouncilMeeting()` and `getCouncilMeetingsForCity()` actually return.

### 2. Drop statistics from this PR

The statistics response schema has significant mismatches with the actual implementation — wrong field names (`topic` vs `topics`, `duration` vs `speakingSeconds`), wrong structure (flat objects vs the `Stat<T>` wrapper in `src/lib/statistics.ts:7-17`), missing top-level `speakingSeconds`. Fixing it properly is a separate effort. Remove the statistics route file from this PR and add it in a follow-up.

### 3. Add route registrations for the missing paths

The 6 paths listed above need `registerPath()` calls so the generated spec covers everything the hand-written one does. This means:
- Extending `cities.ts` to cover `/api/cities/all` and `/api/cities/{cityId}`
- Extending `meetings.ts` to cover `/api/cities/{cityId}/meetings/{meetingId}`
- Creating new route files for parties and people

### 4. Regenerate swagger.yaml

Once all old paths are covered (and statistics is removed), run `npm run generate-openapi` and commit the result. At that point the generated spec should be a superset of the old one, and we can safely replace it.

### 5. Test locally

For each route: run the dev server, open `/docs`, expand the endpoints, and compare what you see against the actual API responses. That's the fastest way to catch mismatches. Also test with different user roles (logged out, logged in, admin) to verify the access-level filtering works.

kdngiorgos:
Thanks for the detailed review and the reference commits — they were very helpful.

Here's what was addressed:

**Schema extraction**: Moved `searchRequestSchema` to `src/lib/zod-schemas/search.ts` and `meetingSchema` to `src/lib/zod-schemas/meeting.ts`. Both are now imported by their respective route handlers and OpenAPI registration files.

**Route coverage**: Added registrations for the 6 missing paths (`/api/cities/all`, `/api/cities/{cityId}`, `/api/cities/{cityId}/meetings/{meetingId}`, parties, people). `UpdateCityRequest` reuses `updateCityFormDataSchema` from `zod-schemas/city` following the same pattern as cities POST. `MeetingSchema` now includes `administrativeBody` to match `CouncilMeetingWithAdminBody`.

**Statistics**: Removed from this PR. Will add in a follow-up once the schema mismatches are sorted out.

**swagger.yaml**: Regenerated — all 10 paths covered, `x-access-level` annotations active.

kouloumos:
Hey @kdngiorgos! 👋 Apologies for the long delay getting back to this — thanks for your patience.

Main had moved on quite a bit (300+ commits) and your branch was out of sync, so I've brought it up to date in two steps:

**1. Rebase onto latest main** — [`e1cfe514..b9701be1`](https://github.com/schemalabz/opencouncil/compare/e1cfe51497f3688a80e8b9c4324089a57214c9d5..b9701be1da0b81fa106d2313c4557659372bc26c)
Replayed your two commits on top of current main. Three conflicts came up, resolved as follows:
- `package.json` / `package-lock.json` — unioned your new dev deps (`@asteasolutions/zod-to-openapi`, `@testcontainers/postgresql`, `@types/aws-sdk`) with the ones main added.
- `src/app/[locale]/(other)/docs/page.tsx` — kept **both** main's new `generateMetadata`/hreflang **and** your role-based access filtering.

**2. Fix two pre-existing type errors** — [`b9701be1..2cabaa7c`](https://github.com/schemalabz/opencouncil/compare/b9701be1da0b81fa106d2313c4557659372bc26c..2cabaa7c3db8809d0d31e6fafc64ed4511606d38)
After rebasing, `tsc` surfaced two issues that predate the rebase:
- `src/lib/zod-schemas/meeting.ts` was missing the `processAgenda` field that the meetings route destructures — restored it.
- `docs/page.tsx` typed `params` synchronously; Next 16 requires `params: Promise<…>` (awaited) — updated to match.

The branch now typechecks cleanly (`tsc --noEmit` passes) and is 0 commits behind main. I squashed the fixes into your first commit, so your history stays as the same two commits.

The original branch state before all this is preserved at [`backup/pr-220-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-220-original) on our fork, so nothing is lost.

I'll continue the review from here — thanks again!

kouloumos:
Hey @kdngiorgos! Since this had been waiting a while, I went through the remaining review items — yours, mine, and Greptile's — and resolved them directly so we can land the groundwork. Pushed on top of your work:

**Folded into your two commits** (kept as-authored):
- **Security:** the spec filter now also prunes `components.schemas` reachable only from hidden operations — previously admin-only schema names/shapes still shipped in the public spec (Greptile's open flag).
- **`meetingId`** back to optional, so `POST …/meetings` auto-generates it as before.
- **Response schemas vs handlers:** real `Role` fields (`name`/`startDate`/… not `title`/`from`/`until`), an honest `search` results shape, `Meeting.videoUrl`/`audioUrl`, `AdministrativeBody.name_en`/`type`, and parties reuse the real `PersonWithRoles`. `swagger.yaml` regenerated.
- **DRY:** shared `MessageSchema` + `cityIdParam`. **Banner:** the public `/docs` note no longer implies a plain sign-in unlocks admin endpoints.

**Two new commits:** `test(api-docs)` (filter + access-level + meeting-schema tests) and `chore(deps)` (dropped two stray devDeps a rebase had re-introduced).

Verified: typecheck + full suite pass, and `swagger.yaml` matches the generator.

**Scope:** this lands the code-first groundwork for the 10 currently-documented paths. Still open under #45 (follow-ups, not blockers here): the statistics route, coverage for the remaining admin/consultations/auth endpoints, and the API contribution guide. One handler-side item I did **not** fold in: `withUserAuthorizedToEdit` throws a plain `Error`, so some auth failures return `500` rather than `401/403` — worth a small separate fix.

Original pre-rebase state remains at [`backup/pr-220-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-220-original). Thanks for the solid groundwork here!

swagger.yaml
```diff
@@ -583,289 +411,1177 @@ components:
...
+          description: Unauthorized — authentication required
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '409':
+          description: Conflict — a meeting with this ID already exists
+          content:
+            application/json:
+              schema:
```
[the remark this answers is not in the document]
kouloumos: Good catch — confirmed the documented `409` was unreachable: an auto-generated id retries on collision (→ `201`), and a client-supplied duplicate currently surfaces as `500`, not `409`. Removed the `409` from the spec so it matches the handler (`201`/`400`/`401`).

Making a client-supplied duplicate return a real `409` (via `ConflictError`) is a small handler-side improvement — tracked as a follow-up along with aligning the other auth failures (`withUserAuthorizedToEdit` currently 500s instead of `401/403`) to their documented codes.

kouloumos:
Another round getting this merge-ready:

**1. Serve the spec in-process — no committed `swagger.yaml`.** Since it's fully generated from the Zod schemas, committing it was just build output that could drift. Both `/docs` and `GET /api` now call `generateSpec()` at runtime (memoized); the file is gitignored and `@asteasolutions/zod-to-openapi` moved to runtime deps. Verified with a production build plus a runtime check that generation/filtering/component-pruning all work. Also removed the `/docs` access banner — it counted *paths* not *operations*, so it read "N of N" for every viewer regardless of role.

**2. Audited every registered operation against its real handler** and folded the schema-vs-handler mismatches into the registrations — same class as the PUT `409`/`processAgenda` items, found systematically:
- **Request:** PUT `/meetings/{id}` no longer advertises `processAgenda`/`meetingId` — it now uses `meetingSchema.omit(...)`, so it stays DRY-derived.
- **Responses now match what the handlers return:** cities `population`/`timezone`, `CityWithGeometry` includes `_count`, meetings list includes `subjects`, create includes `processAgendaStatus`, people POST returns the person **with roles**, and `Person`/`Role` gained `activeFrom`/`activeTo` and `createdAt`/`updatedAt`.

Kept it DRY — reused/derived from existing schemas rather than hand-mirroring. Search was clean.

**Deferred to follow-ups (not blockers):**
- **Request-body validation:** a few write handlers (parties, people, cities message fields) still raw-parse `formData` with no Zod schema, and the meetings-list query params aren't reused in the docs. The DRY fix is to extract shared Zod schemas the handlers `.parse()` through — which also gives them real `400` validation, so it's worth its own PR (it changes runtime behavior).
- **Status codes:** some auth failures surface as `500`/plain-text instead of the documented `401`/`403` (root cause: `withUserAuthorizedToEdit` throws a plain `Error`), and a missing person returns `200 null` instead of `404` — same handler-side follow-up.

kouloumos:
Caught a documentation regression while reviewing: `GET /api/utterance/{utteranceId}/context` was in the old hand-written spec but never got a `registerPath()`, so the code-first regen had silently dropped it. Restored it (mirroring the `UtteranceContext` type), and added a coverage test that asserts the generated spec documents exactly the intended operation set — so a dropped registration fails CI instead of quietly shrinking the docs.

Confirmed the rest of the spec is otherwise a superset of the old one (it *added* all the create/update/delete operations), so this endpoint was the only casualty.

kouloumos:
Hey @kdngiorgos — this is merged! 🎉

First, a genuine apology for how long it took to get over the line — that's on us, not you. Thank you for your patience through the rebases and review rounds, and for the solid groundwork you laid here. The code-first OpenAPI setup, the role-based `/docs` filtering, and the schema reuse are a real step up from the hand-maintained spec, and something we'll keep building on.

It's a genuinely useful contribution — and timely: it feeds directly into **#584 (making OpenCouncil's data usable by agents and third parties)**, which we're picking up next. A clear, auto-generated, role-aware API spec is exactly what makes our Open Data API discoverable and consumable by the assistants and tools people already use — so your work lands at a great moment for us.

We'd love to have you around. If you enjoyed this, a couple of natural follow-ups grew out of the review and would make great next contributions:

- **Request-body validation** — a few write handlers (parties, people, the cities message fields) still parse `formData` manually with no Zod schema, and the meetings-list query params aren't reused in the docs. Extracting shared Zod schemas the handlers `.parse()` through would keep the docs DRY *and* give those endpoints real `400` validation.
- **Handler status codes** — some auth failures surface as `500`/plain-text instead of the documented `401`/`403` (root cause: `withUserAuthorizedToEdit` throws a plain `Error`), and a missing person returns `200 null` instead of `404`.

Both are well-scoped as standalone PRs. And #45 stays open as the umbrella for the rest of the API-docs work (statistics route, the remaining admin/consultations/auth endpoints, a contribution guide).

Thanks again — really appreciate it, and hope to see you back. 🙌

PR #276, opened 2026-03-09
fix(i18n): fill English translation gaps and remove dead src/messages…

### LINKED ISSUE #33: Improve Localization System
### Motivation
Complete the localization infrastructure to properly support both Greek and English throughout the entire application.

### Current State
- **Technology**: We use `next-intl` for internationalization with routing support
- **Configuration**: Two supported locales (`en`, `el`) with Greek as default locale, routing configured in `src/i18n/routing.ts`
- **Translation Files**: Significant gap in translation coverage - Greek translations (~478 lines) vs English translations (~331 lines) in `messages/` directory with translation files split between `messages/en.json` and `src/messages/en.json`
- **Issue**: The codebase contains hardcoded strings throughout components that bypass the translation system, and the split translation files make maintenance difficult

**Examples of hardcoded strings found:**
- UI actions: `"Edit User"`, `"Create User"`, `"Save Changes"`, `"Saving..."` 
- User feedback: `"Loading..."`, `"Search Error"`, `"Show more"`, `"Show less"`
- Mixed language text: `"Απομαγνητοφώνηση"` (Greek) hardcoded in English UI context
- Form elements and buttons without translation keys

### Proposed Outline
1.  **Consolidate Translation Files**: Merge `messages/en.json` and `src/messages/en.json` into a single file in the root `messages/` directory.
1. **Audit hardcoded strings**: Systematically scan all `.tsx` files for hardcoded user-facing text that should use translation keys
2. **Expand English translations**: Add missing translations to `messages/en.json` to achieve parity with Greek translations in `messages/el.json`
3. **Replace hardcoded strings**: Convert all hardcoded text to use `useTranslations()` hooks and translation keys
4. **Ensure translation consistency**: Validate that all translation keys exist in both language files
5. (optional) **Add development tooling**: Implement checks to prevent future hardcoded strings from being committed

> [!NOTE]
> **Note for Contributors**: Before you begin, please read our [CONTRIBUTING.md](CONTRIBUTING.md) guide to understand our development workflow and co-creation process. We look forward to your contribution!

kdngiorgos:
Closes #33 

Fixes #33. The English locale was significantly less complete than Greek, causing raw translation keys to appear in the UI for English-locale users.

Root cause: `src/messages/en.json` (never loaded by next-intl, which only reads from `messages/{locale}.json`) contained keys that were missing from the real `messages/en.json`, giving the false impression those strings were translated.

Changes:
- Delete `src/messages/en.json` (dead file, never loaded by next-intl request config; its presence was misleading)
- Add all missing top-level sections to `messages/en.json`: InputWithDerivatives, Chat, Onboarding, Profile, RolesList, AdministrativeBodiesList
- Add missing keys to existing sections: MeetingCard.beingProcessed City.item / addItem / noItems Person.item / recentSegments / tryDifferentFilter PersonForm: profileUrl, profileUrlDescription, personUpdated, personCreated, personNamePlaceholder, personNameDescription, personNameEnPlaceholder, error, formErrors, roles, success Subject: voting, votingDisclaimer, noVotingUtterances, groupedDiscussion, groupedDiscussionDescription AboutPage: full features / pricing / about sub-trees AddMeetingForm: administrativeBodyType, formErrors OfferForm: recipientName, recipientNameDescription, platformPrice, platformPriceDescription RolesList: from, until (missing from the dead file's version)

Verified: deep recursive key comparison confirms messages/en.json now contains every key present in messages/el.json. UI tested at /en/athens showing correct English labels (Council Meetings, People, Parties, etc.).

<h3>Greptile Summary</h3>

This PR fixes the root cause of raw translation keys appearing for English-locale users: `src/messages/en.json` (a dead file never loaded by next-intl) was giving a false sense of coverage, while the real `messages/en.json` was missing large sections. The fix deletes the dead file, backfills all missing keys into `messages/en.json` to match `messages/el.json`, mirrors the symmetric gaps in `el.json`, and re-enables the translation parity test with a deep bidirectional key comparison.

Safe to merge; the duplicate Profile key from the previous review cycle is resolved and no new structural issues were introduced.

Only P2 findings remain. The previous P1 concerns (duplicate Profile key, silently-dropped allowContact keys) are resolved in the current state of the PR.

src/lib/__tests__/translations.test.ts — the static JSON import has a silent duplicate-key blind spot worth hardening before the next translation-sync issue.

| Filename | Overview |
|----------|----------|
| messages/en.json | Adds all previously missing top-level sections and keys to match el.json; no remaining duplicate keys after this PR resolves the earlier Profile duplication |
| messages/el.json | Adds PersonForm role/party keys, PersonCard section, and Offer section that were already present in en.json |
| src/lib/__tests__/translations.test.ts | Test unskipped and upgraded to deep bidirectional key comparison; static JSON import still has a silent duplicate-key blind spot (P2) |
| src/messages/en.json | Dead file correctly deleted; it was never loaded by next-intl and its presence was misleading developers about translation coverage |
| messages/en/highlights.json | Adds returnToTranscriptDescription key to match the existing entry already present in el/highlights.json |
| messages/el/highlights.json | Adds subjectConnected to highlightCard section; en/highlights.json already had this key |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[next-intl request config] -->|loads| B[messages/en.json ✅]
    A -->|loads| C[messages/el.json ✅]
    A -->|loads| D[messages/en/highlights.json ✅]
    A -->|loads| E[messages/el/highlights.json ✅]
    F[src/messages/en.json] -->|❌ never loaded - DELETED| G[Dead file removed]
    B -->|bidirectional deep key test| H[translations.test.ts]
    C -->|bidirectional deep key test| H
    D -->|modular file test| H
    E -->|modular file test| H
```

<details><summary><h3>Comments Outside Diff (10)</h3></summary>

1. `docker-entrypoint.sh`, line 1 ([link](https://github.com/schemalabz/opencouncil/blob/e383ff5248edf34fab8c9c929544b89c8210a895/docker-entrypoint.sh#L1)) 

   **Unrelated executable-bit removal**

   The only change to this file is a mode change from `100755` (executable) to `100644` (non-executable). This appears to be an accidental side-effect of a local Git/editor operation unrelated to the i18n fix.

   The Dockerfile compensates with an explicit `RUN chmod +x /usr/local/bin/docker-entrypoint.sh` (line 34), so the Docker image itself is unaffected. However, stripping the execute bit in source control means anyone who tries to invoke the script directly (e.g. `./docker-entrypoint.sh` in local dev or a CI step that doesn't go through Docker) will get a `Permission denied` error.

   Consider reverting the mode change to keep the file executable in source control:
   ```
   git update-index --chmod=+x docker-entrypoint.sh
   ```

2. `messages/en.json`, line 430-461 ([link](https://github.com/schemalabz/opencouncil/blob/2cc9f8b8518c0ae20eeec13db874422fd7d756bb/messages/en.json#L430-L461)) 

   **Duplicate `RolesList` key silently drops `noEndDate` translation**

   This PR adds a second top-level `"RolesList"` object at line 430, but one already exists earlier in the file at line 385. JSON does not allow duplicate keys — `JSON.parse` (and next-intl's loader) will silently use the **last** occurrence, meaning the first block (lines 385-417) is completely discarded.

   The critical consequence is that `"noEndDate": "No end date"` present in the first (now-discarded) block is absent from this second block. This key is actively used in `src/components/persons/RolesList.tsx` lines 405 and 408:

   ```tsx
   aria-label={t('noEndDate')}
   // ...
   <div className="font-medium text-foreground text-xs">{t('noEndDate')}</div>
   ```

   English users will see a raw translation key `RolesList.noEndDate` in the UI — exactly the regression this PR is trying to prevent.

   **Fix**: Remove the duplicate second `"RolesList"` block (lines 430-461) and instead merge its unique keys (`from`, `until`) into the first `"RolesList"` block (lines 385-417).

3. `messages/en.json`, line 462-501 ([link](https://github.com/schemalabz/opencouncil/blob/2cc9f8b8518c0ae20eeec13db874422fd7d756bb/messages/en.json#L462-L501)) 

   **Duplicate `AdministrativeBodiesList` key — newly added block is dead code**

   This PR adds an `"AdministrativeBodiesList"` block at line 462, but a more complete one already existed at line 647 of the file. Because JSON parsers use the last occurrence, this newly added block is **silently ignored entirely**.

   The pre-existing block at lines 647-689 (which wins) contains additional keys not present in the new block, including `diavgeiaUnitIds`, `diavgeiaUnitIdsPlaceholder`, and `diavgeiaUnitIdsDescription`. The new block also has some slightly different wording (e.g. `"contactEmailsCC": "CC"` vs `"Carbon Copy (CC)"` and `"submitting": "Saving..."` vs `"Submitting..."`).

   Since the original block wins, these wording differences and missing keys in the new block have no effect — but the new block should be removed to avoid future confusion. If there are keys in the new block that aren't in the original (e.g. different wording choices), those should be reconciled in the single surviving block.

4. `docker-entrypoint.sh`, line 1 ([link](https://github.com/schemalabz/opencouncil/blob/2cc9f8b8518c0ae20eeec13db874422fd7d756bb/docker-entrypoint.sh#L1)) 

   **Executable bit removed from `docker-entrypoint.sh`**

   This commit removes the executable bit from `docker-entrypoint.sh` (mode `100755` → `100644`). Both `Dockerfile` and `Dockerfile.dev` run `chmod +x /usr/local/bin/docker-entrypoint.sh` after copying the file, so Docker builds are unaffected. However, developers who try to run `./docker-entrypoint.sh` locally will get a "Permission denied" error.

   This appears to be an unintentional change (likely from a Git client that doesn't preserve execute bits). Consider reverting it:
   ```bash
   git update-index --chmod=+x docker-entrypoint.sh
   ```

5. `messages/en.json`, line 430-461 ([link](https://github.com/schemalabz/opencouncil/blob/5db56933d10209c38acd3c42372be65363522a70/messages/en.json#L430-L461)) 

   **Duplicate `RolesList` key silently drops `noEndDate` translation**

   This PR adds a second `"RolesList"` object (lines 430–461), but one already exists at lines 385–417. JSON does not support duplicate keys — `JSON.parse` silently uses the **last** occurrence, discarding the first. The first entry (lines 385–417) contains `"noEndDate": "No end date"`, which mirrors the Greek locale (`el.json`: `"noEndDate": "Χωρίς ημερομηνία λήξης"`). The PR-added entry is missing this key, so after parsing, `t("RolesList.noEndDate")` will fall back to the raw key string — exactly the bug this PR intends to fix.

   Beyond the missing key, the two entries also disagree on existing values:
   - `isHeadDescription`: `"Mark if the person is the head for this role"` (original) vs `"Set if the person is head in this role"` (PR)
   - `cityRole`: `"City Council Member"` (original) vs `"City Role"` (PR)
   - `present`: `"present"` (lowercase, original) vs `"Present"` (capitalized, PR)

   The fix is to **remove this duplicate block** and instead merge any corrections and the `from`/`until` keys directly into the existing `RolesList` at lines 385–417 (noting that `from` and `until` are already present in the original). Also add the missing `noEndDate` key:

   
   (Delete lines 430–461 entirely, and update the existing `RolesList` at line 385 if needed.)

6. `messages/en.json`, line 462-501 ([link](https://github.com/schemalabz/opencouncil/blob/5db56933d10209c38acd3c42372be65363522a70/messages/en.json#L462-L501)) 

   **Duplicate `AdministrativeBodiesList` key — PR-added block is unreachable dead code**

   An `"AdministrativeBodiesList"` object already exists at lines 647–691. Because `JSON.parse` uses the **last** occurrence, this block (lines 462–501) is entirely ignored at runtime.

   The pre-existing entry (lines 647–691) is more complete — it includes keys absent from this block: `showUnreviewedTranscript`, `showUnreviewedTranscriptDescription`, `diavgeiaUnitIds`, `diavgeiaUnitIdsPlaceholder`, `diavgeiaUnitIdsDescription`. It also has more precise values for shared keys (e.g. `contactEmailsCC: "Carbon Copy (CC)"` here vs `"CC"`, and `submitting: "Submitting..."` vs `"Saving..."`).

   This block should be **deleted entirely**. Any genuinely missing keys should be merged into the existing entry at lines 647–691.

7. `docker-entrypoint.sh`, line 1 ([link](https://github.com/schemalabz/opencouncil/blob/5db56933d10209c38acd3c42372be65363522a70/docker-entrypoint.sh#L1)) 

   **Unintentional executable-bit removal on entrypoint script**

   This file's mode changed from `100755` (executable) to `100644` (non-executable). A Docker entrypoint script must be executable — removing the bit will prevent the container from starting if the `ENTRYPOINT` references this script directly.

   If this was unintentional (likely, given the PR scope), restore the bit:

   ```
   git update-index --chmod=+x docker-entrypoint.sh
   ```

8. `src/lib/__tests__/translations.test.ts`, line 9-13 ([link](https://github.com/schemalabz/opencouncil/blob/436294d4bf54603b3a8e2cc346ddcfe98aa60fad/src/lib/__tests__/translations.test.ts#L9-L13)) 

   **Test will fail — `Offer` and `PersonCard` missing from `el.json`**

   The test checks both directions of parity (`missingInEn` and `missingInEl` must both be empty), but `messages/en.json` has two top-level sections — `Offer` and `PersonCard` — absent from `messages/el.json`. The assertion `expect(missingInEl).toEqual([])` will fail with `["Offer", "PersonCard"]`, breaking CI immediately.

   The second nested-key test will also fail for `PersonForm`, where EN has nine keys (`party`, `partyDescription`, `role`, `roleDescription`, `roleEn`, `roleEnDescription`, `roleEnPlaceholder`, `rolePlaceholder`, `selectParty`) absent from EL's `PersonForm` section.

   Fix: either add the missing sections/keys to `messages/el.json`, or change the bidirectional check to one-directional (only require that EL keys are present in EN, not the reverse).

9. `messages/en.json`, line 332-351 ([link](https://github.com/schemalabz/opencouncil/blob/d08db56f98f467047a22d7ecd5fe9c2d28929a7a/messages/en.json#L332-L351)) 

   **Duplicate `Profile` key — this entire block is dead code**

   `messages/en.json` already contains a `"Profile"` section at line ~493 with a much more complete set of keys (`phonePlaceholder`, `phoneInvalid`, `emailChangeTooltip`, `allowProductUpdates`, all tab/account/danger-zone keys, etc.). JSON parsers resolve duplicate top-level keys by keeping the **last** occurrence, so this newly-added `Profile` block is silently discarded at runtime and in TypeScript imports.

   The two keys that exist only in this block — `allowContact` and `allowContactDescription` — are consequently dropped and will never be surfaced by next-intl. If those keys are needed, they should be added to the pre-existing `Profile` section (and their Greek equivalents checked in `el.json`). If they are not needed, this block should simply be removed.

10. `src/lib/__tests__/translations.test.ts`, line 3-4 ([link](https://github.com/schemalabz/opencouncil/blob/d10c5f5ca1c11f4245acbc140b8ded3ec99e79b7/src/lib/__tests__/translations.test.ts#L3-L4)) 

    **Test has a blind spot for duplicate top-level keys**

    `import en from '../../../messages/en.json'` uses the JS/TS JSON parser, which silently keeps the **last** value when a key appears twice. `messages/en.json` currently has two `"Profile"` objects (lines ~332 and ~493); the import resolves to the second, so the first block (containing the newly-added `allowContact`/`allowContactDescription`) is invisible to this test. Any future translation key added inside the shadowed first block would never be flagged as missing by this suite.

    A straightforward guard is to parse the raw text and assert no duplicate top-level keys:

    ```ts
    it('en.json should have no duplicate top-level keys', () => {
        const raw = fs.readFileSync(path.join(messagesDir, 'en.json'), 'utf8');
        const keys: string[] = [];
        (raw.match(/"(\w+)"\s*:/g) ?? [])
            .forEach(m => keys.push(m.replace(/[" :]/g, '')));
        const topLevel = keys.filter((k, i) => keys.indexOf(k) !== i);
        expect(topLevel).toEqual([]);
    });
    ```

    (A proper approach would use a streaming JSON parser that errors on duplicates, but even this regex check would have caught the `Profile` duplicate.)
</details>

`````markdown
Path: src/lib/__tests__/translations.test.ts
Line: 25-27

Comment:
**Duplicate-key blind spot in `loadJson` and static imports**

Neither `JSON.parse` (used here for modular files) nor the top-level `import en`/`import el` will detect duplicate top-level keys — both silently keep the last occurrence. The previous `Profile` duplicate was caught by hand, but the suite won't automatically fail if it regresses. A lightweight guard would close the gap:

```ts
function assertNoDuplicateKeys(filePath: string): void {
    const raw = fs.readFileSync(filePath, 'utf8');
    const topLevelKeys = [...raw.matchAll(/"([^"]+)"\s*:/g)]
        .filter(m => raw.slice(0, m.index!).split('{').length === 2)
        .map(m => m[1]);
    const seen = new Set<string>();
    const duplicates = topLevelKeys.filter(k => !seen.has(k) ? !seen.add(k) : true);
    expect(duplicates).toEqual([]);
}
```

Alternatively, switch the main `en`/`el` comparison to `loadJson` so both paths are consistent and easier to extend with a raw-text guard later.

`````

</details>

<sub>Reviews (9): Last reviewed commit: ["test(i18n): enable and strengthen transl..."](https://github.com/schemalabz/opencouncil/commit/2eb705501b8cc9eee1312447cef0af8c37e19ef4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=24513524)</sub>

### CODE REVIEW

File: messages/en.json [hunk 1]
```diff
@@ -283,7 +427,83 @@
         "activeUntil": "Member until",
         "active": "Member"
     },
+    "RolesList": {
```
kouloumos: `RolesList` already exists at line 385. JSON doesn't allow duplicate keys — the parser silently uses the last one and throws away the first. In this case that means:

- `noEndDate` is **lost** (it's in the first block but not this one). It's used in `src/components/persons/RolesList.tsx:405,408` — English users will see the raw key string.
- `cityRole` changes from `"City Council Member"` to `"City Role"` — the original matches the Greek (`"Δημοτικός Σύμβουλος"`) and is more descriptive. This is used as a fallback label in `RolesList.tsx:440`.
- `isHeadDescription` and `present` also get silently overwritten with slightly different values.

Fix: delete this entire duplicate block (lines 430–461). The original at line 385 already has `from` and `until`, so nothing is gained here.

messages/en.json
```diff
@@ -283,7 +427,83 @@
...
+        "update": "Update",
+        "create": "Create",
+        "confirmDelete": "Are you sure you want to delete this role?",
+        "cityRole": "City Role",
+        "head": "Head",
+        "present": "Present",
+        "from": "from",
+        "until": "until"
+    },
+    "AdministrativeBodiesList": {
```
kouloumos: Same issue — `AdministrativeBodiesList` already exists at line 647. Since JSON keeps the last occurrence, this block is entirely dead code. The existing one at 647 is actually more complete (has `diavgeiaUnitIds`, `showUnreviewedTranscript`, and better wording like `"Carbon Copy (CC)"` vs `"CC"`).

Fix: delete this block (lines 462–501).

docker-entrypoint.sh
kouloumos, line 1: This file lost its executable bit (755 → 644). This is a leftover from the `.gitattributes` you added in the first commit and deleted in the second — the `.gitattributes` is gone but the mode change it caused stuck around.

Fix: `git update-index --chmod=+x docker-entrypoint.sh`

kouloumos (commented):
Thanks for working on this @kdngiorgos -- deleting the dead `src/messages/en.json` is the right call, and the bulk of the new translations are correct. But **there are a couple of issues that need fixing before this can merge**, left inline comments on the specifics.

The main one: `messages/en.json` now has duplicate top-level JSON keys (`RolesList` appears twice, `AdministrativeBodiesList` appears twice). JSON doesn't support duplicate keys — `JSON.parse` silently keeps the *last* occurrence and discards the earlier one. The result is that the duplicate `RolesList` actually *drops* the `noEndDate` key, which is used in `src/components/persons/RolesList.tsx` — so English users will see a raw translation key in the UI. **That's exactly the bug this PR is trying to fix**.

**We actually have a test for this**: `src/lib/__tests__/translations.test.ts` checks en/el key parity — it's currently skipped with a TODO saying "enable once el.json and en.json have matching keys." Since that's exactly what this PR achieves, unskipping it (`describe.skip` → `describe`) would be the perfect way to verify your work and prevent future regressions. If you run it right now against your branch (`npm test -- translations`), it will fail because of the missing `noEndDate` — which is how the duplicate key issue would've surfaced.

The other issue is about commit hygiene. Your first commit adds `.gitattributes`, and your second commit deletes it. That's fine -- you changed your mind. But the side effect of `.gitattributes` (it changed `docker-entrypoint.sh` from executable to non-executable) survived the deletion. **When you undo a change, always check whether it had side effects that also need undoing**. And ideally, these two commits should be squashed into one -- there's no reason to keep a "add file" / "delete file" pair in the history.

kouloumos:
@kdngiorgos I pushed two commits on top of your branch with the fixes from the review:

- 0f80d2fb — removes the duplicate `RolesList` and `AdministrativeBodiesList` blocks, restores the executable bit on `docker-entrypoint.sh`
- 436294d4 — unskips the translation key parity test (now that en/el are in sync, this test will catch future regressions)

What's left on your end is squashing the commit history. Right now the branch has 4 commits, but it should be 2: one for the translation fix and one for enabling the test. The "Delete .gitattributes" commit and my fix commit should be folded into your original commit — they're all part of the same change.

Have you done an interactive rebase before? If not, no worries -- it's a useful skill and this is a good place to learn it. We describe the general workflow in our [CONTRIBUTING.md](https://github.com/schemalabz/opencouncil/blob/main/CONTRIBUTING.md#squashing-commits), but here are the exact steps for this branch:

```bash
git rebase -i HEAD~4
```

This opens an editor with 4 lines. Change it to look like this:

```
pick 727a6389 fix(i18n): fill English translation gaps and remove dead src/messages/en.json
squash 5db56933 Delete .gitattributes
squash 0f80d2fb fix: remove duplicate JSON keys and restore docker-entrypoint.sh executable bit
pick 436294d4 test: enable translation key parity test
```

The `squash` lines fold those commits into the one above them. After saving, git will open another editor for you to write a combined commit message — clean it up into a single coherent message (the original message is a good base, just drop the `.gitattributes` parts). Then force-push:

```bash
git push -f
```

kdngiorgos:
Done — squashed to 2 commits as requested.

One thing I noticed while squashing: the test has bidirectional assertions (`expect(missingInEl).toEqual([])` in both cases), but `messages/en.json` has two top-level sections (`Offer`, `PersonCard`) and several `PersonForm` keys that don't exist in `el.json` — so the test would immediately fail in CI. Since the goal of this PR is to ensure EN covers all EL keys (not the reverse), I changed both assertions to one-directional: only assert that `missingInEn` is empty. Both tests now pass.

kouloumos:
Hey! Main has moved forward and your branch was out of sync (86 commits behind). I've rebased your commits on top of the current main and force-pushed the result.

There were two conflicts in `messages/en.json` where new voting-related keys had been added to main — I resolved them by keeping both sides (main's expanded voting keys + your `groupedDiscussion` and `formErrors` additions).

The original branch state before the rebase is preserved at [`backup/pr-276-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-276-original), so nothing is lost.

kouloumos (approved):
All round 1 issues addressed -- duplicate keys gone, permission bit restored, commits squashed cleanly. The translations themselves look good.

There was one more duplicate key hiding in `messages/en.json`: `Profile` appeared twice (your block at line 332 and the pre-existing one further down). The second one is more complete and the two unique keys in yours (`allowContact`, `allowContactDescription`) aren't used anywhere in the codebase, so I removed the duplicate block. Also rewrote the parity test (d08db56f was the original push, now squashed into 2eb70550) -- it's now bidirectional and recursive at all nesting levels, and covers the modular files under `messages/{locale}/` too. That caught a few more gaps: 37 keys in `el.json` had no Greek equivalent (`PersonForm` role fields, `PersonCard`, `Offer`), plus one missing key in each direction in `highlights.json`. All fixed.

Full bidirectional parity across all translation files now, monolithic and modular. Ready to merge.

PR #287, opened 2026-03-20
Fix empty utterance jump

### LINKED ISSUE #164: Editing mode: Unable to write in new text segment in new speaker
https://www.loom.com/share/30efab0967584d0c98dbe872acafd6da

This is the issue that we were discussing, that there are times that we create a new speaker segment, and a new text segment. Then, when attempting to write in this segment, we are redirected somewhere else in the page.

The solutions we have tried is either zooming out and writing in the segment, or refreshing the page.

fyi @thanos

angelospk:
Closes #164

## Summary
This PR fixes transcript UI desynchronization during keyboard navigation, especially after inserting empty speaker segments. The main issue was unstable segment indexing in transcript visibility/scroll tracking, which could cause the UI to jump to the wrong area while audio progression remained correct.

## Changes Made
- Updated `Transcript` visibility and observer tracking to use stable `segment.id` values instead of array indices.
- Updated transcript wrapper keys/IDs to use `segment.id` (`speaker-segment-{segment.id}`) for deterministic DOM targeting after insertions/deletions.
- Added chronological transcript normalization in `CouncilMeetingDataContext` before exposing transcript data to consumers.
- Refactored `VideoProvider` to consume utterances from `CouncilMeetingDataContext` directly (removed `utterances` prop from wrapper/provider contract).
- Hardened `VideoProvider` seek synchronization by updating `currentTime` state immediately on seek operations (`seekTo`, `seekToWithoutScroll`, and `onSeeked`).
- Added/updated tests:
  - `src/components/meetings/__tests__/VideoProvider.test.tsx`
  - `src/components/meetings/transcript/__tests__/Transcript.test.tsx`
- Minor test cleanup: removed unused imports in `VideoProvider.test.tsx`.

## Testing
- `npx jest src/components/meetings/transcript/__tests__/Transcript.test.tsx src/components/meetings/__tests__/VideoProvider.test.tsx --runInBand --watchman=false`
- `npx eslint src/components/meetings/transcript/Transcript.tsx`
- Manual verification on the affected meeting confirmed stable transcript scrolling behavior around inserted empty segments.

<h3>Greptile Summary</h3>

This PR fixes transcript UI desynchronization during keyboard navigation by switching segment tracking from unstable array indices to stable `segment.id` values, adding chronological normalization via `orderedTranscript`, hardening seek synchronization with immediate `currentTime` state updates, and removing the `utterances` prop from `VideoProvider` in favor of direct context consumption.

Most of the issues flagged in previous review rounds have been addressed: the `[0,0]` debounce reset is now bypassed correctly with `cancel()`, `displayedSegments` is back in the observer effect's dep array so new segments are observed, `seekTo` still calls `updateHighlightOnce`, and the stale-ID eviction effect handles deletions.

Safe to merge — all previously flagged P0/P1 issues have been addressed and remaining findings are minor style suggestions.

All blocking concerns from prior rounds are resolved: seekTo calls updateHighlightOnce, displayedSegments is in the observer dep array so new segments are observed, the [0,0] reset bypasses the debounce via cancel(), and the stale-ID eviction effect is in place. The two remaining comments are P2 style improvements that do not affect correctness.

| Filename | Overview |
|----------|----------|
| src/components/meetings/VideoProvider.tsx | Removes utterances prop, consumes transcript from context, adds immediate currentTime state on seeks, and splits initial-time and deep-link effects; seek hardening is correct but the deep-link effect captures utterances/updateHighlightOnce by closure while using only utterances.length as a dep. |
| src/components/meetings/transcript/Transcript.tsx | Replaces index-based segment IDs with stable segment.id keys, introduces visibleSegments React state for renderMock control, eviction effect for stale IDs, and immediate [0,0] reset bypassing debounce; all previously flagged issues addressed. |
| src/components/meetings/CouncilMeetingDataContext.tsx | Adds orderedTranscript useMemo with index-based tiebreaker for equal timestamps; correctly exposes sorted transcript to all consumers and updates downstream memos. |
| src/lib/db/transcript.ts | Adds secondary createdAt sort to getTranscript to provide deterministic DB ordering that matches the client-side tiebreaker assumption. |
| src/lib/utils.ts | Adds cancel() method to debounce return type, used in Transcript to bypass the debounce when resetting scroll interval to [0,0]. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant U as User
    participant T as Transcript
    participant IO as IntersectionObserver
    participant VP as VideoProvider
    participant CTX as CouncilMeetingDataContext

    CTX->>CTX: orderedTranscript (sort by startTimestamp, then index)
    CTX-->>T: transcript (stable, chronological)
    CTX-->>VP: transcript → utterances (flatMap)

    T->>IO: observe(speaker-segment-{id})
    IO-->>T: handleIntersection → setVisibleSegments

    T->>T: calculateTimeInterval(visibleSegments)
    T->>VP: debouncedSetScrollInterval([start, end])

    U->>VP: seekTo(time)
    VP->>VP: currentTimeRef.current = time
    VP->>VP: setCurrentTime(time) [immediate]
    VP->>VP: updateHighlightOnce()
    VP->>T: scrollToUtterance(time) via rAF

    Note over T: segment deleted
    CTX-->>T: displayedSegments updated
    T->>T: evict stale IDs from visibleSegments
    T->>VP: setCurrentScrollInterval([0,0]) [immediate, no debounce]
    IO-->>T: new segment observed → visibleSegments updated
    T->>VP: debouncedSetScrollInterval([new start, new end])
```

<details><summary><h3>Comments Outside Diff (9)</h3></summary>

1. `src/components/meetings/VideoProvider.tsx`, line 112-127 ([link](https://github.com/schemalabz/opencouncil/blob/a3797c022bc7791ce9d51a06da827aa7bd3cbfe9/src/components/meetings/VideoProvider.tsx#L112-L127)) 

   <a href="#">P1</a> **`loadedmetadata` listener re-registered on every transcript mutation**

   The `useEffect` that manages the video duration listener declares `[utterances]` as its dependency, but `utterances` is not referenced anywhere inside the effect body — only `playerRef` is used. Before this PR, `utterances` was derived from a static prop (`memoizedUtterances`), so the array reference never changed at runtime. Now that `utterances` is computed from the live `transcript` context, it produces a new reference on every transcript mutation (inserting/deleting utterances, editing timestamps, etc.). This causes the `loadedmetadata` event listener to be torn down and re-created on every such mutation.

   While the effect defensively calls `updateDuration()` immediately on each run (line 122), there is a brief window between the cleanup removing the old listener and the setup adding the new one where a `loadedmetadata` event could be missed. More importantly, the listener re-registration on hot paths is unnecessary overhead.

   The fix is to remove `utterances` from the dependency array since it is not used in the body:

   

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 112-127

   Comment:
   **`loadedmetadata` listener re-registered on every transcript mutation**

   The `useEffect` that manages the video duration listener declares `[utterances]` as its dependency, but `utterances` is not referenced anywhere inside the effect body — only `playerRef` is used. Before this PR, `utterances` was derived from a static prop (`memoizedUtterances`), so the array reference never changed at runtime. Now that `utterances` is computed from the live `transcript` context, it produces a new reference on every transcript mutation (inserting/deleting utterances, editing timestamps, etc.). This causes the `loadedmetadata` event listener to be torn down and re-created on every such mutation.

   While the effect defensively calls `updateDuration()` immediately on each run (line 122), there is a brief window between the cleanup removing the old listener and the setup adding the new one where a `loadedmetadata` event could be missed. More importantly, the listener re-registration on hot paths is unnecessary overhead.

   The fix is to remove `utterances` from the dependency array since it is not used in the body:

   

   `````
   </details>

2. `src/components/meetings/VideoProvider.tsx`, line 325-351 ([link](https://github.com/schemalabz/opencouncil/blob/a3797c022bc7791ce9d51a06da827aa7bd3cbfe9/src/components/meetings/VideoProvider.tsx#L325-L351)) 

   <a href="#">P2</a> **Context exposes `currentTimeRef.current`, not the `currentTime` state**

   The context `value` object is built with `currentTime: currentTimeRef.current` (line 327) — the ref, not the `useState` value. The `currentTime` state serves purely as a render-trigger: each `setCurrentTime(t)` call (now added to `seekTo`, `seekToWithoutScroll`, and `onSeeked`) causes a re-render, at which point `currentTimeRef.current` (which equals `t`) is snapshotted into the new context value.

   This is correct and the tests confirm it. However, the naming is a source of confusion: there are now two values called `currentTime` — the `useState` variable (render trigger, never read by consumers) and `currentTimeRef.current` (the actual time exposed to context consumers). A short comment clarifying this dual-variable pattern would prevent future developers from accidentally replacing `currentTimeRef.current` with `currentTime` in the context object, which would break the throttled update optimization.

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 325-351

   Comment:
   **Context exposes `currentTimeRef.current`, not the `currentTime` state**

   The context `value` object is built with `currentTime: currentTimeRef.current` (line 327) — the ref, not the `useState` value. The `currentTime` state serves purely as a render-trigger: each `setCurrentTime(t)` call (now added to `seekTo`, `seekToWithoutScroll`, and `onSeeked`) causes a re-render, at which point `currentTimeRef.current` (which equals `t`) is snapshotted into the new context value.

   This is correct and the tests confirm it. However, the naming is a source of confusion: there are now two values called `currentTime` — the `useState` variable (render trigger, never read by consumers) and `currentTimeRef.current` (the actual time exposed to context consumers). A short comment clarifying this dual-variable pattern would prevent future developers from accidentally replacing `currentTimeRef.current` with `currentTime` in the context object, which would break the throttled update optimization.

   `````
   </details>

3. `src/components/meetings/VideoProvider.tsx`, line 142-168 ([link](https://github.com/schemalabz/opencouncil/blob/df8a7ce7186220d761822e2146b4096cc8c94cec/src/components/meetings/VideoProvider.tsx#L142-L168)) 

   <a href="#">P1</a> **URL deep link silently dropped if player unmounted on first effect run**

   `hasAppliedUrlParam.current = true` is set on line 143 before the inner `if (!isNaN(seconds) && playerRef.current)` guard is checked. If `playerRef.current` happens to be `null` when the effect first fires (e.g., the `<video>` element is conditionally rendered), the `?t=` seek is permanently skipped — subsequent runs of this effect are blocked by the already-consumed ref.

   The same issue applies to the `scrollAttempt` closure: it captures `utterances` at call time. If the effect fires with empty `utterances` (before transcript data arrives in an async scenario), `utteranceElement` will always be `undefined` and the retry loop gives up, but `hasAppliedUrlParam.current` is already consumed.

   Move `hasAppliedUrlParam.current = true` inside the inner guard so it is only marked consumed after the seek is actually performed:

   ```ts
   if (timeParam && !hasAppliedUrlParam.current) {
       const seconds = parseInt(timeParam, 10);
       if (!isNaN(seconds) && playerRef.current) {
           hasAppliedUrlParam.current = true; // only mark consumed once seek succeeds
           currentTimeRef.current = seconds;
           setCurrentTime(seconds);
           scrollAttempt(3);
       }
   }
   ```

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 142-168

   Comment:
   **URL deep link silently dropped if player unmounted on first effect run**

   `hasAppliedUrlParam.current = true` is set on line 143 before the inner `if (!isNaN(seconds) && playerRef.current)` guard is checked. If `playerRef.current` happens to be `null` when the effect first fires (e.g., the `<video>` element is conditionally rendered), the `?t=` seek is permanently skipped — subsequent runs of this effect are blocked by the already-consumed ref.

   The same issue applies to the `scrollAttempt` closure: it captures `utterances` at call time. If the effect fires with empty `utterances` (before transcript data arrives in an async scenario), `utteranceElement` will always be `undefined` and the retry loop gives up, but `hasAppliedUrlParam.current` is already consumed.

   Move `hasAppliedUrlParam.current = true` inside the inner guard so it is only marked consumed after the seek is actually performed:

   ```ts
   if (timeParam && !hasAppliedUrlParam.current) {
       const seconds = parseInt(timeParam, 10);
       if (!isNaN(seconds) && playerRef.current) {
           hasAppliedUrlParam.current = true; // only mark consumed once seek succeeds
           currentTimeRef.current = seconds;
           setCurrentTime(seconds);
           scrollAttempt(3);
       }
   }
   ```

   `````
   </details>

4. `src/components/meetings/__tests__/VideoProvider.test.tsx`, line 514-524 ([link](https://github.com/schemalabz/opencouncil/blob/df8a7ce7186220d761822e2146b4096cc8c94cec/src/components/meetings/__tests__/VideoProvider.test.tsx#L514-L524)) 

   <a href="#">P2</a> **Mock side-effect inside a render function**

   `(useCouncilMeetingData as jest.Mock).mockReturnValue(...)` is called directly inside the `ContextProvider` render body. Calling `jest.fn()` helpers as a render-time side effect is an antipattern: React Strict Mode (double-invocation) can call the render function twice, causing the mock to be configured twice per render cycle. It also makes it harder to reason about when the mock is updated relative to a given render.

   Use `beforeEach` / `jest.spyOn` to control mock state, or use a `useRef`/`useMemo` inside the component to initialise the mock once:

   ```tsx
   // Outside the test, or inside beforeEach:
   (useCouncilMeetingData as jest.Mock).mockReturnValue({
       transcript: transcriptData,
       ...
   });

   // And re-configure it before rerender instead of inside ContextProvider
   ```

   `````markdown
   Path: src/components/meetings/__tests__/VideoProvider.test.tsx
   Line: 514-524

   Comment:
   **Mock side-effect inside a render function**

   `(useCouncilMeetingData as jest.Mock).mockReturnValue(...)` is called directly inside the `ContextProvider` render body. Calling `jest.fn()` helpers as a render-time side effect is an antipattern: React Strict Mode (double-invocation) can call the render function twice, causing the mock to be configured twice per render cycle. It also makes it harder to reason about when the mock is updated relative to a given render.

   Use `beforeEach` / `jest.spyOn` to control mock state, or use a `useRef`/`useMemo` inside the component to initialise the mock once:

   ```tsx
   // Outside the test, or inside beforeEach:
   (useCouncilMeetingData as jest.Mock).mockReturnValue({
       transcript: transcriptData,
       ...
   });

   // And re-configure it before rerender instead of inside ContextProvider
   ```

   `````
   </details>

5. `src/components/meetings/VideoProvider.tsx`, line 329 ([link](https://github.com/schemalabz/opencouncil/blob/29acd0f1fd89b5ebbc9c9f506768d1818e350e46/src/components/meetings/VideoProvider.tsx#L329)) 

   <a href="#">P1</a> **`currentTime` in context value reads ref, not state**

   The context object exposes `currentTimeRef.current` directly instead of the `currentTime` state variable. The `currentTime` state (line 91) is purely used as a render trigger — calling `setCurrentTime` causes a re-render, at which point this line re-evaluates `currentTimeRef.current` and consumers see the updated value.

   This creates a fragile implicit invariant: every `setCurrentTime(x)` call **must** be preceded by `currentTimeRef.current = x`, otherwise consumers re-render but see the old ref value. All current call sites in this PR do maintain that order, but future contributors may call `setCurrentTime` to trigger a re-render without realising they also need to update the ref first.

   Consider using the state variable directly to make the contract explicit:

   
   (and ensure `throttledSetCurrentTime` continues to set both the ref and state, as it does today). This removes the hidden dependency between the ref write and the state-triggered re-render.

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 329

   Comment:
   **`currentTime` in context value reads ref, not state**

   The context object exposes `currentTimeRef.current` directly instead of the `currentTime` state variable. The `currentTime` state (line 91) is purely used as a render trigger — calling `setCurrentTime` causes a re-render, at which point this line re-evaluates `currentTimeRef.current` and consumers see the updated value.

   This creates a fragile implicit invariant: every `setCurrentTime(x)` call **must** be preceded by `currentTimeRef.current = x`, otherwise consumers re-render but see the old ref value. All current call sites in this PR do maintain that order, but future contributors may call `setCurrentTime` to trigger a re-render without realising they also need to update the ref first.

   Consider using the state variable directly to make the contract explicit:

   
   (and ensure `throttledSetCurrentTime` continues to set both the ref and state, as it does today). This removes the hidden dependency between the ref write and the state-triggered re-render.

   `````
   </details>

6. `src/components/meetings/__tests__/VideoProvider.test.tsx`, line 521-529 ([link](https://github.com/schemalabz/opencouncil/blob/29acd0f1fd89b5ebbc9c9f506768d1818e350e46/src/components/meetings/__tests__/VideoProvider.test.tsx#L521-L529)) 

   <a href="#">P2</a> **Side effect inside render function**

   `mockReturnValue` is called directly inside the render body of `ContextProvider`. Calling a Jest mock mutator during render is a side effect — React (especially in StrictMode) may invoke the render function multiple times, causing `mockReturnValue` to be called more than once per conceptual render. While the mock is idempotent here, this pattern can introduce subtle ordering issues if the test is extended.

   Consider moving the mock setup into the `beforeEach` block or updating `transcriptData` via `(useCouncilMeetingData as jest.Mock).mockReturnValue(...)` just before the `rerender` call, outside any render function:

   ```ts
   // Before initial render
   (useCouncilMeetingData as jest.Mock).mockReturnValue({ transcript: transcriptData, ... });
   const { rerender } = render(<VideoProvider meeting={mockMeeting}>...</VideoProvider>);

   // Before rerender
   transcriptData = [...];
   (useCouncilMeetingData as jest.Mock).mockReturnValue({ transcript: transcriptData, ... });
   rerender(<VideoProvider meeting={mockMeeting}>...</VideoProvider>);
   ```

   `````markdown
   Path: src/components/meetings/__tests__/VideoProvider.test.tsx
   Line: 521-529

   Comment:
   **Side effect inside render function**

   `mockReturnValue` is called directly inside the render body of `ContextProvider`. Calling a Jest mock mutator during render is a side effect — React (especially in StrictMode) may invoke the render function multiple times, causing `mockReturnValue` to be called more than once per conceptual render. While the mock is idempotent here, this pattern can introduce subtle ordering issues if the test is extended.

   Consider moving the mock setup into the `beforeEach` block or updating `transcriptData` via `(useCouncilMeetingData as jest.Mock).mockReturnValue(...)` just before the `rerender` call, outside any render function:

   ```ts
   // Before initial render
   (useCouncilMeetingData as jest.Mock).mockReturnValue({ transcript: transcriptData, ... });
   const { rerender } = render(<VideoProvider meeting={mockMeeting}>...</VideoProvider>);

   // Before rerender
   transcriptData = [...];
   (useCouncilMeetingData as jest.Mock).mockReturnValue({ transcript: transcriptData, ... });
   rerender(<VideoProvider meeting={mockMeeting}>...</VideoProvider>);
   ```

   `````
   </details>

7. `src/components/meetings/VideoProvider.tsx`, line 130-172 ([link](https://github.com/schemalabz/opencouncil/blob/37b704153d9b0cc831c85844f60ab1744ed41b77/src/components/meetings/VideoProvider.tsx#L130-L172)) 

   <a href="#">P1</a> **URL-param seek can fire again on any transcript mutation**

   The effect depends on `[utterances]`, which changes on every transcript mutation (edit, add, delete). The `hasAppliedUrlParam.current` flag is only set when `playerRef.current` is non-null. If the video element is not yet mounted on the first effect run (conditional render / slow load), `hasAppliedUrlParam.current` remains `false`. The very next transcript mutation — e.g., the user edits an utterance text and `updateUtterance` triggers `setTranscript` → `utterances` changes — re-runs this effect. If `playerRef.current` is available by then, the effect applies the `?t=` deep-link seek, jumping the video to the URL position mid-session and interrupting the user's editing flow.

   The `hasSetInitialTime` concern is separate and can stay in an utterances-dependent effect, but the `?t=` seek should be applied exactly once at mount time. Splitting them avoids the retry-on-mutation footgun:

   ```ts
   // Run once on mount; retry if player not yet attached
   useEffect(() => {
       const timeParam = new URLSearchParams(window.location.search).get('t');
       if (!timeParam || hasAppliedUrlParam.current) return;
       const seconds = parseInt(timeParam, 10);
       if (isNaN(seconds) || !playerRef.current) return;

       hasAppliedUrlParam.current = true;
       currentTimeRef.current = seconds;
       setCurrentTime(seconds);
       // … scroll attempts …
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [playerRef.current]); // or an isMounted flag
   ```

   Keeping the initial-time setter in the `[utterances]` effect is fine; only the URL-param block is risky as-is.

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 130-172

   Comment:
   **URL-param seek can fire again on any transcript mutation**

   The effect depends on `[utterances]`, which changes on every transcript mutation (edit, add, delete). The `hasAppliedUrlParam.current` flag is only set when `playerRef.current` is non-null. If the video element is not yet mounted on the first effect run (conditional render / slow load), `hasAppliedUrlParam.current` remains `false`. The very next transcript mutation — e.g., the user edits an utterance text and `updateUtterance` triggers `setTranscript` → `utterances` changes — re-runs this effect. If `playerRef.current` is available by then, the effect applies the `?t=` deep-link seek, jumping the video to the URL position mid-session and interrupting the user's editing flow.

   The `hasSetInitialTime` concern is separate and can stay in an utterances-dependent effect, but the `?t=` seek should be applied exactly once at mount time. Splitting them avoids the retry-on-mutation footgun:

   ```ts
   // Run once on mount; retry if player not yet attached
   useEffect(() => {
       const timeParam = new URLSearchParams(window.location.search).get('t');
       if (!timeParam || hasAppliedUrlParam.current) return;
       const seconds = parseInt(timeParam, 10);
       if (isNaN(seconds) || !playerRef.current) return;

       hasAppliedUrlParam.current = true;
       currentTimeRef.current = seconds;
       setCurrentTime(seconds);
       // … scroll attempts …
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [playerRef.current]); // or an isMounted flag
   ```

   Keeping the initial-time setter in the `[utterances]` effect is fine; only the URL-param block is risky as-is.

   `````
   </details>

8. `src/components/meetings/VideoProvider.tsx`, line 130-172 ([link](https://github.com/schemalabz/opencouncil/blob/cbef076928349a9d7b65eb57cac0ebed76e1b65f/src/components/meetings/VideoProvider.tsx#L130-L172)) 

   <a href="#">P2</a> **URL-param effect re-runs on every transcript mutation**

   This effect's dependency is `[utterances]`, which is now computed from mutable context state. That means the effect fires on every `addUtteranceToSegment`, `deleteUtterance`, `updateUtterance`, etc. call during a live editing session.

   Both one-shot guards (`hasSetInitialTime.current`, `hasAppliedUrlParam.current`) flip on or before the first transcript mutation, so subsequent runs are functional no-ops. However, each re-run still evaluates `new URLSearchParams(window.location.search)` and checks the guards, and — more importantly — if `playerRef.current` is null when the very first non-empty `utterances` value arrives (e.g., a conditional video render), `hasAppliedUrlParam.current` is never set and the `?t=` deep link will be silently dropped on any subsequent `utterances` change that arrives *after* the player has mounted but without triggering the effect again.

   Consider splitting the two concerns into separate effects:

   ```ts
   // Runs once when the first utterances load
   useEffect(() => {
       if (!hasSetInitialTime.current && utterances.length > 0) {
           hasSetInitialTime.current = true;
           currentTimeRef.current = utterances[0].startTimestamp;
           setCurrentTime(utterances[0].startTimestamp);
       }
   }, [utterances]);

   // Runs once — or retries when playerRef becomes ready — for ?t= deep link
   useEffect(() => {
       const timeParam = new URLSearchParams(window.location.search).get('t');
       if (!timeParam || hasAppliedUrlParam.current) return;
       const seconds = parseInt(timeParam, 10);
       if (isNaN(seconds) || !playerRef.current) return;
       hasAppliedUrlParam.current = true;
       currentTimeRef.current = seconds;
       setCurrentTime(seconds);
       scrollAttempt(3);
       // isPlayerReady would be a boolean state set in the loadedmetadata handler
   }, [isPlayerReady, utterances]);
   ```

   This makes each concern's trigger explicit and avoids the silent-skip risk.

   `````markdown
   Path: src/components/meetings/VideoProvider.tsx
   Line: 130-172

   Comment:
   **URL-param effect re-runs on every transcript mutation**

   This effect's dependency is `[utterances]`, which is now computed from mutable context state. That means the effect fires on every `addUtteranceToSegment`, `deleteUtterance`, `updateUtterance`, etc. call during a live editing session.

   Both one-shot guards (`hasSetInitialTime.current`, `hasAppliedUrlParam.current`) flip on or before the first transcript mutation, so subsequent runs are functional no-ops. However, each re-run still evaluates `new URLSearchParams(window.location.search)` and checks the guards, and — more importantly — if `playerRef.current` is null when the very first non-empty `utterances` value arrives (e.g., a conditional video render), `hasAppliedUrlParam.current` is never set and the `?t=` deep link will be silently dropped on any subsequent `utterances` change that arrives *after* the player has mounted but without triggering the effect again.

   Consider splitting the two concerns into separate effects:

   ```ts
   // Runs once when the first utterances load
   useEffect(() => {
       if (!hasSetInitialTime.current && utterances.length > 0) {
           hasSetInitialTime.current = true;
           currentTimeRef.current = utterances[0].startTimestamp;
           setCurrentTime(utterances[0].startTimestamp);
       }
   }, [utterances]);

   // Runs once — or retries when playerRef becomes ready — for ?t= deep link
   useEffect(() => {
       const timeParam = new URLSearchParams(window.location.search).get('t');
       if (!timeParam || hasAppliedUrlParam.current) return;
       const seconds = parseInt(timeParam, 10);
       if (isNaN(seconds) || !playerRef.current) return;
       hasAppliedUrlParam.current = true;
       currentTimeRef.current = seconds;
       setCurrentTime(seconds);
       scrollAttempt(3);
       // isPlayerReady would be a boolean state set in the loadedmetadata handler
   }, [isPlayerReady, utterances]);
   ```

   This makes each concern's trigger explicit and avoids the silent-skip risk.

   `````
   </details>

9. `src/components/meetings/transcript/Transcript.tsx`, line 54-63 ([link](https://github.com/schemalabz/opencouncil/blob/f22abd7d93f89518ca6579e7c18c8a23661ff8e8/src/components/meetings/transcript/Transcript.tsx#L54-L63)) 

   <a href="#">P2</a> **`calculateTimeInterval` implicitly relies on `displayedSegments` being in chronological order**

   The function picks `validSegments[0]` as the start and `validSegments[validSegments.length - 1]` as the end:

   ```ts
   const firstVisible = validSegments[0];
   const lastVisible = validSegments[validSegments.length - 1];
   return [firstVisible.startTimestamp, lastVisible.endTimestamp];
   ```

   This is correct as long as `displayedSegments` is sorted by `startTimestamp` — which is guaranteed today because `orderedTranscript` sorts on it. However, in non-editable mode `displayedSegments` is derived via `joinTranscriptSegments`, whose ordering is not immediately obvious from this call site. If `joinTranscriptSegments` ever reorders segments, or if a future caller passes an unsorted set, the interval will silently point to the wrong range.

   Consider sorting `validSegments` by `startTimestamp` explicitly before taking the head/tail, or at least adding an assertion comment that this function assumes sorted input.

   `````markdown
   Path: src/components/meetings/transcript/Transcript.tsx
   Line: 54-63

   Comment:
   **`calculateTimeInterval` implicitly relies on `displayedSegments` being in chronological order**

   The function picks `validSegments[0]` as the start and `validSegments[validSegments.length - 1]` as the end:

   ```ts
   const firstVisible = validSegments[0];
   const lastVisible = validSegments[validSegments.length - 1];
   return [firstVisible.startTimestamp, lastVisible.endTimestamp];
   ```

   This is correct as long as `displayedSegments` is sorted by `startTimestamp` — which is guaranteed today because `orderedTranscript` sorts on it. However, in non-editable mode `displayedSegments` is derived via `joinTranscriptSegments`, whose ordering is not immediately obvious from this call site. If `joinTranscriptSegments` ever reorders segments, or if a future caller passes an unsorted set, the interval will silently point to the wrong range.

   Consider sorting `validSegments` by `startTimestamp` explicitly before taking the head/tail, or at least adding an assertion comment that this function assumes sorted input.

   `````
   </details>
</details>

`````markdown
Path: src/components/meetings/VideoProvider.tsx
Line: 217

Comment:
**Stale `utterances` closure in deep-link effect**

The effect body uses `utterances` directly (lines 183, 192–194) but only `utterances.length` is in the dependency array. If utterance timestamps change without changing the count, the closure holds the old `utterances` array. In practice the deep-link flag (`hasAppliedUrlParam.current`) prevents re-runs after the first successful seek, so the stale value is never acted on — but the intent is clearer and the lint warning is avoided if you add `utterances` itself to the deps:

```suggestion
    }, [timeParam, utterances]);
```

---

Path: src/components/meetings/transcript/Transcript.tsx
Line: 211-213

Comment:
**All segments initially render as skeletons**

`visibleSegments` starts empty, so on the very first render every segment — including those already in the viewport — gets `shouldRender = false` and renders the mock skeleton. The `IntersectionObserver` fires almost immediately after mount, but there is a sub-frame flash of skeleton content for visible items before the observer populates `visibleSegments`.

One option is to seed `visibleSegments` so that segments within a reasonable index range (e.g. the first N) start as `true`, letting the observer correct the state rather than starting from all-false.

`````

</details>

<sub>Reviews (8): Last reviewed commit: ["fixup! refactor(meetings): stabilize tra..."](https://github.com/schemalabz/opencouncil/commit/8cd91787dd1a497f7a040e1a887bd575f3cb2826) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=25810929)</sub>


angelospk:
Tried to reproduce it while recording, but couldn't. I remember when I first tried dealing with the issue, it was pretty easy to reproduce the bug with an empty segment and a new speaker below it, like Eliza's loom video. Will try again.
Is it possible that it's fixed after incoming changes from another pr?

kouloumos:
> Is it possible that it's fixed after incoming changes from another pr?

It could be. I am gonna go ahead and close this, and we can revisit when we can reproduce.

PR #307, opened 2026-04-08
User profile page redesign

### LINKED ISSUE #180: Re-design user profile dashboard
Currently, the user profile dashboard shows "you do not have admin rights" for regular users.

### LINKED ISSUE #304: Bug: Phone number field saves "+30" placeholder when left empty in notifications form
## Description
In the `/notifications` form, where users complete their info to receive notifications about subjects they follow, the phone number field is optional and pre-filled with `+30` (Greece's country code) as a default/placeholder value. When a user submits the form without entering a phone number, `+30` is saved to the database instead of `NULL`.

## Impact
This causes multiple downstream issues:

1. **Notification system**: Attempts to send WhatsApp/SMS notifications to `+30`, resulting in failures.
2. **Analytics**: Services that count non-null phone numbers treat these records as having a valid phone number, inflating "completed profile" or registered user metrics.

## Steps to Reproduce
1. Go to the `[municipality]/notifications` onboarding form
2. Continue at the third step and leave the phone number field untouched (shows `+30` by default)
3. Complete the rest and submit the form
4. Check the database — the user's phone number is saved as `+30`

## Expected Behavior
If the user does not enter a phone number beyond the country code prefix, the field should be saved as `NULL` in the database.

## Suggested Fix
Before saving, strip the country code prefix and treat the value as empty if nothing follows it. Save `NULL` instead of the bare country code.

VasiaKoum:
<h3>Greptile Summary</h3>

This PR redesigns the user profile page into a tabbed layout (personal info, communication, notifications, account), adds per-city `notifyByEmail`/`notifyByPhone` channel toggles, and introduces a token-based unsubscribe flow. Two schema migrations rename `allowContact` → `allowProductUpdates`, add `allowPetitionUpdates`, and add the new channel columns to `NotificationPreference`.

- **P1 — delivery ignores channel preferences:** `createNotificationsForMeeting` always creates an email delivery and creates a phone delivery whenever `user.phone` is set, never consulting `userPref.notifyByEmail` or `userPref.notifyByPhone`. Users who toggle off a channel will still receive deliveries through it.

Mostly safe to merge but the channel-preference toggles silently have no effect on actual delivery until the P1 is fixed.

One clear P1: the notifyByEmail/notifyByPhone flags are stored and surfaced in the UI but createNotificationsForMeeting never reads them, so disabling a channel has no effect on delivery. All other findings are P2 or cosmetic. Prior P0/P1 concerns from earlier review rounds (auth bypass, 401 vs 500, optimistic-revert on HTTP error) appear to have been addressed in this revision.

src/lib/db/notifications.ts — createNotificationsForMeeting delivery section (lines 881–904)

<details open><summary><h3>Vulnerabilities</h3></summary>

No security concerns identified. The unsubscribe endpoint is protected by an HMAC-SHA256 signed token with a 30-day TTL, and the public route does not expose user data beyond what the token already encodes. The profile DELETE route now has an explicit auth guard and revokes sessions before deleting the user record. The PATCH route for notification preferences verifies ownership before updating.
</details>

| Filename | Overview |
|----------|----------|
| src/lib/db/notifications.ts | Added updateNotificationPreferenceChannels, disableAll/disableByCity helpers, and getUserNotifications; createNotificationsForMeeting does not honour the new notifyByEmail/notifyByPhone flags — deliveries are always created regardless of preferences. |
| src/components/profile/NotificationPreferencesSection.tsx | New component showing per-city notification preferences with channel toggles (email/SMS), history, and add-city flow; optimistic updates now correctly revert on HTTP errors via explicit throw. |
| src/components/profile/UserInfoForm.tsx | Redesigned profile form with URL-driven tabs; two section headings are hardcoded Greek strings instead of translation keys. |
| src/app/api/unsubscribe/route.ts | New public endpoint processing HMAC-signed unsubscribe tokens; action=all disables product/petition flags but leaves per-city preferences enabled. |
| src/app/api/user/notification-preferences/[id]/route.ts | Added PATCH handler for toggling notifyByEmail/notifyByPhone; both handlers properly authenticate and delegate to db helpers. |
| src/lib/utils/phone.ts | New lightweight phone validation utility with GR full-spec and calling-code-only data for all other countries; validation bypassed for non-GR numbers by design. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[User toggles notifyByEmail / notifyByPhone] --> B[PATCH /api/user/notification-preferences/:id]
    B --> C[updateNotificationPreferenceChannels]
    C --> D[(NotificationPreference DB notifyByEmail / notifyByPhone updated)]
    E[Notification job runs] --> F[createNotificationsForMeeting]
    F --> G{user.phone set?}
    G -- yes --> H[Create phone delivery]
    G -- no --> I[Skip phone delivery]
    F --> J[Create email delivery ALWAYS]
    D -. ignored .-> F
    style D fill:#d4edda,stroke:#28a745
    style J fill:#f8d7da,stroke:#dc3545
    style H fill:#fff3cd,stroke:#ffc107
```

<details><summary><h3>Comments Outside Diff (6)</h3></summary>

1. `src/components/profile/AdminSection.tsx`, line 4-7 ([link](https://github.com/schemalabz/opencouncil/blob/43378779bda040ab7c2a17f9bbf6a70ddc83ce7e/src/components/profile/AdminSection.tsx#L4-L7)) 

   <a href="#">P2</a> **Unused imports and fragmented lucide-react imports**

   `ContactBadge` (line 4) and `Users` (line 6) are imported but never referenced in this component. Additionally, `Flag`, `User`, `Users`, and `Building` are spread across three separate `lucide-react` imports — these should be consolidated.

   

   `````markdown
   Path: src/components/profile/AdminSection.tsx
   Line: 4-7

   Comment:
   **Unused imports and fragmented lucide-react imports**

   `ContactBadge` (line 4) and `Users` (line 6) are imported but never referenced in this component. Additionally, `Flag`, `User`, `Users`, and `Building` are spread across three separate `lucide-react` imports — these should be consolidated.

   

   `````
   </details>

2. `src/lib/db/users.ts`, line 70-74 ([link](https://github.com/schemalabz/opencouncil/blob/80b7a9c5ba63302779b5683f6acdbaff6ce5757d/src/lib/db/users.ts#L70-L74)) 

   <a href="#">P1</a> **Email normalization silently removed**

   The PR removes `normalizeEmail` / `normalizeAdminUserData`, so admin-created and admin-updated users no longer have their email lowercased or trimmed before hitting the database. If an admin enters `User@Example.com`, it is stored as `User@Example.com`. When that person later signs in via a magic link (Auth.js normalises the lookup email to lowercase), the session-lookup may miss the stored record, effectively locking them out or creating a duplicate user.

   

   Apply the same one-liner normalisation in `updateUser` where `userData` is destructured from `data`.

   `````markdown
   Path: src/lib/db/users.ts
   Line: 70-74

   Comment:
   **Email normalization silently removed**

   The PR removes `normalizeEmail` / `normalizeAdminUserData`, so admin-created and admin-updated users no longer have their email lowercased or trimmed before hitting the database. If an admin enters `User@Example.com`, it is stored as `User@Example.com`. When that person later signs in via a magic link (Auth.js normalises the lookup email to lowercase), the session-lookup may miss the stored record, effectively locking them out or creating a duplicate user.

   

   Apply the same one-liner normalisation in `updateUser` where `userData` is destructured from `data`.

   `````
   </details>

3. `src/app/api/dev/seed-users/route.ts`, line 114 ([link](https://github.com/schemalabz/opencouncil/blob/01567735cbc602ed3f78b04821e8682646f1d3ab/src/app/api/dev/seed-users/route.ts#L114)) 

   <a href="#">P1</a> **Stale `allowContact` field breaks seeded user data**

   `allowContact` was renamed to `allowProductUpdates` by the migration in this PR, but this line still uses the old name. Because `userData` is typed `any`, TypeScript won't flag it — Prisma will silently discard the unknown field at runtime, so every seeded user lands with `allowProductUpdates: false` regardless of persona.

   

   The same rename is needed in `src/components/admin/users/seed-users-dialog.tsx` lines 33, 52, 71, and 90 (the `dataPattern.allowContact` display fields).

   `````markdown
   Path: src/app/api/dev/seed-users/route.ts
   Line: 114

   Comment:
   **Stale `allowContact` field breaks seeded user data**

   `allowContact` was renamed to `allowProductUpdates` by the migration in this PR, but this line still uses the old name. Because `userData` is typed `any`, TypeScript won't flag it — Prisma will silently discard the unknown field at runtime, so every seeded user lands with `allowProductUpdates: false` regardless of persona.

   

   The same rename is needed in `src/components/admin/users/seed-users-dialog.tsx` lines 33, 52, 71, and 90 (the `dataPattern.allowContact` display fields).

   `````
   </details>

4. `src/app/api/dev/seed-users/route.ts`, line 92-162 ([link](https://github.com/schemalabz/opencouncil/blob/25f36fe772b5279713e7e385ae50b66b9e5d4c42/src/app/api/dev/seed-users/route.ts#L92-L162)) 

   <a href="#">P1</a> **`lurker` and `newcomer` personas create zero users**

   For `persona === 'lurker'` and `persona === 'newcomer'`, neither `saveNotificationPreferences` nor `savePetition` is called — `userData` is built but never persisted. The loop still increments `totalCreatedUsers`, so the response reports e.g. `"Successfully created 10 lurker users"` when the database is unchanged.

   `saveNotificationPreferences`/`savePetition` are the only paths that call `prisma.user.create`; there is no standalone user-creation step. A direct `prisma.user.create` (or a shared helper) needs to be called for these two personas.

   `````markdown
   Path: src/app/api/dev/seed-users/route.ts
   Line: 92-162

   Comment:
   **`lurker` and `newcomer` personas create zero users**

   For `persona === 'lurker'` and `persona === 'newcomer'`, neither `saveNotificationPreferences` nor `savePetition` is called — `userData` is built but never persisted. The loop still increments `totalCreatedUsers`, so the response reports e.g. `"Successfully created 10 lurker users"` when the database is unchanged.

   `saveNotificationPreferences`/`savePetition` are the only paths that call `prisma.user.create`; there is no standalone user-creation step. A direct `prisma.user.create` (or a shared helper) needs to be called for these two personas.

   `````
   </details>

5. `src/components/profile/UserInfoForm.tsx`, line 1791-1802 ([link](https://github.com/schemalabz/opencouncil/blob/77b2e3a5663c8fe41f71e1d9274a6373edfbaecd/src/components/profile/UserInfoForm.tsx#L1791-L1802)) 

   <a href="#">P1</a> <a href="#">security</a> **Session not invalidated after account deletion**

   `router.push("/")` redirects the user but leaves the session cookie intact. `getCurrentUser()` looks up users by `session.user.email`, so if the same email address is later used to create a new account (via the magic-link flow), the original browser's session will automatically authenticate to that new account without any login step. Use `signOut` from `next-auth/react` to explicitly revoke the session:

   

   `signOut` needs to be imported from `"next-auth/react"`.

   `````markdown
   Path: src/components/profile/UserInfoForm.tsx
   Line: 1791-1802

   Comment:
   **Session not invalidated after account deletion**

   `router.push("/")` redirects the user but leaves the session cookie intact. `getCurrentUser()` looks up users by `session.user.email`, so if the same email address is later used to create a new account (via the magic-link flow), the original browser's session will automatically authenticate to that new account without any login step. Use `signOut` from `next-auth/react` to explicitly revoke the session:

   

   `signOut` needs to be imported from `"next-auth/react"`.

   `````
   </details>

6. `src/lib/db/notifications.ts`, line 881-904 ([link](https://github.com/schemalabz/opencouncil/blob/db0e343ed28e657a3f745beeb69ea6c19f9720df/src/lib/db/notifications.ts#L881-L904)) 

   <a href="#">P1</a> **`notifyByEmail`/`notifyByPhone` preferences ignored during delivery**

   The new per-city channel toggles are wired up in the UI and stored in the database, but `createNotificationsForMeeting` never reads them. An email delivery is always created (the comment even says "always"), and a phone delivery is created whenever `user.phone` is set — regardless of what the user toggled. A user who disables email or SMS for a city will still receive those deliveries.

   `userPref` (already in scope at this point) carries the `notifyByEmail` and `notifyByPhone` fields from the migration. The fix is to gate each delivery on the corresponding flag:

   ```ts
   // Create email delivery only if the user opted in
   if (userPref.notifyByEmail) {
       await prisma.notificationDelivery.create({ data: { ...medium: 'email', ... } });
   }

   // Create message delivery if user has a phone and opted in
   if (user.phone && userPref.notifyByPhone) {
       await prisma.notificationDelivery.create({ data: { ...medium: 'message', ... } });
   }
   ```

   `````markdown
   Path: src/lib/db/notifications.ts
   Line: 881-904

   Comment:
   **`notifyByEmail`/`notifyByPhone` preferences ignored during delivery**

   The new per-city channel toggles are wired up in the UI and stored in the database, but `createNotificationsForMeeting` never reads them. An email delivery is always created (the comment even says "always"), and a phone delivery is created whenever `user.phone` is set — regardless of what the user toggled. A user who disables email or SMS for a city will still receive those deliveries.

   `userPref` (already in scope at this point) carries the `notifyByEmail` and `notifyByPhone` fields from the migration. The fix is to gate each delivery on the corresponding flag:

   ```ts
   // Create email delivery only if the user opted in
   if (userPref.notifyByEmail) {
       await prisma.notificationDelivery.create({ data: { ...medium: 'email', ... } });
   }

   // Create message delivery if user has a phone and opted in
   if (user.phone && userPref.notifyByPhone) {
       await prisma.notificationDelivery.create({ data: { ...medium: 'message', ... } });
   }
   ```

   `````
   </details>

</details>

`````markdown
Path: src/lib/db/notifications.ts
Line: 881-904

Comment:
**`notifyByEmail`/`notifyByPhone` preferences ignored during delivery**

The new per-city channel toggles are wired up in the UI and stored in the database, but `createNotificationsForMeeting` never reads them. An email delivery is always created (the comment even says "always"), and a phone delivery is created whenever `user.phone` is set — regardless of what the user toggled. A user who disables email or SMS for a city will still receive those deliveries.

`userPref` (already in scope at this point) carries the `notifyByEmail` and `notifyByPhone` fields from the migration. The fix is to gate each delivery on the corresponding flag:

```ts
// Create email delivery only if the user opted in
if (userPref.notifyByEmail) {
    await prisma.notificationDelivery.create({ data: { ...medium: 'email', ... } });
}

// Create message delivery if user has a phone and opted in
if (user.phone && userPref.notifyByPhone) {
    await prisma.notificationDelivery.create({ data: { ...medium: 'message', ... } });
}
```

---

Path: src/components/profile/UserInfoForm.tsx
Line: 125

Comment:
**Hardcoded Greek string bypasses i18n**

The heading "Τα στοιχεία σας" is hardcoded in the source rather than going through `t()`. The same applies to "Ρυθμίσεις επικοινωνίας" on line 223. The rest of the component already uses translation keys for all visible text — these two headings should follow the same pattern.

```suggestion
                            <h3 className="font-semibold">{t("personalInfoHeading")}</h3>
```

---

Path: src/app/api/unsubscribe/route.ts
Line: 19-25

Comment:
**`action === 'all'` branch leaves per-city preferences enabled**

When `action` is `'all'`, the code disables `allowProductUpdates`/`allowPetitionUpdates` but skips `disableNotificationPreferenceByCityId` — per-city `notifyByEmail`/`notifyByPhone` flags remain enabled. Users who expect "unsubscribe from everything" to also stop city notifications will continue to receive them.

Consider either calling `disableAllNotificationPreferences(data.userId)` in the `'all'` branch as well, or clarifying in the UI copy that the city-level toggles are unaffected.

`````

</details>

<sub>Reviews (14): Last reviewed commit: ["fix: resolve comments, add missing local..."](https://github.com/schemalabz/opencouncil/commit/db0e343ed28e657a3f745beeb69ea6c19f9720df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27696085)</sub>

### CODE REVIEW

File: src/components/profile/UserInfoForm.tsx [hunk 1]
```diff
@@ -23,30 +29,30 @@ export function UserInfoForm({ user, isOnboarded }: UserInfoFormProps) {
...
+        setIsSubmitting(true);
         try {
             const response = await fetch("/api/profile", {
                 method: "POST",
                 headers: { "Content-Type": "application/json" },
-                body: JSON.stringify({
-                    ...formData,
-                    onboarded: true
-                })
+                body: JSON.stringify({ ...payload, onboarded: true })
```
kouloumos: `saveToApi` always sends `onboarded: true` regardless of which tab calls it. This means saving communication preferences (which only sends `allowProductUpdates` + `allowPetitionUpdates`) marks the user as onboarded even if they never filled in their name on the personal tab. The personal tab has `required` on the name input so it won't submit empty, but the communication tab has no such guard.

Either move `onboarded: true` into `handlePersonalSubmit` only, or add a check that `formData.name` is non-empty before including it.

src/components/profile/UserInfoForm.tsx
```diff
@@ -55,89 +61,282 @@ export function UserInfoForm({ user, isOnboarded }: UserInfoFormProps) {
...
+        try {
+            const response = await fetch("/api/profile", { method: "DELETE" });
+            if (!response.ok) throw new Error("Failed to delete account");
+            router.push("/");
+        } catch (error) {
+            console.error("Failed to delete account:", error);
+        } finally {
+            setIsDeleting(false);
+        }
+    }
```
kouloumos, line 97: Two issues:

1. After deleting the account, `router.push("/")` redirects but never calls `signOut()` from `next-auth/react`. The session cookie survives, so subsequent requests will try to load a deleted user — errors or weird behavior until the cookie expires. Import `signOut` from `next-auth/react` and call it before or instead of the redirect.

2. No user-facing error feedback if the delete fails — it just logs to console. The user clicks "delete", it fails silently, and `setIsDeleting(false)` re-enables the button with no indication anything went wrong.

src/app/api/profile/route.ts
```diff
@@ -34,4 +35,18 @@ export async function POST(request: Request) {
...
     }
+}
+
+export async function DELETE() {
+    try {
+        const user = await getCurrentUser();
+        if (!user) {
+            return new NextResponse("Unauthorized", { status: 401 });
+        }
+        await prisma.user.delete({ where: { id: user.id } });
```
kouloumos: This bypasses the data access layer — there's already a `deleteUser(id)` in `src/lib/db/users.ts` that handles auth checks and error mapping. Use that instead of hitting Prisma directly. Also, account self-deletion should probably clean up the session server-side (delete the session record) rather than relying on the client to sign out.

src/app/api/user/notification-preferences/[id]/route.ts
```diff
@@ -1,6 +1,30 @@
...
+
+    try {
+        const updated = await updateNotificationPreferenceChannels(params.id, currentUser.id, {
+            ...(notifyByEmail !== undefined && { notifyByEmail }),
+            ...(notifyByPhone !== undefined && { notifyByPhone }),
+        });
+        return NextResponse.json(updated);
+    } catch (error) {
+        return NextResponse.json({ error: 'Not found' }, { status: 404 });
+    }
```
Comment (kouloumos) [inline src/app/api/user/notification-preferences/[id]/route.ts:27]: Both catch blocks return 404 for every error — including DB connectivity failures, constraint violations, etc. A Prisma connection error showing up as "Not found" makes debugging painful and makes the client-side optimistic revert misleading.

There's `handleApiError` in `src/lib/api/errors.ts` that does exactly this — maps `ApiError` subclasses to proper HTTP status codes and falls back to 500 for unknown errors. The `updateNotificationPreferenceChannels` function already throws a plain `Error('Notification preference not found')` — you could throw a `NotFoundError` instead and let `handleApiError` handle the rest:

```typescript
} catch (error) {
    return handleApiError(error, 'Failed to update notification preference');
}
```

src/components/profile/NotificationPreferencesSection.tsx
kouloumos, line 1: This component has ~35 hardcoded Greek strings ("Προτιμήσεις ειδοποιήσεων", "Ενημέρωση με email", "Ιστορικό ειδοποιήσεων", "Δήμος", "Θέματα", "Τοποθεσίες", status labels, etc.) while `UserInfoForm.tsx` next to it properly uses `useTranslations("Profile")`. We already have the English keys in `messages/en.json` under `Profile.*` — add `useTranslations` and move all strings to the message files.

src/components/profile/NotificationPreferencesSection.tsx
```diff
@@ -139,232 +138,478 @@ export function NotificationPreferencesSection() {
...
+                                                        </p>
+                                                        <div className="flex flex-col gap-1 text-xs text-muted-foreground">
+                                                            {emailDelivery && (
+                                                                <div className="flex items-center gap-1">
+                                                                    <Mail className="h-3 w-3" />
+                                                                    {emailDelivery.status === 'sent' && <Badge className="bg-green-100 text-green-800 text-xs hover:bg-green-200">Εστάλη</Badge>}
+                                                                    {emailDelivery.status === 'failed' && <Badge className="bg-red-100 text-red-800 text-xs hover:bg-red-200">Αποτυχία</Badge>}
+                                                                    {emailDelivery.status === 'pending' && <Badge className="bg-yellow-100 text-yellow-800 text-xs hover:bg-yellow-200">Εκκρεμεί</Badge>}
```
kouloumos: The email/phone delivery status badge rendering is copy-pasted identically between the mobile and desktop notification history views. Each status (sent/failed/pending) has the same Badge with the same className, repeated for each medium, in both layouts. That's 4 near-identical blocks.

Extract a small helper:
```tsx
function DeliveryStatusBadge({ status }: { status: string }) {
    const styles = {
        sent: 'bg-green-100 text-green-800',
        failed: 'bg-red-100 text-red-800',
        pending: 'bg-yellow-100 text-yellow-800',
    };
    const labels = { sent: t('sent'), failed: t('failed'), pending: t('pending') };
    return <Badge className={`text-xs hover:bg-opacity-80 ${styles[status]}`}>{labels[status]}</Badge>;
}
```

src/components/profile/NotificationPreferencesSection.tsx
```diff
@@ -139,232 +138,478 @@ export function NotificationPreferencesSection() {
...
+                            emptyMessage="Δεν βρέθηκε Δήμος."
+                            loading={loadingCities}
+                            className="w-64"
+                            TriggerComponent={({ isOpen }) => (
+                                <Button variant="outline" onClick={fetchCitiesIfNeeded}>
+                                    <Bell className="mr-2 h-4 w-4" />
+                                    Επιλέξτε Πόλη
+                                </Button>
+                            )}
+                        />
```
kouloumos: The Combobox for selecting a city is configured almost identically in two places (empty state and below the preferences table, see ~L453). Same `items`, `value`, `onChange`, `placeholder`, `getItemLabel`, etc. — only the button label differs. Extract it to a local component:

```tsx
function CitySelector({ label }: { label: string }) { ... }

File: src/components/profile/UserInfoForm.tsx [hunk 7]
```diff
@@ -55,89 +61,282 @@ export function UserInfoForm({ user, isOnboarded }: UserInfoFormProps) {
...
+                        borderBottom: "none",
+                    }}>{t("tabNotifications")}</TabsTrigger>
+                <TabsTrigger value="account"
+                    disabled={!isOnboarded}
+                    className="rounded-none rounded-tl-lg py-3 px-4 data-[state=active]:border-t-[#fc550a] data-[state=active]:border-l-[#fc550a] data-[state=active]:border-r-[#fc550a] data-[state=active]:border-b-white data-[state=active]:border-2"
+                    style={{
+                        borderTopLeftRadius: "0.5rem",
+                        borderTopRightRadius: "0.5rem",
+                        borderBottom: "none",
+                    }}>{t("tabAccount")}</TabsTrigger>
```
kouloumos: Each `TabsTrigger` has the exact same `className` and `style` props copy-pasted. Extract to a constant:

```tsx
const tabTriggerProps = {
    className: "rounded-none rounded-tl-lg py-3 px-4 data-[state=active]:border-t-[#fc550a] ...",
    style: { borderTopLeftRadius: "0.5rem", borderTopRightRadius: "0.5rem", borderBottom: "none" },
};
```

Then use `<TabsTrigger {...tabTriggerProps} value="personal">`.

kouloumos (commented):
The tabbed profile redesign is a nice direction -- the notification preferences table with per-channel toggles and the optimistic updates are well done. The `allowContact` → `allowProductUpdates` rename and `allowPetitionUpdates` addition are clean across the board.

A few things to address before this is mergeable though. First, the branch needs a rebase on main — it's 19 commits behind.

The biggest pattern issue: `NotificationPreferencesSection` has **zero i18n** — every string is hardcoded in Greek. The rest of the profile page uses `useTranslations("Profile")` correctly, so this is inconsistent. There are also a few hardcoded Greek strings in `UserInfoForm.tsx` itself ("Τα στοιχεία σας", "Ρυθμίσεις επικοινωνίας", the placeholder "Βασίλης Παπαδόπουλος"). All of these need to go through `next-intl` — we have English translations set up for this page already in `messages/en.json`.

There's also quite a bit of duplicated code that should be extracted — left inline comments on the specifics, but the theme is: identical rendering blocks repeated across mobile/desktop views, tab styling copy-pasted 4 times, and Combobox configuration duplicated. The project has a strong convention of extracting shared logic (see `src/lib/utils/`, `src/lib/formatters/`) and this PR should follow that.

Other things to fix: `.claude/commands/review.md` was added in commit 80b7a9c5 and shouldn't be part of this PR — please revert that file.

On the phone validation module (`src/lib/utils/phone.ts`): 164 lines of hand-rolled libphonenumber metadata with full validation only for GR and stub entries for ~200 other countries is a lot of code to maintain. Have you considered using `libphonenumber-js`? It's lightweight, well-maintained, and would replace all of this with a few function calls.

src/lib/notifications/tokens.ts
```diff
@@ -1,71 +1,53 @@
...
-        console.error('Error verifying unsubscribe token:', error);
+    } catch {
         return null;
     }
 }
 
+export function buildUnsubscribeUrl(userId: string, cityId: string): string {
+    const token = generateUnsubscribeToken(userId, cityId);
+    return `${env.NEXTAUTH_URL}/el/unsubscribe?token=${encodeURIComponent(token)}`;
+}
```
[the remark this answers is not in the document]
kouloumos: I agree with this one

src/app/api/unsubscribe/route.ts
```diff
@@ -0,0 +1,31 @@
...
+                updateUserProfile(data.userId, { allowProductUpdates: false, allowPetitionUpdates: false }),
+                disableAllNotificationPreferences(data.userId),
+            ]);
+        } else {
+            await disableNotificationPreferenceByCityId(data.userId, data.cityId);
+        }
+        return NextResponse.json({ success: true });
+    } catch {
+        return NextResponse.json({ error: 'Not found' }, { status: 404 });
+    }
```
kouloumos: Same pattern we just fixed in the notification preference routes -- all errors caught as 404. `disableNotificationPreferenceByCityId` already throws `NotFoundError` (notifications.ts:1466), so `handleApiError` would map that correctly to 404 while letting DB errors surface as 500.

```typescript
} catch (error) {
    return handleApiError(error, 'Failed to unsubscribe');
}
```

src/app/[locale]/(other)/admin/users/page.tsx
```diff
@@ -28,7 +28,7 @@ export default function UsersPage() {
     // Calculate enhanced stats
     const totalUsers = users.length
     const onboardedUsers = users.filter(user => user.onboarded).length
-    const contactableUsers = users.filter(user => user.allowContact).length
+    const contactableUsers = users.filter(user => user.allowProductUpdates).length
```
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/users/page.tsx]: Now that we are changing the schema, we need to think how we handle the transition from allowContact to `allowProductUpdates` &  `allowPetitionUpdates`. Who we consider `contactable` in the view we want to have in the admin users page?

prisma/schema.prisma
```diff
@@ -666,7 +666,8 @@ model User {
   emailVerified DateTime?
   phone         String?
   onboarded     Boolean         @default(false)
-  allowContact  Boolean         @default(false)
+  allowProductUpdates  Boolean  @default(false)
+  allowPetitionUpdates Boolean  @default(false)
```
christosporios, line 670: this should be true -- when a user submits a petition, we ask for their consent to email them, so consent has currently been given for all petitions.

christosporios (changes requested):
In addition to this, we should think about how we can let users unsubscribe from product update emails -- and implement product update email sending in a separate PR.

src/lib/db/notifications.ts
```diff
@@ -869,28 +873,10 @@ export async function createNotificationsForMeeting(
...
-                        administrativeBody: meeting.administrativeBody
-                    },
-                    city: {
-                        name_municipality: meeting.city.name_municipality
-                    }
-                });
+                const emailContent = await generateEmailContent(notificationData);
+                const smsBody = await generateSmsContent(notificationData);
 
                 // Create email delivery (always)
```
kouloumos, line 895: The `notifyByEmail` and `notifyByPhone` fields are added to `NotificationPreference`, users can toggle them in the UI, but this function ignores them -- it unconditionally creates an email delivery for every user and a phone delivery for anyone with a phone number. The preference data is already loaded and available via `userPrefMap`, so the fix is straightforward.

src/lib/db/users.ts
```diff
@@ -194,7 +194,17 @@ export async function deleteUser(id: string): Promise<void> {
...
-export type UserProfileUpdateData = Partial<Pick<User, 'name' | 'phone' | 'allowContact' | 'onboarded'>>;
+export async function deleteCurrentUser(): Promise<void> {
+    const user = await getCurrentUser();
+    if (!user) throw new Error('Not authenticated');
+    // Explicitly revoke all sessions server-side before deleting the record,
+    // so any outstanding session tokens are invalidated even if the client
+    // fails to call signOut.
+    await prisma.session.deleteMany({ where: { userId: user.id } });
+    await prisma.user.delete({ where: { id: user.id } });
+}
```
kouloumos: Two things: (1) `session.deleteMany` and `user.delete` aren't wrapped in a transaction -- a crash between them leaves the user logged out but not deleted. (2) The `User` model already has `onDelete: Cascade` on sessions, so the explicit `session.deleteMany` is redundant if `user.delete` succeeds. Either wrap in `prisma.$transaction([...])` or remove the explicit session delete and let the cascade handle it.

src/app/api/unsubscribe/route.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    try {
+        if (action === 'all') {
+            await Promise.all([
+                updateUserProfile(data.userId, { allowProductUpdates: false, allowPetitionUpdates: false }),
+                disableAllNotificationPreferences(data.userId),
+            ]);
+        } else {
+            await disableNotificationPreferenceByCityId(data.userId, data.cityId);
+        }
```
kouloumos: The `action` parameter from the request body isn't validated -- any string that isn't `'all'` silently falls through to the city-specific path.

src/components/profile/NotificationPreferencesSection.tsx
```diff
@@ -76,295 +100,497 @@ export function NotificationPreferencesSection() {
...
+                    TriggerComponent={() => (
+                        <Button variant="outline" size={size} onClick={fetchCitiesIfNeeded}>
+                            <Bell className="mr-2 h-4 w-4" />
+                            {label}
+                        </Button>
+                    )}
+                />
+            </div>
+        );
+    }
```
kouloumos: `CitySelector` is defined as a function declaration inside the component body, so it's re-created on every render. React treats it as a new component type each time, unmounting and remounting the Combobox -- losing any internal state (open/closed, search text). Move it outside the component or extract to its own file.

kouloumos (commented):
New flow and UI looks good! The biggest issue: the new `notifyByEmail`/`notifyByPhone` toggles are never actually checked when notifications are sent -- `createNotificationsForMeeting` still creates deliveries for all channels unconditionally. Users can flip the checkboxes but nothing changes on the delivery side.

On test coverage: `tokens.ts` has zero tests and it's pure HMAC logic that's trivially unit-testable -- that should be the first thing to add. The unsubscribe route and profile DELETE are also untested.

On the unsubscribe flow -- the email path is covered (link in the email body), but **how do we expect users to unsubscribe from phone/SMS notifications?** Right now there's no equivalent mechanism for that channel.

On the phone validation module -- I know you tried `libphonenumber-js` and ran into issues, so the hand-rolled approach makes sense.

Left inline comments on the specifics.

VasiaKoum:
I accidentally removed my commits and it didn't let me to reopen this PR, so I opened a new PR in https://github.com/schemalabz/opencouncil/pull/324

prisma/schema.prisma
VasiaKoum: Answer in https://github.com/schemalabz/opencouncil/pull/324#issuecomment-4281878966.

src/lib/db/notifications.ts
VasiaKoum: Answer in https://github.com/schemalabz/opencouncil/pull/324#issuecomment-4281901490.

src/app/[locale]/(other)/admin/users/page.tsx
Comment (VasiaKoum) [reply src/app/[locale]/(other)/admin/users/page.tsx]: Answer in https://github.com/schemalabz/opencouncil/pull/324#issuecomment-4281908364.

PR #312, opened 2026-04-11
Add topic descriptions and admin CRUD UI

christosporios:
## Summary

- Adds `description` and `deprecated` fields to the `Topic` model with a single migration.
- New superadmin CRUD UI at `/admin/topics` — table shows a unified colored Symbol (icon in colored circle), subject count, description preview, and a prominent "Deprecated" state (amber row + destructive badge + strikethrough).
- Topic dialog supports: icon dropdown sourced from `iconMap` (previews the actual icon), a "Suggest distinct color" button that picks a hue-wheel candidate maximally far from every existing topic color and from previously-suggested colors in the same session, and a deprecated toggle.
- Delete is disabled in the UI when a topic has subjects; server-side `deleteTopic` throws `ConflictError` (409) if a topic is still referenced by subjects, topic labels, or notification preferences.
- New `getActiveTopicsForTasks()` helper: **deprecated topics are never passed to the `summarize` and `processAgenda` task APIs**, so the LLM stops classifying new subjects into retired topics while existing subjects keep their assignment.

## Task API contract change (coordinate with backend)

`topicLabels` on `ProcessAgendaRequest` and `RequestOnTranscript` changes from `string[]` to `{ name: string; description: string }[]`. This affects the `summarize`, `processAgenda`, `fixTranscript`, and `generatePodcastSpec` endpoints. The backend at `TASK_API_URL` needs to accept the new shape and use the descriptions as classification guidance in its prompt templates. The response shape (`SummarizeResult.speakerSegmentSummaries[].topicLabels: string[]`) is unchanged.

## Note

While this branch was being developed, an unrelated `fixup! feat: shared Valkey cache handler for multi-instance deployments` commit (`4bceda2`) got force-pushed onto it. It's included in this PR's diff (`src/app/[locale]/(other)/admin/cache/page.tsx`, `src/components/admin/cache/CacheStats.tsx`). If that belongs to a separate PR, it should be moved before merge.

## Test plan

- [ ] `npx prisma migrate dev` applies `20260411120000_add_topic_description` cleanly; existing topics backfill with empty description and `deprecated = false`.
- [ ] `npm run prisma:generate && npx tsc --noEmit && npm test` all pass locally.
- [ ] Sign in as a superadmin, open `/admin/topics`: create, edit, delete flows all work; the Sparkles button suggests distinct colors on repeated clicks; icon dropdown shows each icon in the currently-selected color.
[...]

PR #313, opened 2026-04-11
Add topic descriptions and admin CRUD UI

christosporios:
Replaces #312 (which had an unrelated Valkey cache fixup swept into its diff).

Goes together with the corresponding backend change https://github.com/schemalabz/opencouncil-tasks/pull/34

## Summary

- Adds `description` and `deprecated` fields to the `Topic` model with a single migration.
- New superadmin CRUD UI at `/admin/topics` — table shows a unified colored Symbol (icon in colored circle), subject count, description preview, and a prominent "Deprecated" state (amber row + destructive badge + strikethrough).
- Topic dialog supports: icon dropdown sourced from `iconMap` (previews the actual icon), a "Suggest distinct color" button that picks a hue-wheel candidate maximally far from every existing topic color and from previously-suggested colors in the same session, and a deprecated toggle.
- Delete is disabled in the UI when a topic has subjects; server-side `deleteTopic` throws `ConflictError` (409) if a topic is still referenced by subjects, topic labels, or notification preferences.
- New `getActiveTopicsForTasks()` helper: **deprecated topics are never passed to the `summarize` and `processAgenda` task APIs**, so the LLM stops classifying new subjects into retired topics while existing subjects keep their assignment.

## Task API contract change (coordinate with backend)

`topicLabels` on `ProcessAgendaRequest` and `RequestOnTranscript` changes from `string[]` to `{ name: string; description: string }[]`. This affects the `summarize`, `processAgenda`, `fixTranscript`, and `generatePodcastSpec` endpoints. The backend at `TASK_API_URL` needs to accept the new shape and use the descriptions as classification guidance in its prompt templates. The response shape (`SummarizeResult.speakerSegmentSummaries[].topicLabels: string[]`) is unchanged.

## Test plan

- [ ] `npx prisma migrate dev` applies `20260411120000_add_topic_description` cleanly; existing topics backfill with empty description and `deprecated = false`.
- [ ] `npm run prisma:generate && npx tsc --noEmit && npm test` all pass locally.
- [ ] Sign in as a superadmin, open `/admin/topics`: create, edit, delete flows all work; the Sparkles button suggests distinct colors on repeated clicks; icon dropdown shows each icon in the currently-selected color.
- [ ] Toggle Deprecated on a topic that has subjects → topic row gains the amber background + "DEPRECATED" badge + strikethrough name; the topic disappears from subsequent `processAgenda`/`summarize` task payloads (verify via the existing dev-server log in `processAgenda.ts`).
- [ ] Non-superadmin users cannot reach `/admin/topics` or `/api/admin/topics` (existing admin-layout gate + explicit `isSuperAdmin` check on the routes).
- [ ] Try to delete a topic that is still referenced by a subject → 409 with a clear message; UI delete button is also disabled for such topics.
- [ ] Mutate a topic via the admin UI → the public `GET /api/topics` returns the change immediately (cache invalidation via `revalidatePath`).
- [ ] Coordinate with the task-backend team on the `topicLabels` shape change before merging.

---

> [!NOTE]
> **Medium Risk**
> Introduces a DB schema change and a breaking `topicLabels` contract change for task payloads, which can impact task-backend integration and topic classification behavior. Also adds new admin write endpoints and delete constraints that could affect existing workflows if misconfigured.
> 
> **Overview**
> Adds `description` and `deprecated` fields to `Topic`, with migration and Prisma schema updates.
> 
> Introduces a superadmin-only Topics admin surface (`/admin/topics`) with full CRUD via new `/api/admin/topics` endpoints, including server-side validation, cache revalidation, and guarded deletion that returns `409` when a topic is still referenced (subjects/topic labels/notification preferences).
> 
> Updates task request payloads so `topicLabels` becomes `{ name, description }[]` and ensures *deprecated topics are excluded* from agenda/summarize task inputs and from topic matching during task result ingestion; onboarding topic selection also hides deprecated topics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51d514e6445ae6b5bd28ed2a4cde498f885bfda1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

Adds `description` and `deprecated` fields to the `Topic` model, introduces a superadmin CRUD admin UI at `/admin/topics`, and updates the task API contract so `topicLabels` becomes `{ name, description }[]` instead of `string[]`. Deprecated topics are excluded from task requests and from new subject ingestion, while the onboarding `TopicSelector` filters them client-side.

All three concerns from the previous review round (non-atomic delete, incomplete `canDelete` guard, missing `topicLabels` assertion in tests) are fully resolved in this revision.

Safe to merge after coordinating the `topicLabels` contract change with the task backend team.

All three previous P1 concerns (non-atomic delete, incomplete canDelete guard, missing topicLabels test assertion) are fully resolved. Auth gating via `withUserAuthorizedToEdit({})` is correctly superadmin-only per auth.ts. Migration is backward-compatible. The only remaining finding is a P2 icon-name validation gap that is low-risk in practice since the UI enforces the constraint via a dropdown.

No files require special attention — the schema, DB layer, API routes, and admin UI components are all solid.

| Filename | Overview |
|----------|----------|
| src/lib/db/topics.ts | Adds CRUD helpers; `deleteTopic` now runs inside a `$transaction` with all three reference counts, resolving the prior atomicity concern. Auth gating is correct (`withUserAuthorizedToEdit({})` with empty params is superadmin-only per auth.ts line 59-60). |
| src/components/admin/topics/topics-table.tsx | Uses all three counts (subjects, topicLabels, notificationPreferences) to compute `canDelete` and render the correct tooltip — resolves the previous gap where label and notification counts were ignored. |
| src/lib/zod-schemas/topic.ts | Well-structured base fields with proper transforms; `createTopicSchema` defaults `deprecated` to `false`, `updateTopicSchema.partial()` leaves absent fields as `undefined` so `updateTopic` skips them correctly. |
| src/lib/db/utils.ts | Switches to `getActiveTopicsForTasks()` for the task request payload and filters deprecated topics from the `validateSubjectPersons` lookup so incoming task results can't silently assign subjects to retired topics. |
| src/lib/tasks/summarize.ts | Adds `deprecated: false` guard in the topic-name lookup within `handleSummarizeResult`; safe fail-open behavior (subject created without topic) if a deprecated name somehow appears in the task response. |
| src/lib/apiTypes.ts | Introduces `TopicLabelInfo` and updates `topicLabels` in both `ProcessAgendaRequest` and `RequestOnTranscript` (which covers `SummarizeRequest`, `FixTranscriptRequest`, `GeneratePodcastSpecRequest` via inheritance). |
| src/lib/__tests__/db-utils.test.ts | Updated to use `getActiveTopicsForTasks`, mock includes `description`, and the new assertion `expect(result.topicLabels).toEqual([{ name: 'Environment', description: '' }])` closes the contract regression gap flagged in the previous review. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Admin UI] --> B[PUT/POST/DELETE /api/admin/topics]
    B --> C{withUserAuthorizedToEdit}
    C -->|not superadmin| D[Throws 401]
    C -->|superadmin| E[DB function]
    E --> F{deleteTopic transaction}
    F --> G[Count subjects + topicLabels + notificationPrefs]
    G -->|any refs exist| H[ConflictError 409]
    G -->|no refs| I[tx.topic.delete]
    E --> J[revalidatePath /api/topics]
    K[processAgenda / summarize task] --> L[getActiveTopicsForTasks]
    L --> M[WHERE deprecated = false]
    M --> N[topicLabels as name+description pairs]
    O[handleSummarizeResult] --> P[findMany WHERE deprecated=false]
    P --> Q[Silently skip deprecated topic names in result]
```

`````markdown
Path: src/lib/zod-schemas/topic.ts
Line: 13

Comment:
**Icon value not validated against `iconMap`**

`icon` accepts any non-empty string, so a direct API call with an arbitrary value like `"FakeIcon"` would be stored. In the admin table, `iconMap["FakeIcon"]` resolves to `undefined`, and in `topic-dialog.tsx` the `SelectItem` would try to render `<IconComponent … />` with `undefined` — a runtime crash. Restricting to `iconMap` keys in the schema closes the gap:

```ts
import { iconMap } from "@/components/icon";
const VALID_ICON_NAMES = Object.keys(iconMap);
// ...
icon: z
    .string()
    .refine((v) => VALID_ICON_NAMES.includes(v), { message: "icon must be a valid icon name" })
    .nullable()
    .optional()
    .transform((val) => val || null),
```

This is only reachable via direct API calls since the UI enforces the constraint through the `Select` dropdown, so risk is low but the schema is the right place to enforce it.

`````

</details>

<sub>Reviews (11): Last reviewed commit: ["refactor: replace manual validation with..."](https://github.com/schemalabz/opencouncil/commit/60cc90d0563a3cdebea3e624587b9bdd22e64fe2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28090680)</sub>

### DIFF
--- a/src/app/api/admin/topics/route.ts
+++ b/src/app/api/admin/topics/route.ts
@@ -1,19 +1,13 @@
 import { NextResponse } from "next/server";
 import { revalidatePath } from "next/cache";
-import { getCurrentUser } from "@/lib/auth";
+import { withUserAuthorizedToEdit } from "@/lib/auth";
 import { createTopic, getAllTopicsWithSubjectCount } from "@/lib/db/topics";
-import { handleApiError } from "@/lib/api/errors";
+import { BadRequestError, handleApiError } from "@/lib/api/errors";
 
 const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
 
-const badRequest = (error: string) =>
-    NextResponse.json({ error }, { status: 400 });
-
 export async function GET() {
-    const user = await getCurrentUser();
-    if (!user?.isSuperAdmin) {
-        return new NextResponse("Unauthorized", { status: 401 });
-    }
+    await withUserAuthorizedToEdit({});
 
     try {
         const topics = await getAllTopicsWithSubjectCount();
@@ -24,32 +18,29 @@ export async function GET() {
 }
 
 export async function POST(request: Request) {
-    const user = await getCurrentUser();
-    if (!user?.isSuperAdmin) {
-        return new NextResponse("Unauthorized", { status: 401 });
-    }
+    await withUserAuthorizedToEdit({});
 
     try {
         const data = await request.json();
         const { name, name_en, colorHex, icon, description, deprecated } = data;
 
         if (typeof name !== "string" || name.trim() === "") {
-            return badRequest("name is required");
+            throw new BadRequestError("name is required");
         }
         if (typeof name_en !== "string" || name_en.trim() === "") {
-            return badRequest("name_en is required");
+            throw new BadRequestError("name_en is required");
         }
         if (typeof colorHex !== "string" || !HEX_REGEX.test(colorHex)) {
-            return badRequest("colorHex must be a valid hex color");
+            throw new BadRequestError("colorHex must be a valid hex color");
         }
         if (typeof description !== "string") {
-            return badRequest("description must be a string");
+            throw new BadRequestError("description must be a string");
         }
         if (icon !== undefined && icon !== null && typeof icon !== "string") {
-            return badRequest("icon must be a string or null");
+            throw new BadRequestError("icon must be a string or null");
         }
         if (deprecated !== undefined && typeof deprecated !== "boolean") {
-            return badRequest("deprecated must be a boolean");
+            throw new BadRequestError("deprecated must be a boolean");
         }
 
         const topic = await createTopic({
--- a/src/app/api/admin/topics/route.ts
+++ b/src/app/api/admin/topics/route.ts
@@ -3,8 +3,7 @@ import { revalidatePath } from "next/cache";
 import { withUserAuthorizedToEdit } from "@/lib/auth";
 import { createTopic, getAllTopicsWithSubjectCount } from "@/lib/db/topics";
 import { BadRequestError, handleApiError } from "@/lib/api/errors";
-
-const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
+import { HEX_REGEX } from "@/lib/utils/colorSuggestion";
 
 export async function GET() {
     await withUserAuthorizedToEdit({});
--- a/src/app/api/admin/topics/route.ts
+++ b/src/app/api/admin/topics/route.ts
@@ -6,6 +6,9 @@ import { handleApiError } from "@/lib/api/errors";
 
 const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
 
+const badRequest = (error: string) =>
+    NextResponse.json({ error }, { status: 400 });
+
 export async function GET() {
     const user = await getCurrentUser();
     if (!user?.isSuperAdmin) {
@@ -31,22 +34,22 @@ export async function POST(request: Request) {
         const { name, name_en, colorHex, icon, description, deprecated } = data;
 
         if (typeof name !== "string" || name.trim() === "") {
-            return new NextResponse("name is required", { status: 400 });
+            return badRequest("name is required");
         }
         if (typeof name_en !== "string" || name_en.trim() === "") {
-            return new NextResponse("name_en is required", { status: 400 });
+            return badRequest("name_en is required");
         }
         if (typeof colorHex !== "string" || !HEX_REGEX.test(colorHex)) {
-            return new NextResponse("colorHex must be a valid hex color", { status: 400 });
+            return badRequest("colorHex must be a valid hex color");
         }
         if (typeof description !== "string") {
-            return new NextResponse("description must be a string", { status: 400 });
+            return badRequest("description must be a string");
         }
         if (icon !== undefined && icon !== null && typeof icon !== "string") {
-            return new NextResponse("icon must be a string or null", { status: 400 });
+            return badRequest("icon must be a string or null");
         }
         if (deprecated !== undefined && typeof deprecated !== "boolean") {
-            return new NextResponse("deprecated must be a boolean", { status: 400 });
+            return badRequest("deprecated must be a boolean");
         }
 
         const topic = await createTopic({
--- a/src/app/api/admin/topics/route.ts
+++ b/src/app/api/admin/topics/route.ts
@@ -4,6 +4,8 @@ import { getCurrentUser } from "@/lib/auth";
 import { createTopic, getAllTopicsWithSubjectCount } from "@/lib/db/topics";
 import { handleApiError } from "@/lib/api/errors";
 
+const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
+
 export async function GET() {
     const user = await getCurrentUser();
     if (!user?.isSuperAdmin) {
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: src/lib/db/utils.ts
Comment (kouloumos) [inline src/lib/db/utils.ts]: This raw query doesn't filter `deprecated: false`. The request-building side correctly uses `getActiveTopicsForTasks()`, but the save side here will still match and link deprecated topic names if they appear in the AI response (stale callback, race with a deprecation toggle, etc.). Should filter or reuse `getActiveTopicsForTasks()` to keep both paths consistent.

File: src/app/api/admin/topics/route.ts [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
+
+const badRequest = (error: string) =>
+    NextResponse.json({ error }, { status: 400 });
+
+export async function GET() {
+    const user = await getCurrentUser();
+    if (!user?.isSuperAdmin) {
+        return new NextResponse("Unauthorized", { status: 401 });
+    }
```
kouloumos: The rest of the admin routes use `await withUserAuthorizedToEdit({})` for superadmin gating (see `consultations/route.ts`, `qr/route.ts`). This uses `getCurrentUser` + manual check instead. Since `createTopic`/`updateTopic`/`deleteTopic` already call `withUserAuthorizedToEdit({})` internally, the route-level check is also redundant — switching to the standard pattern would remove the duplication.

src/app/api/admin/topics/route.ts
```diff
@@ -0,0 +1,71 @@
+import { NextResponse } from "next/server";
+import { revalidatePath } from "next/cache";
+import { getCurrentUser } from "@/lib/auth";
+import { createTopic, getAllTopicsWithSubjectCount } from "@/lib/db/topics";
+import { handleApiError } from "@/lib/api/errors";
+
+const HEX_REGEX = /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/;
```
kouloumos: This regex and the `badRequest` helper are duplicated in `[topicId]/route.ts` and again in `topic-dialog.tsx`. Could extract to a shared module. Alternatively, the existing `BadRequestError` class in `src/lib/api/errors.ts` would let you throw inside the try block and have `handleApiError` return the 400 automatically, which is cleaner than the early-return pattern.

kouloumos (commented):
The migration is clean, the admin UI is solid, and the `deleteTopic` transaction with `ConflictError` is well done. My main concern is that `deprecated` isn't fully wired through -- the request side filters correctly, but a few other code paths still expose deprecated topics.

`TopicSelector.tsx` (the onboarding UI where users pick notification topics) still calls `getAllTopics()` -- so deprecated topics show up as subscribable options even though the LLM will never classify new subjects into them. Similarly, `validateSubjectPersons` in `src/lib/db/utils.ts` uses a raw `prisma.topic.findMany()` (line 156) to resolve topic names on the save path -- a stale callback returning a deprecated topic name would still get linked to a subject (although it might not be an issue with task versions?). Both should filter on `deprecated: false` to be consistent with the request-building side. And `summarize.ts` has a dead `getAllTopics` import (line 6) that should be cleaned up while you're in the area.

Left inline comments on the route-level specifics.

kouloumos (approved):
All the first-round feedback was addressed cleanly. The extra fix in `handleSummarizeResult` (filtering `deprecated: false` on the topic lookup) is a welcome belt-and-suspenders addition.

Also looked at the backend side (opencouncil-tasks#34) — the `TopicLabelInfo` types match, `formatTopicLabels` handles empty descriptions correctly, and the prompts explicitly instruct the LLM to output only the name.

I pushed 5dc41f6 that replaces the manual validation in both topic route files with Zod schemas, following the same pattern we use for cities (`src/lib/zod-schemas/city.ts`). This cuts ~50 lines of duplicated `typeof` checks across the two routes and adds ZodError handling to `handleApiError` so it returns 400 automatically. Includes 20 tests for the schemas.

PR #318, opened 2026-04-15
fix: resolve transcript editing lag from context re-render cascade

kouloumos:
## Summary

- Split `CouncilMeetingDataContext` into a stable main context (functions, lookups, rarely-changing state) and a volatile transcript context (transcript data)
- Stabilize `EditingContext` and `HighlightContext` values during transcript edits using internal refs for lazy transcript access
- Migrate all transcript consumers to `useCouncilMeetingTranscript()`

## Problem

After #289 (scrolling fix), reviewers reported lag in the editing interface — clicking utterances, typing corrections, and adding speakers all felt sluggish. The root cause: every utterance edit called `setTranscript()`, which invalidated the monolithic context value, re-rendering every SpeakerSegment and Utterance (~50-100 segments, hundreds of utterances). The old `renderMock` system had masked this by rendering off-screen segments as lightweight text-only elements.

## Approach

The main context's `useMemo` deps no longer include `transcript` or transcript-derived values. Editing an utterance now only re-renders `Transcript.tsx` + the one affected `SpeakerSegment` + the one affected `Utterance`.

## Test plan

- [ ] Open a large transcript in editing mode — verify lag is gone
- [ ] Edit utterance text and timestamps
- [ ] Move utterances between segments  
- [ ] Create/delete segments, change speaker person/label
- [ ] Shift-click range selection
- [ ] Highlight creation and editing
- [ ] Deep link with `?t=` parameter

---

> [!NOTE]
> **Medium Risk**
> Refactors core meeting/transcript React contexts and many consumers; behavior should be equivalent but there’s moderate risk of stale data or missed re-renders in editing/highlight flows due to the new memoization/ref patterns.
> 
> **Overview**
> **Fixes transcript editing lag** by splitting `CouncilMeetingDataContext` into a *stable* main context (meeting data, speaker tags/highlights, and mutation/lookup APIs) and a *volatile* `CouncilMeetingTranscriptContext` that only holds `transcript` and `getSpeakerSegmentById`.
> 
> Updates transcript-heavy consumers (`Transcript`, `TranscriptControls`, `KeyboardShortcuts`, `EditingContext`, `HighlightContext`, admin/subject views, etc.) to read transcript data via `useCouncilMeetingTranscript()`, and rewrites `EditingContext`/`HighlightContext` to use `useRef`-backed access so their callbacks/context values don’t change on every utterance edit.
> 
> Moves per-speaker segment count computation into `Transcript.tsx` and passes `segmentCount` as a primitive prop to `SpeakerSegment` to improve `React.memo` effectiveness, plus small related stabilizations (memoized provider values in `OptionsContext`, `KeyboardShortcutsContext`, and `ShareContext`, and an updated `Utterance` editor using a textarea ref and memoized context-menu open/close handler).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09007d351de13b09fd8323f1d9e2d0f41c8ba934. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR resolves transcript-editing lag by splitting `CouncilMeetingDataContext` into a stable main context (mutations, lookups) and a volatile `CouncilMeetingTranscriptContext`, then stabilising `EditingContext` and `HighlightContext` callbacks via internal `transcriptRef` refs, and finally switching `Utterance`'s textarea to uncontrolled to eliminate per-keystroke re-renders.

- `deleteEmptySegmentFn` calls `setSpeakerTags` as a side effect *inside* the `setTranscript` functional updater — impure updaters can be called multiple times in StrictMode/concurrent React, scheduling extra `speakerTags` state updates and triggering unnecessary `SpeakerSegment` re-renders; `extractSpeakerSegmentFn` already shows the correct pattern (separate sibling `setState` calls).
- Segment wrappers in `Transcript.tsx` use array index as React key, which causes `SpeakerSegment` local state (`isCollapsed`, `metadataDialogOpen`) to carry over to a different segment after any insert/delete/reorder operation.

Safe to merge with minor fixes — the two P2 findings are quality issues that won't cause data loss or runtime errors in production.

The overall approach is sound and well-executed. The two findings are P2: the impure updater in deleteEmptySegmentFn is a React anti-pattern that only has visible effects in development (StrictMode extra renders), and the index-key issue causes local UI state bleed on segment mutations (cosmetic). Neither is a data-corruption or crash bug. The PR solves a real performance problem cleanly.

src/components/meetings/CouncilMeetingDataContext.tsx (deleteEmptySegmentFn side-effect), src/components/meetings/transcript/Transcript.tsx (segment key strategy)

| Filename | Overview |
|----------|----------|
| src/components/meetings/CouncilMeetingDataContext.tsx | Core of the PR: splits a monolithic context into stable main + volatile transcript contexts; all mutations correctly use functional setTranscript(prev => ...); one anti-pattern where setSpeakerTags is called as a side effect inside a setTranscript updater in deleteEmptySegmentFn. |
| src/components/meetings/EditingContext.tsx | Correctly stabilised using a transcriptRef pattern; callbacks close over the ref instead of the live transcript state; useMemo value only invalidates on user actions (selection, processing). |
| src/components/meetings/HighlightContext.tsx | Migrated to use transcriptRef for stable callbacks; some useCallback functions (goToPreviousHighlight, goToNextHighlight, etc.) still capture highlightUtterances in deps and change whenever editing-highlight state changes — acceptable since those are highlight-specific state changes, not transcript edits. |
| src/components/meetings/transcript/Transcript.tsx | Migrated to transcript context; speakerTagSegmentCounts properly passes primitive segmentCount to SpeakerSegment; IntersectionObserver correctly depends on displayedSegments; wrapper div uses array-index as React key which can bleed local SpeakerSegment state during segment mutations. |
| src/components/meetings/transcript/Utterance.tsx | Switched from controlled textarea to uncontrolled (defaultValue + ref) to eliminate per-keystroke re-renders; handleContextMenuOpenChange wrapped in useCallback; edit flow reads text via textareaRef.current?.value correctly. |
| src/components/meetings/transcript/SpeakerSegment.tsx | Now receives segmentCount as a primitive prop (replaces context lookup); memoizedData correctly deps on segment.speakerTagId and segmentCount; local state for collapse/dialog is correct. |
| src/hooks/useUtteranceData.ts | New hook wrapping transcript context for utterance lookups; subscribes to volatile transcript context and is only used by UtteranceReferenceLink, which is appropriate. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant U as Utterance (uncontrolled)
    participant TC as CouncilMeetingTranscriptContext
    participant MC as CouncilMeetingDataContext (stable)
    participant EC as EditingContext
    participant HC as HighlightContext
    participant SS as SpeakerSegment
    participant T as Transcript

    Note over U: User types (no React state update)
    U->>U: textareaRef captures value
    U->>U: handleEdit on Enter/click
    U->>MC: updateUtterance(segId, uttId, updates)
    MC->>MC: setTranscript(prev => ...)
    MC->>TC: transcriptValue updates
    TC->>T: re-render (displayedSegments changes)
    T->>SS: new segment prop (only affected segment)
    SS->>U: new utterance prop
    Note over EC,HC: transcriptRef.current updated (no re-render cascade)
    TC-->>EC: re-render (cheap, 1 component)
    TC-->>HC: re-render (cheap, 1 component)
    EC->>EC: transcriptRef.current = transcript
    HC->>HC: transcriptRef.current = transcript
    Note over MC: Main context unchanged - no cascade to all consumers
```

`````markdown
Path: src/components/meetings/CouncilMeetingDataContext.tsx
Line: 216-231

Comment:
**Side-effect inside functional state updater**

`setSpeakerTags` is called inside the `setTranscript` functional updater, making the updater impure. React's documentation explicitly requires that state updater functions be pure (no side effects). In React 18 StrictMode, updaters are intentionally invoked twice, scheduling `setSpeakerTags` twice and causing unnecessary re-renders of all `SpeakerSegment` components (which depend on `speakerTags` via `nextUnknownLabel`) — counterproductive for a PR focused on minimizing renders.

`extractSpeakerSegmentFn` (line 249) already does this correctly, calling `setTranscript(...)` and `setSpeakerTags(...)` as separate sibling statements that React 18 will batch automatically.

The simplest fix is to mirror `extractSpeakerSegmentFn`: call `setTranscript` first, then call `setSpeakerTags` after — React 18 will batch the two `setState` calls from the same synchronous frame.

---

Path: src/components/meetings/transcript/Transcript.tsx
Line: 181-194

Comment:
**Array-index key causes local state bleed across segments**

Using array index as React key means `SpeakerSegment`'s local state (`isCollapsed`, `userHasInteracted`, `metadataDialogOpen`) persists on the wrong component after segments are inserted, deleted, or reordered — for example, after `createEmptySegmentBefore` or `moveUtterancesToPrevious`. A segment that was collapsed at index 5 will remain collapsed after another segment is inserted before it, because React reuses the component at that position with the new `segment` prop.

Using `segment.id` as the map key would let React properly unmount/remount components when segment identity changes. The `id` attribute on the wrapper `<div>` (used by the IntersectionObserver) can stay index-based independently.

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: eliminate typing lag with uncontrol..."](https://github.com/schemalabz/opencouncil/commit/09007d351de13b09fd8323f1d9e2d0f41c8ba934) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28471661)</sub>

### DIFF
--- a/src/components/meetings/CouncilMeetingDataContext.tsx
+++ b/src/components/meetings/CouncilMeetingDataContext.tsx
@@ -16,6 +16,7 @@ import type { HighlightWithUtterances } from '@/lib/db/highlights';
 
 interface CouncilMeetingTranscriptContextType {
     transcript: Transcript;
+    getSpeakerSegmentById: (id: string) => Transcript[number] | undefined;
 }
 
 const CouncilMeetingTranscriptContext = createContext<CouncilMeetingTranscriptContextType | undefined>(undefined);
@@ -60,11 +61,6 @@ export interface CouncilMeetingDataContext extends MeetingData {
     updateHighlight: (highlightId: string, updates: Partial<HighlightWithUtterances>) => void;
     removeHighlight: (highlightId: string) => void;
     extractSpeakerSegment: (segmentId: string, startUtteranceId: string, endUtteranceId: string) => Promise<void>;
-
-    // Backward compat — TEMPORARY, remove after all consumers migrate to useCouncilMeetingTranscript()
-    transcript: Transcript;
-    getSpeakerSegmentById: (id: string) => Transcript[number] | undefined;
-    getSpeakerSegmentCount: (tagId: string) => number;
 }
 
 const CouncilMeetingDataContext = createContext<CouncilMeetingDataContext | undefined>(undefined);
@@ -85,17 +81,6 @@ export function CouncilMeetingDataProvider({ children, data }: {
     // === DERIVED MAPS FROM MUTABLE STATE ===
     const speakerTagsMap = useMemo(() => new Map(speakerTags.map(tag => [tag.id, tag])), [speakerTags]);
     const highlightsMap = useMemo(() => new Map(highlights.map(h => [h.id, h])), [highlights]);
-
-    // These are transcript-derived and change on every edit, but needed for backward compat
-    const speakerSegmentsMap = useMemo(() => new Map(transcript.map(segment => [segment.id, segment])), [transcript]);
-    const speakerTagSegmentCounts = useMemo(() => {
-        const counts = new Map<string, number>();
-        transcript.forEach(segment => {
-            const count = counts.get(segment.speakerTag.id) || 0;
-            counts.set(segment.speakerTag.id, count + 1);
-        });
-        return counts;
-    }, [transcript]);
 
     // === HELPERS ===
     const recalculateSegmentTimestamps = useCallback((utterances: Array<{ startTimestamp: number; endTimestamp: number }>) => {
@@ -311,9 +296,11 @@ export function CouncilMeetingDataProvider({ children, data }: {
     }, [recalculateSegmentTimestamps]);
 
     // === TRANSCRIPT CONTEXT VALUE (volatile — changes on every edit) ===
+    const speakerSegmentsMap = useMemo(() => new Map(transcript.map(segment => [segment.id, segment])), [transcript]);
     const transcriptValue = useMemo(() => ({
         transcript,
-    }), [transcript]);
+        getSpeakerSegmentById: (id: string) => speakerSegmentsMap.get(id),
+    }), [transcript, speakerSegmentsMap]);
 
     // === MAIN CONTEXT VALUE (stable during normal utterance editing) ===
     // Deps: data, maps derived from speakerTags/highlights (rarely change), stable callbacks.
@@ -345,10 +332,6 @@ export function CouncilMeetingDataProvider({ children, data }: {
         updateHighlight: updateHighlightFn,
         removeHighlight,
         extractSpeakerSegment: extractSpeakerSegmentFn,
-        // TEMPORARY backward compat — remove after all consumers migrate to useCouncilMeetingTranscript()
-        transcript,
-        getSpeakerSegmentById: (id: string) => speakerSegmentsMap.get(id),
-        getSpeakerSegmentCount: (tagId: string) => speakerTagSegmentCounts.get(tagId) || 0,
     }), [
         data, peopleMap, partiesMap,
         speakerTags, speakerTagsMap,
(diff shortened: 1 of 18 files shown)


kouloumos:
Solved with #337

PR #332, opened 2026-04-24
Send product emails from admin panel

### LINKED ISSUE #322: Send product update emails from admin panel
### Concept

Super-admins need a way to send freeform product update emails to users who have opted in. Currently, the platform sends automated meeting notification emails, but there's no way to manually compose and send announcements, product updates, or other communications to users.

This builds on the `allowProductUpdates` consent field and token-based unsubscribe flow introduced in #307. The email infrastructure (Resend) and unsubscribe mechanism already exist — this issue is about exposing a compose-and-send interface in the admin panel.

### User Story

As a super-admin, I want to compose and send product update emails to opted-in users, so that we can communicate platform updates, announcements, and other relevant information to our user base through our existing email infrastructure.

### Context

- #307 added the `allowProductUpdates` consent field to the User model and a token-based unsubscribe flow (HMAC-SHA256 signed, 30-day TTL)
- Email delivery uses Resend (`src/lib/notifications/deliver.ts`)
- The admin panel already has sections for users, notifications, meetings, etc.
- Only users with `allowProductUpdates: true` should receive these emails
- Each email must include an unsubscribe link — the token-based unsubscribe infrastructure from #307 should be reusable here
- Scope is email only — no WhatsApp/SMS
- Super-admin authorization is required (see `withUserAuthorizedToEdit` pattern)

VasiaKoum:
Fixes #322

<h3>Greptile Summary</h3>

This PR adds a product-update email feature allowing super-admins to compose, preview, and bulk-send emails from the admin panel. It introduces a Resend batch endpoint integration with per-item error recovery (permissive mode), idempotency keys, and per-recipient unsubscribe tokens that carry no city scope. Several issues raised in previous reviews have been addressed: `disableNotificationPreferenceByCityId` is now idempotent via `updateMany`, `userName` is HTML-escaped before substitution, the "send to all" button is disabled after a successful bulk send (`status === 'success'`), and per-item batch failures are correctly parsed from `errors[].index` rather than silently counted as sent.

Safe to merge with minor attention to memory usage on large recipient lists.

All previously-flagged P1s have been addressed in this revision. The one remaining comment is a P2 memory/scalability concern (pre-rendering all recipients concurrently) that won't cause correctness issues at typical list sizes but could become a problem as the user base grows.

src/lib/email/productUpdate.ts — the Promise.all over all recipients before batching

| Filename | Overview |
|----------|----------|
| src/lib/email/productUpdate.ts | New file orchestrating bulk and test product-update sends; uses batching + idempotency keys; renders all recipients in-memory via Promise.all before dispatch |
| src/lib/email/resend.ts | Added sendEmailBatch using Resend's permissive batch mode; correctly parses per-item errors from errors[].index; handles dev override |
| src/lib/email/templates/productUpdateDefault.ts | New template with escapeHtml for userName substitution; handles URL-encoded placeholder forms for marked-processed hrefs |
| src/app/api/admin/product-updates/send/route.ts | New POST route guarded by isSuperAdmin; Zod-validated schema; dispatches to test or bulk send |
| src/components/admin/product-updates/SendProductUpdateDialog.tsx | New dialog for composing and sending product updates; two-step confirm flow; send-all button correctly disabled after success via status === 'success' check |
| src/lib/db/notifications.ts | Made cityId optional in getUnsubscribeContext; disableNotificationPreferenceByCityId now uses updateMany for idempotent no-op on missing preference |
| src/lib/notifications/tokens.ts | cityId made optional throughout; cityId omitted from token payload when not provided; buildUnsubscribeUrl and generateUnsubscribeToken updated accordingly |
| src/lib/db/productUpdates.ts | New DB helpers to fetch opted-in product-update recipients and count; straightforward Prisma queries |
| src/app/api/unsubscribe/route.ts | Added null guards for cityId before invoking city-scoped unsubscribe operations; returns 400 when token has no city scope |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Admin as Admin UI
    participant SendRoute as POST /api/admin/product-updates/send
    participant ProductUpdate as productUpdate.ts
    participant DB as DB (productUpdates.ts)
    participant TokenLib as tokens.ts
    participant ResendBatch as resend.ts (sendEmailBatch)
    participant Resend as Resend API

    Admin->>SendRoute: POST {subject, bodyHtml, tags?}
    SendRoute->>SendRoute: isSuperAdmin check + Zod validation
    alt testEmail provided
        SendRoute->>ProductUpdate: sendProductUpdateTest()
        ProductUpdate->>TokenLib: buildUnsubscribeUrl(adminUserId, undefined)
        TokenLib-->>ProductUpdate: unsubscribeUrl (no city scope)
        ProductUpdate->>Resend: sendEmail (single)
        Resend-->>ProductUpdate: result
    else bulk send
        SendRoute->>ProductUpdate: sendProductUpdateToAll()
        ProductUpdate->>DB: getProductUpdateRecipients()
        DB-->>ProductUpdate: recipients[]
        ProductUpdate->>ProductUpdate: Promise.all – render HTML per recipient
        loop batches of 100
            ProductUpdate->>ResendBatch: sendEmailBatch(batch, idempotencyKey)
            ResendBatch->>Resend: POST /emails/batch (x-batch-validation: permissive)
            Resend-->>ResendBatch: {data:[], errors:[{index,message}]}
            ResendBatch-->>ProductUpdate: {success, failedTos}
            ProductUpdate->>ProductUpdate: sleep 500ms (if not last batch)
        end
    end
    SendRoute-->>Admin: {sent, failed, failedEmails}
```

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `src/app/api/unsubscribe/route.ts`, line 3 ([link](https://github.com/schemalabz/opencouncil/blob/ba471428b4527439d05e1c6d11e61795d6d942b0/src/app/api/unsubscribe/route.ts#L3)) 

   <a href="#">P1</a> **Broken imports of deleted functions — TypeScript build failure**

   `disableAllNotificationPreferences` and `disableNotificationPreferenceByCityId` are imported from `@/lib/db/notifications` on line 3 and called on lines 33, 39, and 56, but this PR deletes both functions from `notifications.ts`. The module no longer exports them, so TypeScript compilation will fail and the entire unsubscribe flow breaks on deploy.

   The two functions need to either be kept in `notifications.ts` or their call-sites in this file need to be rewritten inline before merging.

   `````markdown
   Path: src/app/api/unsubscribe/route.ts
   Line: 3

   Comment:
   **Broken imports of deleted functions — TypeScript build failure**

   `disableAllNotificationPreferences` and `disableNotificationPreferenceByCityId` are imported from `@/lib/db/notifications` on line 3 and called on lines 33, 39, and 56, but this PR deletes both functions from `notifications.ts`. The module no longer exports them, so TypeScript compilation will fail and the entire unsubscribe flow breaks on deploy.

   The two functions need to either be kept in `notifications.ts` or their call-sites in this file need to be rewritten inline before merging.

   `````
   </details>

2. `src/lib/db/notifications.ts`, line 1512-1523 ([link](https://github.com/schemalabz/opencouncil/blob/262ae1b7cb163e3dae96b90c42729a1d1e810fa0/src/lib/db/notifications.ts#L1512-L1523)) 

   <a href="#">P1</a> **`disableNotificationPreferenceByCityId` is not idempotent — unsubscribe errors on repeat clicks**

   This function throws `NotFoundError` when the city preference doesn't exist. The function is called from the unsubscribe route's `'preferences'` action path when `unsubscribeCity === true`. If the city preference has already been deleted (user clicked the link twice, or clicked unsubscribe from a stale email), the second attempt throws, `Promise.all` rejects, `handleApiError` returns an error response, and the user sees an error screen instead of a success.

   For an unsubscribe flow the operation should be idempotent — missing preference should be treated as "already done":

   ```typescript
   export async function disableNotificationPreferenceByCityId(userId: string, cityId: string) {
       const preference = await prisma.notificationPreference.findUnique({
           where: { userId_cityId: { userId, cityId } },
       });

       if (!preference) return; // already unsubscribed — treat as success

       return prisma.notificationPreference.update({
           where: { id: preference.id },
           data: { notifyByEmail: false, notifyByPhone: false },
       });
   }
   ```

   `````markdown
   Path: src/lib/db/notifications.ts
   Line: 1512-1523

   Comment:
   **`disableNotificationPreferenceByCityId` is not idempotent — unsubscribe errors on repeat clicks**

   This function throws `NotFoundError` when the city preference doesn't exist. The function is called from the unsubscribe route's `'preferences'` action path when `unsubscribeCity === true`. If the city preference has already been deleted (user clicked the link twice, or clicked unsubscribe from a stale email), the second attempt throws, `Promise.all` rejects, `handleApiError` returns an error response, and the user sees an error screen instead of a success.

   For an unsubscribe flow the operation should be idempotent — missing preference should be treated as "already done":

   ```typescript
   export async function disableNotificationPreferenceByCityId(userId: string, cityId: string) {
       const preference = await prisma.notificationPreference.findUnique({
           where: { userId_cityId: { userId, cityId } },
       });

       if (!preference) return; // already unsubscribed — treat as success

       return prisma.notificationPreference.update({
           where: { id: preference.id },
           data: { notifyByEmail: false, notifyByPhone: false },
       });
   }
   ```

   `````
   </details>
</details>

`````markdown

---

### Issue 1 of 1
src/lib/email/productUpdate.ts:74-83
**Unbounded `Promise.all` pre-renders all recipient emails in memory at once**

`Promise.all` over the full recipient list renders every per-user HTML string concurrently before a single batch is dispatched. For large opt-in lists (thousands of users), this holds all rendered HTML simultaneously in memory — each expanded template can be 20–50 KB — and exhausts memory or hits the serverless function's memory cap before any email is sent. Consider streaming the render into the batch loop so at most `BATCH_SIZE` strings are live at a time:

```typescript
for (let i = 0; i < recipients.length; i += BATCH_SIZE) {
    const slice = recipients.slice(i, i + BATCH_SIZE);
    const items = await Promise.all(slice.map(async (r) => { /* render + build item */ }));
    // dispatch batch, sleep, etc.
}
```


`````

</details>

<sub>Reviews (11): Last reviewed commit: ["feat(admin): send product-update emails ..."](https://github.com/schemalabz/opencouncil/commit/252e6e82b6ce93f1c86702b851f86d9439fb6db6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29620625)</sub>

### CODE REVIEW

**Comment by VasiaKoum** [comment]
@christosporios Do you want to check the preview? Log in as super-user and then visit `admin/notifications/`, click "Αποστολή ενημέρωσης" to open the email editor. Then, you can complete the subject, there is a default email body (supports md and html) and the recipient field (for test email).

We also discussed with @kouloumos, whether it would be better to move this dialog in a different area, maybe in a new page (?)

Last but not least, I tried the `@react-email/editor` but it was too heavy for compilation, is a TipTap shell that pulls in ProseMirror + ~22 TipTap extensions as a single statically-imported dep. So, I selected a simple, custom md & html editor for our case.

File: src/components/admin/product-updates/SendProductUpdateDialog.tsx [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+                    <Button
+                        variant="outline"
+                        onClick={sendTest}
+                        disabled={isLoading || !testEmailValid}
+                    >
+                        {status === 'sending-test' && <Loader2 className="mr-2 h-4 w-4 animate-spin" />}
+                        {t('sendTestButton')}
+                    </Button>
+                    <Button
+                        onClick={sendAll}
```
kouloumos: This needs a confirmation step before firing. A single click on "Send to all" sends to every opted-in user with no way to undo. Suggest adding an intermediate confirmation that shows the impact: how many recipients out of total users, and how many have opted out. This doubles as useful admin visibility into our opt-out rates.

src/app/api/admin/product-updates/recipients/route.ts
```diff
@@ -0,0 +1,21 @@
...
+    const user = await getCurrentUser();
+    if (!user) {
+        return NextResponse.json({ error: "Authentication required" }, { status: 401 });
+    }
+    if (!user.isSuperAdmin) {
+        return NextResponse.json({ error: "Forbidden" }, { status: 403 });
+    }
+
+    try {
+        const recipients = await getProductUpdateRecipients();
```
kouloumos: Use `prisma.user.count()` instead of loading full records — there's precedent in `cities.ts`, `topics.ts`, `reviews.ts`. Also return `totalUsers` here so the confirmation dialog can show the opt-out ratio (see top-level comment `#1`).

src/lib/email/productUpdate.ts
```diff
@@ -0,0 +1,116 @@
...
+    return result.success;
+}
+
+/**
+ * Send a product-update email to every consenting recipient with rate limiting.
+ * `bodyHtml` is the editor-sanitized HTML with {{userName}}/{{unsubscribeUrl}}
+ * placeholders intact; per-recipient substitution happens here. Sanitization
+ * is handled client-side so this server module stays jsdom-free.
+ */
+export async function sendProductUpdateToAll(params: {
```
kouloumos, line 73: Resend's [batch endpoint](https://resend.com/docs/api-reference/emails/send-batch-emails) sends up to 100 emails per call with built-in rate limiting, idempotency keys, and per-email error reporting. 1,000 recipients = ~10 batch calls = ~5 seconds, vs ~8 minutes with the current sequential approach. See top-level comment `#3` for the suggested approach.

src/lib/email/productUpdate.ts
```diff
@@ -0,0 +1,116 @@
...
+export async function sendProductUpdateTest(params: {
+    subject: string;
+    bodyHtml: string;
+    testEmail: string;
+    testName?: string;
+    adminUserId: string;
+}): Promise<SendProductUpdateResult> {
+    const { subject, bodyHtml, testEmail, testName, adminUserId } = params;
+    try {
+        const unsubscribeUrl = await buildUnsubscribeUrl(adminUserId, undefined, 'el');
```
kouloumos, line 147: nit: Hardcoded `'el'` locale. Add a `// TODO:` for i18n support.

kouloumos (commented):
Νice work on this -- the unsubscribe token changes are clean and the editor is well thought out. A few things to address before merge:

### Must fix

**1. No confirmation before "Send to all"**

This is an irreversible mass-email action behind a single click. I'd like to see a confirmation step that also shows the impact -- something like:

"Send to **142** out of **1,203** registered users.

This serves double duty: it prevents accidental sends, and gives the admin visibility into our opt-out rate. The recipient count is already fetched when the dialog opens -- you'd just need the total user count alongside it.

**2. Recipients route fetches full records just to count them**

`GET /api/admin/product-updates/recipients` calls `getProductUpdateRecipients()` (which loads id, email, name for every matching user) and then returns `{ count: recipients.length }`. Use `prisma.user.count({ where: { ... } })` instead — there's precedent throughout the codebase (e.g., `cities.ts`, `topics.ts`, `reviews.ts`). You could add a `getProductUpdateRecipientCount()` function next to the existing one, and have it also return the total user count for the confirmation prompt above.

**3. Sending will time out for our recipient list — use Resend's batch API**

We expect 1,000+ recipients. At 500ms between individual sends, that's ~8+ minutes. Our DigitalOcean App Platform timeout is 600s (10 min), so it would barely fit — but it ties up a server process the entire time, leaves the admin staring at a spinner with no progress feedback, and any hiccup (slow Resend response, network blip) pushes it past the limit. There's also no deduplication guard, so a retry sends duplicates.

The good news: Resend has a [batch endpoint](https://resend.com/docs/api-reference/emails/send-batch-emails) (`POST /emails/batch`) that sends up to 100 emails per call. 1,000 recipients = 10 batch calls at Resend's 2 req/sec limit = ~5 seconds total. That fits comfortably within a normal API route timeout and eliminates the need for background task infrastructure.

The batch API also supports:
- **Permissive mode**: partial success — returns per-email errors instead of failing the whole batch
- **Idempotency keys**: prevents duplicate sends on retry (retained 24h)

The 500ms sleep between sends is borrowed from `deliver.ts` (which sends one-at-a-time for meeting notifications). With batch sends it's unnecessary — Resend handles rate limiting internally within each batch call.

Suggested approach:
1. Pre-render all emails (per-recipient placeholder substitution + template render)
2. Chunk into batches of 100
3. Send each batch via `POST /emails/batch` with an idempotency key
4. Aggregate results across batches and return to the admin

### Follow-up (not blocking merge if the above are addressed)

**4. Server-side HTML sanitization**

I know you tried this and hit delays — that's almost certainly `isomorphic-dompurify` / `jsdom` overhead (it spins up a full browser DOM simulation, ~2MB, with cold-start cost and memory growth). A lightweight alternative is [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) — it's a string-based parser, no DOM needed, and supports the same tag/attribute allowlists you already have in `SANITIZE_CONFIG`. Would close the gap where a compromised admin session could bypass client-side DOMPurify and inject arbitrary HTML into emails.

src/components/admin/product-updates/SendProductUpdateDialog.tsx
```diff
@@ -0,0 +1,263 @@
...
+        if (!next) {
+            setSubject('');
+            setTestEmail('');
+            reset();
+        }
+    };
+
+    const send = async (
+        extra: Record<string, unknown>,
+        nextStatus: Exclude<Status, 'idle' | 'success' | 'error'>,
```
kouloumos: nit: This type includes `'confirming-all'` which should never be passed to `send()`. Tighten to `'sending-test' | 'sending-all'`.

src/lib/email/resend.ts
```diff
@@ -78,3 +113,84 @@ export async function sendEmail(params: EmailParams) {
...
+            const errText = await response.text();
+            console.error(`Resend batch send failed (${response.status}):`, errText);
+            return {
+                success: false,
+                failedTos: items.map((i) => i.to),
+                error: `Resend batch returned ${response.status}`,
+            };
+        }
+
+        // Even on HTTP 200, Resend's batch response can carry per-item errors
```
kouloumos: The code checks for per-item errors in `data` items, but Resend's strict mode (the default, which is what we're using) is all-or-nothing -- if any email is invalid the entire batch fails with an HTTP error. The per-item error check on `data` items will never trigger.

Switch to **permissive mode** so valid emails in a batch still get sent even if one is invalid. The per-item errors then come in a separate `errors` array (with `index` + `message` fields), not as an `error` field inside `data` -- so the parsing needs updating too. See top-level comment for the response shape.

kouloumos (commented):
Previous feedback addressed -- confirmation step, `prisma.user.count()`, and Resend batch API all look good. The batch implementation (chunking, idempotency keys, dev override) is well structured.

One thing to address before merge: the batch endpoint is running in Resend's default **strict mode**, which means if a single email in a batch is invalid, the entire batch is rejected. With 1,000 recipients split into 10 batches of 100, one bad address silently takes out 99 valid recipients in the same batch -- they never get the email, and the admin just sees "failed: 100" with no way to tell which address caused it.

Resend supports [**permissive mode**](https://resend.com/docs/api-reference/emails/send-batch-emails) for exactly this -- valid emails get sent, invalid ones are reported individually. The response shape changes to include an `errors` array alongside `data`:

```json
{
  "data": [{ "id": "..." }, { "id": "..." }],
  "errors": [{ "index": 2, "message": "The `to` field is missing." }]
}
```

This would require:
1. Adding a batch validation header (or SDK option) to enable permissive mode
2. Parsing `body.errors` (indexed by position) instead of checking for `error` inside `data` items -- the current parsing logic targets the wrong shape.

Since we're already hitting the batch API directly, worth adding [Resend tags](https://resend.com/docs/dashboard/emails/tags) to the batch items. Tags are key/value pairs (up to 75 per email) that show up in the Resend dashboard and webhook events -- they let you filter, group, and associate emails with your application context. Right now all our emails (auth, notifications, product updates) are indistinguishable in the dashboard. This is a good opportunity to start categorizing them.

The rest looks good -- left a couple of nits inline. Make sure to test the batch sending end-to-end since the switch from sequential `sendEmail` to raw `fetch` against Resend's batch endpoint is a significant change.

VasiaKoum:
The last changes:
- Audience: include all users (super-admins are no longer excluded). Only filter is allowProductUpdates: false.
- Batch send: use Resend's batch API in permissive mode, so a single invalid address no longer fails the rest of the batch.
- Tags: every batch email carries a default category=product-update tag; the admin can add extra tags from the dialog. Not very useful in the Resend dashboard today, but ready to filter via the API.
- Reply-to: outgoing product-update emails set replyTo: hello@opencouncil.gr, so any recipient that hits Reply lands in a monitored inbox instead of the no-reply notifications sender.
- Dev override: when `DEV_EMAIL_OVERRIDE` is set, sends in dev/preview are redirected from the original recipient to that address.

PR #337, opened 2026-05-02
perf(transcript): collapse re-render and listener fanout in edit mode

christosporios:
## Summary

Two compounding bottlenecks made the transcript editor unusable on long transcripts. On a 9,283-utterance meeting, every keystroke in the inline editor took **506 ms** to paint (INP). This PR brings it to **36 ms** (14×), eliminates the React re-render cascade on edits, and cuts steady-state DOM size by 35%.

### The two root causes

1. **React-context cascade.** A single `CouncilMeetingDataContext` held the whole transcript. Each edit invalidated its value, cascading through `HighlightProvider` / `EditingProvider` / `VideoProvider` and re-rendering ~9K `UtteranceC` components plus ~46K nested provider re-renders. `React.memo` was useless because every Utterance subscribed to that context.
2. **Per-utterance Radix `<ContextMenu>` listener fanout.** Each `<ContextMenu>` instance registers a `document.addEventListener('keydown', …, { capture: true })` (Radix uses it to track keyboard-vs-mouse usage). With 9K menus mounted, every keystroke fired 9K listeners — measured at ~412 ms of pure JS dispatch, the dominant cost in the input pipeline.

### Fixes

**Context architecture (`CouncilMeetingDataContext.tsx`)**
- Split into three contexts: `Actions` (stable mutations, never changes identity), `Meta` (everything except `transcript`), and `Data` (full backward-compatible shape).
- `Utterance` and small action helpers now use `useCouncilMeetingActions()` — they don't re-render on transcript edits.
- `SpeakerSegment` uses `useCouncilMeetingMeta()` — it bails on transcript-only edits but still updates on speaker-tag changes.
- Stable-identity getters (`getPerson`, `getSpeakerTag`, `getSpeakerSegmentById`, …) via ref + render-time sync, so downstream callbacks (`calculateHighlightData`, `extractSelectedSegment`) keep stable identity.

**Memoize provider values (`EditingContext.tsx`, `HighlightContext.tsx`)**
- Wrap the provider `value` in `useMemo`.
- Mirror transcript-derived state (`allUtterances`, `utteranceMap`, `editingHighlight`, `lastClicked*`) into refs so callbacks have `[]` deps and never churn.
- Move `setIsDirty(true)` out of `setEditingHighlight`'s functional updater (avoids StrictMode double-fire).
- Drop `utteranceMap` from `HighlightContext` value (no external consumers; it would invalidate the value on every transcript change).

**Stable VideoProvider wrappers (`VideoProvider.tsx`)**
- `seekTo`, `seekToAndPlay`, `setIsPlaying`, etc. now ref-backed `useCallback` wrappers. Without this, `HighlightProvider` (which consumes `useVideo()` for `currentTime`) saw new function refs every 2 s during playback, churning all of its `useCallback` deps and re-rendering all Utterances.

**Shared context menu (`UtteranceContextMenu.tsx` — new)**
- A single `<DropdownMenu>` at the Transcript level replaces the 9K per-utterance `<ContextMenu>`s.
- Each utterance span carries `data-utterance-id` / `data-segment-id` / `data-start-timestamp`.
- The wrapper captures `contextmenu`, finds the target via `closest('[data-utterance-id]')`, and opens a virtual-anchored menu at the cursor position. Items, selection-on-open, pending-share-after-close, and the "no menu when no actions" guard are all preserved.

### Results

Measured live on `/thessaloniki/apr1_2026/transcript` (9,283 utterances), edit mode active:

| Metric | Before | After |
|---|---|---|
| Keystroke INP (Chrome) | 506 ms | **36 ms** |
| `UtteranceC` commits per save | ~9,283 | ~1 |
| `Context.Provider` commits per save | ~46,446 | ~5 |
| Heavy commits during playback | 1 every 2s (~1.2 s each) | 0 |
| Total DOM nodes | 59,404 | 38,832 (−35%) |
| Per-keystroke `handleKeyDown` calls | 9,282 | 1 |

### File-by-file

- `CouncilMeetingDataContext.tsx` — split into three contexts; ref-backed getters; stable action callbacks.
- `EditingContext.tsx` — memoized value; ref-mirrored state for `[]`-dep callbacks.
- `HighlightContext.tsx` — memoized value; ref-mirrored state; stable callbacks; functional-updater fix; remove `utteranceMap` from public type.
- `VideoProvider.tsx` — stable function wrappers; memoized value object.
- `SpeakerSegment.tsx` — switched to `useCouncilMeetingMeta()` + `useCouncilMeetingActions()`; inline `memoizedData` (the previous `useMemo` would never recompute with stable getter refs).
- `Utterance.tsx` — strip per-utterance `<ContextMenu>` and the now-unused share/highlight handlers; add data-attrs.
- `Transcript.tsx` — wrap segment list in `<UtteranceContextMenu>`.
- `UtteranceContextMenu.tsx` (new) — shared menu component.
- `useSpeakerSegmentEditor.ts` — switch to `useCouncilMeetingActions()`.

## Test plan

- [x] `npx tsc --noEmit` clean on production code (16 pre-existing test-file matcher errors unrelated to this PR).
- [x] `npm test` — 670/670 pass.
- [x] Verified live on the largest transcript:
  - Typing in an utterance is instant (INP ≤ 36 ms).
  - Right-click on an utterance opens the menu with the correct items per mode (highlight/edit/share).
  - System context menu still shows on segment headers / outside utterances (`hasAnyAction` guard preserved).
  - Edit, save, undo, click-to-select, shift-click range select, highlight create/edit/save/preview, video playback time-tracking — all behave as before.
  - Playback no longer triggers cascading re-renders every 2 s.
- [ ] Re-validate on staging with another long transcript before merging.

---

> [!NOTE]
> **Medium Risk**
> Broad performance refactor across multiple React providers/contexts; while behavior is intended to be preserved, changes to context boundaries and video `setIsPlaying` semantics could cause subtle regressions in editing/highlight/playback flows.
> 
> **Overview**
> **Major performance refactor for transcript edit/highlight mode.** Splits `CouncilMeetingDataContext` into separate `Actions`, `Meta`, and full `Data` contexts with ref-backed stable getters and memoized provider values, so components that only mutate or read non-transcript data no longer re-render on every transcript edit.
> 
> Reworks `EditingContext`, `HighlightContext`, and `VideoProvider` to keep callback/value identities stable (via refs + `useMemo`/`useCallback`) and prevent playback or selection changes from invalidating all `Utterance` components.
> 
> Replaces per-utterance Radix context menus with a single shared `UtteranceContextMenu` wrapper at the `Transcript` level (utterances expose `data-*` attrs), and updates `SpeakerSegment`/`Utterance`/`useSpeakerSegmentEditor` to consume the new actions/meta hooks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffb2689ff9d2b2a9307c1437823da731794765c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

Major performance refactor for long transcripts: splits `CouncilMeetingDataContext` into three stable contexts (`Actions`, `Meta`, full `Data`), adds ref-backed getters and memoized provider values across `EditingContext`, `HighlightContext`, and `VideoProvider`, and replaces ~9K per-utterance Radix `<ContextMenu>` instances with a single transcript-level `<UtteranceContextMenu>` — bringing keystroke INP from 506 ms to 36 ms on a 9,283-utterance transcript. The approach is sound and thoroughly tested; all findings are P2.

Safe to merge after final staging validation; all inline findings are P2 style/optimization suggestions.

No P0 or P1 issues found. The context-splitting strategy is architecturally correct, backward-compatible, and comprehensively tested. The two P2 comments (handleContextMenu dep stability and focus-restoration on virtual-trigger close) are non-blocking. Previous P1 threads have been addressed.

src/components/meetings/transcript/UtteranceContextMenu.tsx — new file; focus-restoration on close and handler stability are worth a follow-up.

| Filename | Overview |
|----------|----------|
| src/components/meetings/CouncilMeetingDataContext.tsx | Split into three contexts (Actions, Meta, Data); ref-backed stable getters; auto-prune effect for orphaned speaker tags. Minor: `data` prop in `metaValue` deps is over-broad. |
| src/components/meetings/EditingContext.tsx | Memoized provider value; ref-mirrored state for stable []-dep callbacks; StrictMode double-fire fix. No issues found. |
| src/components/meetings/HighlightContext.tsx | Memoized value; ref-mirrored state; utteranceMap dropped from public type; speakerTags dep added to editingHighlightData (addresses prior thread). |
| src/components/meetings/VideoProvider.tsx | Ref-backed stable wrappers for all playback functions; separate VideoActionsContext for Utterance; memoized value object. Correct. |
| src/components/meetings/transcript/UtteranceContextMenu.tsx | New shared context menu replacing 9K per-utterance Radix menus. Two P2 issues: handleContextMenu dep stability and focus restoration on close. |
| src/components/meetings/transcript/Utterance.tsx | Stripped per-utterance ContextMenu; added data-attrs; uses stable action hooks. Clean. |
| src/components/meetings/transcript/SpeakerSegment.tsx | Switched to useCouncilMeetingMeta + useCouncilMeetingActions; dropped useMemo for displayData (correct). No issues. |
| src/components/meetings/transcript/Transcript.tsx | Wraps segment list in UtteranceContextMenu; IntersectionObserver uses refs. Clean. |
| src/hooks/useSpeakerSegmentEditor.ts | Migrated to useCouncilMeetingActions. Minimal, correct change. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[CouncilMeetingDataProvider] --> B[CouncilMeetingActionsContext stable for provider lifetime]
    A --> C[CouncilMeetingMetaContext updates on speakerTags/highlights]
    A --> D[CouncilMeetingDataContext updates on transcript + meta]
    B --> E[Utterance useCouncilMeetingActions]
    B --> F[SpeakerSegment useCouncilMeetingActions]
    B --> G[UtteranceContextMenu useCouncilMeetingActions]
    C --> F
    D --> H[Transcript useCouncilMeetingData]
    D --> I[HighlightProvider useCouncilMeetingData]
    D --> J[EditingProvider useCouncilMeetingData]
    K[VideoActionsContext stable seekTo/togglePlayPause] --> E
    L[VideoContext updates every 2s during playback] --> I
    H --> M[UtteranceContextMenu wrapper single shared DropdownMenu]
    M --> N[SpeakerSegment x N]
    N --> O[UtteranceC x M React.memo + data-attrs]
```

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
src/components/meetings/transcript/UtteranceContextMenu.tsx:60-79
**`handleContextMenu` recreated on every selection change**

`handleContextMenu` lists `selectedUtteranceIds` as a dep (line 79). Since each call to `toggleSelection` returns a brand-new `Set`, the handler is recreated on every selection toggle and the wrapping `<div>`'s `onContextMenu` prop changes on every re-render of `UtteranceContextMenu`. The dep is only needed for the `selectedUtteranceIds.has(id)` guard — pulling that read into a ref avoids the churn:

```ts
const selectedUtteranceIdsRef = useRef(selectedUtteranceIds);
selectedUtteranceIdsRef.current = selectedUtteranceIds;

const handleContextMenu = useCallback((e: React.MouseEvent) => {
    if (!hasAnyAction) return;
    const span = (e.target as HTMLElement | null)?.closest<HTMLElement>('[data-utterance-id]');
    if (!span) return;
    const id = span.dataset.utteranceId;
    const segmentId = span.dataset.segmentId;
    if (!id || !segmentId) return;
    const startTimestamp = parseFloat(span.dataset.startTimestamp || '0');
    e.preventDefault();
    setTarget({ id, segmentId, startTimestamp, x: e.clientX, y: e.clientY });
    setOpen(true);
    if (!selectedUtteranceIdsRef.current.has(id)) {
        toggleSelection(id, { shift: false, ctrl: false });
    }
}, [hasAnyAction, toggleSelection]);
```

### Issue 2 of 3
src/components/meetings/CouncilMeetingDataContext.tsx:363-393
**`metaValue` dep on the full `data` prop may invalidate unnecessarily**

`data` is in the dep array for `metaValue` (line 382). `data` contains the initial server-side `speakerTags`, `highlights`, and `transcript` even though the live versions are provided separately as state. If a parent re-renders and passes a structurally identical but new-reference `data` object, `metaValue` invalidates, causing all `useCouncilMeetingMeta()` consumers (every `SpeakerSegment`) to re-render. The `data` reference is typically stable, but for defensive correctness consider depending on individual stable fields (`data.people`, `data.parties`, `data.taskStatus`, etc.) rather than the whole object.

### Issue 3 of 3
src/components/meetings/transcript/UtteranceContextMenu.tsx:165-222
**Focus not restored to the originally focused element after menu closes**

Using a zero-size `aria-hidden` div as the virtual trigger means Radix has no real trigger element to return focus to when the `DropdownMenu` closes. After dismissal, focus lands on `<body>`, breaking keyboard navigation flow. The previous per-utterance `<ContextMenu>` anchored to the utterance `<span>` which naturally held focus.

Tracking the source element and refocusing it on close would fix this:

```ts
const triggerElementRef = useRef<HTMLElement | null>(null);
// in handleContextMenu, after finding span:
triggerElementRef.current = span;
// in handleOpenChange:
if (!next) {
    triggerElementRef.current?.focus({ preventScroll: true });
    // ...flush pending share / clear temp selection...
}
```

`````

</details>

<sub>Reviews (4): Last reviewed commit: ["fixup! perf(transcript): collapse re-ren..."](https://github.com/schemalabz/opencouncil/commit/ffb2689ff9d2b2a9307c1437823da731794765c6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30552085)</sub>

### DIFF
--- a/src/components/meetings/CouncilMeetingDataContext.tsx
+++ b/src/components/meetings/CouncilMeetingDataContext.tsx
@@ -1,5 +1,5 @@
 "use client"
-import React, { createContext, useContext, ReactNode, useMemo, useState, useCallback } from 'react';
+import React, { createContext, useContext, ReactNode, useMemo, useState, useCallback, useRef } from 'react';
 import { Party, SpeakerTag, LastModifiedBy } from '@prisma/client';
 import { updateSpeakerTag } from '@/lib/db/speakerTags';
 import { createEmptySpeakerSegmentAfter, createEmptySpeakerSegmentBefore, moveUtterancesToPreviousSegment, moveUtterancesToNextSegment, deleteEmptySpeakerSegment, updateSpeakerSegmentData, EditableSpeakerSegmentData, extractSpeakerSegment, addUtteranceToSegment } from '@/lib/db/speakerSegments';
@@ -10,12 +10,10 @@ import { PersonWithRelations } from '@/lib/db/people';
 import { getPartyFromRoles } from "@/lib/utils";
 import type { HighlightWithUtterances } from '@/lib/db/highlights';
 
-export interface CouncilMeetingDataContext extends MeetingData {
-    getPerson: (id: string) => PersonWithRelations | undefined;
-    getParty: (id: string) => Party | undefined;
-    getSpeakerTag: (id: string) => SpeakerTag | undefined;
-    getSpeakerSegmentCount: (tagId: string) => number;
-    getSpeakerSegmentById: (id: string) => Transcript[number] | undefined;
+// Actions are mutations + getters that don't depend on transcript identity.
+// They are stable references for the lifetime of the provider, so consumers
+// that only need to mutate (e.g. Utterance) never re-render on data changes.
+export interface CouncilMeetingActions {
     updateSpeakerTagPerson: (tagId: string, personId: string | null) => void;
     updateSpeakerTagLabel: (tagId: string, label: string) => void;
     createEmptySegmentAfter: (afterSegmentId: string) => Promise<void>;
@@ -27,15 +25,30 @@ export interface CouncilMeetingDataContext extends MeetingData {
     addUtteranceToSegment: (segmentId: string) => Promise<string>;
     deleteUtterance: (utteranceId: string) => Promise<void>;
     updateUtterance: (segmentId: string, utteranceId: string, updates: Partial<{ text: string; startTimestamp: number; endTimestamp: number; lastModifiedBy: LastModifiedBy | null }>) => void;
-    getPersonsForParty: (partyId: string) => PersonWithRelations[];
-    getHighlight: (highlightId: string) => HighlightWithUtterances | undefined;
     addHighlight: (highlight: HighlightWithUtterances) => void;
     updateHighlight: (highlightId: string, updates: Partial<HighlightWithUtterances>) => void;
     removeHighlight: (highlightId: string) => void;
     extractSpeakerSegment: (segmentId: string, startUtteranceId: string, endUtteranceId: string) => Promise<void>;
 }
 
+export interface CouncilMeetingDataContext extends MeetingData, CouncilMeetingActions {
+    getPerson: (id: string) => PersonWithRelations | undefined;
+    getParty: (id: string) => Party | undefined;
+    getSpeakerTag: (id: string) => SpeakerTag | undefined;
+    getSpeakerSegmentCount: (tagId: string) => number;
+    getSpeakerSegmentById: (id: string) => Transcript[number] | undefined;
+    getPersonsForParty: (partyId: string) => PersonWithRelations[];
+    getHighlight: (highlightId: string) => HighlightWithUtterances | undefined;
+}
+
 const CouncilMeetingDataContext = createContext<CouncilMeetingDataContext | undefined>(undefined);
+const CouncilMeetingActionsContext = createContext<CouncilMeetingActions | undefined>(undefined);
+// "Meta" context = everything in CouncilMeetingDataContext EXCEPT `transcript`
+// and items derived from it. Lets components like SpeakerSegment subscribe to
+// speakerTags / highlights / getters / static data WITHOUT re-rendering on
+// every transcript edit.
+type CouncilMeetingMeta = Omit<CouncilMeetingDataContext, 'transcript'>;
+const CouncilMeetingMetaContext = createContext<CouncilMeetingMeta | undefined>(undefined);
 
 export function CouncilMeetingDataProvider({ children, data }: {
     children: ReactNode,
@@ -50,7 +63,12 @@ export function CouncilMeetingDataProvider({ children, data }: {
     const speakerSegmentsMap = useMemo(() => new Map(transcript.map(segment => [segment.id, segment])), [transcript]);
     const highlightsMap = useMemo(() => new Map(highlights.map(h => [h.id, h])), [highlights]);
 
-    // Helper function to recalculate segment timestamps based on utterances
+    // cityId/meetingId are constant for the page, so we capture them in
+    // closures rather than reading data.meeting.* (which would force `data`
+    // into action callback deps and break their stable identity).
+    const cityId = data.meeting.cityId;
+    const meetingId = data.meeting.id;
+
     const recalculateSegmentTimestamps = useCallback((utterances: Array<{ startTimestamp: number; endTimestamp: number }>) => {
         if (utterances.length === 0) return null;
         const allTimestamps = utterances.flatMap(u => [u.startTimestamp, u.endTimestamp]);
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/components/meetings/CouncilMeetingDataContext.tsx [hunk 1]
```diff
@@ -93,269 +96,317 @@ export function CouncilMeetingDataProvider({ children, data }: {
...
+    const deleteEmptySegment = useCallback(async (segmentId: string) => {
+        await deleteEmptySpeakerSegment(segmentId, cityId);
+        setTranscript(prev => prev.filter(s => s.id !== segmentId));
+        // Bump speakerTags' identity so SpeakerSegment (subscribed to meta
+        // context) re-renders with the updated segment count for any shared
+        // tag. The auto-prune effect below drops the tag entirely if the
+        // deletion leaves it orphaned — that derivation is robust under
+        // batched deletes that a manual ref-check inside the action wouldn't
+        // get right.
+        setSpeakerTags(prev => [...prev]);
```
kouloumos, line 198: Not critical since deletes are rare, but this re-renders all SpeakerSegments even though only one tag's count changed. Since `Transcript.tsx` already has the transcript, you could compute counts there and pass `segmentCount={counts.get(segment.speakerTag.id) || 0}` as a prop -- `React.memo` skips segments whose count didn't change. `getSpeakerSegmentCount` has no other consumers so it can go. The auto-prune effect below already handles tag removal when the last segment is deleted.

src/components/meetings/VideoProvider.tsx
```diff
@@ -433,51 +433,103 @@ export const VideoProvider: React.FC<VideoProviderProps> = ({ children, meeting,
...
+    const stableScrollToUtterance = useCallback((time: number) => scrollToUtteranceRef.current(time), []);
+    const stableHandleSeeked = useCallback(() => handleSeekedRef.current(), []);
+    const stableHandleSeeking = useCallback(() => handleSeekingRef.current(), []);
+    const stableSetIsPlaying = useCallback(async (shouldPlay: boolean) => {
+        if (shouldPlay) {
+            await playVideoRef.current();
+        } else {
+            await pauseVideoRef.current();
+        }
+    }, []);
```
kouloumos, line 476: nit: The ref-then-stable-callback pattern is repeated 10+ times across 3 files. A small helper would tighten this up:

```ts
function useStableCallback<T extends (...args: any[]) => any>(fn: T): T {
    const ref = useRef(fn);
    ref.current = fn;
    return useCallback((...args: any[]) => ref.current(...args), []) as T;
}
```

Then each wrapper collapses to one line: `const stableSeekTo = useStableCallback(seekTo)`. Not blocking.

kouloumos (approved):
Solid work -- the diagnosis is spot-on (context cascade + listener fanout) and the fix architecture is sound. The three-context split, ref-backed getters, and single shared menu are all the right calls. Left inline comments on a couple of non-blocking suggestions.

The `setSpeakerTags(prev => [...prev])` in `deleteEmptySegment` works but it's broader than it needs to be -- it re-renders every SpeakerSegment in the transcript even though only one tag's count changed. Not critical since segment deletion is rare, but worth noting: `Transcript.tsx` already has the full transcript and re-renders on changes, so computing counts there and passing as a number prop (like #318 did) would let `React.memo` skip segments whose count didn't change. Then you can drop the identity bump and let the auto-prune effect handle orphan cleanup on its own.

PR #341, opened 2026-05-06
Enable inbound messaging for unsusbribe by reply

### LINKED ISSUE #323: Enable inbound WhatsApp messaging for unsubscribe-by-reply
### Concept

Our current messaging integration (Bird) is outbound-only — we send WhatsApp/SMS notifications and poll for delivery status, but we cannot receive messages from users. Enabling inbound messaging is a prerequisite for the Consultation Chatbot (#177) and for basic user interactions like replying to unsubscribe.

The first use case is straightforward: a user replies to a WhatsApp notification (e.g., "STOP") and their mobile notification preferences are disabled. This establishes the inbound infrastructure that #177 will build on for two-way conversational flows.

### User Story

As a citizen who receives WhatsApp notifications, I want to reply to a notification message to unsubscribe, so that I can easily opt out without needing to visit the website.

### Context

- Current Bird integration (`src/lib/notifications/bird.ts`) uses direct message sending + delivery status polling — no inbound support
- A thorough provider evaluation comparing Bird, Vonage, Infobip, Sinch, CM.com, and Messente is documented in #177 ([comment](https://github.com/schemalabz/opencouncil/issues/177#issuecomment-4110753026))
- Key finding: Bird supports inbound via its Conversations API, but our integration currently uses the direct message sending API — we'd need to adopt the Conversations API for webhook-based inbound support
- If Bird's inbound support proves insufficient, Vonage is the recommended alternative (pay-as-you-go, unified Messages API, native failover)
- The current integration is well-isolated (`src/lib/notifications/bird.ts`, direct HTTP calls, no SDK) and the database schema is provider-agnostic, so migration would be straightforward if needed
- WhatsApp Business API restricts outbound messages to pre-approved templates. However, when a user replies to one of our messages, a 24-hour window opens during which we can send freeform (non-template) responses. This is relevant because our confirmation reply (e.g., "You've been unsubscribed") doesn't need a template — it falls within this window naturally.
- Notification preferences and unsubscribe infrastructure from #307 provide the backend for actually disabling a user's mobile notifications

### First steps

The provider decision should come first — confirm that Bird's Conversations API meets our inbound needs or decide to migrate. The research is already done in the #177 comment linked above; this is about making the final call and proceeding with implementation.

VasiaKoum:
<h3>Greptile Summary</h3>

This PR adds inbound Bird webhook handling for the "unsubscribe by reply" flow. It introduces a new `Message` table to track both inbound and outbound messages, a complete Bird webhook route (`/api/webhooks/bird`) with HMAC-SHA256 signature verification, two-stage unsubscribe detection (keyword regex + LLM intent verification), an admin Conversations UI with reply forms, and new `sendAndPersistOutbound` lifecycle helper.

- Adds `Message` schema + two migrations, the `extract.ts` parsing layer (well-typed, well-tested), and `route.ts` orchestration for outbound status updates and inbound STOP detection.
- Introduces `sendAndPersistOutbound` which correctly guards the `input.send()` call in a try/catch so a throw marks the row `failed` instead of leaving it `pending` indefinitely.
- The admin conversations page is entirely new, with tabbed test-send dialog and inline reply form; both use `router.refresh()` to re-fetch after each action.

The webhook handler and unsubscribe flow work correctly for the happy path, but a transient error during unsubscribe processing followed by a Bird retry permanently skips the unsubscribe action — the affected user remains subscribed with no indication anything went wrong.

Several interconnected correctness issues affect the core unsubscribe-by-reply feature this PR is meant to deliver: the retry-idempotency problem in route.ts means a single transient DB or LLM error silently neutralises a valid STOP; welcome-message conversations have no notificationDeliveryId so their replies never enter the unsubscribe flow at all; and SMS STOP messages are intentionally excluded but documented only in passing.

src/app/api/webhooks/bird/route.ts (retry/P2002 interaction with handleUnsubscribeIfApplicable), src/app/[locale]/(other)/admin/conversations/actions.ts (non-atomic notification + delivery creation, releaseNotifications success not checked), src/lib/notifications/bird.ts (participantId set to channelId in sendConversationMessage)

| Filename | Overview |
|----------|----------|
| src/app/api/webhooks/bird/route.ts | New webhook handler with HMAC verification, outbound status reconciliation, and two-stage unsubscribe detection. The P2002-tolerant try/catch wraps both persistMessageRow and handleUnsubscribeIfApplicable together, so a transient failure during the unsubscribe step followed by a Bird retry permanently skips the unsubscribe. |
| src/lib/notifications/unsubscribe.ts | Implements keyword regex gate + LLM intent verification for unsubscribes. Correctly distinguishes confirmed/rejected/failed LLM outcomes and sends appropriate replies. |
| src/app/api/webhooks/bird/extract.ts | Well-typed, pure extraction layer for Bird's variable event shapes; all major paths covered by the accompanying test suite. |
| src/lib/notifications/outbound.ts | sendAndPersistOutbound correctly wraps input.send() in try/catch, marking the row failed before rethrowing — the pending-row orphan issue from the earlier codebase is resolved here. |
| src/app/[locale]/(other)/admin/conversations/actions.ts | New server actions for test sends. sendTestBeforeMeetingNotification creates Notification + NotificationDelivery in two separate non-transactional calls (orphan risk on partial failure), and the releaseNotifications result check only inspects result.failed, not result.success. |
| src/lib/notifications/deliver.ts | sendMessageDelivery now routes through sendAndPersistOutbound, correctly linking outbound Message rows to NotificationDelivery for the inbound webhook's lookup. WhatsApp-then-SMS fallback logic uses finalStatus !== 'failed' guard, avoiding the earlier messageId-based double-delivery bug. |
| src/lib/db/conversations.ts | getConversationSummaries now correctly fetches newest-first (desc) and reverses per-conversation arrays for oldest-first display; the orderBy bug mentioned in an earlier review is addressed. |
| src/lib/notifications/bird.ts | sendConversationMessage sets participantId to the channel ID with participantType: 'flow', which may not correctly identify the sender in Bird's Conversations API and is likely to cause API errors. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Bird
    participant Route as /api/webhooks/bird
    participant Extract as extract.ts
    participant DB as Message (DB)
    participant Unsub as unsubscribe.ts
    participant BirdAPI as Bird Conversations API

    Bird->>Route: POST (HMAC-signed)
    Route->>Route: verifyBirdSignature()
    Route->>Extract: extractMessageFields(event)
    Extract-->>Route: "{direction, phone, body, channel, ...}"

    alt outbound status update
        Route->>DB: findUnique(birdMessageId)
        DB-->>Route: existing row
        Route->>DB: update status/failureReason
        Route-->>Bird: 200 OK
    else inbound message
        Route->>DB: findFirst(conversationId, notificationDeliveryId not null)
        DB-->>Route: notificationDeliveryId (or null)
        Route->>DB: message.create (try/catch P2002)
        alt isUnsubscribeMessage AND whatsapp AND notificationDeliveryId
            Route->>Unsub: verifyUnsubscribeIntent (LLM)
            Unsub-->>Route: "confirmed | rejected | failed"
            alt confirmed
                Route->>DB: updateMany notificationPreference
                Route->>BirdAPI: sendConversationMessage (confirmation)
            else failed
                Route->>BirdAPI: sendConversationMessage (retry text)
            end
        end
        Route-->>Bird: 200 OK
    end
```

<details><summary><h3>Comments Outside Diff (11)</h3></summary>

1. `src/app/api/webhooks/bird/route.ts`, line 698-758 ([link](https://github.com/schemalabz/opencouncil/blob/f177f10bde9264f1f0828d697684d83bb6e8b968/src/app/api/webhooks/bird/route.ts#L698-L758)) 

   <a href="#">P1</a> **No unsubscribe action on inbound "STOP" / "UNSUBSCRIBE" replies**

   The PR title is "Enable inbound messaging for unsubscribe by reply," but the webhook handler only persists the incoming message row — it never inspects the message body for opt-out keywords (e.g. `STOP`, `UNSUBSCRIBE`, `CANCEL`) or updates any user/notification-delivery record to halt future sends. Without that step, a user who replies "STOP" will have their message stored but will continue receiving notifications.

   `````markdown
   Path: src/app/api/webhooks/bird/route.ts
   Line: 698-758

   Comment:
   **No unsubscribe action on inbound "STOP" / "UNSUBSCRIBE" replies**

   The PR title is "Enable inbound messaging for unsubscribe by reply," but the webhook handler only persists the incoming message row — it never inspects the message body for opt-out keywords (e.g. `STOP`, `UNSUBSCRIBE`, `CANCEL`) or updates any user/notification-delivery record to halt future sends. Without that step, a user who replies "STOP" will have their message stored but will continue receiving notifications.

   `````
   </details>

2. `src/app/api/webhooks/bird/route.ts`, line 634-696 ([link](https://github.com/schemalabz/opencouncil/blob/f177f10bde9264f1f0828d697684d83bb6e8b968/src/app/api/webhooks/bird/route.ts#L634-L696)) 

   <a href="#">P2</a> **`any` type usage violates the style guide**

   `extractMessageFields(event: any)` and the `let m: any` block inside use `any` explicitly. The project rule is "Never cast to any, and avoid using any as a type in general." Consider typing `event` as `unknown` and narrowing with type guards, or defining a minimal interface for the expected Bird payload shapes.

   **Context Used:** .cursor/rules/general.mdc ([source](https://app.greptile.com/review/custom-context?memory=dc2172bf-e821-4a8e-b26f-cd165ffe6599))
   
   `````markdown
   Path: src/app/api/webhooks/bird/route.ts
   Line: 634-696
   
   Comment:
   **`any` type usage violates the style guide**
   
   `extractMessageFields(event: any)` and the `let m: any` block inside use `any` explicitly. The project rule is "Never cast to any, and avoid using any as a type in general." Consider typing `event` as `unknown` and narrowing with type guards, or defining a minimal interface for the expected Bird payload shapes.
   
   **Context Used:** .cursor/rules/general.mdc ([source](https://app.greptile.com/review/custom-context?memory=dc2172bf-e821-4a8e-b26f-cd165ffe6599))
   
   `````
   </details>
   
   Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

3. `src/lib/notifications/bird.ts` 

   <a href="#">P2</a> Debug `console.log` logs the full API payload on every conversation message send. This will emit phone numbers and message bodies to server logs in production. Remove before merging.

   

   `````markdown
   Path: src/lib/notifications/bird.ts
   Line: ?

   Comment:
   Debug `console.log` logs the full API payload on every conversation message send. This will emit phone numbers and message bodies to server logs in production. Remove before merging.

   

   `````
   </details>

4. `src/app/[locale]/(other)/admin/conversations/actions.ts`, line 301 ([link](https://github.com/schemalabz/opencouncil/blob/016af83ea50640a9794f2ae30ff562df27151018/src/app/[locale]/(other)/admin/conversations/actions.ts#L301)) 

   <a href="#">P1</a> **`revalidatePath` includes route group — page never revalidates**

   `revalidatePath` expects a URL path, not a filesystem path. Route groups like `(other)` are stripped from URLs, so `/[locale]/(other)/admin/conversations` matches nothing and the conversations page is never revalidated. After `sendTestTemplate` or `sendTestBeforeMeetingNotification`, the admin sees stale data and has no feedback that the send succeeded. The same bug appears on line 397 for `sendTestBeforeMeetingNotification`. Compare to `sendTestReply` (line 215) and `sendTestSms` (line 253) which correctly omit the group.

   `````markdown
   Path: src/app/[locale]/(other)/admin/conversations/actions.ts
   Line: 301

   Comment:
   **`revalidatePath` includes route group — page never revalidates**

   `revalidatePath` expects a URL path, not a filesystem path. Route groups like `(other)` are stripped from URLs, so `/[locale]/(other)/admin/conversations` matches nothing and the conversations page is never revalidated. After `sendTestTemplate` or `sendTestBeforeMeetingNotification`, the admin sees stale data and has no feedback that the send succeeded. The same bug appears on line 397 for `sendTestBeforeMeetingNotification`. Compare to `sendTestReply` (line 215) and `sendTestSms` (line 253) which correctly omit the group.

   `````
   </details>

5. `src/lib/notifications/deliver.ts`, line 186-255 ([link](https://github.com/schemalabz/opencouncil/blob/0555d86a9c94122830c2aff2bd7b1f3733a7e5ee/src/lib/notifications/deliver.ts#L186-L255)) 

   <a href="#">P1</a> **WhatsApp success + missing `messageId` silently triggers SMS double-delivery**

   `createOrUpdateConversation` can return `{ success: true, conversationId: "...", messageId: undefined }` — for example when `createConversation` builds `messageId` from `data?.initialMessage?.id ?? data?.lastMessage?.id ?? data?.messageId` and none of those paths are present in Bird's response body. The guard `whatsappResult.success && whatsappResult.messageId` treats this as a failure: the `else` branch logs a misleading "WhatsApp create-conversation failed" message and falls through to `sendSMSMessage`, so the recipient receives the notification twice (once via WhatsApp, once via SMS) and the WhatsApp send is never recorded in the `Message` table.

   `````markdown
   Path: src/lib/notifications/deliver.ts
   Line: 186-255

   Comment:
   **WhatsApp success + missing `messageId` silently triggers SMS double-delivery**

   `createOrUpdateConversation` can return `{ success: true, conversationId: "...", messageId: undefined }` — for example when `createConversation` builds `messageId` from `data?.initialMessage?.id ?? data?.lastMessage?.id ?? data?.messageId` and none of those paths are present in Bird's response body. The guard `whatsappResult.success && whatsappResult.messageId` treats this as a failure: the `else` branch logs a misleading "WhatsApp create-conversation failed" message and falls through to `sendSMSMessage`, so the recipient receives the notification twice (once via WhatsApp, once via SMS) and the WhatsApp send is never recorded in the `Message` table.

   `````
   </details>

6. `src/app/[locale]/(other)/admin/conversations/actions.ts`, line 623-630 ([link](https://github.com/schemalabz/opencouncil/blob/c2357e8a7ab5cd65ee96a49484d090c321becdfa/src/app/[locale]/(other)/admin/conversations/actions.ts#L623-L630)) 

   <a href="#">P1</a> **Silent success when `releaseNotifications` fails at the outer level**

   `releaseNotifications` catches all top-level errors (e.g., `getPendingDeliveries` throws) and returns `{ success: false, failed: 0 }`. Because only `result.failed > 0` is checked here, that outer-failure case falls through to `return { success: true }` — the admin sees a green "sent" confirmation while no message was actually dispatched and no delivery row was ever transitioned from `pending`. `result.success` must also be consulted before returning success.

   `````markdown
   Path: src/app/[locale]/(other)/admin/conversations/actions.ts
   Line: 623-630

   Comment:
   **Silent success when `releaseNotifications` fails at the outer level**

   `releaseNotifications` catches all top-level errors (e.g., `getPendingDeliveries` throws) and returns `{ success: false, failed: 0 }`. Because only `result.failed > 0` is checked here, that outer-failure case falls through to `return { success: true }` — the admin sees a green "sent" confirmation while no message was actually dispatched and no delivery row was ever transitioned from `pending`. `result.success` must also be consulted before returning success.

   `````
   </details>

7. `src/app/[locale]/(other)/admin/conversations/actions.ts`, line 590-621 ([link](https://github.com/schemalabz/opencouncil/blob/c2357e8a7ab5cd65ee96a49484d090c321becdfa/src/app/[locale]/(other)/admin/conversations/actions.ts#L590-L621)) 

   <a href="#">P1</a> **Non-atomic notification + delivery creation leaves a stuck orphan on partial failure**

   The `Notification` and `NotificationDelivery` rows are created in two separate `prisma.create` calls inside the same `try` block but without a transaction. If the `notification.create` succeeds but the `notificationDelivery.create` fails (any error other than the `userId` P2002), the function rethrows — leaving an orphaned `Notification` with no `NotificationDelivery`. On the admin's next retry, `notification.create` hits the unique-constraint violation, the catch recognises it as the `userId` P2002, and returns "A beforeMeeting notification already exists for this user + meeting" — even though no delivery was ever created. The test notification can never be triggered again without manual DB cleanup. Wrapping both creates in a `prisma.$transaction` eliminates the stuck state.

   `````markdown
   Path: src/app/[locale]/(other)/admin/conversations/actions.ts
   Line: 590-621

   Comment:
   **Non-atomic notification + delivery creation leaves a stuck orphan on partial failure**

   The `Notification` and `NotificationDelivery` rows are created in two separate `prisma.create` calls inside the same `try` block but without a transaction. If the `notification.create` succeeds but the `notificationDelivery.create` fails (any error other than the `userId` P2002), the function rethrows — leaving an orphaned `Notification` with no `NotificationDelivery`. On the admin's next retry, `notification.create` hits the unique-constraint violation, the catch recognises it as the `userId` P2002, and returns "A beforeMeeting notification already exists for this user + meeting" — even though no delivery was ever created. The test notification can never be triggered again without manual DB cleanup. Wrapping both creates in a `prisma.$transaction` eliminates the stuck state.

   `````
   </details>

8. `src/app/api/webhooks/bird/route.ts`, line 1319-1394 ([link](https://github.com/schemalabz/opencouncil/blob/9ae6a9d9d116b1c3917eab125cea293fd4d18dcf/src/app/api/webhooks/bird/route.ts#L1319-L1394)) 

   <a href="#">P1</a> **Unsubscribe logic permanently skipped after transient error + Bird retry**

   The `try/catch` block wraps both `prisma.message.create` and the entire unsubscribe pipeline. If `message.create` succeeds but any downstream step throws a transient error (e.g., `findUserByNotificationDeliveryId`, `verifyUnsubscribeIntent`, or `unsubscribeUserPhoneFromAllCities` hits a brief DB timeout), the handler returns 500. Bird retries the event; on retry, `message.create` throws P2002 (row already exists), which the catch block swallows and returns 200 — so the unsubscribe is never processed and the user remains subscribed forever.

   The fix is to separate the two concerns: wrap only `prisma.message.create` in the P2002-tolerant try/catch, then run the unsubscribe detection outside it (with its own error handling that returns 500 on failure so Bird can retry just that step).

   `````markdown
   Path: src/app/api/webhooks/bird/route.ts
   Line: 1319-1394

   Comment:
   **Unsubscribe logic permanently skipped after transient error + Bird retry**

   The `try/catch` block wraps both `prisma.message.create` and the entire unsubscribe pipeline. If `message.create` succeeds but any downstream step throws a transient error (e.g., `findUserByNotificationDeliveryId`, `verifyUnsubscribeIntent`, or `unsubscribeUserPhoneFromAllCities` hits a brief DB timeout), the handler returns 500. Bird retries the event; on retry, `message.create` throws P2002 (row already exists), which the catch block swallows and returns 200 — so the unsubscribe is never processed and the user remains subscribed forever.

   The fix is to separate the two concerns: wrap only `prisma.message.create` in the P2002-tolerant try/catch, then run the unsubscribe detection outside it (with its own error handling that returns 500 on failure so Bird can retry just that step).

   `````
   </details>

9. `src/lib/notifications/deliver.ts`, line 3299-3313 ([link](https://github.com/schemalabz/opencouncil/blob/9ae6a9d9d116b1c3917eab125cea293fd4d18dcf/src/lib/notifications/deliver.ts#L3299-L3313)) 

   <a href="#">P1</a> **Successful SMS with no `messageId` incorrectly marks delivery as `failed`**

   The guard `smsResult.success && smsResult.messageId` causes `persistAndPollOutbound` to be skipped when `smsResult.success` is `true` but `messageId` is `undefined` (e.g., Bird's Messaging API response omits `id`). Control then falls through to the "Both failed" path which calls `updateDeliveryStatus(delivery.id, 'failed', undefined)` — the user receives the SMS but the delivery row is permanently marked failed, and any retry logic upstream will re-send.

   `````markdown
   Path: src/lib/notifications/deliver.ts
   Line: 3299-3313

   Comment:
   **Successful SMS with no `messageId` incorrectly marks delivery as `failed`**

   The guard `smsResult.success && smsResult.messageId` causes `persistAndPollOutbound` to be skipped when `smsResult.success` is `true` but `messageId` is `undefined` (e.g., Bird's Messaging API response omits `id`). Control then falls through to the "Both failed" path which calls `updateDeliveryStatus(delivery.id, 'failed', undefined)` — the user receives the SMS but the delivery row is permanently marked failed, and any retry logic upstream will re-send.

   `````
   </details>

10. `src/app/api/webhooks/bird/route.ts`, line 1307-1342 ([link](https://github.com/schemalabz/opencouncil/blob/179e88cc49c1c41f7eead2b1e13597fe2ff763ae/src/app/api/webhooks/bird/route.ts#L1307-L1342)) 

    <a href="#">P1</a> **Unsubscribe silently skipped for replies to welcome messages**

    The `notificationDeliveryId` lookup only succeeds when the conversation already contains at least one outbound message that has a `notificationDeliveryId` set. Welcome messages are recorded without a delivery link (`notificationDeliveryId: null`) because there is no `NotificationDelivery` row for the welcome flow. A user who receives only a welcome message and replies "STOP" will have `notificationDeliveryId = null`, the unsubscribe guard `notificationDeliveryId && isUnsubscribeMessage(...)` will evaluate to `false`, and the user will not be unsubscribed — despite the PR's goal of "unsubscribe by reply."

    To catch this case, consider falling back to a phone-based user lookup when `notificationDeliveryId` is null: look up the `User` by `fields.phone` directly and proceed with the unsubscribe flow if a matching user exists.

    `````markdown
    Path: src/app/api/webhooks/bird/route.ts
    Line: 1307-1342

    Comment:
    **Unsubscribe silently skipped for replies to welcome messages**

    The `notificationDeliveryId` lookup only succeeds when the conversation already contains at least one outbound message that has a `notificationDeliveryId` set. Welcome messages are recorded without a delivery link (`notificationDeliveryId: null`) because there is no `NotificationDelivery` row for the welcome flow. A user who receives only a welcome message and replies "STOP" will have `notificationDeliveryId = null`, the unsubscribe guard `notificationDeliveryId && isUnsubscribeMessage(...)` will evaluate to `false`, and the user will not be unsubscribed — despite the PR's goal of "unsubscribe by reply."

    To catch this case, consider falling back to a phone-based user lookup when `notificationDeliveryId` is null: look up the `User` by `fields.phone` directly and proceed with the unsubscribe flow if a matching user exists.

    `````
    </details>

11. `src/app/[locale]/(other)/admin/conversations/actions.ts`, line 617-622 ([link](https://github.com/schemalabz/opencouncil/blob/07b3ecdd5be50f877f0d12fbfd443c783b8d2d0f/src/app/[locale]/(other)/admin/conversations/actions.ts#L617-L622)) 

    <a href="#">P1</a> **`result.success` not checked — outer `releaseNotifications` failure returns silent success**

    `releaseNotifications` catches top-level errors (e.g. `getPendingDeliveries` throws) and returns `{ success: false, failed: 0 }`. Because only `result.failed > 0` is checked here, that outer-failure case falls through to `return { success: true }` — the admin sees a green confirmation while no message was dispatched and no delivery row was ever transitioned from `pending`. Add `!result.success` to the failure guard.

    

    `````markdown
    Path: src/app/[locale]/(other)/admin/conversations/actions.ts
    Line: 617-622

    Comment:
    **`result.success` not checked — outer `releaseNotifications` failure returns silent success**

    `releaseNotifications` catches top-level errors (e.g. `getPendingDeliveries` throws) and returns `{ success: false, failed: 0 }`. Because only `result.failed > 0` is checked here, that outer-failure case falls through to `return { success: true }` — the admin sees a green confirmation while no message was dispatched and no delivery row was ever transitioned from `pending`. Add `!result.success` to the failure guard.

    

    `````
    </details>
</details>

`````markdown

---

### Issue 1 of 1
src/app/[locale]/(other)/admin/conversations/actions.ts:244-245
The `catch (error: any)` annotation violates the project style guide ("Never cast to any"). `error` is `unknown` by default in TypeScript 4.0+ catch clauses, and Prisma exposes a proper typed error class that can be used with `instanceof` or a targeted cast instead.

```suggestion
    } catch (error) {
        const prismaError = error as Prisma.PrismaClientKnownRequestError | undefined;
        if (prismaError?.code === 'P2002' && prismaError?.meta?.target?.includes('userId')) {
```


`````

</details>

<sub>Reviews (15): Last reviewed commit: ["test(notifications): add unit tests for ..."](https://github.com/schemalabz/opencouncil/commit/867daf552abe86e838e23b92406d4a68b6e43143) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30965110)</sub>

**Context used:**

- Context used - .cursor/rules/general.mdc ([source](https://app.greptile.com/review/custom-context?memory=dc2172bf-e821-4a8e-b26f-cd165ffe6599))

### DIFF
--- a/docs/bird-setup.md
+++ b/docs/bird-setup.md
@@ -1,0 +1,178 @@
+# Bird Messaging Setup (WhatsApp + SMS)
+
+This guide walks you through configuring [Bird](https://bird.com) for OpenCouncil's notification system: **outbound** WhatsApp/SMS (meeting reminders, welcome messages) and **inbound** WhatsApp replies (the unsubscribe-by-reply flow). By the end you'll have all the `BIRD_*` environment variables filled in and a local ngrok tunnel that lets Bird POST inbound events back to your machine.
+
+For the canonical list of variables and their default values, see [`.env.example`](../.env.example) and [`environment-variables.md`](./environment-variables.md).
+
+## Prerequisites
+
+- A Bird account — sign up at [app.bird.com](https://app.bird.com/sign-up).
+- A WhatsApp Business account (Bird's onboarding flow walks you through linking one). For local development, Bird's sandbox WhatsApp number is enough.
+- [`openssl`](https://www.openssl.org/) on your machine (for generating the webhook secret).
+- [`ngrok`](https://ngrok.com/download) (only needed for testing inbound locally).
+
+## Step 1: Create a Bird workspace
+
+1. Sign in at [app.bird.com](https://app.bird.com). On first sign-in Bird creates a workspace for you; otherwise create one from the workspace switcher in the top-left.
+2. Once inside the workspace, copy the **workspace UUID** from the URL — it's the segment after `/workspaces/`:
+
+   ```
+   https://app.bird.com/workspaces/<workspace-uuid>/...
+                                    ^^^^^^^^^^^^^^^^
+                                    BIRD_WORKSPACE_ID
+   ```
+
+   See [Bird — How to find a workspace ID](https://docs.bird.com/applications/settings/account/organization-settings/how-to-find-a-workspace-id) if you need help locating it.
+
+## Step 2: Set up the WhatsApp channel
+
+The channel is what Bird uses to deliver WhatsApp messages and receive inbound replies.
+
+1. In the workspace sidebar go to **Channels** → **Add channel** → choose **WhatsApp**. (Bird's [supported channels](https://docs.bird.com/api/channels-api/supported-channels) page lists every option.)
+2. Follow the WhatsApp Business onboarding (link a phone number, verify ownership). For a local-only setup, use Bird's sandbox WhatsApp number — no business verification required.
+3. Once the channel is active, open it from **Channels** and copy the **channel ID** from the URL — it's the UUID at the end:
+
+   ```
+   https://app.bird.com/workspaces/.../channels/whatsapp/<channel-uuid>
+                                                          ^^^^^^^^^^^^^^
+                                                          BIRD_WHATSAPP_CHANNEL_ID
+   ```
+
+## Step 3 (optional): Set up the SMS channel
+
+Repeat Step 2 picking **SMS** instead. Only needed if you want SMS fallback for users without WhatsApp.
+
+> ⚠️ **Inbound SMS is not supported by OpenCouncil.** Bird does support it, but it requires extra account setup (see [Receiving inbound SMS](https://docs.bird.com/connectivity-platform/receiving-sms/setting-your-account-up-to-receive-inbound-sms)) and OpenCouncil's webhook handler currently only routes inbound WhatsApp through the unsubscribe flow. SMS in OpenCouncil is outbound-only.
+
+Copy the channel ID into `BIRD_SMS_CHANNEL_ID`.
+
+## Step 4: Create the WhatsApp templates
+
+WhatsApp Business restricts outbound messages outside a 24-hour reply window to **pre-approved templates**. OpenCouncil uses three:
+
+| Env var | Used for | Variables it must accept |
+|---|---|---|
+| `BIRD_WHATSAPP_TEMPLATE_WELCOME` | First message after a user signs up | user name, city name |
+| `BIRD_WHATSAPP_TEMPLATE_BEFORE_MEETING` | Notification sent before a meeting | meeting title, date, link |
+| `BIRD_WHATSAPP_TEMPLATE_AFTER_MEETING` | Notification sent after a meeting | meeting title, link |
+
+For each template:
+
+1. Open **Bird Studio** (sidebar → **Studio**) and create a new WhatsApp template. See [Bird — WhatsApp templates](https://docs.bird.com/api/channels-api/supported-channels/programmable-whatsapp) for syntax and approval requirements.
+2. Submit it for WhatsApp approval (usually a few minutes).
+3. Once approved, copy the **template project ID (UUID)** from the URL or the template detail panel.
+4. Map each UUID into its env var (`BIRD_WHATSAPP_TEMPLATE_*`).
+
+> The actual template wording is yours to write — but match the variable list above so the substitution code in `src/lib/notifications/bird.ts` doesn't break.
+
+## Step 5: Generate an API key
+
+1. In the workspace sidebar go to **Settings** → **Developers** → **API access** (or **Access keys**, depending on Bird's UI version).
+2. Click **Create access key**, give it a descriptive name (e.g. `opencouncil-local`), and grant it the **Conversations** and **Channels** scopes.
+3. **Copy the key value immediately** — Bird only shows it once. Save it as `BIRD_API_KEY` in `.env`.
+
+## Step 6: Generate the webhook signing secret
+
+This is the shared secret Bird uses to sign inbound webhook events, so the `/api/webhooks/bird` handler can verify they actually came from Bird (not a forged request to your public URL).
+
+Generate 32 random bytes, base64-encoded:
+
+```sh
+openssl rand -base64 32
+```
+
+Save the **same value** in two places:
+
+- Locally as `BIRD_WEBHOOK_SECRET` in `.env`.
+- On Bird's side as the `signingKey` of the webhook subscription you'll create in [Step 8](#step-8-register-the-webhook-in-bird).
+
+> **Treat it like a password.** Anyone with this value can forge requests that look like they came from Bird. Don't commit it; generate a different one per environment (dev, staging, prod) so a leaked dev key doesn't compromise prod. The HMAC scheme itself (HMAC-SHA256 over `timestamp \n url \n sha256(body)`, base64-encoded) is documented in [Bird — Verifying a webhook subscription](https://docs.bird.com/api/notifications-api/api-reference/webhook-subscriptions/verifying-a-webhook-subscription).
+
+## Step 7: Fill in `.env`
+
+After Steps 1–6 you should have all eight variables. Add them to your `.env`:
+
+```bash
+# Bird API for WhatsApp/SMS notifications
+BIRD_WORKSPACE_ID=<workspace-uuid>
+BIRD_API_KEY=<your-api-key>
+BIRD_WHATSAPP_CHANNEL_ID=<whatsapp-channel-uuid>
+BIRD_SMS_CHANNEL_ID=<optional>
+BIRD_WHATSAPP_TEMPLATE_WELCOME=<uuid>
+BIRD_WHATSAPP_TEMPLATE_BEFORE_MEETING=<uuid>
+BIRD_WHATSAPP_TEMPLATE_AFTER_MEETING=<uuid>
+BIRD_WEBHOOK_SECRET=<openssl-output-from-step-6>
+```
+
+At this point **outbound** sends will work — you can use the admin `/conversations` page to fire test WhatsApp messages and they'll reach a real phone. **Inbound** still requires the next two steps.
+
+## Step 8: Expose the webhook locally with ngrok
+
+Bird needs a publicly reachable URL to POST inbound events to. In production that's `https://opencouncil.gr/api/webhooks/bird`; locally we tunnel with ngrok.
+
+1. Start the dev server:
+
+   ```sh
+   npm run dev
+   ```
+
+2. In a separate terminal, start an ngrok tunnel pointing at your local Next.js port (default `3000`):
+
+   ```sh
+   ngrok http 3000
+   ```
+
+3. ngrok prints a forwarding URL like:
+
+   ```
+   Forwarding  <ngrok-url> -> http://localhost:3000
+   ```
+
+   Your webhook URL is that forwarding URL plus `/api/webhooks/bird`:
+
+   ```
+   <ngrok-url>/api/webhooks/bird
+   ```
+
+> Free ngrok URLs change every restart. If you restart ngrok, update the webhook subscription URL in Bird (Step 9) — otherwise Bird will keep POSTing to a dead tunnel.
+
+## Step 9: Register the webhook in Bird
+
+1. In the workspace sidebar go to **Developers** → **Webhooks** → **New webhook subscription** (Bird's UI may also call this "Event subscription").
+2. Fill in:
+
+   | Field | Value |
+   |---|---|
+   | **URL** | The ngrok URL from Step 8, e.g. `<ngrok-url>/api/webhooks/bird` |
+   | **Signing key** | The same string you put in `BIRD_WEBHOOK_SECRET` (Step 6) |
+   | **Service** | `Conversations` |
+   | **Events** | `conversation.created`, `conversation.updated` |
+
+3. Save. Bird will start POSTing matching events to your tunnel.
+
+> The two events together cover the inbound path: `conversation.created` fires when a contact replies to one of your messages for the first time, `conversation.updated` fires for every subsequent message in that thread (which is where the inbound WhatsApp body and unsubscribe-detection logic actually run).
+
+## Step 10: Verify the full inbound path
+
+1. Open `http://localhost:3000/admin/conversations` (you need to be logged in as an admin).
+2. Click **Send test message** and pick one of the templates (e.g. welcome). Send it to a phone number you control that has WhatsApp installed.
+3. On that phone, reply to the message — try `STOP` to exercise the unsubscribe flow.
+4. Watch your dev server logs. A successful inbound looks like:
+
+   ```
+   Bird webhook: disabled phone notifications for user <id> across all cities (delivery <id>)
+   ```
+
+   Refresh `/admin/conversations` and the inbound message row should appear in the thread.
+
+If signature verification fails you'll see a warning like:
+
+```
+Bird webhook: signature verification failed — signature mismatch
+```
+
+The most common causes are:
+
+- The signing key in Bird's webhook subscription doesn't match `BIRD_WEBHOOK_SECRET` in `.env` (re-paste both).
+- ngrok was restarted and the URL on the Bird subscription is stale (update it).
+- Your `.env` was loaded before you set `BIRD_WEBHOOK_SECRET` — restart `npm run dev`.
--- a/src/app/[locale]/(other)/admin/conversations/actions.ts
+++ b/src/app/[locale]/(other)/admin/conversations/actions.ts
@@ -1,0 +1,260 @@
+"use server";
+
+import { withUserAuthorizedToEdit } from '@/lib/auth';
+import prisma from '@/lib/db/prisma';
+import {
+    sendConversationMessage,
+    createOrUpdateConversation,
+    sendSMSMessage,
+} from '@/lib/notifications/bird';
+import { sendAndPersistOutbound } from '@/lib/notifications/outbound';
+import { generateSmsContent, generateWelcomeSmsContent } from '@/lib/notifications/content';
+import { releaseNotifications } from '@/lib/notifications/deliver';
+import type { OutboundSendResult } from '@/lib/notifications/types';
+import type { MessageChannel } from '@prisma/client';
+
+export type SendReplyResult = Pick<OutboundSendResult, 'success' | 'error'>;
+
+/**
+ * Send a free-form text reply into an existing Bird conversation via the
+ * Conversations API, persisting the local Message row through the lifecycle
+ * (pending → sent / failed → polled terminal status).
+ */
+export async function sendTestReply(input: {
+    conversationId: string;
+    phone: string;
+    text: string;
+    channel: MessageChannel;
+}): Promise<SendReplyResult> {
+    await withUserAuthorizedToEdit({});
+
+    const text = input.text.trim();
+    if (!text) return { success: false, error: 'Message body is required' };
+    if (!input.conversationId) return { success: false, error: 'Conversation ID is required' };
+    if (!input.phone) return { success: false, error: 'Phone number is required' };
+
+    const result = await sendAndPersistOutbound({
+        channel: input.channel,
+        phone: input.phone,
+        body: text,
+        conversationId: input.conversationId,
+        send: () => sendConversationMessage({
+            conversationId: input.conversationId,
+            channel: input.channel,
+            text,
+            recipientPhone: input.phone,
+        }),
+    });
+
+    if (result.success && result.finalStatus !== 'failed') {
+        return { success: true };
+    }
+    return { success: false, error: result.finalReason ?? result.error };
+}
+
+/**
+ * Send a free-form SMS test to a phone via the Messaging API.
+ * Persists an outbound Message row in the same shape as `sendTestTemplate`.
+ */
+export async function sendTestSms(input: {
+    phone: string;
+    body: string;
+}): Promise<SendReplyResult> {
+    await withUserAuthorizedToEdit({});
+
+    const phone = input.phone.trim();
+    const body = input.body.trim();
+    if (!phone) return { success: false, error: 'Phone number is required' };
+    if (!body) return { success: false, error: 'Message body is required' };
+
+    const result = await sendAndPersistOutbound({
+        channel: 'sms',
+        phone,
+        body,
+        send: () => sendSMSMessage(phone, body),
+    });
+
+    if (result.success && result.finalStatus !== 'failed') {
+        return { success: true };
+    }
+    return { success: false, error: result.finalReason ?? result.error };
+}
+
+/**
+ * Send a WhatsApp welcome template to a phone via the Messaging API. Used by
+ * the admin "Send test message" dialog on `/admin/conversations` to seed a
+ * conversation for testing — kicks off a thread the user can reply to.
+ */
+export async function sendTestTemplate(input: {
+    phone: string;
+    userName?: string;
+    cityName?: string;
+}): Promise<SendReplyResult> {
+    await withUserAuthorizedToEdit({});
+
+    const phone = input.phone.trim();
+    if (!phone) return { success: false, error: 'Phone number is required' };
+    const userName = input.userName?.trim() || 'Friend';
+    const cityName = input.cityName?.trim() || 'Athens';
+
+    // Route through createOrUpdateConversation so the welcome message lands in
+    // the same Bird thread we reuse for later notifications.
+    const result = await sendAndPersistOutbound({
+        channel: 'whatsapp',
+        phone,
+        body: '[welcome template]',
+        send: () => createOrUpdateConversation({
+            phone,
+            notificationType: 'welcome',
+            params: { userName, cityName },
+        }),
+    });
+    if (result.success && result.finalStatus !== 'failed') {
+        return { success: true };
+    }
+
+    // WhatsApp failed — fall back to SMS. Either the send failed, or
+    // reconciliation flagged it failed post-send (24h window, blocked
+    // recipient, etc.).
+    const waError = result.finalReason ?? result.error;
+    const smsBody = await generateWelcomeSmsContent(userName, cityName);
+    const smsResult = await sendAndPersistOutbound({
+        channel: 'sms',
+        phone,
+        body: smsBody,
+        send: () => sendSMSMessage(phone, smsBody),
+    });
+    if (smsResult.success && smsResult.finalStatus !== 'failed') {
+        return { success: true };
+    }
+    const smsError = smsResult.finalReason ?? smsResult.error;
+    return { success: false, error: `WhatsApp: ${waError}; SMS: ${smsError}` };
+}
+
+// ---------------------------------------------------------------------------
+// Admin "Before-meeting" test send. Creates a real Notification +
+// NotificationDelivery row for a chosen user/city/meeting (mirroring the prod
+// flow that `createNotificationsForMeeting` runs after the agenda task), then
+// calls `releaseNotifications` so it goes through the same conversation +
+// message-row pipeline as a real notification.
+// ---------------------------------------------------------------------------
+
+export interface CityOption { id: string; name: string }
+export interface MeetingOption { id: string; name: string; dateTime: string }
+
+export async function listCitiesForTest(): Promise<CityOption[]> {
+    await withUserAuthorizedToEdit({});
+    const cities = await prisma.city.findMany({
+        select: { id: true, name: true },
+        orderBy: { name: 'asc' },
+    });
+    return cities;
+}
+
+export async function listMeetingsForTest(cityId: string): Promise<MeetingOption[]> {
+    await withUserAuthorizedToEdit({});
+    if (!cityId) return [];
+    const meetings = await prisma.councilMeeting.findMany({
+        where: { cityId },
+        select: { id: true, name: true, dateTime: true },
+        orderBy: { dateTime: 'desc' },
+        take: 20,
+    });
+    return meetings.map((m) => ({ ...m, dateTime: m.dateTime.toISOString() }));
+}
+
+export async function sendTestBeforeMeetingNotification(input: {
+    phone: string;
+    cityId: string;
+    meetingId: string;
+}): Promise<SendReplyResult> {
+    await withUserAuthorizedToEdit({});
+
+    const phone = input.phone.trim();
+    if (!phone) return { success: false, error: 'Phone number is required' };
+    if (!input.cityId) return { success: false, error: 'City is required' };
+    if (!input.meetingId) return { success: false, error: 'Meeting is required' };
+
+    const user = await prisma.user.findFirst({ where: { phone } });
+    if (!user) {
+        return { success: false, error: `No user found with phone ${phone}` };
+    }
+
+    const firstSubject = await prisma.subject.findFirst({
+        where: { councilMeetingId: input.meetingId, cityId: input.cityId },
+    });
+    if (!firstSubject) {
+        return { success: false, error: 'Meeting has no subjects to attach' };
+    }
+
+    let notificationId: string;
+    try {
+        // Include the relations `generateSmsContent` needs so we can
+        // pre-render the SMS body inline — same path the production flow
+        // takes in `createNotificationsForMeeting`.
+        const notification = await prisma.notification.create({
+            data: {
+                userId: user.id,
+                cityId: input.cityId,
+                meetingId: input.meetingId,
+                type: 'beforeMeeting',
+                subjects: {
+                    create: [{ subjectId: firstSubject.id, reason: 'generalInterest' }],
+                },
+            },
+            include: {
+                subjects: { include: { subject: { include: { topic: true } } } },
+                meeting: { include: { administrativeBody: true } },
+                city: true,
+            },
+        });
+        notificationId = notification.id;
+
+        const smsBody = await generateSmsContent({
+            id: notification.id,
+            userId: notification.userId,
+            cityId: notification.cityId,
+            type: 'beforeMeeting',
+            subjects: notification.subjects.map((ns) => ({
+                id: ns.subject.id,
+                name: ns.subject.name,
+                description: ns.subject.description,
+                topic: ns.subject.topic
+                    ? { name: ns.subject.topic.name, colorHex: ns.subject.topic.colorHex }
+                    : null,
+            })),
+            meeting: {
+                dateTime: notification.meeting.dateTime,
+                administrativeBody: notification.meeting.administrativeBody
+                    ? { name: notification.meeting.administrativeBody.name }
+                    : null,
+            },
+            city: { name_municipality: notification.city.name_municipality },
+        });
+
+        await prisma.notificationDelivery.create({
+            data: {
+                notificationId: notification.id,
+                medium: 'message',
+                status: 'pending',
+                phone,
+                body: smsBody,
+            },
+        });
+    } catch (error: any) {
+        if (error?.code === 'P2002' && error?.meta?.target?.includes('userId')) {
+            return {
+                success: false,
+                error: 'A beforeMeeting notification already exists for this user + meeting',
+            };
+        }
+        throw error;
+    }
+
+    const result = await releaseNotifications([notificationId]);
+
+    if (result.failed > 0) {
+        return { success: false, error: `Release failed (${result.failed} failed deliveries)` };
+    }
+    return { success: true };
+}
--- a/src/app/[locale]/(other)/admin/conversations/page.tsx
+++ b/src/app/[locale]/(other)/admin/conversations/page.tsx
@@ -1,0 +1,389 @@
+import { Metadata } from 'next';
+import type { Message } from '@prisma/client';
+import { getConversationSummaries } from '@/lib/db/conversations';
+import { formatNumericDateTime } from '@/lib/formatters/time';
+import {
+    Table,
+    TableBody,
+    TableCell,
+    TableHead,
+    TableHeader,
+    TableRow,
+} from '@/components/ui/table';
+import { ExpandableTableRow } from '@/components/ui/expandable-table-row';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+import { Badge } from '@/components/ui/badge';
+import { MessageCircle, MessageSquare, ChevronRight } from 'lucide-react';
+import { FaWhatsapp } from 'react-icons/fa';
+import { ReplyForm } from '@/components/admin/conversations/ReplyForm';
+import { SendTemplateDialog } from '@/components/admin/conversations/SendTemplateDialog';
+import { ScrollableThread } from '@/components/admin/conversations/ScrollableThread';
+
+export const metadata: Metadata = {
+    title: 'Conversations - Admin',
+    description: 'Inbound and outbound message threads, grouped by participant.',
+};
+
+const PREVIEW_MAX = 120;
+
+function truncate(s: string, max = PREVIEW_MAX) {
+    return s.length <= max ? s : `${s.slice(0, max).trimEnd()}…`;
+}
+
+function statusBadgeVariant(status: Message['status']) {
+    return status === 'failed' ? 'destructive' : 'outline';
+}
+
+function directionBadgeVariant(direction: Message['direction']) {
+    return direction === 'inbound' ? 'default' : 'secondary';
+}
+
+/**
+ * WhatsApp keeps its brand mark (in-house green); SMS falls back to a generic
+ * chat-bubble icon. Both carry an aria-label for screen readers since they're
+ * the sole indicator that a "channel" badge would have shown.
+ */
+function ChannelIcon({ channel }: { channel: Message['channel'] }) {
+    if (channel === 'whatsapp') {
+        return <FaWhatsapp className="h-4 w-4 text-[#25D366]" aria-label="WhatsApp" />;
+    }
+    return <MessageSquare className="h-4 w-4 text-muted-foreground" aria-label="SMS" />;
+}
+
+/**
+ * Cells for the visible row. Same layout for plain rows and expandable rows
+ * so columns stay aligned regardless of which renderer the row uses.
+ */
+function ConversationCells({
+    phone,
+    lastMessage,
+    messageCount,
+}: {
+    phone: string;
+    lastMessage: Message;
+    messageCount: number;
+}) {
+    return (
+        <>
+            <TableCell className="font-mono">{phone}</TableCell>
+            <TableCell className="hidden lg:table-cell">
+                <ChannelIcon channel={lastMessage.channel} />
+            </TableCell>
+            <TableCell className="max-w-md break-words">{truncate(lastMessage.body)}</TableCell>
+            <TableCell>
+                <Badge variant={directionBadgeVariant(lastMessage.direction)}>
+                    {lastMessage.direction}
+                </Badge>
+            </TableCell>
+            <TableCell>
+                <Badge variant={statusBadgeVariant(lastMessage.status)}>
+                    {lastMessage.status}
+                </Badge>
+            </TableCell>
+            <TableCell className="hidden lg:table-cell tabular-nums">{messageCount}</TableCell>
+            <TableCell className="text-right tabular-nums text-muted-foreground">
+                {formatNumericDateTime(lastMessage.createdAt)}
+            </TableCell>
+        </>
+    );
+}
+
+/**
+ * Stacked card content for one conversation. Used on mobile in place of a
+ * table row. Layout matches the spec: phone + channel inline at top, body
+ * below, direction + status inline, timestamp at the bottom.
+ */
+function ConversationMobileCardContent({
+    phone,
+    lastMessage,
+    messageCount,
+}: {
+    phone: string;
+    lastMessage: Message;
+    messageCount: number;
+}) {
+    return (
+        <div className="flex flex-col gap-2 text-left">
+            <div className="flex items-center justify-between gap-3">
+                <div className="flex items-center gap-2 min-w-0">
+                    <span className="font-mono text-sm truncate">{phone}</span>
+                    <ChannelIcon channel={lastMessage.channel} />
+                </div>
+                {messageCount > 1 && (
+                    <span className="text-xs text-muted-foreground tabular-nums shrink-0">
+                        {messageCount} messages
+                    </span>
+                )}
+            </div>
+            <p className="text-sm break-words">{truncate(lastMessage.body)}</p>
+            <div className="flex items-center justify-between gap-3 flex-wrap">
+                <div className="flex items-center gap-2">
+                    <Badge variant={directionBadgeVariant(lastMessage.direction)}>
+                        {lastMessage.direction}
+                    </Badge>
+                    <Badge variant={statusBadgeVariant(lastMessage.status)}>
+                        {lastMessage.status}
+                    </Badge>
+                </div>
+            </div>
+            <span className="text-xs text-muted-foreground tabular-nums">
+                {formatNumericDateTime(lastMessage.createdAt)}
+            </span>
+        </div>
+    );
+}
+
+/**
+ * Mobile card wrapper. Multi-message threads use a native `<details>` element
+ * for expansion (oldest-first thread view inside) — no client component or
+ * state needed. Single-message threads render as a plain styled card.
+ */
+function ConversationMobileCard({
+    phone,
+    messages,
+    lastMessage,
+    messageCount,
+}: {
+    phone: string;
+    messages: Message[];
+    lastMessage: Message;
+    messageCount: number;
+}) {
+    const replyTarget = lastMessage.conversationId
+        ? {
+            conversationId: lastMessage.conversationId,
+            phone,
+            channel: lastMessage.channel,
+        }
+        : null;
+    const expandable = messageCount > 1 || replyTarget !== null;
+
+    if (expandable) {
+        return (
+            <details className="group rounded-md border bg-background overflow-hidden">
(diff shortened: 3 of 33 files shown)

### CODE REVIEW

**Comment by VasiaKoum** [comment]
### Implemented

- **Admin UI for testing inbound/outbound** — `/admin/conversations` page lists threads grouped by `(phone, channel)` with a chat-style view. Backed by server actions for sending WhatsApp templates, free-form SMS, and replays of the "before-meeting" notification flow, so the full inbound path can be driven end-to-end without waiting on real users.
- **Inbound message handling via Bird webhook** at `POST /api/webhooks/bird`, with **HMAC signature verification** following Bird's documented scheme:
  - Algorithm: HMAC-SHA256 over `timestamp \n url \n sha256(body)`
  - Signature header: `messagebird-signature` (base64-encoded)
  - Timestamp header: `messagebird-request-timestamp`, with a 5-minute replay window
  - URL is reconstructed from `x-forwarded-host` / `x-forwarded-proto`, so it works behind ngrok in dev and the load balancer in prod
  - Handler fails closed with `500` in production if `BIRD_WEBHOOK_SECRET` is unset
- **Two-stage unsubscribe detection** on inbound WhatsApp:
  1. Regex first-pass for `STOP`, `ΣΤΟΠ`, `unsubscribe`, `διακοπή`, `απεγγραφή`
  2. LLM second-pass (Claude Haiku 4.5) verifies the message in context, so `"stop the meeting"` or `"θα stop περάσω αύριο"` doesn't trigger an accidental unsubscribe
  3. Three-state result: `confirmed` → unsubscribe, `rejected` → no action, `failed` → reply with a retry prompt instead of silently dropping a valid STOP
- **Unsubscribe scope: phone-only, all cities.** A confirmed STOP flips `notifyByPhone=false` on every `NotificationPreference` row for that user. Email preferences and `allowProductUpdates` / `allowPetitionUpdates` are untouched — replying STOP on a phone channel is a phone-channel signal, not a global "leave the platform" signal.
- **Phone-ownership guard** before any preference write: the inbound phone must match the user resolved from the conversation's notification delivery.

### Wasn't Implemented in this PR

- **Inbound SMS.** SMS is sent through Bird's Messaging API (no Bird-side conversation thread), so there's no `conversationId` to walk back to a user. Receiving inbound SMS at all has Bird-side setup requirements — see [Bird docs — Receiving inbound SMS](https://docs.bird.com/connectivity-platform/receiving-sms/setting-your-account-up-to-receive-inbound-sms). We had an issue to register a numerical phone number based in Greece to be able to receive the inbound messages (or we keep the NL number, we couldn't add Greece as destination country)
- **Channel interaction events** (`unsubscribe-request`, `reported-as-spam`) from Bird's [Message Status & Interactions](https://docs.bird.com/api/channels-api/message-status-and-interactions#message-interactions). We currently only consume `conversation.updated`; native interaction events would be a more reliable unsubscribe signal than keyword + LLM.

## Rolling out to staging and production

The webhook code needs no per-env config — `signedUrl` is built from forwarded headers at request time. What does need per-environment setup:

- **One Bird webhook subscription per environment**, pointing at that env's public URL:
  - Staging: `https://staging.opencouncil.gr/api/webhooks/bird`
  - Production: `https://opencouncil.gr/api/webhooks/bird`
- **A unique `BIRD_WEBHOOK_SECRET` per environment** — generate with `openssl rand -base64 32`, set as Bird's `signingKey` on the subscription and as `BIRD_WEBHOOK_SECRET` in the env. Don't reuse keys across envs;
- **Sandbox channels for non-prod.** `BIRD_WHATSAPP_CHANNEL_ID` / `BIRD_SMS_CHANNEL_ID` / `BIRD_WHATSAPP_TEMPLATE_*` should ideally point to a sandbox channel in staging so test sends don't hit real users' phones.
- **Previews.** Per-PR URLs change too often to register Bird subscriptions for each one.

File: docs/bird-setup.md [hunk 1]
```diff
@@ -0,0 +1,197 @@
...
+| Variable | What it is |
+|---|---|
+| `BIRD_WORKSPACE_ID` | UUID of your Bird workspace |
+| `BIRD_API_KEY` | API key scoped to your workspace |
+| `BIRD_WHATSAPP_CHANNEL_ID` | UUID of the WhatsApp channel |
+| `BIRD_SMS_CHANNEL_ID` | UUID of the SMS channel (optional, only if you want SMS fallback) |
+| `BIRD_WHATSAPP_TEMPLATE_WELCOME` | UUID of the welcome template |
+| `BIRD_WHATSAPP_TEMPLATE_BEFORE_MEETING` | UUID of the "before meeting" template |
+| `BIRD_WHATSAPP_TEMPLATE_AFTER_MEETING` | UUID of the "after meeting" template |
+| `BIRD_WEBHOOK_SECRET` | Shared secret used to verify HMAC signatures on inbound webhook events |
```
kouloumos: Let's stick to `docs/environment-variables.md` and `env.example` as the source of truth for env vars. We currently have partial bird setup instructions in `docs/environment-variables.md` -- I suggest removing the instructions from there and keep only the variables listed and point a reference link to the full Bird setup instead.

docs/bird-setup.md
```diff
@@ -0,0 +1,197 @@
...
+
+- The signing key in Bird's webhook subscription doesn't match `BIRD_WEBHOOK_SECRET` in `.env` (re-paste both).
+- ngrok was restarted and the URL on the Bird subscription is stale (update it).
+- Your `.env` was loaded before you set `BIRD_WEBHOOK_SECRET` — restart `npm run dev`.
+
+## Where to go next
+
+- **[notifications.prd.md](./notifications.prd.md)** — Product spec for the notification system this plugs into.
+- **[environment-variables.md](./environment-variables.md)** — Full reference for every env var, including the rest of the OpenCouncil stack.
+- **[admin-alerts.md](./admin-alerts.md)** — Discord alerts for delivery batches and other notification events.
```
kouloumos: remove this, it creates maintenance burden.

src/app/[locale]/(other)/admin/conversations/page.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+const PREVIEW_MAX = 120;
+
+function truncate(s: string, max = PREVIEW_MAX) {
+    return s.length <= max ? s : `${s.slice(0, max).trimEnd()}…`;
+}
+
+function formatTimestamp(d: Date) {
+    return formatNumericDateTime(d);
+}
```
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/page.tsx]: why not use `formatNumericDateTime` directly?

src/lib/notifications/bird.ts
```diff
@@ -133,21 +126,26 @@ async function makeBirdRequest(
         if (!response.ok) {
             const errorText = await response.text();
             console.error(`${logPrefix} API error:`, response.status, errorText);
```
kouloumos, line 149: On non-2xx, the response body is read as text for logging but never parsed or returned. Bird's 409 `ContactAlreadyInConversation` includes the existing `conversationId` in `details.conversationId` — `createConversation` needs that to recover. Parse the error body (at least for JSON responses) and return it so callers can act on structured errors.

src/lib/notifications/bird.ts
```diff
@@ -37,89 +38,81 @@ function sleep(ms: number): Promise<void> {
...
 
-/**
- * Make Bird API request with error handling
- */
 async function makeBirdRequest(
     url: string,
     payload: unknown,
     logPrefix: string
-): Promise<{ success: boolean; error?: string; messageId?: string }> {
+): Promise<{ success: boolean; error?: string; messageId?: string; conversationId?: string; data?: any }> {
```
kouloumos: `any` disables type checking for everything downstream — any caller that touches `result.data` can access arbitrary properties without TypeScript catching typos or wrong assumptions (e.g. `data?.converationId` compiles fine). At minimum use `Record<string, unknown>` so callers are forced to narrow before accessing fields. Even better: since `makeBirdRequest` is called for a few specific endpoints, make it generic (`makeBirdRequest<T>`) and define lightweight interfaces for each response shape (`CreateConversationResponse`, `SendMessageResponse`, etc.) so the data is properly typed at each call site.

src/lib/db/messages.ts
```diff
@@ -0,0 +1,55 @@
...
+    const recent = await prisma.message.findFirst({
+        where: {
+            phone: { in: [formatted, phone] },
+            conversationId: { not: null },
+        },
+        orderBy: { createdAt: 'desc' },
+        select: { conversationId: true },
+    });
+    return recent?.conversationId ?? null;
+}
```
kouloumos: This returns the most recent conversationId for a phone regardless of channel. If a phone has both WhatsApp and SMS messages, `createOrUpdateConversation` (which is WhatsApp-only) could route into an SMS Bird conversation. Accept a `channel` parameter so it works correctly as the system grows.

Also, phone normalization here (querying both `+phone` and `phone`) diverges from `formatPhoneNumber` in `bird.ts` and the inline lambda in `route.ts:214`. Extract a single `normalizePhone` utility.

src/app/api/webhooks/bird/route.ts
```diff
@@ -0,0 +1,285 @@
...
+    phone?: string;
+    body: string;
+    channel: MessageChannel;
+    status: MessageStatus;
+    failureReason?: string;
+} {
+    const eventName = String(event?.event ?? '');
+    const payload = event?.payload ?? event?.data ?? event ?? {};
+
+    let m: any;
```
kouloumos: `extractMessageFields(event: any)` and `let m: any` — project rule is no `any`. Use `unknown` with narrowing, or define a minimal interface for the Bird payload shapes.

src/app/[locale]/(other)/admin/conversations/actions.ts
```diff
@@ -0,0 +1,271 @@
...
+    const { status: finalStatus, reason } = await pollForDeliveryStatus(channelId, birdMessageId);
+    if (!finalStatus) return;
+
+    const mapped = mapBirdMessageStatus(finalStatus);
+    await updateMessage(localMessageId, {
+        status: mapped,
+        // Only persist a reason on failure; clear it on success/in-flight.
+        failureReason: mapped === 'failed' ? reason ?? null : null,
+    });
+}
```
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/actions.ts]: `reconcileAfterSend` (actions.ts:23-40) and `persistAndPollOutbound` (deliver.ts:121-154) do the same resolve-channel -> poll -> map-status -> update-message cycle. Extract to a shared helper in `lib/notifications/`.

src/components/admin/conversations/ScrollableThread.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { useLayoutEffect, useRef, type ReactNode } from 'react';
+
+export function ScrollableThread({ children }: { children: ReactNode }) {
+    const ref = useRef<HTMLOListElement>(null);
+
+    useLayoutEffect(() => {
+        const el = ref.current;
+        if (el) el.scrollTop = el.scrollHeight;
+    }, []);
```
kouloumos: Triggers a React SSR warning on this server-rendered admin page. Scroll-to-bottom doesn't need synchronous layout measurement — `useEffect` works fine here.

src/app/[locale]/(other)/admin/conversations/actions.ts
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/actions.ts:1]: The `revalidatePath` calls are unnecessary.
These server actions are called via `startTransition` from client components (`ReplyForm`, `SendTemplateDialog`). In Next.js App Router, server actions invoked through `startTransition` automatically trigger a re-render of the current route's server components -- the RSC payload gets refreshed without explicit revalidation. Since the conversations page is a dynamic route that fetches data via a direct Prisma query (no `fetch` caching), the query re-executes on re-render and the data is always fresh. Remove all four calls — they're dead code (and inconsistent with each other on top of that).

kouloumos (commented):
I tested locally with the Bird env vars configured. Tried the before-meeting test send from the admin page -- Bird returned a 409 `ContactAlreadyInConversation` because a conversation already exists for my phone (created during your earlier testing). The 409 response includes the existing `conversationId` in `details`, but `makeBirdRequest` treats all non-2xx as generic failures and discards the body, so that conversationId is lost.

This is the main issue with the PR right now: `createOrUpdateConversation` checks our local DB for an existing conversation (`findRecentConversationIdByPhone`), but when the local DB has no record and Bird already has one, we hit a wall. This isn't just a test issue — it'll happen in production whenever Bird's state and our DB diverge (deployments, DB resets, conversations created outside our system). `createConversation` needs to handle the 409 by extracting the conversationId from the error response and retrying via `sendConversationMessage`.

Two more things I noticed while testing this:

1. **Inconsistent fallback behavior.** The before-meeting flow (`releaseNotifications` -> `sendMessageDelivery`) has WA->SMS fallback, so when the 409 hits it falls through to SMS. But the welcome template flow (`sendTestTemplate`) calls `createOrUpdateConversation` directly with no fallback — it just shows the 409 error in the UI. These should be consistent.

2. **SMS fallback sends placeholder text.** When the before-meeting flow fell back to SMS, I received `[beforeMeeting template]` as the message body — that's the literal placeholder stored in `NotificationDelivery.body`, not the actual template content (which Bird assembles server-side from the template parameters). Same issue with `welcome.ts` — the SMS fallback there calls `sendWelcomeSMS` which sends a proper Greek message, but no `Message` row is persisted for it, so it's invisible in the admin conversations view.

Beyond the testing issues: phone normalization is fragile (three different implementations across `bird.ts`, `route.ts`, and `messages.ts` — left inline comment), and there are several `any` type violations. Also, `mapBirdMessageStatus`, `isUnsubscribeMessage`, and `extractMessageFields` are pure functions with zero tests — the first two are trivially testable and correctness-relevant, please add tests for at least those.

One thing to think about for a follow-up: now that the webhook handles outbound status updates (route.ts:145-161), the synchronous `pollForDeliveryStatus` loop (5 attempts x 2s = up to 10s blocking per message) is doing redundant work. In the delivery pipeline this adds up — `releaseNotifications` processes sequentially, so 10 recipients means ~105s of wall time mostly spent sleeping. The polling pre-dates this PR so it's not something to fix here, but the webhook you've built makes it possible to drop the polling and let status reconciliation happen asynchronously. The only thing polling still gates is the WA→SMS fallback decision, which would need to move to a webhook-triggered path.

Left inline comments on the specifics.

VasiaKoum:
## New changes

1. **409 from Bird's create-conversation** is now recovered by extracting the existing `conversationId` from the error body and sending into that conversation via `sendConversationMessage` — fixes the cross-env conflict the reviewer hit and any prod scenario where Bird's state and our DB diverge.
2. **Phone normalization** consolidated into `phone.ts:normalizePhone`, replacing three drifting inline implementations across `bird.ts`, `route.ts`, and `messages.ts`.
3. **Status reconciliation** (resolve-channel → poll → map → update) was duplicated between `actions.ts` and `deliver.ts`; extracted into `reconcile.ts:reconcileMessageStatus` so both delegate.
4. **SMS fallback** now fires only on a genuine WhatsApp failure — the 409 case no longer triggers it (recovered upstream), and a 2xx-with-no-`messageId` response no longer drops to SMS either (was silently double-sending).
5. **SMS bodies** render proper Greek text from the new `sms-templates.ts` instead of the literal placeholder `[beforeMeeting template]` we were sending before; admin `sendTestTemplate` now mirrors this prod fallback flow.
6. **Unsubscribe regex** pre-normalizes input via `stripDiacritics` (NFD + combining-mark strip) and uses `απεγγρα[φψ]` to cover both Greek φ/ψ conjugation stems, so `Απεγγράψτε με` and accented past-tense forms now match.
7. **LLM prompt** restructured with explicit Greek false-positive examples (noun `στοπ` as a road sign, past-tense narration like `Απεγγράφτηκα προχτές και θέλω να ξαναεγγραφώ`) so the verifier correctly rejects ambiguous regex matches.
8. **"Already unsubscribed" reply** — `unsubscribeUserPhoneFromAllCities` returns the count of actually-changed rows, and when zero we send the new `UNSUBSCRIBE_ALREADY_TEXT` (with the same re-subscribe link) instead of a misleading fresh confirmation.
9. **Unit tests** added — `isUnsubscribeMessage` regex coverage, `verifyUnsubscribeIntent` with a mocked LLM (including the outage path), and `mapBirdMessageStatus` across all four output buckets.
10. **Inline-comment cleanups**: dropped all `any` types from the Bird module (`makeBirdRequest<T>` + typed webhook payloads), removed four dead `revalidatePath` calls, swapped `useLayoutEffect` → `useEffect` to fix an SSR warning, deleted two thin wrappers, added a required `channel` parameter to `findRecentConversationIdByPhone`, and consolidated Bird env-var docs into a single source of truth.

src/app/api/webhooks/bird/route.ts
```diff
@@ -0,0 +1,358 @@
...
+    const phone =
+        m?.sender?.contact?.identifierValue ??
+        m?.sender?.contact?.platformAddress ??
+        fromObj?.identifierValue ??
+        fromObj?.phone ??
+        fromStr ??
+        m?.sender?.identifierValue ??
+        m?.contact?.identifierValue ??
+        m?.sender?.phone ??
+        m?.participant?.identifierValue;
```
kouloumos: For outbound `conversation.updated` events, `sender.type` is `"flow"` (us) with no phone. The phone lives in two places the function doesn't check:
- `lastMessage.recipients[0].identifierValue`
- `payload.featuredParticipants` → the participant with `type: "contact"` → `contact.identifierValue`

Add these as fallbacks in the phone extraction chain. Without this, the outbound branch at lines 240-257 (status reconciliation) is dead code — it never executes because the event gets dropped at the `!fields.phone` guard on line 235.

src/app/api/webhooks/bird/route.ts
```diff
@@ -0,0 +1,358 @@
...
+
+    // Idempotent insert (inbound, or outbound where the row didn't exist).
+    try {
+        await prisma.message.create({
+            data: {
+                channel: fields.channel,
+                direction: fields.direction,
+                birdMessageId: fields.birdMessageId,
+                conversationId: fields.conversationId ?? null,
+                phone: fields.phone,
```
kouloumos: `fields.phone` is written to the DB as-is, but every other write path goes through `normalizePhone()`. Bird sends `+306940158931` in practice, but normalizing here defensively keeps the data consistent. Same on lines 321 and 339.

src/app/[locale]/(other)/admin/conversations/actions.ts
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/actions.ts:1]: The same `record→send→update→reconcile lifecycle` cycle repeats across `sendTestReply`, `sendTestSms`, `sendTestTemplate` (twice for the SMS fallback), plus `sendUnsubscribeReply` in `unsubscribe.ts:114-132` and both branches in `welcome.ts:58-82`. Extract a shared helper that owns this: `recordOutboundMessage` → call Bird → `updateMessage` with result → optionally `reconcileMessageStatus`. Each call site just provides the channel, phone, body, and a send function.

src/app/[locale]/(other)/admin/conversations/actions.ts
```diff
@@ -0,0 +1,281 @@
...
+    });
+
+    // Route through createOrUpdateConversation so the welcome message lands in
+    // the same Bird thread we reuse for later notifications.
+    const result = await createOrUpdateConversation({
+        phone,
+        notificationType: 'welcome',
+        params: {
+            userName: input.userName?.trim() || 'Friend',
+            cityName: input.cityName?.trim() || 'Athens',
```
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/actions.ts]: `userName` and `cityName` defaults are computed identically in both places. Extract once at the top of `sendTestTemplate`:
```typescript
const userName = input.userName?.trim() || 'Friend';
const cityName = input.cityName?.trim() || 'Athens';
```
Then pass to both `createOrUpdateConversation` and `sendWelcomeSMS`.

kouloumos (commented):
Nice work on the updates -- the 409 recovery, `normalizePhone` extraction, typed `makeBirdRequest<T>`, `reconcileMessageStatus` dedup, SMS templates, and the tests all address what we discussed. The Bird setup guide is thorough too.

I tested the full flow locally with a real Bird webhook (ngrok tunnel) and it mostly works well. The inbound path + unsubscribe chain is solid: keyword detection, LLM verification, preference update, and confirmation reply all work end-to-end. A couple of things came up during testing that need attention.

**Outbound webhook events can't extract the phone.** `extractMessageFields` looks for the phone on `m.sender.contact.identifierValue` and similar paths, which works for inbound (sender is `type: "contact"`). But for outbound events, the sender is `type: "flow"` (us) — no phone there. The phone is on `lastMessage.recipients[0].identifierValue` and `payload.featuredParticipants[0].contact.identifierValue`, neither of which the function checks. Every outbound `conversation.updated` event hits "missing required fields" and gets dropped. This means outbound message status never gets reconciled via the webhook — statuses stay at whatever the polling caught (often `sent` even after delivery). Left an inline comment with the specific paths to check.

**Page doesn't refresh after sending from the admin UI.** The `revalidatePath` calls were removed (which I asked for), but the page doesn't actually re-render after a server action in this case. After sending a reply or test message, the success indicator shows but the conversation list stays stale until manual refresh. Add `revalidatePath('/admin/conversations', 'page')` back in each action, or call `router.refresh()` from the client components after the action resolves.

**Unsubscribe detection is keyword-only, not semantic.** The regex + LLM two-layer approach works well, but the LLM only runs if the regex matches first. A message like "δεν θέλω άλλα μηνύματα" won't trigger unsubscribe because there's no keyword match — even though the LLM prompt lists that exact phrase as an example of unsubscribe intent. This is fine as a deliberate conservative choice (running the LLM on every inbound message would add cost and latency), but add a comment on the regex explaining that semantic-only messages are intentionally not caught and why.

**The record→send→update→reconcile cycle is repeated in 7 places** across `actions.ts`, `unsubscribe.ts`, `welcome.ts`, and `deliver.ts`. Each does: `recordOutboundMessage` → send via Bird → `updateMessage` with status → optionally `reconcileMessageStatus`. The parameters differ but the lifecycle is identical. A shared helper (e.g. `sendAndPersistOutbound({ channel, phone, body, send })`) that owns this cycle would cut significant duplication. Left an inline comment on the first occurrence.

**Confirmation message copy.** Minor: "ξαναενεργοποιήσετε" is a bit clunky — "ενεργοποιήσετε ξανά" reads more naturally. Also the "already unsubscribed" message ("Έχετε απεγγραφεί") implies the user previously unsubscribed — for someone who is not currently subscribed, something like "Δεν λαμβάνετε ειδοποιήσεις μέσω τηλεφώνου" would be more accurate.

Left inline comments on the specifics.

src/app/[locale]/(other)/admin/conversations/actions.ts
Comment (VasiaKoum) [reply src/app/[locale]/(other)/admin/conversations/actions.ts:1]: Extracted a shared helper `sendAndPersistOutbound` in a new file `src/lib/notifications/outbound.ts` that owns the entire lifecycle and reuse it.

```
export async function sendAndPersistOutbound(
    input: SendAndPersistInput,
): Promise<SendAndPersistResult> {
    const row = await recordOutboundMessage({ ... });   // row-first

    const result = await input.send();                  // Bird call

    await updateMessage(row.id, {                       // status update
        status: result.success ? 'sent' : 'failed',
        birdMessageId: result.messageId ?? null,
        ...(input.conversationId == null && result.conversationId
            ? { conversationId: result.conversationId }
            : {}),
    });

    let finalStatus: MessageStatus | null | undefined;
    if (input.reconcile !== false && result.success && result.messageId) {
        finalStatus = await reconcileMessageStatus({    // optional poll
            localMessageId: row.id,
            channel: input.channel,
            birdMessageId: result.messageId,
        });
    }

    return { ...result, rowId: row.id, finalStatus };
}
```

src/lib/notifications/bird.ts
```diff
@@ -157,218 +206,348 @@ async function makeBirdRequest(
...
+ * Send a message (text or template) INTO an existing Bird conversation via
+ * `POST /workspaces/{ws}/conversations/{id}/messages`.
+ */
+export async function sendConversationMessage(input: {
+    conversationId: string;
+    channel?: 'whatsapp' | 'sms';
+    text?: string;
+    template?: TemplateBody;
+    recipientPhone?: string;
+}): Promise<{ success: boolean; error?: string; messageId?: string; conversationId?: string }> {
```
kouloumos: `sendConversationMessage` and `sendSMSMessage` both define `Promise<{ success: boolean; error?: string; messageId?: string; conversationId?: string }>` inline — which is exactly `OutboundSendResult` from `outbound.ts`. The interface exists and is already used to type the `send` callback in `sendAndPersistOutbound`, but these functions don't reference it. Use `Promise<OutboundSendResult>` so a shape change is one edit, not four.

Same pattern for `createConversation` (lines 357-362) and `createOrUpdateConversation` (lines 456-461) — both extend the base shape with one boolean flag (`alreadyExisted` / `created`). Define a shared base and extend it:
```typescript
interface ConversationResult extends OutboundSendResult {
    alreadyExisted?: boolean;
    created?: boolean;
}
```

src/lib/notifications/bird.ts
```diff
@@ -10,12 +12,6 @@ interface WhatsAppTemplateParams {
     notificationId: string;
 }
```
kouloumos: `MeetingSmsParams` is `WhatsAppTemplateParams` minus `notificationId`. Both are built from the same `templateParams` object in `deliver.ts`. Define the base once and extend:
```typescript
// shared
interface MeetingNotificationParams { date, cityName, subjectsSummary, adminBody }
// bird.ts
interface WhatsAppTemplateParams extends MeetingNotificationParams { notificationId: string }
```

Same for `WelcomeTemplateParams` (bird.ts:216) and `WelcomeSmsParams` (sms-templates.ts:9) — both are `{ userName: string; cityName: string }`, defined separately.

src/app/[locale]/(other)/admin/conversations/actions.ts
```diff
@@ -0,0 +1,232 @@
...
+} from '@/lib/notifications/bird';
+import { sendAndPersistOutbound } from '@/lib/notifications/outbound';
+import { renderWelcomeSms } from '@/lib/notifications/sms-templates';
+import { releaseNotifications } from '@/lib/notifications/deliver';
+import type { MessageChannel } from '@prisma/client';
+
+export interface SendReplyResult {
+    success: boolean;
+    error?: string;
+}
```
Comment (kouloumos) [inline src/app/[locale]/(other)/admin/conversations/actions.ts]: `SendReplyResult` is `{ success: boolean; error?: string }` — exactly `Pick<OutboundSendResult, 'success' | 'error'>`. Make the relationship explicit so it's obvious they're connected and stays in sync.

src/components/admin/conversations/ReplyForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+import { useState, useTransition } from 'react';
+import { useRouter } from 'next/navigation';
+import type { MessageChannel } from '@prisma/client';
+import { Button } from '@/components/ui/button';
+import { Textarea } from '@/components/ui/textarea';
+import { Loader2, Send, CheckCircle, XCircle } from 'lucide-react';
+import { sendTestReply } from '@/app/[locale]/(other)/admin/conversations/actions';
+
+type Status = 'idle' | 'sending' | 'sent' | 'error';
```
kouloumos: `type Status = 'idle' | 'sending' | 'sent' | 'error'` is identical here and in `SendTemplateDialog.tsx:38`. Extract to a shared location.

kouloumos (approved):
Tested the full flow locally with a real Bird webhook (ngrok tunnel): outbound sends, 409 recovery, inbound persistence, unsubscribe (both already-unsubscribed and first-time), outbound status reconciliation via webhook -- all working! The `sendAndPersistOutbound` extraction cleaned things up nicely.

One thing I'd like addressed before merging: there's a type consistency issue across the notification layer that will hurt maintainability as this grows. After that, rebase on main and it's good to go. Left an inline comment with the details.

The short version: `OutboundSendResult` is defined in `outbound.ts` but the Bird functions that produce that exact shape (`sendConversationMessage`, `sendSMSMessage`) define it inline instead of using it. Same pattern repeats with template params and the UI status type. When a type is defined once but the same shape is written out separately in 4+ other places, a field change means hunting across files instead of updating one interface. **This is an important habit for this codebase -- we centralize types so they stay in sync.** The full list is in the inline comment.

### Follow-up items (not blocking, but worth tracking):

**Performance: drop synchronous polling in the delivery pipeline.** Now that the webhook reconciles outbound statuses (and we verified it works), the 10s polling per message in `reconcileMessageStatus` is redundant for the production path. For 100 recipients that's ~17 min of wall time mostly sleeping. `sendUnsubscribeReply` already sets `reconcile: false` — the delivery pipeline should do the same. The one thing to solve: the WA→SMS fallback currently relies on the polling to detect post-send failures synchronously, so that decision would need to move to a webhook-triggered async path.

**Conversations page: individually addressable conversations.** Right now threads are inline-expanded in the list. Each conversation should have its own URL (e.g. `/admin/conversations/:conversationId`) so we can link to it from user profiles (phone → conversation cross-reference) and vice versa. This also naturally solves the data loading — the list page only needs summaries, the full thread loads on demand. Along with that: filtering by channel, status, date range.

**Refactor `extractMessageFields`.** This function is 80+ lines of interleaved fallback chains handling event unwrapping, direction detection, phone extraction, body extraction, and channel detection all in one place. Hard to maintain — when the outbound phone extraction was missing, it was buried in a long chain. Break it into focused steps (unwrap event → extract direction → extract phone → extract body → determine channel), each small and testable. The inbound vs outbound phone paths are fundamentally different lookups and should be clearly separated rather than crammed into one fallback chain. This ties into the next point.

**Tests for `extractMessageFields`.** The most complex pure function in the PR — handles ~10 different Bird payload shapes. `mapBirdMessageStatus` and `isUnsubscribeMessage` got tests, this one should too. Breaking it up (above) makes it straightforward to test each step.

**`SendTemplateDialog`: align with codebase form conventions.** The dialog uses 13 `useState` calls with manual reset and validation. The rest of the codebase uses React Hook Form + Zod (`PersonForm`, `PartyForm`, `AddMeetingForm` etc.) with the `Form`/`FormField` components from `src/components/ui/form.tsx`. Not urgent since it's an admin test tool, but worth aligning — `form.reset()` replaces the manual 8-variable reset, Zod replaces the inline regex validation, and the duplicated submit handler pattern (`setStatus('sending')` → `startTransition` → `if (result.success)`) can be shared.

**`listCitiesForTest` / `listMeetingsForTest` duplicate existing DAL methods.** `getAllCitiesMinimal()` in `cities.ts` and `getCouncilMeetingsForCity()` in `meetings.ts` already do the same queries. The only difference is Date serialization for the server action boundary, which is a one-liner `.map()` at the call site.

VasiaKoum:
After-review changes:

* OutboundSendResult / ConversationResult are now the canonical Bird-call return shapes; bird.ts wrappers reference them instead of redefining inline
* MeetingNotificationParams base + WhatsAppTemplateParams extends it with notificationId
* SendReplyResult = Pick<OutboundSendResult, 'success' | 'error'> in admin actions
* SendStatus extracted to src/components/admin/conversations/types.ts (shared by ReplyForm + SendTemplateDialog)
* All notification-domain types moved into one src/lib/notifications/types.ts
* Split the functionality of `extractMessageFields` into 7 small pure functions: `unwrapEvent` / `extractDirection` / `extractInboundPhone` / `extractOutboundPhone` / `extractPhone` / `extractBody` / `extractChannel` and move them to src/app/api/webhooks/bird/extract.ts
* Add also 48 unit tests for these functions in __tests__/extract.test.ts
* Fix `sendAndPersistOutbound` to return `finalReason` instead of just `status`
* Split the functionality of `sendAndPersistOutbound` into 5 small pure functions: `verifyRequest` / `updateOutboundMessage` / `getNotificationDeliveryForMessage` / `persistMessageRow` / `handleUnsubscribeIfApplicable`
* Add `router.refresh()` runs in both success AND failure branches (so failed rows appear in the thread)
* Replace actions.ts sendTestBeforeMeetingNotification to call `generateSmsContent` inline (no more '[beforeMeeting template]' placeholder)
* Fix `ScrollableThread` to re-scroll on children-count change (was mount-only)

kouloumos (approved):
LGTM! 🥳  Ready to merge!
Follow-up items from my previous comment still apply.

PR #346, opened 2026-05-12
fix(elasticsearch): fix deletion propagation and strip REF links via ingest pipeline

kouloumos:
## Problem

When meetings are reprocessed, subjects are deleted and recreated in PostgreSQL. The DELETE events did not propagate to Elasticsearch — deleted subjects remained searchable as ghost documents. We run a cleanup script to purge them since we spotted this.

## Root cause: `base_tables: ["Subject"]` on a child view

`SubjectSearchView` stripped markdown REF links from `Subject.description` before indexing. In `schema.json` it declared `base_tables: ["Subject"]` — the root table of the sync tree. PGSync then routed Subject DELETE events to the child view node instead of the root node: it tried to re-sync the description, found no row, and never deleted the ES document.

**Rule: a child view must never list the root table in `base_tables`.**

## Fix (commit 1)

Remove `SubjectSearchView` and read `description` directly from the root table. `views.sql` now drops the view explicitly, because `CREATE OR REPLACE` does not remove old views.

## REF links move to an ingest pipeline (commit 2)

Removing the view meant REF links (`[text](REF:TYPE:ID)`) reached the index raw. Measured impact on the production cluster:
- **BM25: unaffected.** The Greek analyzer keeps each link target as one token (`ref:person:abc123`); our queries are term-based with no phrase matching.
- **Semantic: a small, systematic cost.** `semantic_text` embeds the raw field value. On 300 staging subjects, markup-heavy documents gain similarity against every query, and clean documents lose the top result in ~4% of self-retrieval queries.

The fix that needs no view on any table: a `strip-refs` **Elasticsearch ingest pipeline** (`elasticsearch/pipeline.json`), attached by PGSync through the top-level `"pipeline"` key in `schema.json`. It runs before analysis and before `semantic_text` inference, so tokens and embeddings both come out clean. Deletes do not use the pipeline. The database keeps the raw markdown; the frontend renders REF links from PostgreSQL.

## Verification

Full E2E with `pgsync-test.sh` (local PG + real cluster), against the exact production pgsync 7.0.1 image:

| pgsync | schema | insert | delete |
|---|---|---|---|
| 7.0.1 | old (main) | ✅ | ❌ — notification accepted, no ES operation (the misroute) |
| 7.0.1 | this PR | ✅ | ✅ — document deleted in ~6s |
| 7.0.1 | this PR + pipeline | ✅ stripped | ✅ |

Bootstrap documents and live updates both index stripped text (nested `speaker_contributions.text` included).

Found along the way: pgsync < 7.x installs a `table_notify()` trigger function that drops every root DELETE notification (`indices` missing from the DELETE branch). Production's function is already the fixed 7.x version (verified read-only), so this bug is not active on prod — it is documented in the README FAQ because any environment bootstrapped with an older pgsync has it.

## Deployment

1. Create the pipeline: `curl -X PUT "$ELASTICSEARCH_URL/_ingest/pipeline/strip-refs" ... -d @elasticsearch/pipeline.json` (must exist before PGSync references it)
2. Apply the views: `psql "$PG_URL" < elasticsearch/views.sql`
3. Update the schema gist that `SCHEMA_URL` points to
4. Run a PGSync `--bootstrap`, then start the daemon. The bootstrap applies the pipeline to all existing documents and removes the accumulated ghost documents.

<h3>Greptile Summary</h3>

The PR removes the root-table child view that interfered with Subject deletion propagation and moves REF-link normalization into an Elasticsearch ingest pipeline.
- Reads Subject descriptions directly from the root table so PGSync can process root deletes normally.
- Adds a `strip-refs` pipeline for descriptions, contribution text, and segment summaries.
- Updates view validation and operational documentation for the four remaining SQL views and pipeline deployment.

The PR appears safe to merge.

The ingest pipeline completes the prior semantic-text fix while the removal of the root-backed child view restores normal Subject deletion handling, and no blocking failure remains.

| Filename | Overview |
|----------|----------|
| elasticsearch/schema.json | Removes the unsafe SubjectSearchView child node, indexes the root description directly, and attaches the REF-stripping ingest pipeline. |
| elasticsearch/pipeline.json | Defines guarded processors that strip REF markdown from the indexed fields implicated by the prior semantic-embedding comment. |
| elasticsearch/views.sql | Drops the legacy root-backed view and preserves raw contribution text for normalization at Elasticsearch ingest time. |
| elasticsearch/validate-views.sql | Updates validation for the four-view configuration and uses the current contribution view primary-key column. |
| elasticsearch/README.md | Documents pipeline creation order, re-indexing behavior, deletion propagation constraints, and the revised deployment flow. |

<sub>Reviews (3): Last reviewed commit: ["feat(elasticsearch): strip REF links at ..."](https://github.com/schemalabz/opencouncil/commit/fa4748fa5b37f0563f93d79477074c8945c12dbe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31855910)</sub>

### DIFF
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -160,6 +160,7 @@ PGSync requires helper views to denormalize complex relationships and handle Pos
 1. **LocationSearchView** - Converts PostGIS geometry to GeoJSON format
 2. **IntroducedByPartyView** - Resolves party affiliation through the `Role` table
 3. **SubjectSpeakerSegmentSearchView** - Denormalizes speaker segments with concatenated utterances
+4. **SpeakerContributionSearchView** - Denormalizes speaker contributions with party resolution
 
 ### Create the Views
 
@@ -170,7 +171,7 @@ psql "$DATABASE_URL" < elasticsearch/views.sql
 ```
 
 The script will:
-- Create all three required views
+- Create all four required views
 - Run verification checks on each view
 - Display sample data to confirm everything works
 - Show statistics about data coverage
@@ -723,7 +724,13 @@ A: PGSync tracks its position in the WAL stream using Redis. When it restarts, i
 **Q: How do I update the schema?**  
 A: See <https://pgsync.com/advanced/re-indexing>
 
-**Q: Why use views instead of direct table joins in PGSync?**  
+**Q: Why are REF links (`[text](REF:TYPE:ID)`) indexed in Elasticsearch instead of being stripped?**
+A: This is intentional. The Greek analyzer strips punctuation during tokenization, so `[Ο Δήμαρχος](REF:PERSON:abc123)` tokenizes to the same search terms as `Ο Δήμαρχος`. The extra tokens (`ref`, `person`, `abc123`) are harmless noise that no user would search for. More importantly, **a view with `base_tables: ["Subject"]` would intercept Subject DELETE events in PGSync**, causing deletions to never propagate to Elasticsearch. By reading `description` directly from the Subject table (the root node), we ensure PGSync correctly processes root-table deletions. The app's frontend handles REF link rendering via `FormattedTextDisplay` and `stripMarkdown()`.
+
+**Q: Why must views never list the root table in `base_tables`?**
+A: PGSync builds an internal `base_table_to_node` mapping from all views' `base_tables`. If a view declares `base_tables: ["Subject"]` (the root table), PGSync treats Subject WAL events as child-node changes for that view, instead of root-table operations. This means DELETE events on Subject are never processed as document deletions — PGSync tries to "re-sync" the parent document instead, which fails silently because the row is already gone. **Rule: only use `base_tables` for tables that are NOT the root node.**
+
+**Q: Why use views instead of direct table joins in PGSync?**
 A: Views handle complex logic (PostGIS conversion, role-based party resolution, utterance concatenation) in PostgreSQL where it's more efficient. PGSync sees views as simple tables, keeping the sync configuration clean.
 
 **Q: Can I combine semantic search with traditional text search?**  
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -9,10 +9,11 @@ This document describes how OpenCouncil uses Elasticsearch to provide powerful s
 2. [Codebase Structure](#codebase-structure)
 3. [Set up Elasticsearch](#set-up-elasticsearch)
 4. [Configure PostgreSQL Views](#configure-postgresql-views)
-5. [Set up PGSync](#set-up-pgsync)
-6. [Sync Data](#sync-data)
-7. [Search Examples](#search-examples)
-8. [Best Practices & FAQ](#best-practices--faq)
+5. [Configure the Ingest Pipeline](#configure-the-ingest-pipeline)
+6. [Set up PGSync](#set-up-pgsync)
+7. [Sync Data](#sync-data)
+8. [Search Examples](#search-examples)
+9. [Best Practices & FAQ](#best-practices--faq)
 
 
 ### Architecture Overview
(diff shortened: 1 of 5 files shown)


elasticsearch/README.md
```diff
@@ -737,7 +738,13 @@ A: PGSync tracks its position in the WAL stream using Redis. When it restarts, i
 **Q: How do I update the schema?**  
 A: See <https://pgsync.com/advanced/re-indexing>
 
-**Q: Why use views instead of direct table joins in PGSync?**  
+**Q: Why are REF links (`[text](REF:TYPE:ID)`) indexed in Elasticsearch instead of being stripped?**
+A: This is intentional. The Greek analyzer strips punctuation during tokenization, so `[Ο Δήμαρχος](REF:PERSON:abc123)` tokenizes to the same search terms as `Ο Δήμαρχος`. The extra tokens (`ref`, `person`, `abc123`) are harmless noise that no user would search for. More importantly, **a view with `base_tables: ["Subject"]` would intercept Subject DELETE events in PGSync**, causing deletions to never propagate to Elasticsearch. By reading `description` directly from the Subject table (the root node), we ensure PGSync correctly processes root-table deletions. The app's frontend handles REF link rendering via `FormattedTextDisplay` and `stripMarkdown()`.
```
christosporios: I understand that this is the easiest way and that it makes sense -- but let's be honest here. It probably strips punctuation only, so `[Ο Δήμαρχος](REF:PERSON:abc123)` tokenizes the same as `ο δήμαρχος ref person abc123`, not `ο δήμαρχος`

elasticsearch/views.sql
christosporios, line 1: we never drop the removed view?

christosporios (approved):
LGTM, pending these two changes

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/98f8c3196e02791181b9d39e654ee35a83326158..fa4748fa5b37f0563f93d79477074c8945c12dbe) to address both review comments and to complete the REF-link story:

- `views.sql` now drops the legacy `SubjectSearchView` (58d95d2).
- The README REF-link answer now states measured facts instead of the hand-wavy tokenization claim — details in the inline reply.
- New commit fa4748f: REF links are stripped at index time by a `strip-refs` ingest pipeline (versioned in `elasticsearch/pipeline.json`, attached via the `pipeline` key in schema.json). This restores the stripping the view used to do, without any view touching the root table.

Everything is verified E2E with `pgsync-test.sh` against the production pgsync 7.0.1 image — matrix in the updated PR description. The deploy now needs a one-time pipeline `PUT` and a bootstrap (which also sweeps the existing ghost documents).

elasticsearch/README.md
kouloumos: Correct — that claim was wrong, and the truth is a third thing. I measured it with `_analyze` on the production cluster: the Greek analyzer keeps the whole target as **one** token, so `[Ο Δήμαρχος](REF:PERSON:abc123)` becomes `ο δήμαρχος ref:person:abc123`, not `ο δήμαρχος ref person abc123`. BM25 term queries are unaffected (our queries use `multi_match`/`match` with no phrase matching).

The real cost is in semantic search: `semantic_text` embeds the raw field value, so the markup enters the embeddings. I measured this on 300 staging subjects with the e5 endpoint: markup-heavy documents gain similarity against every query, and clean documents lose the top spot in ~4% of self-retrieval queries. Small, but systematic.

I rewrote the README answer with the measured facts, and commit fa4748f in this PR strips REF links with an ingest pipeline (`pipeline` key in schema.json — pgsync supports it natively, verified E2E with pgsync-test.sh). That cleans tokens and embeddings in one place, with no view touching the root table.

elasticsearch/views.sql
kouloumos: Good catch — fixed in 58d95d2. `views.sql` now starts with `DROP VIEW IF EXISTS "SubjectSearchView"`.

PR #368, opened 2026-05-20
feat(transcript): add find & replace tool to editing mode

angelospk:
Closes #301 

## Description

Adds a Find & Replace tool to the transcript editor. Until now, editors cleaning up council-meeting transcripts had no way to mass-fix recurring errors like misspelled names (e.g. `Μαυρογιώργης` → `Μαυρογεώργης`). On meetings with ~9K utterances, doing this one utterance at a time was slow and easy to get wrong.

The tool works like Cmd+F in Word or VS Code: a floating panel at the top-right of the transcript with search and replace inputs, a case-sensitive toggle, prev/next navigation, single replace, and Replace All. All matches in the loaded transcript are highlighted yellow; the active match is orange.

If you have some text selected before pressing the keyboard shortcut, the search term comes prepopulated in the floating panel, similar to the browser's search tool.

## UI look
image

## Changes

New server action `replaceAllInUtterances(cityId, meetingId, term, replacement, caseSensitive)` in `src/lib/db/utterance.ts`. Runs all updates and audit-trail rows inside one `prisma.$transaction`. Scoped by `meetingId` so it can't bleed into other meetings in the same city. Returns `{ utteranceCount, occurrenceCount }` for the toast.

`FindReplaceContext` computes the matches list with `useMemo` over the loaded transcript. The per-utterance match map is exposed via `useSyncExternalStore`, so `UtteranceC` (memoised, ~9K instances) only re-renders when its own match set changes. Typing in the search box does not invalidate the whole transcript.

`FindReplacePanel` is the floating UI. Confirmation dialog before Replace All, toast with the replaced count, scrolls the active match into view.

`Utterance.tsx` splits the rendered text into `<mark>` spans for matches when the utterance is not being edited. The outer span keeps its id, data-attrs, and click handler, so context-menu and click-to-edit still work.

Cmd/Ctrl+F shortcut in `Transcript.tsx`, only when editing mode is active. Outside edit mode the browser's native find still works. Pre-fills the search input from `window.getSelection()` or the active textarea's selection; if pre-filled, focus jumps straight to the replace input.

`EditingGuideDialog` documents the new shortcut.

## Out of scope

- Regex search.
- Matches that span utterance boundaries.
- Tonos-insensitive Greek matching (would need the Postgres `unaccent` extension).
- Updating the `Word` table. Single-utterance edits don't touch it either.

## Testing

11 integration tests in `tests/integration/replace-all-in-utterances.test.ts`. They cover literal replace across utterances, multiple occurrences per utterance, case-sensitive vs insensitive, Greek text, audit-trail creation, no-op when nothing matches, regex special chars treated as literal, isolation between cities, isolation between meetings in the same city (regression test for an early version that scoped only by `cityId`), and empty-term rejection.

Run with:

```
DOCKER_HOST=… TESTCONTAINERS_RYUK_DISABLED=true npm run test:integration -- replace-all-in-utterances
```

`npm run build`, `npm run lint`, `npm test` all green.

Manual: open a meeting in editing mode, press `Cmd+F`, and the search/replace works. The toast count matches the highlighted count.

<h3>Greptile Summary</h3>

This PR adds a Cmd/Ctrl+F find & replace panel to the transcript editor, addressing the practical need to bulk-fix recurring name misspellings across ~9K utterances. The implementation is well-structured, with a thoughtful `useSyncExternalStore`-based per-utterance subscription pattern to avoid re-rendering the full transcript on every keystroke.

- **New server action** `replaceAllInUtterances` in `utterance.ts`: meeting-scoped, auth-checked, chunked transactions, function-form replacement to prevent `$`-token expansion; backed by 11 integration tests.
- **`FindReplaceContext`**: computes match positions via `useMemo`, diffs and notifies only changed utterances via a ref-based external store; correctly clears highlights when the panel closes.
- **`Utterance.tsx`**: renders `<mark>` spans for matches and fixes `getCaretOffsetFromClick` with a TreeWalker to correctly accumulate preceding text-node lengths. The `renderedDisplay` path uses the raw stored text (`localUtterance.text`) rather than the `localize()`-transliterated form — for the Serbian locale this means marks appear in the wrong script and character offsets into the transliterated display are misaligned (Cyrillic→Latin digraphs `љ→lj`, `њ→nj`, `џ→dž` shift subsequent offsets).

- Safe to merge for Greek and most locales; Serbian locale users will see mixed-script mark rendering and potentially misaligned highlight boundaries until the localize() bypass in renderedDisplay is fixed.
- The find-and-replace logic, server action, and subscription architecture are solid and well-tested. The one functional gap is in Utterance.tsx: `renderedDisplay` renders marks against the raw stored text instead of the transliterated display text. For Serbian (sr/sr-Latn), Cyrillic→Latin transliteration is not character-count-preserving (digraphs like љ→lj expand one stored character to two display characters), so mark boundaries computed on the stored form land at the wrong positions in the displayed Latin text. Every other aspect of the feature is correct.
- src/components/meetings/transcript/Utterance.tsx — the renderedDisplay block should apply localize() to the base text and ideally compute match offsets against the localized form for Serbian locale correctness.

| Filename | Overview |
|----------|----------|
| src/components/meetings/transcript/Utterance.tsx | Adds per-utterance match highlighting and fixes `getCaretOffsetFromClick` for multi-node spans. The `renderedDisplay` path bypasses `localize()`, causing wrong script and misaligned mark boundaries for Serbian locale users. |
| src/components/meetings/transcript/FindReplaceContext.tsx | New context using `useSyncExternalStore` for per-utterance subscriptions. Correctly guards against highlights persisting after panel close. Implementation is sound. |
| src/components/meetings/transcript/FindReplacePanel.tsx | Floating panel with search/replace inputs, navigation, and Replace All confirmation dialog. `aria-label` hardcoded in English (P2). Core logic is correct. |
| src/lib/db/utterance.ts | Adds `replaceAllInUtterances` server action with auth check, meeting-scoped query, and chunked transactions. JSDoc still says "across every utterance in a city" (P2). |
| src/lib/utils/findReplace.ts | New utility with `escapeRegExp` and `literalReplaceAll` (function-form replacement to prevent `$`-token expansion). Clean and well-tested. |
| src/components/meetings/transcript/Transcript.tsx | Wraps transcript in `FindReplaceProvider`, adds `FindShortcut` (document-level Cmd+F with pre-fill), and renders `FindReplacePanel`. Changes are clean. |
| tests/integration/replace-all-in-utterances.test.ts | 11 integration tests covering literal match, multi-occurrence, case sensitivity, Greek text, audit trail, no-op, regex metacharacters, city isolation, and meeting isolation. Good coverage. |

<sub>Reviews (9): Last reviewed commit: ["feat(transcript): add find &amp; replace too..."](https://github.com/schemalabz/opencouncil/commit/b7c662a433c636bd94650a2c6e862ff104a5c481) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33000728)</sub>


angelospk:
Pushed a batch of fixes from the review and a few extra things I caught while poking at it.

- Replace All was passing the user-provided replacement string straight to `String.replace(regex, str)`, which interprets `$&`, `$1`, `$$` and friends as backref tokens. Replacing with `$5.00` came out wrong. Both the server action and the client-side optimistic update now go through a `literalReplaceAll` helper that wraps the replacement in a function callback, so the dollar tokens go in verbatim. Regression tests cover `$&`, `$1`, `$'`, `` $` ``, `$$`, `$5.00`, plus regex metacharacters in the search term.
- Click-to-edit was placing the caret in the wrong spot once the utterance had `<mark>` children. `caretRangeFromPoint` / `caretPositionFromPoint` return an offset relative to the deepest text node, not the whole utterance. The handler now walks the span's text nodes in document order and sums their lengths.
- Closing the panel via Esc or ✕ used to leave yellow highlights on every matched utterance. The matches `useMemo` only short-circuited on empty search term, not on `isOpen`. Fixed.
- Replace All was scoped by `cityId` only, so it would have touched utterances in every other meeting in the same city. Now scoped by `cityId` + `meetingId`. New test covers it.
- Navigation (prev/next) was re-rendering every `UtteranceC` because a global active-index subscriber fired on each press. Dropped it. The per-utterance store carries both the match list and the active index, and the provider only notifies the two affected ids.
- Typing in the search box was triggering `scrollIntoView` on every keystroke (the effect depended on `fr.matches` identity, which changes constantly). A ref keyed by `utteranceId:start` now gates the scroll so it only fires when the active match actually moves.
- Replace All on a 9K-utterance meeting with a common term could push thousands of ops into one `$transaction`. Chunked at 500 utterances per transaction to stay well under Postgres parameter limits.
- Panel labels, confirmation dialog, toast, and the shortcut entry in the keyboard guide all go through `useTranslations` now. Added a `transcript.findReplace` namespace and a `shortcuts.editing.findReplace` key to `messages/en/*` and `messages/el/*`.
- `escapeRegExp` lived in three files. Now imported from `src/lib/utils/findReplace.ts` along with `literalReplaceAll`.
- Focus effect destructures `isOpen`, `pendingFocus`, `clearPendingFocus` from the context object so the dep array no longer includes the whole context (which changed identity on every keystroke).

Compare: [`4daa50e..305b316`](https://github.com/schemalabz/opencouncil/compare/4daa50e7f80b1f1ac46c91a43952d0df4ddcfbd5..305b3163b34e5afc44a14abac362b12923d98bf3)

christosporios:
Ευχαριστούμε -- όμως δε φαίνεται να δουλεύει στο preview. Τα edits δε φαίνεται να γίνονται ποτέ.

angelospk:
> Ευχαριστούμε -- όμως δε φαίνεται να δουλεύει στο preview. Τα edits δε φαίνεται να γίνονται ποτέ.

In my local dev test, it worked fine.
https://www.youtube.com/watch?v=JhzLXex8HwU

PR #379, opened 2026-05-24
Feat/fisheye transcript

### LINKED ISSUE #109: Fish‑Eye Transcript View for Council Meetings
### Summary
Add an alternative “Fish‑Eye” transcript view that centers the currently active `SpeakerSegment` with rich detail, while progressively abstracting nearby segments into compact summaries and distant segments into minimal stripes. This mirrors fisheye/zoomed context patterns to let users see both the “answer” and its surroundings at once.

Reference: Fish-eye lens concept by Amelia Wattenberger: [Fish Eye](https://wattenberger.com/thoughts/fish-eye)

### Motivation
- Users often need detail and context simultaneously: what’s being said now, what came right before and after, and how this fits into the meeting’s overall flow/subjects/speakers.
- We already have the data, UI hooks, and a timeline visualization; a fisheye mode is a natural extension that improves comprehension and navigation without adding new backend complexity.

### Scope
- Add a Fish‑Eye view toggle on the meeting transcript page.
- Center the active segment (full detail), show adjacent segments as summaries, and render distant segments as thin stripes or collapsed blocks.
- Optionally display peripheral context: subject strip (subject of the center segment if any), and meeting-level summary ribbon.
- Should be responsive and performant on large transcripts.

### Out of Scope
- No changes to backend schema or APIs.
- No brand-new summarization pipeline; reuse existing summaries where available. If missing, gracefully fall back to truncated text.
- No editing features beyond what already exists in transcript edit mode.

### UX Outline
- Toggle: “Fish‑Eye” on/off control on the transcript view. Persist via query param (e.g., mode=fisheye) or in `OptionsContext`.
- Center segment: full rendering as today (speaker, party color, text, time, controls).
- Near context (N segments before/after): render compact summaries (first sentence or existing segment summaries).
- Far context: render minimal placeholders/stripes to convey structure and position.
- Optional periphery:
  - Subject strip above the center (if the segment is linked to a subject).
  - Meeting summary ribbon at the top for quick orientation.
- Smooth transitions when center changes (scroll or time updates).

### Technical Notes
- Likely files to touch:
  - `src/app/[locale]/(city)/[cityId]/(meetings)/[meetingId]/transcript/page.tsx` (add view toggle / query param)
  - `src/components/meetings/transcript/Transcript.tsx` (extract/compose fisheye rendering or wrap)
  - `src/components/meetings/transcript/SpeakerSegment.tsx` (support compact/summary modes)
  - `src/components/meetings/TranscriptControls.tsx` (optional lens-style timeline density)
  - `src/components/meetings/CouncilMeetingDataContext` (existing data access)
  - `src/components/meetings/Summary.tsx` (meeting-level summary ribbon if reused)
- Existing hooks/utilities you can leverage:
  - IntersectionObserver already in `Transcript.tsx` to detect visible segments; reuse to determine the “center.”
  - `useVideo()` (currentTime), `useCouncilMeetingData()` (transcript, subjects, speakers), `joinTranscriptSegments()` for grouped display.
- Rendering approach:
  - Track “centerIndex” by the segment closest to viewport center or nearest to current playback time.
  - Calculate distance from center for each segment and map to rendering modes: full → summary → stripe.
  - Keep DOM light: only fully render the center and neighbors; collapse distant segments into lightweight placeholders to preserve performance.
- Performance:
  - Continue using lazy rendering and visibility checks; avoid re-rendering the entire list on scroll.
  - Memoize mapping from index distance to render mode; avoid heavy computations in scroll handlers.
- Accessibility:
  - Preserve keyboard navigation within the center and neighbor segments.
  - Ensure color contrast remains accessible when dimming.
- Internationalization:
  - Add strings to `messages/en.json` and `messages/el.json` for the toggle label and any UI text.
- Authorization guardrails:
  - The transcript view is public. If adding any admin-only overlays, wrap with:
    - `await isUserAuthorizedToEdit({ cityId })` for conditionals, or
    - `await withUserAuthorizedToEdit({ cityId })` in server actions that must throw if unauthorized.

### Acceptance Criteria
- Fish‑Eye toggle appears on the transcript page and switches modes without page reload.
- Active segment renders fully; adjacent segments render as summaries; far segments render as stripes.
- Scrolling or playing the video updates the centered segment smoothly.
- Optional subject strip and meeting summary ribbon can be toggled in code/config; they don’t break if missing data.
- No regressions to the existing transcript view; performance is comparable or better due to collapses.
- Mobile and desktop layouts are responsive.
- i18n strings added for any new labels.

### Nice-to-Haves (Optional)
- Density‑aware timeline: widen center interval and compress distant intervals in `TranscriptControls.tsx`.
- Smooth CSS transitions between render modes.
- URL-shareable center anchor (e.g., `?t=123` syncs to the centered segment).

### Testing Notes
- Manually verify with a long meeting (many segments) that virtualization still holds and scrolling remains smooth.
- Verify toggling modes retains scroll position and current time link to video.
- Verify i18n and RTL safety (if applicable).

### Implementation Freedom
- You may implement as a wrapper component (e.g., `FishEyeTranscript.tsx`) or by extending `Transcript.tsx` with a mode switch—whichever yields cleaner separation and minimal duplication.
- Rendering styles for summary/stripe modes are up to you; aim for clarity and subtlety. Avoid heavy third-party dependencies.
- You can choose whether to drive the center by scroll position, current video time, or whichever signal yields better UX; just keep it consistent and intuitive.

### References
- Fish‑eye lens concept by Amelia Wattenberger: [Fish Eye](https://wattenberger.com/thoughts/fish-eye)

angelospk:
## Summary

Adds an opt-in "fish-eye" view mode for the transcript page. When it's on, only the segment that's currently playing and its two immediate neighbors render in full; everything else collapses to a one-line row with the speaker's surname, a short preview, and a timestamp. The focus window only moves when playback moves. Scrolling never shifts it. Clicking any compact row seeks to that segment and starts playing, which is the only way to change what's in focus.

The point is to make long meetings scannable. You can see what's playing without losing the shape of the rest of the agenda below.

The preview text in compact rows is the verbatim utterance for short speeches (≤ ~2 sentences, ≤ 160 chars). For longer ones it falls back to the AI summary, with a small Bot icon so it's obvious which rows are synthetic and which are real speech.

[597261238-b65abfb3-49f2-461b-949a-528c0325117b.webm](https://github.com/user-attachments/assets/26732cee-1b92-4e0d-b613-5c60c1f72dc8)

## Changes

- New `useViewMode` hook on `useSyncExternalStore` so the toggle button and the transcript read from the same state cell. A `?mode=fisheye` URL param wins per-link; otherwise the choice is saved to localStorage and synced across tabs via the storage event.
- New `FisheyeToggle` button. On mobile it sits next to the orange show/hide-controls button and animates together with it when the controls bar collapses.
- Controls visibility moved out of `TranscriptControls` and into `LayoutContext` so the toggle can react to it.
- New pure mapper `fisheyeModeForDistance` and a `SegmentContext` compact row component.
- i18n keys added to `messages/{en,el}/transcript.json`. These override the monolithic file via the shallow merge in `src/i18n/request.ts`, so the keys go there, not in `messages/{en,el}.json`.
- `PersonBadge` gets an optional `lastNameOnly` prop for the inline variant.

## Test plan

- [ ] Turn fish-eye on. Only the active segment and one row on each side render fully; everything else is compact.
- [ ] Let playback advance across segments. Focus follows without a scroll jump.
- [ ] Scroll up and down with fish-eye on. The focus window does NOT move.
- [ ] Click a compact row. It seeks to that segment and starts playing.
- [ ] Reload the page. The chosen mode persists.
- [ ] Open a `?mode=fisheye` link in a tab where the saved mode is default. URL wins.
- [ ] Open the same page in two tabs and toggle in one. The other updates.
- [ ] A short speech shows the verbatim text. A long one shows the AI summary with a Bot icon.
- [ ] Mobile: hide and show the controls bar. The fish-eye button animates with it.
- [ ] `npm test` passes locally (703).

Closes #109

<h3>Greptile Summary</h3>

This PR adds an opt-in "fish-eye" view mode for the transcript page: the active segment and its two immediate neighbors render fully while everything else collapses to a compact one-line row. The architecture is solid — `useSyncExternalStore` with eager module-level hydration, a stable `VideoActionsContext` to avoid per-tick re-renders on compact rows, and a lifted `LayoutContext` for controls visibility.

- **`useViewMode`** is backed by a module-level store that seeds from URL/localStorage before the first render, eliminating the layout-shift on `?mode=fisheye` links, and syncs across tabs via the storage event.
- **`SegmentContext`** renders the compact row and correctly subscribes only to `useVideoActions()` (not the hot `VideoContext`), so it avoids per-frame re-renders during playback.
- **`handleActivate`** only calls `seekToWithoutScroll` — the video stays paused if it was already paused — which contradicts the PR description and test plan item 3 ("seeks to that segment and starts playing"). `setIsPlaying` was added to `VideoActionsContext` in this PR but is never invoked from the click handler.

Safe to merge after confirming whether clicking a compact row should start playback — the code explicitly keeps the user paused, while the PR description and test plan say it starts playing.

The click handler in SegmentContext only seeks without starting playback, directly contradicting test plan item 3. If the intent is 'seek only' (the code comment says so), the description and test plan need updating. If the intent is 'seek and play', setIsPlaying(true) is missing from handleActivate.

src/components/meetings/transcript/SegmentContext.tsx — the handleActivate function and whether it should call setIsPlaying(true)

| Filename | Overview |
|----------|----------|
| src/components/meetings/transcript/SegmentContext.tsx | New compact context-row component; now correctly uses useVideoActions() to avoid per-frame re-renders, but handleActivate does not call setIsPlaying(true) despite the PR description and test plan requiring it |
| src/hooks/useViewMode.ts | New module-level store backed by useSyncExternalStore; eagerly hydrates from URL/localStorage at module-eval time to avoid layout shift, correctly handles cross-tab sync via storage event |
| src/hooks/useFisheyeCenter.ts | Thin passthrough hook that returns activeSegmentIndex unchanged; kept as a hook so the 'center follows playback, not scroll' invariant is documented in one place |
| src/components/meetings/transcript/Transcript.tsx | Adds fisheye rendering path: 4Hz setInterval polling currentTimeRef drives activeSegmentIndex; fisheyeModeForDistance gates SpeakerSegment vs SegmentContext per row |
| src/components/meetings/VideoProvider.tsx | Adds setIsPlaying to the stable VideoActionsContext so consumers can trigger playback without subscribing to the hot VideoContext |
| src/components/meetings/CouncilMeetingWrapper.tsx | Lifts isControlsVisible/setIsControlsVisible into LayoutContext; resize handler now only forces-show controls on landscape and never force-hides on narrow, preserving the user's toggle choice |
| src/components/persons/PersonBadge.tsx | Adds optional lastNameOnly prop; correctly extracts first token of formatSurnameFirst output (which uses space, not comma, as separator) |
| src/lib/utils/fisheye.ts | Pure utility mapping distance-from-center to FisheyeMode; well-tested with symmetric and boundary cases |
| src/components/meetings/transcript/FisheyeToggle.tsx | Simple toggle button that reads/writes from useViewMode; correctly suppressed in editable mode |

</details>

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `src/components/meetings/transcript/SegmentContext.tsx`, line 333-336 ([link](https://github.com/schemalabz/opencouncil/blob/f667a96960da0323e15e4869c3e46011ac9de638/src/components/meetings/transcript/SegmentContext.tsx#L333-L336)) 

   <a href="#">P1</a> **Unhandled rejection in `handleActivate`**

   `setIsPlaying` is backed by `stableSetIsPlaying`, which is an `async` function (it `await`s `playVideoRef.current()`). The interface exposes it as `(isPlaying: boolean) => void`, hiding the async nature. If the underlying `playVideo` call rejects (e.g., media not ready, browser autoplay policy), the rejected promise is silently swallowed here — no error boundary or console output, and the seek has already been committed. This leaves the player seeked to the new position but not playing, with no feedback to the user.

   Consider wrapping in a `.catch` so failures are at least logged, or aligning the interface signature with the async reality.

   `````markdown
   Path: src/components/meetings/transcript/SegmentContext.tsx
   Line: 333-336

   Comment:
   **Unhandled rejection in `handleActivate`**

   `setIsPlaying` is backed by `stableSetIsPlaying`, which is an `async` function (it `await`s `playVideoRef.current()`). The interface exposes it as `(isPlaying: boolean) => void`, hiding the async nature. If the underlying `playVideo` call rejects (e.g., media not ready, browser autoplay policy), the rejected promise is silently swallowed here — no error boundary or console output, and the seek has already been committed. This leaves the player seeked to the new position but not playing, with no feedback to the user.

   Consider wrapping in a `.catch` so failures are at least logged, or aligning the interface signature with the async reality.

   `````
   </details>
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Asrc%2Fcomponents%2Fmeetings%2Ftranscript%2FSegmentContext.tsx%3A65-71%0A**Clicking%20a%20compact%20row%20does%20not%20start%20playback**%0A%0A%60handleActivate%60%20only%20calls%20%60seekToWithoutScroll%60%20%E2%80%94%20it%20never%20calls%20%60setIsPlaying%28true%29%60.%20The%20PR%20description%20states%20%22clicking%20any%20compact%20row%20seeks%20to%20that%20segment%20and%20**starts%20playing**%22%20and%20test%20plan%20item%203%20says%20%22It%20seeks%20to%20that%20segment%20and%20**starts%20playing**.%22%20With%20this%20implementation%2C%20a%20paused%20user%20who%20clicks%20a%20row%20stays%20paused.%20The%20%60setIsPlaying%60%20function%20was%20added%20to%20%60VideoActionsContext%60%20in%20this%20same%20PR%20%28presumably%20for%20exactly%20this%20call-site%29%20but%20is%20never%20invoked%20here%20%E2%80%94%20the%20test%20plan%20check%20would%20fail%20as-written.%0A%0A&repo=schemalabz%2Fopencouncil&pr=379&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=3">Fix All in Claude Code</picture></a>

`````markdown

---

### Issue 1 of 1
src/components/meetings/transcript/SegmentContext.tsx:65-71
**Clicking a compact row does not start playback**

`handleActivate` only calls `seekToWithoutScroll` — it never calls `setIsPlaying(true)`. The PR description states "clicking any compact row seeks to that segment and **starts playing**" and test plan item 3 says "It seeks to that segment and **starts playing**." With this implementation, a paused user who clicks a row stays paused. The `setIsPlaying` function was added to `VideoActionsContext` in this same PR (presumably for exactly this call-site) but is never invoked here — the test plan check would fail as-written.


`````

</details>

<sub>Reviews (12): Last reviewed commit: ["fix(transcript): apply view mode before ..."](https://github.com/schemalabz/opencouncil/commit/359a77b032fe68350fd271a4a1d2db1158d0da5a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33653311)</sub>

### DIFF
--- a/messages/el.json
+++ b/messages/el.json
@@ -895,5 +895,6 @@
         "successSummary": "Εστάλησαν: {sent}, απέτυχαν: {failed}.",
         "failedListWithCount": "Διευθύνσεις που απέτυχαν ({count})",
         "errorTitle": "Η αποστολή απέτυχε"
-    }
+    },
+    "transcript": {}
 }
--- a/messages/el/transcript.json
+++ b/messages/el/transcript.json
@@ -76,5 +76,15 @@
         "button": "Αντιγραφή τμήματος",
         "segmentCopied": "Το τμήμα αντιγράφηκε",
         "copySourceAttribution": "Αντιγράφηκε από το OpenCouncil · opencouncil.gr"
+    },
+    "viewModes": {
+        "toggleLabel": "Εναλλαγή προβολής fish-eye",
+        "default": "Προεπιλεγμένη προβολή",
+        "fisheye": "Προβολή fish-eye",
+        "expand": "Επέκταση τμήματος",
+        "collapse": "Σύμπτυξη",
+        "seekHere": "Αναπαραγωγή από εδώ",
+        "seekHereDetailed": "Αναπαραγωγή από {speaker} στις {time}",
+        "aiSummary": "Περίληψη από ΤΝ"
     }
 }
--- a/messages/el/transcript.json
+++ b/messages/el/transcript.json
@@ -83,8 +83,8 @@
         "fisheye": "Προβολή fish-eye",
         "expand": "Επέκταση τμήματος",
         "collapse": "Σύμπτυξη",
-        "seekHere": "Αναπαραγωγή από εδώ",
-        "seekHereDetailed": "Αναπαραγωγή από {speaker} στις {time}",
+        "seekHere": "Μετάβαση εδώ",
+        "seekHereDetailed": "Μετάβαση σε {speaker} στις {time}",
         "aiSummary": "Περίληψη από ΤΝ"
     }
 }
--- a/src/components/meetings/transcript/SegmentContext.tsx
+++ b/src/components/meetings/transcript/SegmentContext.tsx
@@ -47,12 +47,12 @@ function buildPreview(segment: TranscriptType[number]): Preview {
 
 /**
  * Compact one-line view used for every segment outside the active 3-item
- * focus window. Click the whole row to seek+play that segment — playback
- * advance is the only way to change which segments are in focus.
+ * focus window. Click the row to move the playhead there — playback advance
+ * is what changes which segments are in focus.
  */
 const SegmentContext = React.memo(({ segment }: SegmentContextProps) => {
     const { getPerson, getSpeakerTag } = useCouncilMeetingMeta();
-    const { seekToWithoutScroll, setIsPlaying } = useVideoActions();
+    const { seekToWithoutScroll } = useVideoActions();
     const t = useTranslations('transcript.viewModes');
 
     const speakerTag = getSpeakerTag(segment.speakerTagId);
@@ -63,13 +63,11 @@ const SegmentContext = React.memo(({ segment }: SegmentContextProps) => {
     const preview = useMemo(() => buildPreview(segment), [segment]);
 
     const handleActivate = () => {
+        // Move the playhead only — never force playback. The 4Hz poll in
+        // Transcript picks up the new position and expands this row within
+        // ~250ms. If the user was already playing they continue from here;
+        // if paused, they stay paused. Expanding a row must not start audio.
         seekToWithoutScroll(segment.startTimestamp);
-        // setIsPlaying is typed as void-returning but internally awaits
-        // playerRef.current.play(), which can reject (autoplay policy, media
-        // not ready). Surface those so the user-visible "seeked but not
-        // playing" state at least leaves a console trail.
-        Promise.resolve(setIsPlaying(true) as unknown as void | Promise<void>)
-            .catch(err => console.error('Fish-eye seek: failed to start playback', err));
     };
 
     const speakerName = person?.name ?? speakerTag?.label ?? null;
--- a/src/components/meetings/transcript/SegmentContext.tsx
+++ b/src/components/meetings/transcript/SegmentContext.tsx
@@ -1,0 +1,119 @@
+"use client";
+import React, { useMemo } from 'react';
+import { useCouncilMeetingMeta } from "../CouncilMeetingDataContext";
+import { useVideoActions } from "../VideoProvider";
+import { Transcript as TranscriptType } from '@/lib/db/transcript';
+import { PersonBadge } from '@/components/persons/PersonBadge';
+import { Bot } from "lucide-react";
+import { formatTimestamp, getPartyFromRoles } from "@/lib/utils";
+import { stripMarkdown } from '@/lib/formatters/markdown';
+import { useTranslations } from 'next-intl';
+
+const SHORT_UTTERANCE_MAX_CHARS = 160;
+const SHORT_UTTERANCE_MAX_SENTENCES = 2;
+
+interface SegmentContextProps {
+    segment: TranscriptType[number];
+}
+
+interface Preview {
+    text: string;
+    isAI: boolean;
+}
+
+/**
+ * Picks what to render below the speaker name in context mode.
+ * Short speeches (≤ ~2 sentences) display verbatim — readers don't need a
+ * summary for a one-liner. Longer speeches use the AI summary when available;
+ * otherwise we clip the verbatim text to the short-speech budget and append
+ * an ellipsis so the row visually signals it's a truncated excerpt rather
+ * than a complete short utterance. The isAI flag drives a small icon so the
+ * reader can tell at a glance which rows show synthetic vs verbatim text.
+ */
+function buildPreview(segment: TranscriptType[number]): Preview {
+    const joined = segment.utterances.map(u => u.text).join(' ').trim();
+    const sentenceCount = joined ? (joined.match(/[.!?]+(?:\s|$)/g)?.length ?? 1) : 0;
+    const isShort = joined.length > 0
+        && joined.length <= SHORT_UTTERANCE_MAX_CHARS
+        && sentenceCount <= SHORT_UTTERANCE_MAX_SENTENCES;
+    if (isShort) return { text: joined, isAI: false };
+
+    const summaryText = segment.summary?.text;
+    if (summaryText) return { text: stripMarkdown(summaryText).trim(), isAI: true };
+    if (!joined) return { text: '', isAI: false };
+    const clipped = joined.slice(0, SHORT_UTTERANCE_MAX_CHARS).trimEnd();
+    return { text: `${clipped}…`, isAI: false };
+}
+
+/**
+ * Compact one-line view used for every segment outside the active 3-item
+ * focus window. Click the whole row to seek+play that segment — playback
+ * advance is the only way to change which segments are in focus.
+ */
+const SegmentContext = React.memo(({ segment }: SegmentContextProps) => {
+    const { getPerson, getSpeakerTag } = useCouncilMeetingMeta();
+    const { seekToWithoutScroll, setIsPlaying } = useVideoActions();
+    const t = useTranslations('transcript.viewModes');
+
+    const speakerTag = getSpeakerTag(segment.speakerTagId);
+    const person = speakerTag?.personId ? getPerson(speakerTag.personId) : undefined;
+    const party = person ? getPartyFromRoles(person.roles) : null;
+    const borderColor = party?.colorHex || '#D3D3D3';
+
+    const preview = useMemo(() => buildPreview(segment), [segment]);
+
+    const handleActivate = () => {
+        seekToWithoutScroll(segment.startTimestamp);
+        // setIsPlaying is typed as void-returning but internally awaits
+        // playerRef.current.play(), which can reject (autoplay policy, media
+        // not ready). Surface those so the user-visible "seeked but not
+        // playing" state at least leaves a console trail.
+        Promise.resolve(setIsPlaying(true) as unknown as void | Promise<void>)
+            .catch(err => console.error('Fish-eye seek: failed to start playback', err));
+    };
+
+    const speakerName = person?.name ?? speakerTag?.label ?? null;
+    const timestamp = formatTimestamp(segment.startTimestamp);
+    // The button's aria-label fully replaces its inner text for screen readers,
+    // so the AI indicator has to live here too — the Bot SVG below is
+    // decorative for AT.
+    const baseLabel = speakerName
+        ? t('seekHereDetailed', { speaker: speakerName, time: timestamp })
+        : t('seekHere');
+    const ariaLabel = preview.isAI ? `${baseLabel} · ${t('aiSummary')}` : baseLabel;
+
+    return (
+        <button
+            type="button"
+            onClick={handleActivate}
+            aria-label={ariaLabel}
+            className="my-1 w-full text-left rounded-r-md border-l-[3px] sm:border-l-4 bg-muted/20 hover:bg-muted/40 transition-colors px-2.5 sm:px-4 py-1.5 flex items-baseline gap-2 overflow-hidden"
+            style={{ borderLeftColor: borderColor }}
+        >
+            <PersonBadge
+                person={person}
+                speakerTag={speakerTag}
+                variant="inline"
+                lastNameOnly
+                className="shrink-0"
+            />
+            {preview.text && (
+                <span className="text-xs sm:text-sm text-muted-foreground truncate flex-1 min-w-0 flex items-center gap-1">
+                    {preview.isAI && (
+                        <Bot
+                            className="h-3 w-3 shrink-0 opacity-70"
+                            aria-hidden="true"
+                        />
+                    )}
+                    <span className="truncate">{preview.text}</span>
+                </span>
+            )}
+            <span className="text-[10px] text-muted-foreground tabular-nums shrink-0">
+                {timestamp}
+            </span>
+        </button>
+    );
+});
+SegmentContext.displayName = 'SegmentContext';
+
+export default SegmentContext;
--- a/messages/en.json
+++ b/messages/en.json
@@ -895,5 +895,6 @@
         "successSummary": "Sent: {sent}, failed: {failed}.",
         "failedListWithCount": "Failed addresses ({count})",
         "errorTitle": "Send failed"
-    }
+    },
+    "transcript": {}
 }
--- a/messages/en/transcript.json
+++ b/messages/en/transcript.json
@@ -76,5 +76,15 @@
     "button": "Copy segment",
     "segmentCopied": "Segment copied",
     "copySourceAttribution": "Copied from OpenCouncil · opencouncil.gr"
+  },
+  "viewModes": {
+    "toggleLabel": "Toggle fish-eye view",
+    "default": "Default view",
+    "fisheye": "Fish-eye view",
+    "expand": "Expand segment",
+    "collapse": "Collapse",
+    "seekHere": "Play from here",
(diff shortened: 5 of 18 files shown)

### CODE REVIEW

**Comment by angelospk** [comment]
Addressed all four greptile bot's comments:

* `CouncilMeetingWrapper` resize handler now only force-shows the controls on wide viewports. Narrow resizes — and incidental height changes like a soft keyboard popping up — no longer wipe out a manual toggle.
* `SegmentContext` switched from `useVideo()` to `useVideoActions()`. I added `setIsPlaying` to `VideoActionsContextType` (it's just the existing `stableSetIsPlaying`), so compact rows stop re-rendering on every playback tick.
* `FisheyeToggle` tooltip ternary: desktop now gets `top`, mobile keeps `left`. The previous one was a no-op.
* `messages/el/transcript.json`: `aiSummary` re-indented to actually nest under `viewModes`. Matches the English file now.

**Comment by VasiaKoum** [comment]
Overall I like this approach — the compact version makes the transcript much easier to read, and it's convenient for parsing through the whole thing and actually reaching the end. A few pieces of feedback:

- **Auto-play on expand.** When I click a fisheye row, it expands correctly — but it also starts playing the video/council meeting even if I hadn't started playback before. This is a problem if the user doesn't want audio at that particular moment. Expanding a row shouldn't trigger playback.

- **Unclear collapse behavior.** Once a fisheye row is open, it's not obvious how to collapse it. The fisheye visually resembles an accordion (expand/collapse), so users will expect accordion-like behavior — but it doesn't behave that way, which is confusing.

- **Different content shown for summarized segments.** When a speaker segment has AI-summarization content (the speaker contribution), the fisheye shows that summary instead of the actual speaker segment. See the attached screenshot for an example.
Screenshot 2026-06-03 at 17 55 41

- **Floating action button (minor).** I'm not a fan of the floating button icon — it's not clear what the action does. We could instead place this action above the transcript. Right now there's nothing in that area, so we have room to take advantage of it.

Let me know your thoughts. Again, I really like the compact version overall — it's a nice improvement for readability.

**Comment by angelospk** [comment]
Thanks for the thorough review, and for the kind words. Glad the compact view is landing well. Here's where we landed on each point.

**1. Auto-play on expand.** Agreed, that's a bug. Expanding a row shouldn't touch playback. We're going a bit further than just removing the auto-play: clicking a row will only expand it for reading. Moving playback to that point becomes a separate, explicit action (the timestamp, or a play control). That keeps "I want to read ahead" and "I want to jump there" apart. If you were paused, you stay paused.

**2. Collapse behavior.** This is the real design question. The view isn't actually an accordion, it's a fisheye: the three segments nearest the playhead expand automatically (current one in the middle), everything else stays compact. There's no manual collapse because the focus window just follows playback. I can see why it reads as an accordion though. Two things to fix that:
- Give the currently-playing rows a distinct "now playing" look so they don't read as generic expanded panels.
- Let you manually expand/collapse any other row to peek ahead, with a clear chevron. That's the accordion behavior you'd expect, just for the rows outside the auto-focus window.

We might also add a one-time hint on first open, but as a bonus rather than the main fix.

**3. Summary vs actual segment.**  Today the rule is: short speeches show verbatim, longer ones show the AI summary when one exists, and longer ones with no summary fall back to clipped verbatim. So summaries aren't available for every segment, which means "show summaries everywhere" isn't reliably possible. The actual problem is that summary rows aren't clearly marked as summaries (there's only a small bot icon that's easy to miss), so it looks like we're hiding the real text. Fix is to mark summarized rows clearly, so it reads as "here's the gist, expand for the full words" instead of swapped-out content. Curious whether you think gist-first is even the right call: I think it scans better for long speeches, but happy to be talked out of it.

**4. Floating button.** Agreed, the icon's unclear. Moving it to a labeled control above the transcript, where there's room anyway.

I'll probably split this so the quick fixes (auto-play, summary labeling, the toggle) land in this PR and the manual expand/collapse follows up. Let me know if you'd rather see it all in one go.

File: messages/el.json [hunk 1]
```diff
@@ -895,5 +895,6 @@
         "successSummary": "Εστάλησαν: {sent}, απέτυχαν: {failed}.",
         "failedListWithCount": "Διευθύνσεις που απέτυχαν ({count})",
         "errorTitle": "Η αποστολή απέτυχε"
-    }
-}
\ No newline at end of file
+    },
+    "transcript": {}
```
VasiaKoum, line 899: I think we can remove this (remove this from the `messages/en.json` as well)

src/components/meetings/transcript/SegmentContext.tsx
```diff
@@ -0,0 +1,117 @@
...
+ * Short speeches (≤ ~2 sentences) display verbatim — readers don't need a
+ * summary for a one-liner. Longer speeches use the AI summary when available;
+ * otherwise we clip the verbatim text to the short-speech budget and append
+ * an ellipsis so the row visually signals it's a truncated excerpt rather
+ * than a complete short utterance. The isAI flag drives a small icon so the
+ * reader can tell at a glance which rows show synthetic vs verbatim text.
+ */
+function buildPreview(segment: TranscriptType[number]): Preview {
+    const joined = segment.utterances.map(u => u.text).join(' ').trim();
+    const sentenceCount = joined ? (joined.match(/[.!?]+(?:\s|$)/g)?.length ?? 1) : 0;
```
VasiaKoum, line 35: Greek question sentences end with `;` (U+037E / `;`), so Greek questions won't count as sentence boundaries.

messages/el/transcript.json
```diff
@@ -76,5 +76,15 @@
         "button": "Αντιγραφή τμήματος",
         "segmentCopied": "Το τμήμα αντιγράφηκε",
         "copySourceAttribution": "Αντιγράφηκε από το OpenCouncil · opencouncil.gr"
+    },
+    "viewModes": {
+        "toggleLabel": "Εναλλαγή προβολής fish-eye",
```
VasiaKoum, line 81: Let's use the text "Εναλλαγή συμπιεσμένης προβολής" or "Εναλλαγή συμπιεσμένης προβολής (fish-eye)" in order to educate the user that this is called fish-eye pattern.

VasiaKoum (commented):
Tested the fixes, thank you for your quick turnaround! On each point:

**2.** Let's hold off on the collapse functionality for now. We need to test the current behavior with our power users first, to see whether they can navigate the transcript easily or need something more.

One small bug I found: on https://pr-379.preview.opencouncil.gr/zografou/apr24_2026/transcript?mode=fisheye, if you click the 4th element ("Έχω εξηγήσει και στη Δημοτική Επιτροπή που πέρασε..."), it isn't shown as the middle (focused) element — it appears as the last one instead.

**3.** I'd suggest showing the speaker segment for all utterances. Then, if the user clicks an utterance, they can read both the AI-generated summary and the actual speaker segment. I don't feel strongly about it, but I think it keeps things consistent (active vs. passive voice, the flow of the discussion). What do you think?

Also, a comment on the implementation:

Right now, opening a shared link like `…/zografou/apr24_2026/transcript?mode=fisheye&t=5308` doesn't just render fisheye for that page view — `hydrate()` writes `fisheye` into localStorage (`writeToStorage(fromUrl)` in useViewMode.ts). So from then on, **every** transcript that person opens in that browser defaults to fisheye, with no `?mode=` in the URL anymore.

The problem is reversibility for non-technical users. Someone who was simply sent a link now sees every meeting in an unfamiliar layout, and there's no obvious mental model for why or how to undo it. The only way back is to recognize it's a "mode" and find the toggle, which most casual readers won't do. Effectively a single shared link silently and durably changes their default.

I don't think we need that. Issue #109 only asks to "persist via query param (e.g., mode=fisheye) or in OptionsContext". So I'd suggest:

- Treat `?mode=` as an **ephemeral, per-link/per-session override** — render fisheye for that visit, but **don't** write it to localStorage.

@kouloumos we'd also like to share your input on this.

kouloumos:
Appreciate the effort on this PR! I'm still a fan of the fish-eye concept, but the UX in this current implementation doesn't provide the utility we need for the end user.

image

Because this needs a fundamental rethink of the design, I’m going to close this PR for now. Let’s head back to the original issue to brainstorm and nail down the UX before we commit more engineering and review time to it.

PR #384, opened 2026-05-27
fix(highlights): delete orphaned highlight when canceling without saving

### LINKED ISSUE #353: Canceling a freshly-created highlight leaves an orphan unnamed highlight
From a user-feedback session in Chalandri (2026-05-13): the user clicked to create a new highlight, then clicked "Cancel" — and was redirected to `/highlights` where a new, unnamed highlight had appeared in the list.

### Root cause

In [`HighlightContext.tsx:509-540`](src/components/meetings/HighlightContext.tsx#L509-L540), `createHighlight()` does a `POST /api/highlights` *before* the user enters edit mode. When the user then cancels, [`exitEditMode()`](src/components/meetings/HighlightContext.tsx#L440-L444) just routes away — the freshly-created record is never cleaned up.

### Fix options

1. **Defer creation until first save** — keep the new highlight in component state only; POST on save. Cleanest but a larger refactor.
2. **DELETE on cancel-if-pristine** — when `exitEditMode()` runs and the highlight is still empty/unnamed, call `DELETE /api/highlights/{id}` before navigating. Smaller change.

Option 2 is the minimal fix; option 1 is the right long-term shape.

angelospk:
Closes #353

## Description

When a user clicked "+ Highlight" and then immediately clicked "Cancel" without making any edits, the freshly-created highlight was left behind in the database and showed up as an unnamed entry on `/highlights`.

The root cause is in `HighlightContext.tsx`: `createHighlight()` POSTs to `/api/highlights` up front (before edit mode is even entered), but `exitEditMode()` only navigates away, so the orphan record is never cleaned up. Reported during a user-feedback session in Chalandri on 2026-05-13.

This PR applies the minimal fix (Option 2 from the issue): DELETE the highlight on cancel when it was never saved. A deferred-creation refactor (Option 1) would be cleaner long-term but is out of scope here.

## Changes Made

- `src/components/meetings/HighlightContext.tsx`:
  - Added a `unsavedNewHighlightIdRef` ref to track a highlight created in the current session that has not yet been saved.
  - `createHighlight()` sets the ref to the new highlight's ID after the POST succeeds.
  - `saveHighlight()` clears the ref on a successful PUT, so cancels after a save do not delete the highlight.
  - `exitEditMode()` checks the ref; if it matches the current editing highlight, it calls `removeHighlight()` locally and fires `DELETE /api/highlights/{id}` before navigating.
  - The navigate-away cleanup `useEffect` mirrors the same logic so back-button or URL-change exits behave the same way.
  - The ref is cleared synchronously before `router.push`, so the cleanup effect that fires post-navigation does not issue a second DELETE.

### Network / request tracing

The lifecycle endpoints were verified in the browser Network tab.

**1. Fresh creation and cancel**
1. Clicking "+ Highlight" triggers:
   - `POST http://localhost:3000/api/highlights` → `201 Created`
   - Response includes a new ID (e.g. `cmpnrnxi60001oeujy8t6kn9l`)
2. Clicking "Cancel" immediately triggers:
   - `DELETE http://localhost:3000/api/highlights/cmpnrnxi60001oeujy8t6kn9l` → `200 OK`
   - Highlight count stays at `0`. No orphaned "Unnamed Highlight" appears.

**2. Save and regression check**
1. Creating a highlight and toggling an utterance enables the save button.
2. Saving triggers `PUT http://localhost:3000/api/highlights/cmpnrpc600002oeuj48nud7sc` → `200 OK`.
3. Exiting edit mode after a save (with or without further modifications) does NOT fire a `DELETE` for the saved highlight.
4. The saved highlight stays intact in the database and the list views.

### Results

- Orphan prevention works: freshly-created highlights that are cancelled without ever being saved get removed from both the local client state and the Postgres DB.
- REST pairing verified: `POST /api/highlights` (201) is paired with `DELETE /api/highlights/{id}` (200) on cancel.
- No regression on saved highlights: existing highlights are left untouched when edit sessions are cancelled.
- Confirmation prompt still works: discarding changes to an existing highlight shows the native confirmation dialog, and accepting it discards local changes without deleting the highlight from the DB.
- Console clean: no uncaught browser exceptions during any step.

<h3>Greptile Summary</h3>

This PR fixes the orphaned-highlight bug where clicking \"+ Highlight\" immediately followed by \"Cancel\" left an unnamed DB record behind. The fix adds a `unsavedNewHighlightIdRef` in `HighlightContext` to track freshly-created (not yet saved) highlights, then fires a `DELETE` on cancel, on SPA navigation away, and on page-hide/browser-unload.

- **Core fix** (`HighlightContext.tsx`): `createHighlight` sets the ref; `saveHighlight` clears it; `exitEditMode` and the navigation `useEffect` both call the new `deleteUnsavedHighlightIfNeeded` helper, which fires DELETE only on a successful response before calling `removeHighlight` locally.
- **UX upgrade** (`HighlightModeBar.tsx`): replaces the native `confirm()` call with a custom `<Dialog>` that distinguishes \"discard a new, never-saved highlight\" from \"discard unsaved edits to an existing one,\" with tailored copy for each case.
- **Guard fix** (`Transcript.tsx`): adds `prevHighlightParamRef` to prevent the URL-sync effect from re-entering edit mode when `exitEditMode` clears `editingHighlight` while the `?highlight` param is still being removed mid-navigation.

Safe to merge — the fix is targeted and well-guarded, with no changes to shared data paths.

All three exit paths (in-app Cancel, SPA navigation away, page unload) consistently use the same ref-guarded helper, preventing double-DELETE. The local state update is deferred until the server confirms, so a network failure leaves the orphan visible with a toast rather than silently corrupting local state. The Transcript.tsx guard correctly prevents re-entering edit mode during the brief window between state clear and URL update.

| Filename | Overview |
|----------|----------|
| src/components/meetings/HighlightContext.tsx | Core fix — adds unsavedNewHighlightIdRef, deleteUnsavedHighlightIfNeeded helper, pagehide/beforeunload handlers, and reworks exitEditMode to stay on transcript. Logic is sound: double-DELETE is prevented via synchronous ref clear, removeHighlight only fires on server confirmation, and the navigation cleanup effect correctly uses segment-aware isTranscriptPath. |
| src/components/meetings/HighlightModeBar.tsx | Replaces native confirm() with a custom Dialog; correctly differentiates 'delete new highlight' vs 'discard unsaved edits' paths; Cancel button is disabled during save to prevent interleaved state. |
| src/components/meetings/transcript/Transcript.tsx | Adds prevHighlightParamRef to gate the URL-sync effect, preventing re-entry into edit mode when exitEditMode clears state while the ?highlight param is being removed. |
| messages/en/highlights.json | Adds i18n keys for the new custom Dialog and error toast; removes obsolete confirmExit and returningToHighlights keys. |
| messages/el/highlights.json | Greek translations matching the English key additions/removals; copy is consistent with the English strings. |

</details>

<sub>Reviews (12): Last reviewed commit: ["fixup! fix(highlights): delete orphaned ..."](https://github.com/schemalabz/opencouncil/commit/0349096d3ee51da6d4c279dc65045fc06d7c1341) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34141635)</sub>


VasiaKoum:
Nice work! I have two comments about the functionality:

https://github.com/user-attachments/assets/99fde03d-42d8-4153-bbb5-2c6eedbb29d4

1. As you can see in the screen recording, I had to click the highlights button twice before I could choose which utterances to include in the highlight (i.e. to enable edit mode). It should only take one click.

2. If I start building a highlight and then decide to exit, it navigates me to /highlights and discards everything. Instead, it should first ask the user to confirm — warning them that exiting will cancel the highlight flow (along with the utterances they've already chosen) — and only hide the yellow banner if they confirm.

angelospk:
> Nice work! I have two comments about the functionality:
> 
>  Screen.Recording.2026-06-03.at.17.07.11.mov 
> 1. As you can see in the screen recording, I had to click the highlights button twice before I could choose which utterances to include in the highlight (i.e. to enable edit mode). It should only take one click.
> 2. If I start building a highlight and then decide to exit, it navigates me to /highlights and discards everything. Instead, it should first ask the user to confirm — warning them that exiting will cancel the highlight flow (along with the utterances they've already chosen) — and only hide the yellow banner if they confirm.

thanks for the review!
Regarding 2, it has a native window dialog before exiting. So, should I keep the same functionality but change the native window dialog to a UI styled Alert Dialog instead?

angelospk:
@greptile review

src/components/meetings/HighlightModeBar.tsx
```diff
@@ -76,15 +82,22 @@ export function HighlightModeBar() {
   };
 
   const handleCancel = () => {
-    if (hasUnsavedChanges) {
-      const confirmed = confirm(t('exit.confirmExit'));
```
VasiaKoum, line 80: If nothing else references `exit.confirmExit`, we can remove it from locale files.

VasiaKoum (commented):
Nice! One small bug, though: if I start choosing utterances for an unsaved highlight and then refresh the page, the chosen utterances get deselected and an orphaned highlight with no utterances shows up on `/highlights`.

angelospk:
I think this happens because when you select certain phrases the system automatically creates a new highlight, and if we don’t discard it ourselves it may be kept. I’m not sure how to handle this when we refresh or leave the page. Do you have any suggestions?

VasiaKoum:
I think it would be a good idea to warn before losing the changes — but heads up on one browser limitation: for a hard exit (refresh, tab close, leaving the site) we **can't reuse our custom Dialog**. `beforeunload` is synchronous, and our Dialog is async (set state → re-render → wait for a click), so it can't pause the unload. Browsers also block custom UI here for security. The most we can do is ask the browser to show its **own native confirmation alert** (the generic "Leave site? Changes you made may not be saved.").

So the split is:
- **In-app Cancel** → our nice custom Dialog (what we already built — no native alert needed there).
- **Hard exit (refresh/close)** → fall back to the browser's native alert, plus a best-effort cleanup if they do leave.

One important detail: the cleanup DELETE must run on `pagehide`, **not** `beforeunload` — `beforeunload` fires even when the user clicks "Stay", which would wrongly delete a highlight they're still editing. `pagehide` only fires once the page is actually unloading.

```ts
useEffect(() => {
  // Hard exit can't use our Dialog (sync + browser security), so we trigger
  // the browser's native "unsaved changes" alert to warn the user.
  const onBeforeUnload = (e: BeforeUnloadEvent) => {
    if (!unsavedNewHighlightIdRef.current) return;
    e.preventDefault();
    e.returnValue = ''; // required to trigger the native alert
  };

  // Runs only on an actual unload (not when the user clicks "Stay"), so this is
  // the safe place for the best-effort cleanup. keepalive lets the DELETE outlive
  // the document; we call DELETE directly since sendBeacon can only POST.
  const onPageHide = () => {
    const id = unsavedNewHighlightIdRef.current;
    if (!id) return;
    unsavedNewHighlightIdRef.current = null;
    fetch(`/api/highlights/${id}`, { method: 'DELETE', keepalive: true });
  };

  window.addEventListener('beforeunload', onBeforeUnload);
  window.addEventListener('pagehide', onPageHide);
  return () => {
    window.removeEventListener('beforeunload', onBeforeUnload);
    window.removeEventListener('pagehide', onPageHide);
  };
}, []);
```
Do you want to try it at your implementation?

angelospk:
Thanks! Pushed the hard-exit handling you suggested.

- Refresh / tab close / leaving the site now triggers the browser's native unsaved-changes alert when there's an unsaved highlight or dirty edits. The custom Dialog can't run during a synchronous unload, so this is the fallback.
- Cleanup runs on `pagehide`, not `beforeunload`, so clicking "Stay" never deletes a highlight you're still editing. It only deletes a brand-new unsaved highlight; existing ones are left alone.
- Listeners attach only while editing so they don't hurt bfcache.
- Also disabled Cancel while a save is in flight (it could race the exit) and dropped the now-unused `exit.confirmExit` string.

The double-click-to-edit and the in-app confirm-before-discard were already fixed in earlier commits on the branch.

PR #410, opened 2026-06-02
fix(notifications): batch release emails so login emails aren't starved

### LINKED ISSUE #380: Login emails delayed while bulk notification emails are being sent
## Summary

Users attempting to log in experience delayed verification emails when the system is actively sending bulk notification emails (e.g., notifications for Athens council meetings). The auth and notification email paths share a single Resend API key, and the sequential notification sending loop saturates the rate limit, starving auth emails.

## Impact
- **Affected**: Any user trying to log in while a notification batch is being released
- **Since**: Notification system launch (2025-11-27 rate-limit delay added, but the contention existed from day one)
- **Severity**: Degraded experience — login is blocked for the duration of the notification send (potentially minutes)

## Background

The platform sends emails through Resend for two distinct purposes:
1. **Auth emails** — magic-link login, sent via direct `fetch` to `https://api.resend.com/emails` in `src/auth.config.ts:39-52`
2. **Notification emails** — bulk alerts about new council meetings, sent sequentially in `src/lib/notifications/deliver.ts:28-56` with a 500ms delay between each send

Both paths use the same `RESEND_API_KEY`. Resend enforces per-key rate limits, so when one path is sending at capacity, the other is throttled.

## Root Cause

`releaseNotifications()` in `src/lib/notifications/deliver.ts` processes deliveries **one at a time in a loop**, with an explicit 500ms sleep between sends (line 49). For a city like Athens with many subscribers, this loop can run for minutes, continuously consuming the shared Resend rate limit.

Meanwhile, auth emails hit the same API key. When the rate limit is exhausted by notification sends, Resend throttles or queues auth requests, causing login email delays.

The 500ms delay was introduced in `b8f84452` (2025-11-27) as a quick fix to avoid hitting Resend rate limits during notification sends. A `sendEmailBatch()` function was later added in `252e6e82` (2026-04-29) for the product-update feature, which sends up to 100 emails per API call — but `releaseNotifications()` was never updated to use it.

<details>
<summary>

## Evidence

</summary>

**Observed in production** (2026-05-25): Login email delayed while Athens notification batch was being sent (reported by kouloumos in Discord).

**Code path — notifications** (`src/lib/notifications/deliver.ts:28-56`):
- Sequential loop over deliveries
- Line 49: `await new Promise(resolve => setTimeout(resolve, 500));` — 500ms forced delay
- Each email sent individually via `sendEmail()` → `resend.emails.send()`

**Code path — auth** (`src/auth.config.ts:39-52`):
- Direct `fetch` to `https://api.resend.com/emails`
- No queue, no retry, no priority mechanism

**Shared resource**: Both use `env.RESEND_API_KEY` — single rate limit pool at Resend's end.

**Batch capability exists but unused by notifications**:
- `src/lib/email/resend.ts:139-222` — `sendEmailBatch()` supports up to 100 emails per call
- Currently only used by `src/lib/email/productUpdate.ts`

</details>

## Fix Options

| # | Approach | Effort | Addresses |
|---|----------|--------|-----------|
| 1 | Migrate `releaseNotifications()` to use `sendEmailBatch()` | Low | Cause — reduces API calls by ~100x, freeing rate limit headroom for auth emails |
| 2 | Separate Resend API keys for auth vs notifications | Low | Cause — isolates rate limit pools entirely |
| 3 | Priority queue with auth emails taking precedence | High | Symptom — auth emails skip ahead but rate limit is still shared |

Option 1 is the lowest-effort fix since `sendEmailBatch()` already exists and is proven in the product-update flow. Option 2 could complement it for full isolation.

<details>
<summary>

## Investigation Trail

</summary>

1. Observed login email delay during Athens notification send (Discord report, 2026-05-25)
2. Checked auth email path — direct fetch to Resend API in `src/auth.config.ts`, no queuing
3. Checked notification path — sequential loop with 500ms delay in `src/lib/notifications/deliver.ts`
4. Confirmed both paths use the same `RESEND_API_KEY` via `src/env.mjs`
5. Found `sendEmailBatch()` in `src/lib/email/resend.ts` — supports 100 emails/call, but only used by product updates
6. Checked git history: batch function introduced in `252e6e82` (2026-04-29) for product updates; notification code predates it and was never updated
7. The 500ms delay (`b8f84452`, 2025-11-27) was a rate-limit workaround — batch sending would eliminate the need for it

</details>

angelospk:
Closes #380

## Description

Magic-link login emails and bulk notification emails share one Resend API key. `releaseNotifications()` sent notifications one at a time with a 500ms gap between each, which saturated Resend's per-key rate limit for minutes during a release and left login emails waiting in line behind them.

This moves the email path to Resend's batch endpoint (`sendEmailBatch`, up to 100 per call), so a release of N emails takes a handful of API calls instead of N sequential sends. The auth path no longer gets starved.

Bird (SMS/WhatsApp) messages keep the 500ms-per-message spacing. Bird has no batch endpoint, and its rate-limit concern is a separate problem.

## Changes Made

- `src/lib/notifications/deliver.ts`:
  - Partition pending deliveries into email vs message up front.
  - Emails are chunked into batches of 100 and sent via `sendEmailBatch`, replacing the old per-recipient loop with its 500ms delay.
  - Split the old `sendEmailDelivery` into a pure `buildEmailPayload` (subject/html/to) plus the per-delivery status writes, so a batch response can be mapped back to individual delivery rows.
  - Failure attribution is by delivery slot rather than by email address. When two deliveries go to the same recipient and Resend lists that address once in `failedTos`, a small failure budget marks only one row failed instead of both.
  - Deterministic idempotency key per batch (from the sorted delivery ids) so replaying the same set doesn't double-send.
  - The `Release complete: X emails, Y messages, Z failed` log line stays.
- `src/lib/notifications/__tests__/deliver.test.ts`: new unit tests for the batching path.

## Testing

I don't have Resend credentials, so I couldn't run the full UI-to-inbox release flow locally. Instead I covered the change with unit tests against a mocked `sendEmailBatch`:

- single batch when count is 100 or fewer, with no per-email delay
- chunking into batches of 100
- per-recipient failures mapped back from the batch response
- deliveries missing email/title/body marked failed before the batch send
- deterministic idempotency key for the same delivery set
- batch call skipped entirely when there are no email deliveries
- the duplicate-recipient case: two deliveries on the same address, one failure marks only one row

All 7 pass. `npx tsc --noEmit` is clean and the existing notification tests still pass.

A reviewer with Resend access can confirm the end-to-end behaviour: trigger a release of 100 or more pending email deliveries, watch the Resend logs for about N/100 batch calls instead of a per-email loop, then request a magic-link sign-in a second or two into the release. It should arrive within seconds rather than waiting for the release to drain.

Splitting auth and notification traffic onto separate Resend keys (`RESEND_AUTH_API_KEY`) would give full isolation, but that's a possible follow-up and out of scope here.

<h3>Greptile Summary</h3>

The PR batches notification emails through Resend while retaining sequential pacing for Bird messages, reducing contention with authentication emails.
- Sends email deliveries in chunks of 100 with deterministic idempotency keys.
- Adds 500 ms pacing between successive Resend batch calls.
- Maps provider failures back to individual delivery rows using occurrence-counted recipient failures.
- Adds regression coverage for batching, invalid payloads, idempotency, and duplicate-recipient failure handling.

The PR appears safe to merge.

No blocking failure remains; both previously reported issues are addressed by inter-batch pacing and explicit duplicate-recipient whole-batch regression coverage.

| Filename | Overview |
|----------|----------|
| src/lib/notifications/deliver.ts | Implements paced Resend batching, deterministic idempotency, validation, and per-delivery status attribution; the previously reported missing inter-batch delay is resolved. |
| src/lib/notifications/__tests__/deliver.test.ts | Adds focused batching tests, including the previously requested whole-batch duplicate-recipient failure case. |

<sub>Reviews (6): Last reviewed commit: ["fix(notifications): correct failure attr..."](https://github.com/schemalabz/opencouncil/commit/10167ea815020cd56b077da58a821034ce5fea21) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35166866)</sub>

### DIFF
--- a/src/lib/notifications/deliver.ts
+++ b/src/lib/notifications/deliver.ts
@@ -1,6 +1,7 @@
 "use server";
 
-import { sendEmail } from '@/lib/email/resend';
+import { createHash } from 'crypto';
+import { sendEmailBatch } from '@/lib/email/resend';
 import { getPendingDeliveries, updateDeliveryStatus } from '@/lib/db/notifications';
 import {
     createOrUpdateConversation,
@@ -9,8 +10,32 @@ import {
 import { sendAndPersistOutbound } from './outbound';
 import { env } from '@/env.mjs';
 
+const FROM_ADDRESS = 'OpenCouncil <notifications@opencouncil.gr>';
+const EMAIL_BATCH_SIZE = 100;
+const MESSAGE_DELAY_MS = 500;
+// Space successive batch calls so a multi-batch release stays under Resend's
+// per-key request rate limit. A 429 maps the whole chunk to failedTos, which
+// would drop ~100 notifications at once, so this is cheap insurance.
+const BATCH_DELAY_MS = 500;
+
+interface EmailPayload {
+    from: string;
+    to: string;
+    subject: string;
+    html: string;
+}
+
 /**
- * Release notifications by sending all pending deliveries
+ * Release notifications by sending all pending deliveries.
+ *
+ * Emails go through Resend's batch endpoint in chunks of 100, so a release of
+ * thousands of recipients consumes only a handful of API calls instead of
+ * thousands of sequential per-recipient sends with 500ms gaps. That sequential
+ * loop used to saturate the Resend rate limit for our shared API key, starving
+ * auth magic-link emails for minutes — see issue #380.
+ *
+ * Bird (WhatsApp/SMS) messages still go one-at-a-time with a 500ms gap; their
+ * API doesn't expose a batch endpoint and the rate concern there is separate.
  */
 export async function releaseNotifications(notificationIds: string[]): Promise<{
     success: boolean;
@@ -28,34 +53,32 @@ export async function releaseNotifications(notificationIds: string[]): Promise<{
 
         console.log(`Releasing ${pendingDeliveries.length} pending deliveries for ${notificationIds.length} notifications`);
 
-        // Process each delivery
-        for (const delivery of pendingDeliveries) {
+        // Partition by medium up front so emails can be batched and messages
+        // can keep their per-delivery loop.
+        const emailDeliveries = pendingDeliveries.filter((d: any) => d.medium === 'email');
+        const messageDeliveries = pendingDeliveries.filter((d: any) => d.medium === 'message');
+
+        // ---- Emails: batch send via Resend ----
+        const emailResult = await sendEmailDeliveriesBatched(emailDeliveries);
+        emailsSent += emailResult.sent;
+        failed += emailResult.failed;
+
+        // ---- Messages: sequential with rate-limit delay (unchanged) ----
+        for (const delivery of messageDeliveries) {
             try {
-                if (delivery.medium === 'email') {
-                    const result = await sendEmailDelivery(delivery);
-                    if (result) {
-                        emailsSent++;
-                    } else {
-                        failed++;
-                    }
-                } else if (delivery.medium === 'message') {
-                    const result = await sendMessageDelivery(delivery);
-                    if (result) {
-                        messagesSent++;
-                    } else {
-                        failed++;
-                    }
+                const result = await sendMessageDelivery(delivery);
+                if (result) {
+                    messagesSent++;
+                } else {
+                    failed++;
                 }
-
-                // Add a small delay to avoid rate limiting
-                // 500ms delay allows for ~2 requests per second, which is a safe limit for most services
-                await new Promise(resolve => setTimeout(resolve, 500));
-
             } catch (error) {
                 console.error(`Error sending delivery ${delivery.id}:`, error);
                 await updateDeliveryStatus(delivery.id, 'failed');
                 failed++;
             }
+            // Keep the 500ms spacing for Bird/SMS — only emails moved to batch.
+            await new Promise(resolve => setTimeout(resolve, MESSAGE_DELAY_MS));
         }
 
         console.log(`Release complete: ${emailsSent} emails, ${messagesSent} messages, ${failed} failed`);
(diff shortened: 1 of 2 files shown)


angelospk:
Both addressed.

The first one (no delay between successive batch calls): I added a 500ms pause between batch sends, just not before the first or after the last. A large release used to fire every chunk back-to-back, so a single 429 on Resend's per-key request limit would have failed a whole 100-email chunk at once. The pause keeps multi-batch releases under that limit and still costs a fraction of the old per-email loop.

The second one (duplicate recipient on a whole-batch failure): good catch, that was a real gap in the tests. There's now a case where the entire batch fails and `failedTos` lists the same address twice, asserting that both delivery rows get marked failed. The earlier "only one of the two fails" case is still there, so the failure-budget logic is pinned from both sides.

8/8 unit tests pass and `tsc --noEmit` is clean.

src/lib/notifications/deliver.ts
```diff
@@ -85,37 +112,102 @@ export async function releaseNotifications(notificationIds: string[]): Promise<{
...
+                await updateDeliveryStatus(delivery.id, 'failed');
+                console.error(`Failed to send email to ${payload.to}`);
+                failed++;
+            } else {
+                await updateDeliveryStatus(delivery.id, 'sent');
+                sent++;
+            }
+        }
+
+        if (!result.success) {
```
kouloumos: This is where the batch blast radius bites, and I think it is worth a small change before merge.

`sendEmailBatch` returns two very different kinds of failure in the same shape:

- **per-recipient** — permissive mode `errors[]`, so `failedTos` holds only the addresses Resend actually rejected, and `error` is **not** set.
- **transport-level** — a non-2xx (429, 5xx) or a thrown network error, where `failedTos` is **every address in the chunk** and `error` **is** set (`resend.ts:189-196`, and the `catch` at 212).

The loop above treats them identically, so one 429 marks up to 100 deliveries `failed`.

That is terminal. `getPendingDeliveries` selects `status: 'pending'` only, so a re-run will skip them, and there is no other retry path in the codebase. Before this change a transient blip cost one notification; now it costs the whole chunk.

Your `BATCH_DELAY_MS` comment says exactly this and calls the delay cheap insurance. It is — but it lowers the probability without changing the cost.

Since `error` is set only in the transport case, distinguishing them is easy:

```suggestion
        if (result.error) {
            // Transport-level failure (429, 5xx, network): every address in the
            // chunk is in failedTos, but none was individually rejected. Leave
            // these deliveries 'pending' so a re-run retries them — marking them
            // 'failed' is terminal, because getPendingDeliveries only selects
            // 'pending' and nothing else re-sends a failed delivery.
            console.error(`Notification email batch failed at transport level:`, result.error);
        } else if (!result.success) {
            console.error(`Notification email batch had per-recipient failures:`, result.failedTos);
        }
```

The suggestion above only fixes the logging. To make it real, the per-delivery loop needs to skip the status write in the transport case — happy to leave the shape of that to you, or to take it as it is if you would rather land the speedup now and treat retries as a separate issue. Either way I would like the two cases distinguished, because right now they are silently the same.

src/lib/notifications/deliver.ts
```diff
@@ -85,37 +112,102 @@ export async function releaseNotifications(notificationIds: string[]): Promise<{
...
+        const payloads = chunk.map((p) => p.payload);
+        const idempotencyKey = makeBatchIdempotencyKey(chunk.map((p) => p.delivery.id));
+
+        const result = await sendEmailBatch(payloads, { idempotencyKey });
+
+        // Resend reports failures as a list of `to` addresses (count == #fails).
+        // If two deliveries target the same address and only one fails, the
+        // address appears once. Consume the list as a multiset so we mark the
+        // right number of rows failed instead of every match.
+        const failureBudget = new Map<string, number>();
```
kouloumos: This is the part of the PR I would point at if someone asked whether it was carefully done.

Resend reports failures as addresses, not indices into your input, so two deliveries to the same address with one failure would naively mark both failed. Treating `failedTos` as a multiset with a per-address budget is correct, and you wrote the two tests that pin it.

No change requested. Recording it because it is the kind of detail that usually only shows up later as a bug report.

PR #411, opened 2026-06-03
Fix/task callback resilience

### LINKED ISSUE #360: Task response is lost when callback processing fails
## Summary

When the task callback processing throws an error (e.g. a Prisma failure), `handleTaskUpdate` overwrites the `responseBody` field with the error message, destroying the original task server response. This means even if the processing bug is fixed, the task must be re-run from scratch (and pay for the LLM call again) instead of replaying the existing valid response.

## Impact
- **Affected**: Any task type where the task server succeeds but callback processing fails
- **Severity**: Wasted resources — valid LLM responses are lost, requiring paid re-runs

## Root Cause

In `handleTaskUpdate` (`src/lib/tasks/tasks.ts:228-234`), the error handler overwrites `responseBody` with a combined string:

```typescript
const errorDetail = `Processing error: ${(error as Error).message}\n\n--- Original task server response ---\n${originalResponse}`;
await prisma.taskStatus.update({
    where: { id: taskId },
    data: { status: 'failed', responseBody: errorDetail, version: update.version }
});
```

The original response is embedded after a text marker, but not as clean replayable JSON. There's no mechanism to extract and replay it.

## Fix Options

| # | Approach | Effort | Addresses |
|---|----------|--------|-----------|
| 1 | Store the error in a separate field (e.g. `processingError`) and leave `responseBody` untouched | Medium | Clean separation, responseBody stays valid JSON |
| 2 | Only write to `responseBody` on success; on failure, store error details elsewhere | Medium | Same outcome, different implementation |
| 3 | Add a replay endpoint/utility that parses the embedded response from the error string | Low | Quick workaround but fragile |

angelospk:
Closes #360

## Description
This pull request changes how task failures are recorded and shown in the admin dashboard.

Currently, if the post-callback results processor throws an error, the application overwrites the task's `responseBody` with the error details. This prevents developers from replaying the task using the original payload after fixing the processor.

To fix this:
- I added a `processingError` column to the `TaskStatus` database model to store processor errors separately.
- I updated `src/lib/tasks/tasks.ts` to write the raw task server payload to `responseBody` before running the processor. If the processor fails, the task status becomes `failed` and the stack trace is saved in `processingError` without overwriting the original response.
- I updated the task status component in `src/components/meetings/admin/TaskStatus.tsx` to show the human-readable `processingError` stack trace in the expanded task card view.
- I added unit tests to verify the error handling and state persistence logic in `src/lib/tasks/__tests__/handleTaskUpdateProcessingError.test.ts`.

## Testing
I ran the following local tests:
- **Database & Migrations**: Removed an empty leftover migration directory (`20260518203511_enable_unaccent_extension`) to allow migrations to run. Applied the new migration (`20260527120000_task_status_processing_error`) to add the `processingError` column.
- **Fail Path End-to-End**: Inserted a mock pending `transcribe` task and sent a malformed payload via `curl` to the callback route. Verified that the task status updated to `failed`, the raw payload stayed intact in `responseBody` (byte-for-byte matching), and `processingError` stored the error stack trace.
- **Admin UI Display**: Logged in as a superadmin and opened the admin dashboard. Expanded the failed task status card and confirmed that it shows the human-readable stack trace instead of raw JSON.
- **Replay & Recovery**: Re-posted the malformed payload to test idempotency, then POSTed a corrected payload. Confirmed that the task transitioned to `succeeded` and the `processingError` was cleared.
- **Unit Tests**: Ran `npm test` to verify all 793 unit tests pass.

No regressions or unexpected issues were encountered.

<h3>Greptile Summary</h3>

This PR preserves raw task callback payloads separately from post-processing failures and delays terminal success until result processing completes.
- Adds nullable `processingError` persistence through the Prisma schema and migration.
- Records and clears processing failures across success, replay, and backend-error transitions.
- Exposes complete processing errors in the administrative task details dialog.
- Adds unit coverage for payload preservation, failure normalization, stale-error clearing, and terminal-state ordering.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/tasks/tasks.ts | Separates processor errors from callback payloads, safely handles non-Error rejections, clears stale errors, and marks success only after processing completes. |
| src/components/meetings/admin/TaskStatus.tsx | Prioritizes processing errors for failed tasks and adds the complete value to the expanded metadata dialog. |
| src/lib/tasks/__tests__/handleTaskUpdateProcessingError.test.ts | Covers payload preservation, delayed terminal success, clean retries, backend failures, and non-Error processor rejections. |
| prisma/schema.prisma | Adds the nullable `processingError` field to `TaskStatus`. |
| prisma/migrations/20260527120000_task_status_processing_error/migration.sql | Adds the corresponding nullable text column without rewriting existing task records. |

<sub>Reviews (6): Last reviewed commit: ["fix(tasks): mark a task succeeded only a..."](https://github.com/schemalabz/opencouncil/commit/2cac62ccbe65e6d7600b4e91c629779bf241bba9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35316110)</sub>

### DIFF
--- a/src/lib/tasks/tasks.ts
+++ b/src/lib/tasks/tasks.ts
@@ -244,7 +244,7 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
                     where: { id: taskId },
                     data: {
                         status: 'failed',
-                        processingError: `${err.message}\n${err.stack ?? ''}`.trim(),
+                        processingError: (err.stack ?? err.message).trim(),
                         version: update.version,
                     }
                 });
@@ -282,7 +282,7 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
     } else if (update.status === 'error') {
         await prisma.taskStatus.update({
             where: { id: taskId },
-            data: { status: 'failed', responseBody: update.error, version: update.version }
+            data: { status: 'failed', responseBody: update.error, processingError: null, version: update.version }
         });
 
         // Send Discord admin alert for task failure
--- a/src/lib/tasks/tasks.ts
+++ b/src/lib/tasks/tasks.ts
@@ -195,9 +195,17 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
     const sendGenericAlerts = getDiscordAlertMode(task.type) !== 'none';
 
     if (update.status === 'success') {
+        // Persist the raw task server payload BEFORE running our processor.
+        // If processing throws, responseBody stays intact and the task can be
+        // replayed without re-running the (paid) backend job.
         const updatedTask = await prisma.taskStatus.update({
             where: { id: taskId },
-            data: { status: 'succeeded', responseBody: JSON.stringify(update.result), version: update.version }
+            data: {
+                status: 'succeeded',
+                responseBody: JSON.stringify(update.result),
+                processingError: null,
+                version: update.version,
+            }
         });
 
         if (update.result) {
@@ -227,11 +235,18 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
                 }
             } catch (error) {
                 console.error(`Error processing result for task ${taskId}:`, error);
-                const originalResponse = JSON.stringify(update.result);
-                const errorDetail = `Processing error: ${(error as Error).message}\n\n--- Original task server response ---\n${originalResponse}`;
+                const err = error as Error;
+                // Keep responseBody untouched — it already holds the raw task
+                // server payload from the pre-processing write above, so the
+                // task can be replayed via processTaskResponse once the bug
+                // in the processor is fixed.
                 await prisma.taskStatus.update({
                     where: { id: taskId },
-                    data: { status: 'failed', responseBody: errorDetail, version: update.version }
+                    data: {
+                        status: 'failed',
+                        processingError: `${err.message}\n${err.stack ?? ''}`.trim(),
+                        version: update.version,
+                    }
                 });
 
                 // Send Discord admin alert for processing failure
--- a/src/lib/tasks/tasks.ts
+++ b/src/lib/tasks/tasks.ts
@@ -238,7 +238,14 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
                 }
             } catch (error) {
                 console.error(`Error processing result for task ${taskId}:`, error);
-                const err = error as Error;
+                // A processor can reject with a non-Error (a string, a plain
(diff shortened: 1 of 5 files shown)


src/lib/tasks/tasks.ts
```diff
@@ -198,15 +198,36 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
     const sendGenericAlerts = getDiscordAlertMode(task.type) !== 'none';
 
     if (update.status === 'success') {
+        // Persist the raw task server payload BEFORE running our processor.
+        // If processing throws, responseBody stays intact and the task can be
+        // replayed without re-running the (paid) backend job.
+        //
+        // The status stays non-terminal while the processor runs. Tasks in the
```
kouloumos: The reasoning here is sound and I agree with the trade — a sibling's terminal hook must not report a clean batch over a processor that is still running.

One consequence worth a conscious decision rather than a silent one: this widens the window in which a crash leaves a task non-terminal. If the process dies between the payload write above and the `succeeded` write below, the task stays `processing` indefinitely. On main it would have been marked `succeeded` immediately — wrongly, but at least terminally.

I searched for a stale-task reaper and did not find one, so nothing currently recovers such a task.

I think honest-and-stuck is the right choice over wrong-and-terminal, so I am not asking you to change it. I would like it acknowledged, and a follow-up issue for a reaper is probably the real answer. Say if you would rather I open that.

src/lib/tasks/tasks.ts
```diff
@@ -230,11 +251,25 @@ export const handleTaskUpdate = async <T>(taskId: string, update: TaskUpdate<T>,
                 }
             } catch (error) {
                 console.error(`Error processing result for task ${taskId}:`, error);
-                const originalResponse = JSON.stringify(update.result);
-                const errorDetail = `Processing error: ${(error as Error).message}\n\n--- Original task server response ---\n${originalResponse}`;
+                // A processor can reject with a non-Error (a string, a plain
+                // object, undefined). Reading .stack/.message off one of those
+                // throws, which would skip the failed-status write and the alert
+                // below and leave the task marked succeeded.
+                const failureMessage = error instanceof Error ? error.message : String(error);
```
kouloumos: I checked this one against a real database because the comment above it makes a strong claim, and I initially thought it was too broad: a string's `.message` is `undefined`, not a throw.

For `undefined` it is exactly right. On main, a processor that rejects with `undefined` makes `handleTaskUpdate` throw at `(error as Error).message`, which skips the failed-status write and the alert, and leaves the task marked **succeeded** with nothing processed.

So this is not defensive tidying — it closes a silent false-success path. Worth keeping the comment, and worth the test you wrote for it.

PR #418, opened 2026-06-04
feat(meetings): redirect YouTube meeting links to OpenCouncil via URL prefix

### LINKED ISSUE #342: Redirect YouTube meeting links to OpenCouncil via URL prefix
### Concept

When people share council meeting videos, they typically share the YouTube link — often with a timestamp. We should provide a way for users to transform that YouTube URL into an OpenCouncil URL that takes them to the corresponding meeting page at the right point in the transcript/video.

This follows the pattern used by services like 12ft.io or github1s.com, where prefixing or modifying a URL redirects you through a different platform. For example, someone could take `youtube.com/watch?v=xxx&t=1234` and turn it into something that lands them on the OpenCouncil meeting page with the video positioned at that timestamp.

This makes OpenCouncil the natural destination for anyone discussing council meetings online, even when the starting point is a YouTube link.

### User Story

As a citizen who encounters a YouTube link to a council meeting (e.g. shared on social media or in a group chat), I want to easily get to the OpenCouncil page for that meeting at the same timestamp, so that I can access the transcript, agenda context, and other structured information around that moment.

### Context

- Meetings already store their YouTube video URL in the database, so matching an incoming YouTube link to a meeting is feasible.
- `src/lib/utils/youtube.ts` already handles parsing and validating multiple YouTube URL formats (watch, live, shorts, youtu.be).
- YouTube timestamps (`&t=` parameter) should be extracted and used to position the user at the corresponding point in the meeting.
- The exact URL prefix/redirect mechanism is open to implementation — could be a route on the main domain, a subdomain, or another approach.

angelospk:
Closes #342

## What & why

When someone shares a council meeting video, they share the YouTube link, usually with a timestamp. There was no way to turn that link into the OpenCouncil page for the same meeting at the same point.

This adds a 12ft.io / github1s.com-style prefix: put `opencouncil.gr/yt/` in front of any YouTube meeting link and you get a 307 redirect to the matching meeting transcript, with the timestamp preserved.

```
opencouncil.gr/yt/https://www.youtube.com/watch?v=abc123&t=90
  -> /<cityId>/<meetingId>/transcript?t=90
```

Meetings already store their raw `youtubeUrl`, so matching is a lookup by video id. The implementation follows the existing `/qr/[code]` route handler (DB lookup + redirect, i18n bypass in middleware) and the `/api/utterance/[utteranceId]` redirect (the transcript `?t=` deep-link format).

## Changes

- `src/lib/utils/youtube.ts`: `extractYouTubeVideoId(url)` pulls the video id via the existing `YOUTUBE_URL_REGEX` capture group. `extractYouTubeTimestamp(url)` parses the `t`/`start` param, supporting plain seconds (`90`, `90s`) and `1h2m3s` notation, and returns null otherwise.
- `src/lib/db/meetings.ts`: `findCouncilMeetingByYouTubeVideoId(videoId)` does a `findFirst` over released meetings whose `youtubeUrl` contains the id, most-recent-first, and returns `{ cityId, id } | null`.
- `src/app/yt/[...rest]/route.ts` (new): catch-all handler. It recovers the full embedded YouTube URL from `request.url` so the embedded `?v=…&t=…` query survives, tolerates percent-encoded links, validates with `isValidYouTubeUrl`, looks up the meeting, and 307s to `/<cityId>/<meetingId>/transcript[?t=<seconds>]`. Returns 404 when the link isn't a valid YouTube URL or no meeting matches, and redirects to `/` when the target is empty.
- `src/middleware.ts`: adds `yt\/` to the i18n bypass regex (next to `qr\/`) so the handler runs without locale-prefix injection.
- `src/lib/utils/__tests__/youtube.test.ts` (new): 25 unit tests for the validator and the two parsers.

## Testing

- `npx tsc --noEmit`: clean.
- `npm test -- src/lib/utils/__tests__/youtube.test.ts`: 25/25 pass.
- Full `npm run build` not run (scoped, additive change).
- Manual QA plan in `plans/qa/issue-342-qa.md`: valid link to correct meeting, timestamp notations, alternate URL formats, unknown link to 404, invalid URL to 404, URL-encoding edge cases, i18n bypass.

## Risks / Notes

- Released-only: the lookup filters `released: true`, so an unreleased meeting's video id won't leak its page.
- No proxying: the handler only redirects to internal pages. It never fetches or proxies YouTube content.
- Multiple matches: when several meetings share a video (re-uploads), the most recent one (`dateTime desc`) wins.
- Base URL: redirects are built against `NEXTAUTH_URL`, so PR previews work. No `NEXT_PUBLIC_*` baked-in URL.
- No DB index on `CouncilMeeting.youtubeUrl`. The `contains` lookup is a substring scan, fine at current volume (<10k meetings). A functional index can be added later if needed.
- Deferred (optional in the plan): no homepage/footer "paste a YouTube link" form and no separate `?url=` endpoint. The prefix route is the single entry point for v1.

<h3>Greptile Summary</h3>

This follow-up adds a prefix route that resolves shared YouTube links to released OpenCouncil meeting transcripts while preserving bounded timestamps.
- Adds the `/yt/` redirect handler and bypasses locale rewriting for that route.
- Resolves meetings by validated 11-character YouTube video IDs and verifies database candidates by parsing their stored URLs.
- Adds timestamp parsing, a 24-hour upper bound, and tests covering supported URL shapes and prior false-positive cases.

The PR appears safe to merge.

No blocking failure remains; the current implementation fixes the previously reported short-ID, substring-match, and oversized-timestamp cases.

| Filename | Overview |
|----------|----------|
| src/app/yt/[...rest]/route.ts | Adds the public YouTube-prefix redirect flow with URL validation, released-meeting lookup, and timestamp preservation. |
| src/lib/db/meetings.ts | The revised lookup validates exact-length IDs and post-parses candidates, resolving the previously reported short-ID and substring false matches. |
| src/lib/utils/youtube.ts | Adds exact video-ID comparison and bounded timestamp parsing; the previously reported oversized timestamp is rejected. |
| src/lib/utils/__tests__/youtube.test.ts | Covers supported URL formats, exact candidate matching, false-positive exclusions, and timestamp bounds. |
| src/proxy.ts | Excludes `/yt/` from application locale and script routing so the direct route handler can resolve it. |

<sub>Reviews (5): Last reviewed commit: ["fix(meetings): verify the stored YouTube..."](https://github.com/schemalabz/opencouncil/commit/1c054375b559be00c1d760fd002dec175b9c0386) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35612134)</sub>

### DIFF
--- a/src/app/yt/[...rest]/route.ts
+++ b/src/app/yt/[...rest]/route.ts
@@ -1,0 +1,71 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { env } from '@/env.mjs';
+import { findCouncilMeetingByYouTubeVideoId } from '@/lib/db/meetings';
+import {
+    isValidYouTubeUrl,
+    extractYouTubeVideoId,
+    extractYouTubeTimestamp,
+} from '@/lib/utils/youtube';
+
+/**
+ * URL-prefix redirect: prepend `opencouncil.gr/yt/` to any YouTube meeting link
+ * to land on the corresponding OpenCouncil meeting transcript at the same
+ * timestamp (12ft.io / github1s.com style).
+ *
+ * Example:
+ *   /yt/https://www.youtube.com/watch?v=abc123&t=90
+ *     -> /<cityId>/<meetingId>/transcript?t=90
+ *
+ * The embedded YouTube URL keeps its own query string (?v=…&t=…). A catch-all
+ * segment only captures the path, so we recover the full URL from request.url.
+ */
+export async function GET(request: NextRequest) {
+    try {
+        // Recover everything after the literal "/yt/" prefix, including the
+        // embedded YouTube URL's query string.
+        const marker = '/yt/';
+        const markerIndex = request.url.indexOf(marker);
+        if (markerIndex === -1) {
+            return NextResponse.redirect(new URL('/', env.NEXTAUTH_URL));
+        }
+
+        const rawTarget = request.url.slice(markerIndex + marker.length);
+        if (!rawTarget) {
+            return NextResponse.redirect(new URL('/', env.NEXTAUTH_URL));
+        }
+
+        // Tolerate links that were URL-encoded when prefixed.
+        let youtubeUrl = rawTarget;
+        if (!/^https?:\/\//i.test(youtubeUrl)) {
+            try {
+                youtubeUrl = decodeURIComponent(rawTarget);
+            } catch {
+                youtubeUrl = rawTarget;
+            }
+        }
+
+        if (!isValidYouTubeUrl(youtubeUrl)) {
+            return new NextResponse('Not a valid YouTube URL', { status: 404 });
+        }
+
+        const videoId = extractYouTubeVideoId(youtubeUrl);
+        if (!videoId) {
+            return new NextResponse('Could not extract YouTube video id', { status: 404 });
+        }
+
+        const meeting = await findCouncilMeetingByYouTubeVideoId(videoId);
+        if (!meeting) {
+            return new NextResponse('No OpenCouncil meeting found for this YouTube link', {
+                status: 404,
+            });
+        }
+
+        const timestamp = extractYouTubeTimestamp(youtubeUrl);
+        const target = `/${meeting.cityId}/${meeting.id}/transcript${timestamp != null ? `?t=${timestamp}` : ''}`;
+
+        return NextResponse.redirect(new URL(target, env.NEXTAUTH_URL));
+    } catch (error) {
+        console.error('Error redirecting YouTube link:', error);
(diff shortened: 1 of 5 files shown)


angelospk:
Pushed 25da998 to address the review notes on the video-ID lookup.

`findCouncilMeetingByYouTubeVideoId` now checks the 11-char base64url format before it queries, so a crafted single-char ID can no longer substring-match nearly every stored meeting. The lookup is also anchored to `v=<id>` and `/<id>`, which keeps an ID that only shows up inside a playlist param from resolving to the wrong meeting. And timestamp parsing is capped at 24h so an oversized `t=` value can't seek the player somewhere nonsensical.

Added three tests around the 24h boundary and out-of-range values. 28/28 pass, tsc clean.

PR #419, opened 2026-06-04
feat(search): emphasize matched terms in subject result cards

### LINKED ISSUE #404: Improve citizen search — relevance ranking and emphasis of key fields
In a UX test on the public search, a user got many results but still couldn't find the subject they were after, and several low-relevance hits were mixed in with the relevant ones. Separately, a user wanted important fields (e.g. **"Αδειοδότηση"** / licensing) to stand out so results can be scanned quickly.

Two improvements on `src/components/search/SearchPage.tsx` and the result card (`src/components/subject-card.tsx`): (1) tune ranking/relevance for citizen-style queries so the best matches surface first, and (2) emphasize the matched field/term (e.g. bold) in each result card.

Related: #373 (search returns no results when filtering by a specific person).

angelospk:
Closes #404

## What & why

A UX test on the public search found that the matched term wasn't called out anywhere in the result cards, so there was no quick way to scan a page of results and spot the relevant one. This PR bolds the matched terms in each card.

## Changes

- `src/lib/search/constants.ts` (new): alphanumeric sentinel tags instead of `<em>`. The client splits on plain strings, so it never parses markup or touches `dangerouslySetInnerHTML`. Alphanumeric matters because `stripMarkdown` would mangle `<em>` or `__...__` style tags.
- `src/lib/search/query.ts`: a `highlight` block on `name`/`description` using those tags, gated on the (previously unused) `enableHighlights` flag.
- `src/lib/search/types.ts`: optional `nameHighlight` / `descriptionHighlight` on `SearchResultLight`.
- `src/lib/search/core.ts`: carries `hit.highlight` across the retrieval/hydration seam, since only the query knows what matched.
- `src/lib/search/highlight.tsx` (new): `renderHighlighted()` builds React nodes and wraps matched spans in `<strong>`. For descriptions it strips markdown on the whole fragment first, which keeps the spacing around matched terms intact. Greek-safe, falls back to plain text when there is no highlight.
- `src/lib/search/highlight.test.tsx` (new): unit tests for Greek and English terms, the no-highlight fallback, adjacent spans, markdown stripping, and a malformed (unterminated) fragment.
- `src/components/subject/SubjectListContainer.tsx`, `src/components/subject-card.tsx`, `src/components/subject/SubjectCardContent.tsx`: thread the highlights through and render them.
- `src/components/search/SearchPage.tsx`: turns highlights on.

## Notes

- The ranking tuning this PR originally carried (phrase boosts, `fuzziness: AUTO`, `minimum_should_match`) is gone. `main` landed equivalent changes in its search redesign on 17-20 Aug, so the rebase kept only the highlighting.
- RRF means `_score` isn't pure BM25, so bold text won't always line up with rank order. That's expected.
- Highlights only cover `name` and `description`. A result that ranks because of a speaker-segment match shows no bold on its card.

## Testing

`src/lib/search/highlight.test.tsx` covers the render paths above.

<h3>Greptile Summary</h3>

The PR adds matched-term emphasis to public-search subject cards by carrying Elasticsearch highlight fragments through hydration and rendering marked spans safely as React nodes.
- Enables whole-field highlighting for searchable subject fields.
- Threads title, description, and location matches into both subject-card variants.
- Adds coverage for query generation, hydration, marker parsing, markdown stripping, localization transforms, and rendered emphasis.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/search/query.ts | Adds opt-in Elasticsearch highlighting, while the previously reported threshold is now set to `'2<75%'`. |
| src/lib/search/core.ts | Preserves highlighted field fragments while converting and hydrating Elasticsearch hits. |
| src/lib/subjectText.tsx | Selects highlighted or plain subject text, applies existing transforms, and renders matched spans as React markup. |
| src/lib/search/matches.ts | Splits sentinel-delimited fragments into matched and unmatched text segments. |
| src/components/subject/SubjectRow.tsx | Renders highlighted title, description, and location content in the row search-result variant. |
| src/components/subject-card.tsx | Renders highlighted subject fields in the card search-result variant. |

<sub>Reviews (8): Last reviewed commit: ["feat(search): emphasize a matched locati..."](https://github.com/schemalabz/opencouncil/commit/9c5ba922c677114ac8ecec57c7ff98f3e72aae8b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35613977)</sub>


angelospk:
Greptile caught one real issue: the `minimum_should_match: '3<75%'` threshold was off by one. In Elasticsearch's notation, `3<75%` only applies the 75% requirement when there are more than 3 terms, so 3-word queries kept full recall and never got the weak-hit filtering this PR was after.

Fixed it to `2<75%` in 1ddc5f90, which applies the 75% rule once a query has more than 2 terms, i.e. at 3+ terms, matching the documented intent. I also expanded the comment to spell out the ES semantics so the off-by-one doesn't creep back in.

Verified with `npx jest src/lib/search` (7 tests pass) and `npx tsc --noEmit` (clean). Greptile re-reviewed and the PR is now at 5/5 confidence with no unresolved comments.

kouloumos:
Rebased your commits on top of the current main and force-pushed the result.

Main refactored `SubjectCard` into `SubjectCardContent`/`SubjectCardFooter` since you opened this PR. I ported the highlight-emphasis feature onto the new structure to resolve the conflict.

The rebase also exposed two type errors in the feature commit: `EsHit` was missing a `highlight` field, and the new `match_phrase` blocks in `query.ts` used `mergedRequest.query` instead of the already-trimmed `queryText`. I fixed both and folded them into your first commit.

The original branch state before the rebase is preserved at [`backup/pr-419-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-419-original), so nothing is lost.

kouloumos (commented):
@angelospk Thank you for your patience on this one.

## Your code was correct when you wrote it

When you opened this PR, `SubjectListContainer` rendered every result as a `SubjectCard`, and you wired the emphasis into `SubjectCard`. That was the right place.

On 18 August, `da10436f` added `SubjectRow` and moved `/search` to `variant="row"`. The row path never received the fragments. Elasticsearch kept marking up the fields, and nothing rendered them. Your branch did not conflict, so the rebase stayed clean while the feature went dark.

You also had no way to notice. Search reads the production Elasticsearch index, so it does not run locally. A preview deployment on a fork PR needs a maintainer to approve it, and nobody approved one here until yesterday.

## What we changed

We fixed it forward in `0d3e8e1d` rather than ask you to re-architect around a change you were never told about. **Your commit is unchanged; ours sits on top.**

- The match now travels on the subject object, the way `statistics` already does. `SubjectCard` and `SubjectRow` read it through one shared helper. Every list variant shows the emphasis, and a new variant cannot miss it silently.
- The markers are private-use code points. The ASCII markers did not survive Serbian transliteration: `oc7H1ghL1ghtStart9q` became `оц7Х1гхЛ1гхтСтарт9q`. The new markers pass through `localize(stripMarkdown(...))` unchanged, which also restores that order for every card outside search.
- The emphasis element is `<mark>`, and it sets no colour. Both title elements already use `font-semibold`, so the previous `<strong>` was invisible inside a title. It also pinned `text-foreground`, which stopped the matched word from following the hover colour.
- The renderer is a split on one capture group.

New tests cover the Elasticsearch request, the retrieval and hydration seam, and both list variants. The variant test fails without the wiring above.

## How it looks

Captured against the production index.

**One term, every inflected form** — `προϋπολογισμός`, 14 of 14 results marked.
search results with every inflected form of the query marked

**Two terms, marked independently** — `άδεια καταστήματος`, 15 of 15 results.
search results with two query terms marked independently

**A different domain** — `κυκλοφοριακές ρυθμίσεις`.
search results for a traffic-regulation query

**Mobile, 390 px** — no title overflows the two-line clamp.
search results on a 390px mobile viewport

## Cost

Measured against the production index with the app's own query builder, 60 samples per arm, interleaved.

| | Without | With | Delta |
|---|---|---|---|
| Elasticsearch `took`, median | 15 ms | 24 ms | +9 ms |
| Elasticsearch `took`, p90 | 17 ms | 31 ms | +14 ms |
| Response to the browser | 666 KB | 697 KB | +31 KB |
| Client CPU to split the markup | — | — | 0.041 ms per ~4 pages |

A production search takes 1.2 s to 2.6 s end to end, so the emphasis adds about 0.4% of that.

Whole-field markup costs the same as a single snippet (24 ms either way), and dropping a field saves 1 ms. The cost is a fixed charge for enabling the highlight phase, so there is no cheaper configuration available. The faster vector highlighter would need term vectors, which the index does not carry and which would cost more to store than they save.

Only the search page asks for markup. The public API and the MCP layer do not, so nothing else in the product pays for this.

## Known limits

- A result that matches only through the semantic clauses shows no emphasis. Only `name` and `description` carry markup.
- The marked text comes from the Elasticsearch index, so it can differ from the database row between an edit and the next reindex.

@christosporios what do you think about this addition?

kouloumos (approved):
Approving. The emphasis works on the preview, the tests cover the wiring, and the cost is negligible.

## One addition

We also mark a matched **location**. The card already prints the subject's location, and the query already matched on it, so a place-name search had a visible reason for matching that stayed unmarked. `location_text` now joins `MATCH_FIELDS`, and the Elasticsearch request, the hit mapping and the `SearchMatches` type all derive from that list — a fourth field needs one line.

To be accurate about what this buys: it adds coherence, not coverage. Across 179 results in twelve place-name queries, it marked 85 cards but rescued **zero** that were otherwise unmarked. Every card whose location matched also matched in its title or description.

search for λυκαβηττού, with the term marked in the title, the description and the location line

search for πλατεία, with the term marked in each result's location

## How well it covers real searches

We replayed **32 real queries taken from the `SearchQuery` log**, rather than queries we invented: parks, budget, rubbish, cycle lanes, a councillor's name, a street, accent-dropped spellings, typos, and one off-topic search.

**387 result cards, 363 marked — 93.8%.** 25 of the 30 queries that returned results were marked in full.

## What stays unmarked, and why

The query matches on more fields than the card shows: the nested transcript (`speaker_contributions`), the introducer's name, and the semantic vectors. A result that matched only there has nothing on the card to mark.

The misses are not spread evenly. They collect in four queries, and they are all the same kind of search:

| Query | Marked | Why the rest miss |
|---|---|---|
| `παλαιστίνη` | 3 of 9 | matched in the transcript |
| `βαρδή πολυξένη` | 9 of 15 | a councillor — matched by who spoke |
| `νεα δημοκρατια` | 4 of 9 | a party — the same |
| `Netflix σειρές` | 0 of 1 | semantic only |

**People and parties are what go unmarked.** A subject matches a person because they spoke in it, and the card shows speakers as avatars, not as text. We confirmed this against the index: for one place-name search, three unmarked results carried the term only in their transcripts, while their titles, descriptions and locations named different places entirely.

This is not new. It is how search already behaves on `main`. The emphasis makes it visible by contrast.

## Follow-up

Explaining an unmarked result needs design, not code. The card would have to show *why* it matched — a matched line from the transcript, a highlighted speaker avatar, or an expanded state. That is a product decision about result explainability, and it deserves its own issue rather than a widening of this PR.

PR #420, opened 2026-06-04
feat(webhooks): id-based replay protection for Bird webhooks

### LINKED ISSUE #391: Bird webhooks: replace timestamp window with id-based replay protection
## Context

Follow-up to [#389](https://github.com/schemalabz/opencouncil/pull/389), which widened the Bird webhook timestamp freshness window from 300s → 24h after we discovered Bird's retry queue re-delivers webhooks with the *original* event timestamp (observed 3–4h stale on real traffic). The widening was a hotfix to stop silently dropping inbound user messages (including \`STOP\` unsubscribe requests) — but it weakens replay protection: an attacker who captures a legitimate signed webhook now has up to 24h to replay it instead of 5 minutes.

The right long-term fix is **id-based replay protection** — track Bird's per-event \`id\` and reject duplicates. Time stops being part of the security model.

## Proposal

After signature verification passes in [src/app/api/webhooks/bird/route.ts](https://github.com/schemalabz/opencouncil/blob/main/src/app/api/webhooks/bird/route.ts), check whether we've already processed the event's \`id\` (extracted in [src/app/api/webhooks/bird/extract.ts](https://github.com/schemalabz/opencouncil/blob/main/src/app/api/webhooks/bird/extract.ts)):

1. **Already seen** → return 200 OK with no further work. Bird treats this as success and stops retrying.
2. **Not seen** → record the id atomically, then proceed to processing.

Once dedupe is in place, the time window can be shrunk (e.g. back to a few minutes) or removed entirely. The signature still proves authenticity; the id check stops replays.

## Storage options

Two reasonable backends — both work, pick one:

| Option | Pros | Cons |
|---|---|---|
| **Postgres table** (\`WebhookSeenId(birdId, seenAt)\`) | Durable, ACID — \`INSERT ... ON CONFLICT DO NOTHING\` is naturally atomic and tells us in one round-trip whether the id was new | One extra DB write per webhook; needs a periodic cleanup job for \`seenAt < now - retention\` |
| **Valkey set with TTL** (\`SET birdId 1 NX EX 604800\`) | Fast, fits existing cache infra, auto-expiry — no cleanup job needed | Ephemeral; a Valkey wipe means we'd briefly accept replays |

Recommendation: **Valkey with TTL** for speed and zero cleanup ops; security degrades gracefully (worst case is a brief replay window after a Valkey reset, no different from the current 24h state).

Retention: 7 days. Generous margin over Bird's observed 3–4h retry delays; fits inside DO Valkey's storage easily.

## Acceptance criteria

- [ ] Webhook handler does an atomic check-and-set on Bird's event id immediately after signature verification.
- [ ] Duplicate ids are dropped with 200 OK + a single info-level log line (\`webhook id <id> already processed — skipping\`).
- [ ] First-time ids proceed through the existing extract → persist → unsubscribe path unchanged.
- [ ] [\`REPLAY_WINDOW_SECONDS\`](https://github.com/schemalabz/opencouncil/blob/main/src/app/api/webhooks/bird/route.ts) is tightened (e.g. 1h) or removed once id dedupe is the primary defense.
- [ ] Comment block at the top of \`route.ts\` updated to reflect the new model.

## Threat model recap (for whoever picks this up)

| Threat | Current defense | After this issue |
|---|---|---|
| Spoofing (no secret) | HMAC signature | HMAC signature (unchanged) |
| Tampering in transit | HMAC signature | HMAC signature (unchanged) |
| Replay of captured webhook | Time window — was 5 min, currently 24h, weakening | id dedupe — exact, no false positives, time-window-independent |

## References

- [#389](https://github.com/schemalabz/opencouncil/pull/389) — interim fix (verify signature first, widen window)
- [Bird API docs — verifying a webhook subscription](https://docs.bird.com/api/notifications-api/api-reference/webhook-subscriptions/verifying-a-webhook-subscription)
- Bird's newer JWT signing scheme uses a \`jti\` claim for exactly this kind of dedupe — worth considering as part of the same migration ([messagebird-nodejs/lib/webhook-signature-jwt.js](https://github.com/messagebird/messagebird-nodejs/blob/master/lib/webhook-signature-jwt.js))

angelospk:
Closes #391

## What & why

PR #389 widened the Bird webhook timestamp freshness window from 300s to 24h. The reason was real: Bird's retry queue re-delivers webhooks carrying the *original* event timestamp, sometimes 3-4h stale, and the tight window was silently dropping legitimate user messages. The trade-off was that replay protection then leaned entirely on a 24h time window, so a captured signed payload could be replayed for most of a day.

This PR moves replay protection off the clock and onto the event id. After signature verification we record each event's id in Valkey with an atomic `SET NX EX` and drop exact replays. Time is no longer part of the security model.

## Changes

- **New `src/lib/notifications/webhookDedupe.ts`**
  - `markWebhookSeen(id)` does an atomic `SET bird:webhook:seen:<id> 1 NX EX 604800` (7-day TTL). A non-null reply means `'new'`, null means `'duplicate'`. When `CACHE_URL` is unset or Valkey is unreachable it returns `'unknown'` and fails open.
  - `clearWebhookSeen(id)` rolls back a marker (best-effort) so a transient processing failure doesn't poison the retry.
  - Reuses the Valkey client pattern already proven in `admin/cache/stats` (module singleton, `pingInterval`, `valkeys://` to `rediss://`).
- **`src/app/api/webhooks/bird/route.ts`**
  - Right after signature verification and field extraction, compute a dedupe id and call `markWebhookSeen`.
  - `'duplicate'` returns `200 { ok: true }` plus one info log, and Bird stops retrying.
  - `'new'` / `'unknown'` take the existing extract, persist, unsubscribe path, unchanged.
  - On a transient (non-P2002) persist error that returns 500, call `clearWebhookSeen` so the retry isn't dropped as a duplicate.
  - Rewrote the header comment to describe the new model.
- **Unit tests** for the dedupe module: new vs duplicate, status-progression keys staying distinct, the `NX/EX` args, fail-open on no `CACHE_URL` and on client error, and rollback via `clearWebhookSeen`.

### On the dedupe key and the replay window

The dedupe key is `payload.id : birdMessageId : status`, not the message id alone. Bird emits several lifecycle events for one message (`sent`, `delivered`, `failed`) under the same message id, so a message-only key would suppress legitimate progressions. Composing the stable identifiers means only an exact replay collides.

`REPLAY_WINDOW_SECONDS` stays at 24h rather than being tightened to 1h. A 1h window would reject Bird's 3-4h-stale retries at signature verification, before the dedupe layer ever runs, which is exactly the bug #389 fixed. So the window's *role* changes (sanity ceiling, not replay defense) while its value stays generous. Id dedupe is now the only replay defense.

## Testing

- `npx jest src/lib/notifications/__tests__/webhookDedupe.test.ts src/app/api/webhooks/bird/__tests__/extract.test.ts`: 54 passing.
- `npx tsc --noEmit` clean for the changed files. (There's a pre-existing unrelated `diavgeiaSignerIds`/TS2322 error on the branch base, not touched here.)

## Risks / notes

- **Threat model**: signature verification proves the payload came from Bird untampered; id dedupe stops the same signed payload being processed twice. A real attacker would still need a valid signature (they don't have the secret), so this mainly hardens against Bird's own retries and accidental re-delivery.
- **Fail-open**: if Valkey is down, dedupe degrades to the existing DB guards: the `birdMessageId` unique constraint (P2002 swallow) for inbound and `isForwardProgression` for outbound. We lose replay *rejection*, not correctness, which matches today's behaviour. Worth metering Valkey errors so a fail-open storm is visible.
- **Backend choice**: Valkey over a Postgres `WebhookSeenId` table because Valkey is already wired via `CACHE_URL`, `SET NX` is a single atomic op, and the TTL handles cleanup. No migration.
- Did not refactor `admin/cache/stats` to share the new client, to keep the change surgical.

<h3>Greptile Summary</h3>

The PR adds Valkey-backed, ID-based replay protection for signed Bird webhooks while retaining retry behavior after transient processing failures.
- Builds dedupe identifiers from the event ID, Bird message ID, and normalized status.
- Uses atomic `SET NX EX` markers with fail-open behavior when Valkey is unavailable.
- Clears first-seen markers after transient failures so Bird retries can be processed.
- Addresses the previous delimiter-collision feedback and adds the requested fallback and precedence tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/app/api/webhooks/bird/extract.ts | Extracts the composite dedupe-ID builder and replaces the ambiguous colon delimiter as previously requested. |
| src/app/api/webhooks/bird/__tests__/extract.test.ts | Adds coverage for ID precedence, missing-ID fallbacks, status separation, and the previously reported delimiter collision. |
| src/app/api/webhooks/bird/route.ts | Integrates first-seen checks, duplicate acknowledgment, and marker rollback into the signed webhook processing flow. |
| src/lib/notifications/webhookDedupe.ts | Implements atomic Valkey replay markers, fail-open handling, singleton connection reuse, and best-effort rollback. |
| src/lib/notifications/__tests__/webhookDedupe.test.ts | Covers atomic marker behavior, expiration options, unavailable-cache degradation, status-specific keys, and rollback. |

<sub>Reviews (4): Last reviewed commit: ["fix(webhooks): use | as dedupe-key separ..."](https://github.com/schemalabz/opencouncil/commit/b8a9cd4a0d743757fd35ce54ecf3b29c652e5ac6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35615094)</sub>

### DIFF
--- a/src/lib/notifications/webhookDedupe.ts
+++ b/src/lib/notifications/webhookDedupe.ts
@@ -1,0 +1,115 @@
+import { createClient } from 'redis';
+import { env } from '@/env.mjs';
+
+/**
+ * Id-based replay protection for incoming webhooks (Bird).
+ *
+ * Atomic check-and-set against Valkey: `SET <key> 1 NX EX <ttl>`. A non-null
+ * reply means the key did not exist (first time we've seen this event); a null
+ * reply means it already existed (a replay/retry of the exact same event).
+ *
+ * Backend choice: Valkey is already wired via `CACHE_URL` (DO Valkey,
+ * `valkeys://`) and proven in `src/app/api/admin/cache/stats/route.ts`. `SET NX`
+ * is a single atomic op, so there is no read-then-write race. The 7-day TTL
+ * comfortably covers Bird's retry horizon without unbounded key growth.
+ *
+ * Degradation: when `CACHE_URL` is unset (single-instance dev) or Valkey is
+ * unreachable, the dedupe call FAILS OPEN — it returns `'unknown'` so the
+ * webhook is still processed. Correctness is then guarded by the existing DB
+ * layer (the `birdMessageId` unique constraint + forward-only status
+ * progression), so failing open only loses replay *rejection*, not integrity.
+ */
+
+const KEY_PREFIX = 'bird:webhook:seen:';
+const TTL_SECONDS = 7 * 24 * 60 * 60; // 604800 — covers Bird's retry horizon
+
+export type DedupeResult = 'new' | 'duplicate' | 'unknown';
+
+// Module-level singleton — reused across requests, mirrors the pattern in
+// src/app/api/admin/cache/stats/route.ts.
+let client: ReturnType<typeof createClient> | null = null;
+let connectingPromise: Promise<ReturnType<typeof createClient> | null> | null = null;
+
+async function getClient(): Promise<ReturnType<typeof createClient> | null> {
+    if (client?.isReady) return client;
+    if (connectingPromise) return connectingPromise;
+
+    const cacheUrl = env.CACHE_URL;
+    if (!cacheUrl) return null;
+
+    connectingPromise = (async () => {
+        if (client) {
+            client.disconnect().catch(() => {});
+        }
+        // The `redis` npm package uses rediss:// for TLS; DO Valkey uses valkeys://
+        const normalizedUrl = cacheUrl.replace(/^valkeys:\/\//, 'rediss://');
+        const next = createClient({ url: normalizedUrl, pingInterval: 60_000 });
+        next.on('error', (error) => {
+            console.error('[webhook-dedupe] Valkey client error:', error.message);
+        });
+        await next.connect();
+        client = next;
+        return client;
+    })();
+
+    try {
+        return await connectingPromise;
+    } finally {
+        connectingPromise = null;
+    }
+}
+
+/**
+ * Atomically record that we've seen this webhook event id.
+ *
+ * @returns
+ *   - `'new'`       — first time we've seen this id; caller should process it.
+ *   - `'duplicate'` — already processed; caller should drop it (200 OK).
+ *   - `'unknown'`   — dedupe unavailable (no CACHE_URL / Valkey down); fail open
+ *                     and process, relying on the DB idempotency layer.
+ */
+export async function markWebhookSeen(id: string): Promise<DedupeResult> {
+    try {
+        const redisClient = await getClient();
+        if (!redisClient) return 'unknown';
+
+        const reply = await redisClient.set(`${KEY_PREFIX}${id}`, '1', {
+            NX: true,
+            EX: TTL_SECONDS,
+        });
+        // redis@4 returns 'OK' when set, null when NX prevented the write.
+        return reply === null ? 'duplicate' : 'new';
+    } catch (error) {
+        // Reset a broken client so the next request reconnects.
+        if (client && !client.isReady) {
+            client.disconnect().catch(() => {});
+            client = null;
+        }
+        console.warn(
+            '[webhook-dedupe] dedupe check failed, failing open:',
+            error instanceof Error ? error.message : error,
+        );
+        return 'unknown';
+    }
+}
+
+/**
+ * Remove a previously-recorded seen marker. Called when processing of a
+ * first-seen event fails transiently (e.g. a DB error that returns 500 and
+ * makes Bird retry) — without this, the dedupe entry would cause the retry to
+ * be dropped as a duplicate and the event would be lost permanently.
+ *
+ * Best-effort: failures are swallowed (the entry will expire via its TTL).
+ */
+export async function clearWebhookSeen(id: string): Promise<void> {
+    try {
+        const redisClient = await getClient();
+        if (!redisClient) return;
+        await redisClient.del(`${KEY_PREFIX}${id}`);
+    } catch (error) {
+        console.warn(
+            '[webhook-dedupe] failed to clear seen marker (will expire via TTL):',
+            error instanceof Error ? error.message : error,
+        );
+    }
+}
--- a/src/app/api/webhooks/bird/__tests__/extract.test.ts
+++ b/src/app/api/webhooks/bird/__tests__/extract.test.ts
@@ -1,5 +1,6 @@
 /** @jest-environment node */
 import {
+    buildDedupeId,
     extractBody,
     extractChannel,
     extractDirection,
@@ -9,6 +10,7 @@ import {
     extractPhone,
     unwrapEvent,
     type BirdMessageLike,
+    type ExtractedMessageFields,
 } from '../extract';
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


angelospk:
Both Greptile P2s are addressed in 1eea459d.

1. Dedupe-key separator: switched from `:` to `|`. A `:` can show up inside Bird payload or message ids, so `a:b` + `c` and `a` + `b:c` collapsed to the same key. `|` never appears in Bird ids, so the split stays unambiguous.

2. buildDedupeId test coverage: moved `buildDedupeId` out of the route handler into `extract.ts` (pure and exported) and added unit tests for the fallback paths: missing payload/data id (the `-` fallback), missing message id, status progressions staying distinct, and a case proving the old `:` collision no longer happens.

61/61 webhook tests pass, `npx tsc --noEmit` clean. Re-review came back with 0 new comments; resolved both threads.

src/lib/notifications/webhookDedupe.ts
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+const KEY_PREFIX = 'bird:webhook:seen:';
+const TTL_SECONDS = 7 * 24 * 60 * 60; // 604800 — covers Bird's retry horizon
+
+export type DedupeResult = 'new' | 'duplicate' | 'unknown';
+
+// Module-level singleton — reused across requests, mirrors the pattern in
+// src/app/api/admin/cache/stats/route.ts.
+let client: ReturnType<typeof createClient> | null = null;
```
kouloumos: **This line is why Nix Checks fails.**

```
webhookDedupe.ts(51,9): error TS2322:
  Type 'RedisClientType<{}, {}, {}, 3, {}>' is not assignable to
  type 'RedisClientType<RedisModules, RedisFunctions, RedisScripts, RespVersions, TypeMapping> | null'
```

`createClient` is generic, so `ReturnType<typeof createClient>` resolves its type parameters to their *unresolved* forms (`RedisModules`, `RespVersions`, …). A real call infers them concretely (`{}, {}, {}, 3, {}`). The two are not assignable, so the write at line 51 is rejected.

Worth knowing: `src/lib/cache/valkey.ts` on main uses the *same* annotation and compiles fine, because it assigns the call result straight to the annotated variable, where contextual typing settles the parameters. Your version goes through an intermediate `const next` at line 46, which types independently first and then will not assign. Genuinely subtle.

A small non-generic factory makes the annotation and the call the same type, with no cast and no `any`:

```suggestion
// `createClient` is generic, so `ReturnType<typeof createClient>` resolves its type
// parameters to their unresolved forms, which a real call never produces — a call
// infers them concretely. Deriving the type from a concrete factory keeps the
// annotation and the call identical.
function createCacheClient(url: string) {
    return createClient({ url, pingInterval: 60_000 });
}
type CacheClient = ReturnType<typeof createCacheClient>;

let client: CacheClient | null = null;
```

The rest follows: `connectingPromise` and `getClient`'s return type become `CacheClient | null`, and line 46 becomes `const next = createCacheClient(normalizedUrl);`.

I have this committed locally and verified — `tsc` clean, full suite 1,885 tests passing. Say the word and Andreas can push it to your branch, or take the suggestion above and I will re-check.

src/lib/notifications/webhookDedupe.ts
```diff
@@ -0,0 +1,115 @@
...
+ * layer (the `birdMessageId` unique constraint + forward-only status
+ * progression), so failing open only loses replay *rejection*, not integrity.
+ */
+
+const KEY_PREFIX = 'bird:webhook:seen:';
+const TTL_SECONDS = 7 * 24 * 60 * 60; // 604800 — covers Bird's retry horizon
+
+export type DedupeResult = 'new' | 'duplicate' | 'unknown';
+
+// Module-level singleton — reused across requests, mirrors the pattern in
```
kouloumos: This comment is accurate for when you wrote it, and it is worth recording why it is now out of date — through no fault of yours.

`src/app/api/admin/cache/stats/route.ts` was the only precedent on 4 June. On **30 June** main gained `src/lib/cache/valkey.ts` (in `8b3ea3e8`), which is a shared client with the same singleton, the same connect-dedupe, the same URL normalization, the same `pingInterval`, and a doc comment that literally says it is *"Used for one-shot dedup markers"*.

It also has `disconnectCache()`, and the lack of an equivalent here is not theoretical: my probe against a real Valkey hung the test runner, because this module keeps a client open with no way to close it. Your mocked tests never open a socket, so they could not have caught it.

If you take the refactor, the shape I would suggest is to add two primitives to `valkey.ts` — `cacheSetIfAbsent(key, ttlSeconds)` returning whether the write landed, and `cacheDelete(key)` — and have `markWebhookSeen` / `clearWebhookSeen` call those. Everything from here down to `getClient` then goes away.

Not a blocker for the CI fix, and not something you should have to do to get this merged if Andreas would rather it were a follow-up.

christosporios:
now dead, closing

Issue #51, opened 2025-07-08
Back button doesn't work properly


kouloumos:
**Problem**
When navigating to a meeting page and then pressing the browser’s back button, the user is redirected to the homepage (`/`) instead of the expected city page (`/chania`).

https://github.com/user-attachments/assets/1983b3a0-ff3a-42aa-8339-33e2265758af

[iPhone 16 Pro using Google Chrome]

**Steps to Reproduce**

1. Visit [https://opencouncil.gr/](https://opencouncil.gr).
2. Click on **Χανιά**.
3. Select any meeting and wait for the page to fully load.
4. Press the browser's back button.

**Expected Behavior**
User should be taken back to `/chania`.

**Actual Behavior**
User is taken back to `/`.

kouloumos:
fixed by 5dc1033d09aad7f8800c833c6930ebc2232b1a0c

PR #422, opened 2026-06-04
feat(navigation): surface subject context and add accessible back/home affordance

### LINKED ISSUE #405: Subject & search pages: breadcrumb is the only context/back affordance and it's under-discovered
The breadcrumb in `src/components/layout/Header.tsx` quietly carries too much weight. On a subject page it's the **only** place that shows which council/meeting the subject belongs to, so users miss that context entirely. It's also the only way "back" — one user didn't realize this and clicked the logo to escape. And on the search page there's no obvious "home" affordance, so a user resorted to editing the URL by hand.

Proposal: surface the subject's context (council/meeting) on the page itself rather than relying solely on the breadcrumb, and add a clear, accessible back/home affordance on subject and search pages.

Related (both closed): #51 (back-button redirected to `/`), #293 (WCAG audit).

angelospk:
Closes #405

The header breadcrumb was doing too much. On a subject page it was the only thing telling you which meeting/council the subject belonged to, and the only way back, and people missed both. One user clicked the logo to get out because they never realized the breadcrumb was clickable. On the search page there was no home link at all, so a user ended up hand-editing the URL to escape.

This puts the subject's context (meeting + council) on the page body itself and adds a back/home link on the subject and search pages. The breadcrumb works exactly as before.

Back navigation is a real `<Link>` to the meeting page, not `history.back()`, so it can't fall back to `/` the way the old behavior did in #51, and it still works when the page is opened cold with no history. The links carry `aria-label`s and visible keyboard focus rings, which covers the WCAG points from #293.

## Changes
- `src/components/meetings/subject/subject.tsx` — read `city` alongside `meeting` from the existing `useCouncilMeetingData()` context (no refetch), then render a context `<nav>` at the top of the content column: a back link (arrow + meeting name) to `/<cityId>/<meetingId>`, the city name linking to `/<cityId>`, and the meeting date.
- `src/components/search/SearchPage.tsx` — add a keyboard-focusable back-to-home `<Link href="/">` at the top of the page.
- `messages/el.json`, `messages/en.json` — add `Subject.backToMeeting`, `Subject.backToMeetingNamed`, `Subject.partOf`, and `Common.backToHome` for both locales.

## Testing
- `npx tsc --noEmit` is clean.
- Ran the QA manually in a browser, both locales: the context row shows up on the page body, the back link lands on the meeting page and not `/`, the search home link works, keyboard focus and screen-reader labels behave, and the breadcrumb is untouched.

## Notes
- This is additive UI. The breadcrumb behavior doesn't change. The on-page context is deliberate redundancy for discoverability, kept light so it doesn't fight the breadcrumb.
- `meeting` and `city` are always there in the server-fetched `MeetingData`, and the component already `notFound()`s when the subject is missing, so there's no `/undefined/undefined` link to worry about.
- The other search-page strings are still hard-coded Greek (they were before); only the new home-link string goes through next-intl. Rewriting the rest is out of scope.
- Related closed issues: #51 (back must not redirect to `/`), #293 (WCAG audit).

<h3>Greptile Summary</h3>

This PR surfaces subject context (meeting name + city + date) directly on the subject page body and adds back/home links on the subject and search pages to address discoverability issues where users missed the header breadcrumb or had no obvious escape path.

- `subject.tsx`: reads `city` from the existing context (no extra fetch) and renders a `<nav>` with a back link to the meeting page, a city link, and the meeting date — `formatDate` correctly passes `locale` as the third argument.
- `SearchPage.tsx`: adds a `Link href=\"/\"` home affordance using the i18n-aware router; translation key correctly targets the `Common` namespace.
- `messages/en.json` + `messages/el.json`: adds `Common.backToHome` and `Subject.backToMeetingNamed` / `Subject.partOf` for both locales.

Purely additive UI — no existing behavior is modified, data comes from already-fetched context, and all navigation targets use real Link components with typed hrefs.

Both `meeting` and `city` are non-nullable in `MeetingDataCore`, the component already guards against a missing subject with `notFound()`, and the locale is correctly threaded through to `formatDate`. The translation keys are consistently named across both locale files and match exactly what the components consume.

| Filename | Overview |
|----------|----------|
| src/components/meetings/subject/subject.tsx | Adds an on-page context nav (back link + city link + meeting date) using already-available meeting/city data from context; locale is correctly passed to formatDate. |
| src/components/search/SearchPage.tsx | Adds a home affordance link at the top using the i18n-aware Link component; translation key correctly maps to the Common namespace. |
| messages/en.json | Adds backToHome to Common and backToMeetingNamed/partOf to Subject — keys match exactly what the components consume. |
| messages/el.json | Greek translations added for the same three keys, consistent with the English counterparts. |

</details>

<sub>Reviews (3): Last reviewed commit: ["chore(i18n): drop unused Subject.backToM..."](https://github.com/schemalabz/opencouncil/commit/8271ef4cda02c388fd039213a0a844216dc99f1b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35662721)</sub>

### DIFF
--- a/messages/el.json
+++ b/messages/el.json
@@ -16,7 +16,8 @@
         "adminBodyType_committee": "Επιτροπές",
         "adminBodyType_community": "Κοινότητες",
         "allPeople": "Όλα",
-        "allBodies": "Όλες"
+        "allBodies": "Όλες",
+        "backToHome": "Πίσω στην αρχική σελίδα"
     },
     "MeetingCard": {
         "noVideo": "Χωρίς μέσα",
@@ -778,6 +779,9 @@
     },
     "Subject": {
         "agendaItem": "Θέμα #{index}",
+        "backToMeeting": "Πίσω στη συνεδρίαση",
+        "backToMeetingNamed": "Πίσω στη συνεδρίαση: {meeting}",
+        "partOf": "Μέρος της συνεδρίασης",
         "parties": "Παρατάξεις",
         "minutes": "λεπτά",
         "speakers": "{count, plural, =1 {1 ομιλητής} other {# ομιλητές}}",
--- a/messages/el.json
+++ b/messages/el.json
@@ -779,7 +779,6 @@
     },
     "Subject": {
         "agendaItem": "Θέμα #{index}",
-        "backToMeeting": "Πίσω στη συνεδρίαση",
         "backToMeetingNamed": "Πίσω στη συνεδρίαση: {meeting}",
         "partOf": "Μέρος της συνεδρίασης",
         "parties": "Παρατάξεις",
--- a/messages/en.json
+++ b/messages/en.json
@@ -16,7 +16,8 @@
         "adminBodyType_committee": "Committees",
         "adminBodyType_community": "Communities",
         "allPeople": "All",
-        "allBodies": "All"
+        "allBodies": "All",
+        "backToHome": "Back to home"
     },
     "PilotPage": {
         "title": "OpenCouncil is running a pilot with some of the latest city council meetings of the Municipality of Athens.",
@@ -799,6 +800,9 @@
     },
     "Subject": {
         "agendaItem": "Subject #{index}",
+        "backToMeeting": "Back to meeting",
+        "backToMeetingNamed": "Back to meeting: {meeting}",
+        "partOf": "Part of the meeting",
         "parties": "Parties",
         "minutes": "minutes",
         "speakers": "{count, plural, =1 {1 speaker} other {# speakers}}",
--- a/messages/en.json
+++ b/messages/en.json
@@ -800,7 +800,6 @@
     },
     "Subject": {
         "agendaItem": "Subject #{index}",
-        "backToMeeting": "Back to meeting",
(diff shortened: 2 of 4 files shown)


angelospk:
Both Greptile P2s are fixed in 445e1959.

1. Locale-aware date: the on-page subject date called `formatDate` without the active locale, so English-locale users got a Greek-formatted date. I actually hit this during the manual QA, where the `/en` subject page showed "26 Φεβρουαρίου 2026". It now passes the `useLocale()` value.

2. Redundant aria-label: the search home link had an `aria-label` that just repeated its visible span text. The accessible name already comes from the inner text, so the `aria-label` added nothing and could drift out of sync with the translation. Removed it.

tsc and eslint are clean. Re-review is back to 5/5 with both threads resolved.

angelospk:
Dropped the unused `Subject.backToMeeting` key from both `en.json` and `el.json` in 8271ef4c. Only `backToMeetingNamed` is actually used (the back-link aria-label), so the bare string was dead weight.

VasiaKoum (approved):

PR #262, opened 2026-03-08
fix: enable dev login tools in preview deployments

MaxwellCalkin:
Fixes #250

## Summary

The Quick Login component and related dev tools are gated on `IS_DEV` (`NODE_ENV === 'development'`), but preview deployments run with `NODE_ENV=production`. This makes it impossible to log in as Admin/City Admin on preview URLs for role-based testing.

This PR:
- Adds `IS_PREVIEW` (detects `*.preview.*` in `NEXTAUTH_URL`) and `IS_DEV_OR_PREVIEW` constants to `src/lib/utils.ts`
- Updates all dev tool access gates to use `IS_DEV_OR_PREVIEW`:
  - `QuickLogin` component
  - `DevelopmentSection` profile component
  - `MobilePreviewReporter` component
  - `/api/dev/quick-login` route
  - `/api/dev/seed-test-users` route
  - `/api/dev/mobile-preview` route
- Preserves `IS_DEV` for cookie configuration in the quick-login route (preview deployments correctly use `__Secure-` prefixed cookies over HTTPS)

## Why this is safe

Per the [issue discussion](https://github.com/schemalabz/opencouncil/issues/250#issuecomment-2694785655):
- Preview databases use PII-free staging data
- Task server integration is disabled by default in previews
- The detection is scoped to `NEXTAUTH_URL` containing `.preview.`, which is only set by the NixOS preview module — production deployments are unaffected

## Files changed

| File | Change |
|------|--------|
| `src/lib/utils.ts` | Add `IS_PREVIEW`, `IS_DEV_OR_PREVIEW` |
| `src/components/dev/QuickLogin.tsx` | `IS_DEV` → `IS_DEV_OR_PREVIEW` |
| `src/components/dev/MobilePreviewReporter.tsx` | `IS_DEV` → `IS_DEV_OR_PREVIEW` |
| `src/components/profile/DevelopmentSection.tsx` | `IS_DEV` → `IS_DEV_OR_PREVIEW` |
| `src/app/api/dev/quick-login/route.ts` | `IS_DEV` → `IS_DEV_OR_PREVIEW` (access), keep `IS_DEV` (cookies) |
| `src/app/api/dev/seed-test-users/route.ts` | `IS_DEV` → `IS_DEV_OR_PREVIEW` |
| `src/app/api/dev/mobile-preview/route.ts` | `IS_DEV` → `IS_DEV_OR_PREVIEW` |

## Test plan

- [ ] Deploy as a preview and verify the Quick Login floating button appears
[...]

PR #428, opened 2026-06-10
fix(dev): enable quick login on preview deployments

### LINKED ISSUE #250: Unable to login as Admin/City Admin in Preview Deployments
Currently, it is not possible to log in as an administrator, city admin, or other specific user roles on preview deployment pages. This occurs because the component providing this quick login capability is conditionally rendered to appear only in the local development environment (dev env), making role-based testing on preview URLs impossible.

angelospk:
Closes #250

## What & why

Reviewers can't log in as admin / city admin / other roles on PR preview deployments (`pr-N.preview.opencouncil.gr`). The Quick Login dev tool and its API routes are gated on `IS_DEV` (`NODE_ENV === 'development'`), but previews are production builds (`NODE_ENV=production`), so the tool never renders and the APIs return 403. That's exactly the role-based QA the issue asks for, and right now it's blocked.

## How this differs from #262

#262 was closed pending discussion, and the approach there detected previews by string-matching `*.preview.*` inside `NEXTAUTH_URL`. This PR reuses the existing `IS_PREVIEW=true` env var instead — the preview systemd service already sets it, and `src/lib/email/resend.ts` already reads it (it went in for the email-redirect feature). So no new env var, no brittle URL parsing, and no `flake.nix` / `env.mjs` / `.env.example` changes. The plumbing is already in the tree.

It also tries to answer your open question from the issue thread ("are there any downsides?") — see the security tradeoff section below.

## Changes

- `src/lib/utils.ts` — add `IS_PREVIEW` and `DEV_TOOLS_ALLOWED = IS_DEV || IS_PREVIEW` next to `IS_DEV`.
- `src/app/[locale]/layout.tsx` — load `QuickLogin` when `NODE_ENV === 'development'` OR `IS_PREVIEW === 'true'`. Both stay literal `process.env.X` comparisons so the bundler still dead-code-eliminates QuickLogin on real production.
- `src/components/dev/QuickLogin.tsx` — gate on `DEV_TOOLS_ALLOWED`; show a PREVIEW badge and "Preview Quick Login" title on previews.
- `src/components/profile/DevelopmentSection.tsx` — gate on `DEV_TOOLS_ALLOWED`; PREVIEW ONLY badge on previews.
- `src/app/api/dev/quick-login/route.ts` and `src/app/api/dev/seed-test-users/route.ts` — guard with `DEV_TOOLS_ALLOWED`. The HTTPS secure-cookie logic in `quick-login` stays keyed on `IS_DEV`, since previews are HTTPS and need the secure cookie name like production does.
- `src/lib/__tests__/dev-tools-allowed.test.ts` — unit tests for the gate.

`MobilePreviewReporter`, `seed-users`, `mobile-preview`, and `lan-info` stay `IS_DEV`-only, since they're LAN/HMR dev features that don't make sense on remote previews.

## Testing

- `npx jest src/lib/__tests__/dev-tools-allowed.test.ts` — 4/4 pass (false on production, true on preview, true in dev, false for a non-`"true"` value).
- `npx tsc --noEmit` — no new errors. Two pre-existing errors in `cities/[cityId]/meetings/route.ts` are unrelated and untouched.
- Manual QA scenarios in the QA doc, including one that confirms Quick Login is absent and stripped from the bundle on a production build.

## Security tradeoff (the "any downsides?" answer)

- On previews, Quick Login lets anyone with the URL log in as super-admin against the shared staging DB. I think that's acceptable: previews are non-production and already expose staging data and dev tooling (the email redirect, for one). Test-user emails are confined to the known `*@test.com` set, and the staging DB is PII-free. The task-server integration is also off by default unless a preview is explicitly wired to an opencouncil-tasks preview.
- Real production is protected by two independent layers: it sets neither `NODE_ENV=development` nor `IS_PREVIEW=true`, so QuickLogin is dead-code-eliminated from the bundle and the API routes return 403.
- The one residual risk: if someone manually set `IS_PREVIEW=true` on the production host, Quick Login would activate. That flag is only ever set by the preview systemd unit, and production uses a separate deployment path, so this leans on deployment hygiene rather than hard-coding a production domain into the code. I left out an optional hardening (reject when `NEXTAUTH_URL` is the production domain) to keep the change small, but happy to add it if you'd rather have the belt-and-suspenders version.

<h3>Greptile Summary</h3>

This PR enables the Quick Login dev tool on preview deployments (`pr-N.preview.opencouncil.gr`) by introducing `IS_PREVIEW` and `DEV_TOOLS_ALLOWED` constants and threading `isPreview` as a boolean prop from server components down to the two client components that gate the UI.

- `src/lib/utils.ts` gains `IS_PREVIEW` and `DEV_TOOLS_ALLOWED`; API routes swap their `IS_DEV` guards for `DEV_TOOLS_ALLOWED` (server-side, so `process.env.IS_PREVIEW` is readable at runtime).
- `QuickLogin.tsx` and `DevelopmentSection.tsx` receive `isPreview` as a serialized prop from their respective server-side parents (`layout.tsx`, `profile/page.tsx`) instead of reading the env var directly — correctly working around the `NEXT_PUBLIC_` requirement for client bundles.
- Unit tests cover production, preview, dev, and non-`"true"` values for the gate.

Safe to merge — the change is well-scoped to dev tooling, real production is protected by two independent gates (neither NODE_ENV=development nor IS_PREVIEW=true is set there), and the client-bundle env-var limitation is correctly handled via server-to-client prop passing.

The implementation correctly threads isPreview as a serialized prop from server components to client components, avoiding the NEXT_PUBLIC_ requirement. API route guards are enforced server-side via DEV_TOOLS_ALLOWED. The secure-cookie logic intentionally stays keyed on IS_DEV (previews run over HTTPS). Tests cover all four gate combinations with proper module isolation. No logic errors or security regressions found.

| Filename | Overview |
|----------|----------|
| src/lib/utils.ts | Adds IS_PREVIEW and DEV_TOOLS_ALLOWED exports alongside IS_DEV; straightforward and correct. |
| src/app/[locale]/layout.tsx | Server component reads process.env.IS_PREVIEW at runtime and passes it as a prop to QuickLogin; dynamic require guard extended with the IS_PREVIEW branch. |
| src/components/dev/QuickLogin.tsx | Correctly uses IS_DEV (NODE_ENV-based, available client-side) combined with the isPreview prop; does not attempt to read IS_PREVIEW directly in the client bundle. |
| src/components/profile/DevelopmentSection.tsx | Same correct pattern as QuickLogin; isPreview prop from server, IS_DEV from bundle-inlined NODE_ENV. |
| src/app/api/dev/quick-login/route.ts | Guard updated to DEV_TOOLS_ALLOWED; secure cookie logic intentionally kept on IS_DEV (correct, previews are HTTPS). |
| src/app/api/dev/seed-test-users/route.ts | Both POST and GET handlers updated to DEV_TOOLS_ALLOWED; no other changes. |
| src/lib/__tests__/dev-tools-allowed.test.ts | Covers all four gate combinations; uses jest.isolateModules for correct module-reload behavior. |
| src/app/[locale]/(other)/profile/page.tsx | Server component correctly passes IS_PREVIEW env var as isPreview prop to DevelopmentSection. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(dev): pass IS\_PREVIEW from server so..."](https://github.com/schemalabz/opencouncil/commit/0690c48cafd66808f373fb5506c8d3171c97b0f3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36655958)</sub>

### DIFF
--- a/src/app/[locale]/(other)/profile/page.tsx
+++ b/src/app/[locale]/(other)/profile/page.tsx
@@ -14,7 +14,7 @@ export default async function ProfilePage() {
     return (
         <div className="container max-w-2xl py-8 space-y-8 !px-3 sm:!px-8">
             <h1 className="text-3xl font-bold">{t("title")}</h1>
-            <DevelopmentSection />
+            <DevelopmentSection isPreview={process.env.IS_PREVIEW === 'true'} />
             {user.onboarded && (user.isSuperAdmin || user.administers.length > 0) && <AdminSection user={user} t={t} />}
             <UserInfoForm user={user} isOnboarded={!!user.onboarded} />
         </div>
--- a/src/app/[locale]/layout.tsx
+++ b/src/app/[locale]/layout.tsx
@@ -40,7 +40,7 @@ export default async function LocaleLayout({
             {children}
 
             <Toaster />
-            {QuickLogin && <QuickLogin />}
+            {QuickLogin && <QuickLogin isPreview={process.env.IS_PREVIEW === 'true'} />}
             {MobilePreviewReporter && <MobilePreviewReporter />}
         </NextIntlClientProvider>
     );
--- a/src/app/[locale]/layout.tsx
+++ b/src/app/[locale]/layout.tsx
@@ -6,8 +6,11 @@ import { routing } from '@/i18n/routing';
 import { notFound } from "next/navigation";
 import { Toaster } from "@/components/ui/toaster";
 
-// Only import in development — excluded from production bundles entirely
-const QuickLogin = process.env.NODE_ENV === 'development'
+// Only import in development or on preview deployments — excluded from real
+// production bundles entirely. Both branches use literal `process.env.X === '...'`
+// comparisons so the bundler can dead-code-eliminate QuickLogin when neither flag
+// is set (i.e. real production).
+const QuickLogin = process.env.NODE_ENV === 'development' || process.env.IS_PREVIEW === 'true'
     ? require("@/components/dev/QuickLogin").default
     : null;
 const MobilePreviewReporter = process.env.NODE_ENV === 'development'
--- a/src/app/api/dev/quick-login/route.ts
+++ b/src/app/api/dev/quick-login/route.ts
@@ -1,11 +1,11 @@
 import { NextRequest, NextResponse } from 'next/server'
 import { prisma } from '@/lib/db/prisma'
 import { PrismaAdapter } from "@auth/prisma-adapter"
-import { IS_DEV } from '@/lib/utils'
+import { IS_DEV, DEV_TOOLS_ALLOWED } from '@/lib/utils'
 
 export async function POST(request: NextRequest) {
-  // Only allow in development environment
-  if (!IS_DEV) {
+  // Only allow in development or on preview deployments — never on real production
+  if (!DEV_TOOLS_ALLOWED) {
     return NextResponse.json({ error: 'Not allowed in production' }, { status: 403 })
   }
 
@@ -41,8 +41,10 @@ export async function POST(request: NextRequest) {
       })
     }
 
-    // Set the session cookie with the same name Next-Auth uses
-    // In dev, use port-specific cookie name to allow multiple instances
+    // Set the session cookie with the same name Next-Auth uses.
+    // In local dev, use a port-specific non-secure cookie to allow multiple instances.
+    // Previews run over HTTPS in production mode, so they use the secure cookie name
+    // (this intentionally keys off IS_DEV, not DEV_TOOLS_ALLOWED).
     const port = process.env.APP_PORT || '3000'
     const cookieName = !IS_DEV
       ? '__Secure-authjs.session-token'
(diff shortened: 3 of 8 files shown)


angelospk:
Pushed 0690c48c to fix the client-bundle issue Greptile flagged. Good catch.

The problem: `IS_PREVIEW` isn't a `NEXT_PUBLIC_` var, and it's only set at runtime by the preview systemd unit, so `process.env.IS_PREVIEW` is `undefined` in the client bundle. That made the client-side gate always false on previews. The API routes worked (server-side), but `QuickLogin` / `DevelopmentSection` rendered `null` after hydration, so the UI never actually showed up.

Exposing it via `next.config.mjs` `env` or `NEXT_PUBLIC_` wouldn't help either: the build is environment-agnostic and `IS_PREVIEW` only exists at runtime, so build-time inlining would just capture `undefined`. Instead I pass the runtime value down as an `isPreview` prop from the two server components that already render these (`layout.tsx`, `profile/page.tsx`), and gate on `IS_DEV || isPreview` on the client. `NODE_ENV` (`IS_DEV`) is still inlined into the client bundle by Next, so that part stays correct. The server-side gates (the `layout` require and the `/api/dev/*` routes) keep using `DEV_TOOLS_ALLOWED` unchanged.

`npx tsc --noEmit` is clean apart from the pre-existing `meetings/route.ts` errors, and the existing 4 unit tests still pass.

kouloumos:
> The task-server integration is also off by default unless a preview is explicitly wired to an opencouncil-tasks preview.

fyi staging task server is always connected in all preview deployments. This is fine for now, but we should find a better solution down the line.

PR #451, opened 2026-06-11
fix(notifications): announce upcoming meeting when agenda has no items

### LINKED ISSUE #308: Inconsistent notification behavior when processing an agenda with no items (λογοδοσία)
When processing an agenda for a meeting with no items, the notification behavior is inconsistent across different cases.

Λογοδοσία Athens (20/04): Agenda was processed with no items → no notifications were sent
Λογοδοσία Chania (22/04): Agenda was processed with no items → notifications were sent

Since both meetings had no agenda items, the behavior should be the same in both cases. The inconsistency suggests a bug in the notification logic.

Expected Behavior:
Notification behavior should be consistent across all meetings. 

Actual Behavior:
Notifications are sent in some cases and not in others, with no clear distinction between them.

Desired Behavior:
When an agenda of Λογοδοσία is processed with no items, a notification is still sent to subscribers informing them of the upcoming meeting (such as in Chania).

angelospk:
## What

Fixes inconsistent notification behavior when a meeting's agenda is processed with no items, which is typical for Λογοδοσία (accountability) sessions.

Two such meetings behaved differently. Athens (20/04) sent no notifications; Chania (22/04) sent them. Same situation, different outcome.

The cause: when an agenda has zero subjects, `createNotificationsForMeeting` builds an empty matching list, `matchUsersToSubjects` returns nothing, and the function returns early with `notificationsCreated: 0`. So the upcoming-meeting announcement gets silently dropped.

## Change

- For `beforeMeeting` notifications with zero subjects, notify every user subscribed to the city with a subject-less announcement. This follows the existing `topicImportance: 'high'` path, which already notifies all city subscribers. The `(userId, cityId, meetingId, type)` unique constraint keeps re-runs idempotent.
- `afterMeeting` is left alone, since a post-meeting summary with no subjects has nothing to report.
- The email template gets an announcement-only branch that drops the "Θέματα που σας αφορούν" heading.
- SMS drops the "N νέα θέματα" phrasing and announces the session instead.

One note on where this lives: the agenda PDF parser in opencouncil-tasks reports an empty agenda as an ordinary success (`{ subjects: [] }`) with no distinguishing signal, so the no-items behavior is handled here in the frontend rather than upstream.

## Tests

- `content.test.ts` — SMS announcement copy
- `NotificationEmail.test.tsx` — announcement-only render
- `integration/notifications.creation.test.ts` — no-items meeting produces notifications end to end

Closes #308

<h3>Greptile Summary</h3>

The PR adds subject-less upcoming-meeting announcements for city subscribers when an agenda contains no items, while preserving the no-op behavior for post-meeting notifications.
- Creates idempotent `beforeMeeting` notifications without subject matches for eligible city subscribers.
- Adds type-specific announcement copy for email and SMS.
- Adds unit and integration coverage for rendering, delivery preferences, post-meeting behavior, and reruns.

The PR appears safe to merge.

No blocking failure remains; both previously reported type-agnostic announcement branches are now restricted to `beforeMeeting`.

| Filename | Overview |
|----------|----------|
| src/lib/db/notifications.ts | Adds the zero-subject `beforeMeeting` fan-out while retaining city subscription, channel preference, notification policy, and idempotency behavior. |
| src/lib/email/templates/NotificationEmail.tsx | Renders announcement-only text specifically for zero-subject `beforeMeeting` emails; the previous type-agnostic branch has been corrected. |
| src/lib/notifications/content.ts | Generates subject-count-free SMS copy specifically for zero-subject `beforeMeeting` notifications; the previous type-agnostic branch has been corrected. |
| tests/integration/notifications.creation.test.ts | Covers empty-agenda fan-out, channel opt-ins, subject-less records, rerun idempotency, and the `afterMeeting` no-op contract. |
| src/lib/email/templates/__tests__/NotificationEmail.test.tsx | Verifies announcement-only email rendering and unchanged subject-list rendering. |
| src/lib/notifications/__tests__/content.test.ts | Verifies zero-subject announcement SMS copy and unchanged subject-bearing SMS content. |

<sub>Reviews (4): Last reviewed commit: ["fix(notifications): restrict no-items an..."](https://github.com/schemalabz/opencouncil/commit/0dd13f1995c557f8b78bd9699c19f55bed625acc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36929848)</sub>

### DIFF
--- a/src/lib/db/notifications.ts
+++ b/src/lib/db/notifications.ts
@@ -795,16 +795,34 @@ export async function createNotificationsForMeeting(
             }
         }
 
-        // Use shared matching logic to determine which users should be notified
-        const userSubjectMatches = await matchUsersToSubjects(
-            subjectsForMatching,
-            usersWithPreferences,
-            effectiveOverrides
-        );
+        // Determine which users should be notified.
+        //
+        // Special case — meeting announcement with no agenda items:
+        // Some sessions (notably Λογοδοσία/accountability meetings) are processed with
+        // zero subjects. Subject-based matching would notify nobody, silently dropping the
+        // upcoming-meeting announcement. For `beforeMeeting` we instead notify every user
+        // with a preference for this city, creating a subject-less announcement
+        // notification. This mirrors the existing `topicImportance: 'high'` behavior, which
+        // already notifies all city subscribers. `afterMeeting` keeps the old behavior:
+        // a post-meeting summary with no subjects has nothing to report.
+        const isMeetingAnnouncement = subjectsForMatching.length === 0 && type === 'beforeMeeting';
 
-        // Filter out users with no matching subjects
-        const usersToNotify = Array.from(userSubjectMatches.entries())
-            .filter(([_, subjects]) => subjects.size > 0);
+        let usersToNotify: Array<[string, Set<{ subjectId: string; reason: 'proximity' | 'topic' | 'generalInterest' }>]>;
+        if (isMeetingAnnouncement) {
+            console.log('Meeting has no subjects — creating beforeMeeting announcement for all city subscribers');
+            usersToNotify = usersWithPreferences.map(user => [user.userId, new Set()]);
+        } else {
+            // Use shared matching logic to determine which users should be notified
+            const userSubjectMatches = await matchUsersToSubjects(
+                subjectsForMatching,
+                usersWithPreferences,
+                effectiveOverrides
+            );
+
+            // Filter out users with no matching subjects
+            usersToNotify = Array.from(userSubjectMatches.entries())
+                .filter(([_, subjects]) => subjects.size > 0);
+        }
 
         if (usersToNotify.length === 0) {
             console.log('No users matched any subjects');
--- a/src/lib/db/notifications.ts
+++ b/src/lib/db/notifications.ts
@@ -758,16 +758,34 @@ export async function createNotificationsForMeeting(
             }
         }
 
-        // Use shared matching logic to determine which users should be notified
-        const userSubjectMatches = await matchUsersToSubjects(
-            subjectsForMatching,
-            usersWithPreferences,
-            effectiveOverrides
-        );
+        // Determine which users should be notified.
+        //
+        // Special case — meeting announcement with no agenda items:
+        // Some sessions (notably Λογοδοσία/accountability meetings) are processed with
+        // zero subjects. Subject-based matching would notify nobody, silently dropping the
+        // upcoming-meeting announcement. For `beforeMeeting` we instead notify every user
+        // with a preference for this city, creating a subject-less announcement
+        // notification. This mirrors the existing `topicImportance: 'high'` behavior, which
+        // already notifies all city subscribers. `afterMeeting` keeps the old behavior:
+        // a post-meeting summary with no subjects has nothing to report.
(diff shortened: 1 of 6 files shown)


angelospk:
Pushed da7b2cf: added a `type === 'beforeMeeting'` guard to the zero-subject branch in both the SMS and email paths, so a zero-subject `afterMeeting` notification can't render an "upcoming session" message for a meeting that already happened. Matches the guard already in `createNotificationsForMeeting`. Unit tests still green.

src/lib/db/notifications.ts
```diff
@@ -758,16 +758,34 @@ export async function createNotificationsForMeeting(
...
+        //
+        // Special case — meeting announcement with no agenda items:
+        // Some sessions (notably Λογοδοσία/accountability meetings) are processed with
+        // zero subjects. Subject-based matching would notify nobody, silently dropping the
+        // upcoming-meeting announcement. For `beforeMeeting` we instead notify every user
+        // with a preference for this city, creating a subject-less announcement
+        // notification. This mirrors the existing `topicImportance: 'high'` behavior, which
+        // already notifies all city subscribers. `afterMeeting` keeps the old behavior:
+        // a post-meeting summary with no subjects has nothing to report.
+        const isMeetingAnnouncement = subjectsForMatching.length === 0 && type === 'beforeMeeting';
```
kouloumos: The reasoning here is sound and the comment is unusually good — it explains the case, the precedent, and why `afterMeeting` is excluded. I am not asking for a change, but I want one consequence written down rather than left implicit.

`subjectsForMatching.length === 0` is both a **real state** (a Λογοδοσία with genuinely no agenda items) and a **failure state** (an agenda PDF that could not be parsed, or a run that produced nothing). The response to it changes from "do nothing" to "notify every subscriber in the city".

I checked the two things that bound the risk rather than assuming them:

- The call site is `processAgenda.ts:99`, *after* the agenda task has run — so zero subjects is a completed result, not a pending one.
- Notifications only auto-send when the administrative body is `NOTIFICATIONS_AUTO`; otherwise they queue for approval.

So the actual exposure is narrow: a body on AUTO whose agenda failed to parse would mail everyone. That is small enough that I would merge it, and specific enough that it should be a decision.

If we wanted a floor later, the obvious one is to require that the meeting has an agenda URL but produced no items — that separates a real empty agenda from a failed parse. Not for this PR.

src/lib/db/notifications.ts
```diff
@@ -758,16 +758,34 @@ export async function createNotificationsForMeeting(
...
+
+        let usersToNotify: Array<[string, Set<{ subjectId: string; reason: 'proximity' | 'topic' | 'generalInterest' }>]>;
+        if (isMeetingAnnouncement) {
+            console.log('Meeting has no subjects — creating beforeMeeting announcement for all city subscribers');
+            usersToNotify = usersWithPreferences.map(user => [user.userId, new Set()]);
```
kouloumos: Confirming this line does what it says, because it is the one that decides the blast radius.

In my run it created **6** notifications while only **5** belonged to the users I had created — the sixth reached a city subscriber already in the database that my test knew nothing about. So this really is every subscriber for the city, not just the ones the test set up.

That is the intended behaviour and it matches the `topicImportance: 'high'` precedent you cite. Recording the measurement because "notifies all city subscribers" is the kind of claim that deserves a number behind it.

PR #454, opened 2026-06-11
feat(voiceprint): manual segment selection with audio + transcript preview

### LINKED ISSUE #88: voiceprints: Manual Segment Selection
### Concept
Allow administrators to manually select or choose from multiple speaker segments when generating voiceprints, instead of relying solely on the automated longest-segment heuristic.

### User Story
As a city administrator, I want to manually select the best speaker segment for voiceprint generation so that I can ensure high-quality voiceprints when the automated selection chooses a poor-quality audio segment.

## Current State
- **Technology**: We use `findLongestSpeakerSegmentForPerson()` in `src/lib/tasks/generateVoiceprint.ts` to automatically select the longest speaker segment for voiceprint generation
- **Configuration**: The system takes a 30-second snippet centered on the midpoint of the selected segment (`VOICEPRINT_DURATION = 30`)
- **Issue**: The automated selection sometimes chooses segments with poor audio quality, background noise, or unclear speech, leading to suboptimal voiceprints

## Technical Considerations
- **Database**: Leverage existing `SpeakerSegment` model with `speakerTag.personId` relationship
- **Audio Preview**: Use existing meeting audio/video URLs with timestamp-based seeking
- **UI Integration**: Extend existing admin interface in `src/components/meetings/admin/`
- **Backward Compatibility**: Maintain current automated selection as default behavior
- **Authorization**: Ensure manual selection respects existing city-level permissions via `withUserAuthorizedToEdit()`

## Alternative Implementation (Simpler)
If full manual selection is too complex, implement a "top N segments" approach:
1. **Modify `findLongestSpeakerSegmentForPerson()`** to return top 3-5 longest segments instead of just the longest
2. **Add UI component** to display these alternatives with basic metadata
3. **Allow selection** from the presented options while keeping automated selection as default

## Leveraging Existing Transcript Interface
A more elegant approach would leverage the existing transcript page and highlight selection patterns:

### **Admin Transcript View Enhancement**
- **Speaker Status Indicators**: Show which speakers already have voiceprints generated (visual indicator on speaker segments)
- **Voiceprint Playback**: Allow listening to existing voiceprints directly from the transcript interface
- **Voiceprint Generation Mode**: Similar to highlight editing mode, but for voiceprint segment selection

### **Selection Workflow**
1. **Enter Voiceprint Generation Mode**: Similar to highlight editing mode, but focused on single speaker
2. **Utterance Selection**: Use existing utterance selection UI (same as highlights) to choose segments
3. **Duration Constraint**: Enforce 30-second maximum selection with visual feedback
4. **Preview & Generate**: Preview selected segment and proceed to voiceprint generation

### **Human Review Integration**
- **Speaker Identification Aid**: During human review, display voiceprint snippets for uncertain speaker identifications
- **Quick Reference**: Allow reviewers to quickly listen to voiceprint samples to verify speaker assignments
- **Contextual Help**: Show voiceprint availability status during transcript editing workflow

This approach reuses existing UI patterns from the highlights system (`HighlightContext`, utterance selection, transcript controls) while providing a familiar interface for administrators.

> [!NOTE]
> **Note for Contributors**: Before you begin, please read our [CONTRIBUTING.md](CONTRIBUTING.md) guide to understand our development workflow and co-creation process. We look forward to your contribution!

angelospk:
## What

Admins can now manually pick which speaker segment a person's voiceprint is built from, instead of always relying on the automatic longest-segment heuristic. This is the "simpler alternative" path described in #88: the automatic selection sometimes lands on a noisy or unclear segment, and there was no way to override it.

## How it works

In the admin People page, the voiceprint dialog for a person without a voiceprint now offers "Choose segment manually" next to the existing auto-select. Expanding it lists the person's longest eligible (>=30s) segments, and for each one the admin can:

- Listen to the exact 30-second window that will be used, via an inline audio player seeked to that window (`#t=start,end`) on the meeting media.
- Read the transcript of that same 30s window inline, to follow along, plus the full segment transcript behind a collapsible.

"What you hear" equals "what the voiceprint job consumes": a single `computeVoiceprintWindow` helper computes the centered 30s window, shared by both task dispatch and the candidate list.
## How it looks:
Screenshot 2026-06-11 at 4 35 38 PM

## Notes

- Auto-select stays the default; manual selection is opt-in.
- Server actions validate the chosen segment belongs to the person, enforce the 30s minimum, and authorize the caller for the segment's city, the same path as the automatic flow.
- No backend (opencouncil-tasks) change: the generate task already accepts an explicit segment plus window.
- Scope: this is the focused "manual single-segment selection" from #88, not the larger in-transcript voiceprint mode the issue also sketches.

## Tests

`generateVoiceprint.test.ts` covers `computeVoiceprintWindow` (centered/edge/invalid), candidate filtering and sorting, the enriched payload (mediaUrl, preview window, windowText vs fullText), and segment validation. Verified end to end against a real meeting: the audio loads with a 206 range request to the correct window and the transcript aligns.

Closes #88

<h3>Greptile Summary</h3>

The PR adds manual voiceprint segment selection with synchronized audio and transcript previews while retaining automatic selection.
- Adds a reusable helper for computing the centered voiceprint window.
- Adds candidate loading, selection, retry, and request-lifecycle handling to the admin dialog.
- Validates manually selected segments and dispatches the existing voiceprint task with an explicit segment and window.
- Adds coverage for candidate construction, window calculation, and manual dispatch validation.

The PR is not yet safe to merge because the manual dispatch path can still create a voiceprint for a person outside the administrator's authorized city.

The ownership check accepts a matching person ID while dispatch authorization checks only the segment's city, and the result is subsequently persisted against the caller-provided person ID; the previously reported cross-city authorization failure therefore remains.

**Files Needing Attention:** src/lib/tasks/generateVoiceprint.ts

| Filename | Overview |
|----------|----------|
| src/components/admin/people/voiceprint-actions.tsx | Adds the manual candidate picker and correctly addresses the previously reported dialog-reset, loading-state, duration-label, and language inconsistencies. |
| src/lib/tasks/generateVoiceprint.ts | Adds candidate retrieval and explicit-segment dispatch, but the previously reported cross-city authorization boundary remains unenforced. |
| src/lib/tasks/voiceprintWindow.ts | Centralizes finite, validated computation of the voiceprint preview and dispatch window. |
| src/lib/__tests__/generateVoiceprint.test.ts | Covers window computation, candidate enrichment, ownership rejection, minimum duration, and task payload construction, but not the outstanding cross-city authorization case. |

<sub>Reviews (10): Last reviewed commit: ["fix(voiceprint): reset loading state and..."](https://github.com/schemalabz/opencouncil/commit/45431b659463cd2b92baed43fbb1d3a5366c6c1c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36957394)</sub>

### DIFF
--- a/src/components/admin/people/voiceprint-actions.tsx
+++ b/src/components/admin/people/voiceprint-actions.tsx
@@ -344,7 +344,7 @@ export function VoiceprintActions({ personId, personName, voicePrint }: Voicepri
                                                 ) : (
                                                     <>
                                                         <p className='text-xs text-slate-500'>
-                                                            Pick the segment with the clearest audio. A 30-second window
+                                                            Pick the segment with the clearest audio. A {VOICEPRINT_DURATION}-second window
                                                             centered on the segment will be used.
                                                         </p>
                                                         <ul className='max-h-[45vh] space-y-2 overflow-y-auto pr-1'>
--- a/src/components/admin/people/voiceprint-actions.tsx
+++ b/src/components/admin/people/voiceprint-actions.tsx
@@ -20,6 +20,7 @@ import {
     getCandidateSegmentsForVoiceprint,
     type VoiceprintCandidateSegment,
 } from "@/lib/tasks/generateVoiceprint";
+import { VOICEPRINT_DURATION } from "@/lib/tasks/voiceprintWindow";
 import { deleteTaskStatus, getVoiceprintTasksForPerson } from "@/lib/db/tasks";
 import { useToast } from "@/hooks/use-toast";
 import { useRouter } from "next/navigation";
@@ -48,6 +49,19 @@ export function VoiceprintActions({ personId, personName, voicePrint }: Voicepri
     const [candidatesLoaded, setCandidatesLoaded] = useState(false);
     const [selectedSegmentId, setSelectedSegmentId] = useState<string | null>(null);
     const { toast } = useToast();
+
+    // Reset the manual picker whenever the dialog closes so reopening starts clean
+    // (no stale candidates expanded, no previously highlighted segment that could be
+    // generated from by accident).
+    const handleDialogOpenChange = useCallback((open: boolean) => {
+        setIsDialogOpen(open);
+        if (!open) {
+            setIsManualOpen(false);
+            setSelectedSegmentId(null);
+            setCandidates([]);
+            setCandidatesLoaded(false);
+        }
+    }, []);
 
     const fetchTaskStatuses = useCallback(async () => {
         try {
@@ -238,7 +252,7 @@ export function VoiceprintActions({ personId, personName, voicePrint }: Voicepri
     const canGenerateVoiceprint = !voicePrint && !latestPendingTask;
 
     return (
-        <Dialog open={isDialogOpen} onOpenChange={setIsDialogOpen}>
+        <Dialog open={isDialogOpen} onOpenChange={handleDialogOpenChange}>
             <DialogTrigger asChild>
                 <Button variant='outline' size='sm' className={voicePrint ? "border-blue-300 text-blue-700" : ""}>
                     <Volume2 className='mr-2 h-4 w-4' />
@@ -325,7 +339,7 @@ export function VoiceprintActions({ personId, personName, voicePrint }: Voicepri
                                                     </div>
                                                 ) : candidates.length === 0 ? (
                                                     <p className='text-sm text-slate-600'>
-                                                        No segments of at least 30 seconds are available for this person.
+                                                        No segments of at least {VOICEPRINT_DURATION} seconds are available for this person.
                                                     </p>
                                                 ) : (
                                                     <>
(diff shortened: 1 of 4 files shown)


angelospk:
Pushed two follow-up commits addressing the review.

**Hardening (f1646098):**
- Guard the ownership check against a missing `speakerTag` relation, so an unexpected null raises the "doesn't belong" error instead of a TypeError. The FK is required by the schema, so this is purely defensive.
- Reset the manual picker (expanded state, loaded candidates, selection) when the dialog closes, so reopening can't generate from a stale segment.
- Switched the audio `aria-label` and the transcript summary to English, to match the rest of the admin component.

**Query and copy (5263e622):**
- `getCandidateSegmentsForVoiceprint` no longer loads every segment's full transcript up front. It scans segments cheaply (timestamps and meeting metadata only), keeps the longest eligible ones, then fetches utterances only for those. I didn't use the suggested `orderBy: endTimestamp desc` cap, because `endTimestamp` doesn't track duration: it would bias toward late-in-meeting segments and could drop the actual longest ones. Capping the cheap first pass keeps the result bounded without that tradeoff.
- The picker hint text now interpolates `VOICEPRINT_DURATION` instead of a hardcoded "30-second".

Unit tests cover all of it, and I checked the picker end to end against a real meeting: candidates load, the audio seeks to the right window, and the transcript lines up with it.

Issue #69, opened 2025-07-23
Consider upgrading to Next.js 15


christosporios:
Evaluate the pros/cons of upgrading to v15; and do the upgrade if needed.

kouloumos:
## Caching makes this upgrade urgent

We've been running a shared Valkey cache (`@neshca/cache-handler` v1.9.0) for multi-instance cache sharing, and it's fundamentally broken — `revalidateTag` silently fails, leaving stale data for users. See #358 for the full investigation.

The library is unmaintained and only supports Next.js 13-14. The alternatives that actually work in production all require Next.js 15+:

- **[nextjs-turbo-redis-cache](https://github.com/trieb-work/nextjs-turbo-redis-cache)** (Next.js 15+): Uses Redis pub/sub for cross-instance coordination + timestamp-based lazy invalidation
- **[@fortedigital/nextjs-cache-handler](https://github.com/fortedigital/nextjs-cache-handler) v2+** (Next.js 15+): Evolved fork of `@neshca`, fixes the sharedTags memory leak
- **Next.js 16's native `cacheHandlers` API**: Built-in distributed coordination with `refreshTags()` (called before every request) and `updateTags()` (writes to shared storage)

## Recommended upgrade path: 14 → 16 directly

Next.js 16 is stable (current: 16.2.6). The `npx @next/codemod@canary upgrade` tool supports jumping from 14 to 16 directly — it collects all codemods from both v15 and v16 and applies them in one shot.

### Major breaking changes to absorb

**From v15:**
- Async request APIs: `cookies()`, `headers()`, `params`, `searchParams` become async (biggest change — the codemod handles most of it but needs manual review)
- Caching defaults flipped: `fetch()` and GET route handlers no longer cached by default
- React 19

**From v16:**
- `middleware.ts` renamed to `proxy.ts` (affects next-intl and Auth.js integration)
- Turbopack becomes default bundler (use `--webpack` flag to opt out)
- `revalidateTag()` signature changes (requires second argument)
- Node.js 20.9+ required
- `@neshca/cache-handler` must be replaced (incompatible)

### Library compatibility
- **next-intl**: Will need upgrade to v4.x for Next.js 16 (ESM-only, handles proxy.ts rename)
- **Auth.js**: Supports Next.js 16, needs proxy.ts adaptation
- **Prisma**: Unaffected
- **Tailwind CSS**: v3 still works, v4 optional

### Suggested approach
[...]

PR #470, opened 2026-06-15
fix(auth): configurable sign-in email sender + clearer Resend errors

### LINKED ISSUE #56: [sign-in] Local email sign-in fails due to resend test email and domain restrictions
### Summary
When setting up the project for local development, contributors encounter two major issues with the Resend email provider integration for authentication. These issues are not clearly addressed in the documentation and result in confusing errors that block local testing of the sign-in flow.

#### Error Scenarios
1. Test Email Restriction (Status 422)
- Steps to Reproduce:
  - Set up the project and a valid Resend API key in `.env`.
  - Attempt to sign in using a test/placeholder email (e.g., example@example.com).
- UI:
  - Shows "Server Error".
- Console:
  - `GET http://localhost:3000/api/auth/error?error=Configuration 500 (Internal Server Error)`
- Container Logs:
```bash
[auth][error] Error: Resend error: {"statusCode":422,"message":"Invalid `to` field. Please use our testing email address instead of domains like `example.com`. See our documentation for more information.","name":"validation_error"}
```
- Explanation:
  - Resend blocks sending to generic/test domains unless you use a verified/test recipient.

2. Unverified Domain Restriction (Status 403)

- Steps to Reproduce:
  - Set up the project and a valid Resend API key in `.env`.
  - Attempt to sign in using a real email (e.g., youremail@gmail.com).
- UI:
`Shows "Server error. There is a problem with the server configuration."`
- Console:
  - `GET http://localhost:3000/api/auth/error?error=Configuration 500 (Internal Server Error)`
- Container Logs:
```bash
[auth][error] Error: Resend error: {"statusCode":403,"message":"The opencouncil.gr domain is not verified. Please, add and verify your domain on https://resend.com/domains","name":"validation_error"}
```
- Explanation:
The default sender is auth@opencouncil.gr, but this domain is not verified in Resend, so all emails are blocked.

### Root Causes
- The documentation does not explain that contributors must use a test domain (e.g., resend.dev) for the "from" address in local development.
- The code does not make the "from" address easily configurable via environment variables.
- Error messages from Resend are not surfaced in a user-friendly way in the UI.
- Contributors may not realize that using `example.com` or an unverified sender domain will always fail.

### Expected Behavior
Contributors should be able to:
- Use a test domain (e.g., resend.dev) for the "from" address in local development.
- Configure the "from" address via .env without code changes.
- Test sign-in with their own email address as the recipient.
- Documentation should clearly explain these requirements and steps.

angelospk:
## What & why

New contributors and self-hosters couldn't sign in locally. Two Resend failures, both opaque:

- 403 — the sign-in `from` address was hardcoded to the production domain (`auth@opencouncil.gr`), which isn't verified on a fork's Resend account, so every send was blocked.
- 422 — sending magic links to placeholder recipients like `example@example.com` is rejected by Resend.

The `from` address wasn't configurable, and Resend's error was rethrown as an unhelpful JSON blob, so contributors just saw a generic server error.

## Changes

- `src/env.mjs`: add `AUTH_EMAIL_FROM` server env var (default `OpenCouncil <auth@opencouncil.gr>`, so production is unchanged when unset), wired into `runtimeEnv`.
- `src/auth.config.ts`: use `env.AUTH_EMAIL_FROM` for the Resend provider's `from`; on a failed send, log an actionable message (status code + Resend message + a fix hint) under an `[Auth][Resend]` prefix and rethrow it.
- `src/lib/email/authError.ts` (new): pure helper that builds the error message (403 → sender-domain hint, 422 → recipient hint, fallback to statusText), so it's unit-testable.
- `src/lib/email/__tests__/authError.test.ts` (new): 3 tests for the 403/422/fallback paths.
- `.env.example`: add `AUTH_EMAIL_FROM="OpenCouncil <onboarding@resend.dev>"` with a note about Resend test mode.
- `docs/environment-variables.md`: document `AUTH_EMAIL_FROM` and a "Resend setup for local development" section (resend.dev sender + own-email recipient rule).

## Testing

- `npx jest src/lib/email/__tests__/authError.test.ts` — 3 passing.
- `npx tsc --noEmit` — no errors in the changed files.
- Manual 403/422 reproduction steps are in the QA notes.

## Risks / notes

- Default keeps the current production sender, so leaving it unset is a no-op in prod.
- The sign-in page still shows the existing generic message; the richer detail lands in the server logs, where someone debugging locally will look. A user-facing error-page rework overlaps with #69 and is out of scope here.

Closes #56

<h3>Greptile Summary</h3>

This PR fixes two common local-dev/self-hosting sign-in failures caused by a hardcoded `from` address and opaque Resend error messages. It makes the sender address configurable via `AUTH_EMAIL_FROM` (defaulting to the existing production value so deployed environments are unaffected) and replaces the raw JSON blob rethrow with structured, hint-bearing error messages.

- **`AUTH_EMAIL_FROM` env var** added to `src/env.mjs` with a production-safe default; `src/auth.config.ts` uses it in place of the hardcoded string.
- **Improved Resend error handling** in `auth.config.ts`: JSON body is parsed defensively, a structured log line is emitted under `[Auth][Resend]`, and the thrown error carries a human-readable message built by the new `buildResendAuthErrorMessage` helper.
- **New pure helper + tests** in `src/lib/email/authError.ts` and its test file cover the 403 (unverified domain), 422 (blocked recipient), and fallback paths; trailing-period stripping prevents `..` artifacts in the output string.

- Safe to merge — the change is backward-compatible (production sender is unchanged when AUTH_EMAIL_FROM is unset) and the error-handling path only affects failed sends.
- All changed code paths are focused and low-risk: the env default keeps production behavior identical, the Resend error branch only fires on already-failed sends, and the new helper is pure and covered by unit tests. No data, auth, or session logic is altered.
- No files require special attention.

| Filename | Overview |
|----------|----------|
| src/env.mjs | Adds AUTH_EMAIL_FROM server env var with sensible default; properly wired into runtimeEnv. Validation is a simple non-empty string check, which is sufficient since Resend will report format errors at send time. |
| src/auth.config.ts | Replaces hardcoded "from" with env.AUTH_EMAIL_FROM and wraps the Resend error path with defensive JSON parsing, structured logging, and a rethrown actionable error. The catch(() => ({})) guard on res.json() is correct. |
| src/lib/email/authError.ts | New pure helper builds actionable error messages for 403/422/fallback Resend failures. Trailing-period stripping via .replace(/\.+$/, '') correctly prevents double-period artifacts. |
| src/lib/email/__tests__/authError.test.ts | Three tests cover the 403 (domain hint), 422 (recipient hint), and fallback (statusText) paths. The 403 test explicitly guards against the double-period artifact. The 500 fallback test uses toBe for exact match. |
| .env.example | Adds AUTH_EMAIL_FROM with the onboarding@resend.dev test sender and a clear note about the Resend test-mode delivery restriction. |
| docs/environment-variables.md | Documents AUTH_EMAIL_FROM in the table and adds a "Resend setup for local development" section explaining the 403 and 422 failure modes with clear fix instructions. |

<sub>Reviews (5): Last reviewed commit: ["fix(auth): configurable sign-in email se..."](https://github.com/schemalabz/opencouncil/commit/34f79deb92b7eba948fee96a3585778c2e77c02f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37570712)</sub>

### DIFF
--- a/src/auth.config.ts
+++ b/src/auth.config.ts
@@ -4,6 +4,7 @@ import { AuthEmail } from "./lib/email/templates/AuthEmail"
 import { renderReactEmailToHtml } from "./lib/email/render"
 import { env } from "./env.mjs"
 import { isTestUserEmail } from "./lib/dev/test-users"
+import { buildResendAuthErrorMessage } from "./lib/email/authError"
 
 // In development, use port-specific session cookie names to allow multiple
 // instances on different ports to have independent sessions. Without this,
@@ -22,7 +23,7 @@ export default {
         },
     } : undefined,
     providers: [Resend({
-        from: 'OpenCouncil <auth@opencouncil.gr>',
+        from: env.AUTH_EMAIL_FROM,
         apiKey: env.RESEND_API_KEY,
         sendVerificationRequest: async (params) => {
             const { identifier: to, provider, url, theme } = params
@@ -51,8 +52,18 @@ export default {
                 }),
             })
 
-            if (!res.ok)
-                throw new Error("Resend error: " + JSON.stringify(await res.json()))
+            if (!res.ok) {
+                const body = await res.json().catch(() => ({})) as { name?: string; message?: string }
+                const errorMessage = buildResendAuthErrorMessage({
+                    status: res.status,
+                    from: provider.from,
+                    to: emailTo,
+                    body,
+                    statusText: res.statusText,
+                })
+                console.error(`[Auth][Resend] Failed to send sign-in email${body.name ? ` (${body.name})` : ''}: ${errorMessage}`)
+                throw new Error(errorMessage)
+            }
         }
     })],
 } satisfies NextAuthConfig
--- a/src/auth.config.ts
+++ b/src/auth.config.ts
@@ -7,6 +7,7 @@ import { isTestUserEmail } from "./lib/dev/test-users"
 import { signInUrlForRequest } from "./lib/auth/signInUrl"
 import { localeForRequest } from "./lib/auth/requestLocale"
 import { devSessionCookieName } from "./lib/auth/sessionMirror"
+import { buildResendAuthErrorMessage } from "./lib/email/authError"
 
 // In development, use port-specific session cookie names to allow multiple
 // instances on different ports to have independent sessions. Without this,
@@ -25,7 +26,7 @@ export default {
         },
     } : undefined,
     providers: [Resend({
-        from: 'OpenCouncil <auth@opencouncil.gr>',
+        from: env.AUTH_EMAIL_FROM,
         apiKey: env.RESEND_API_KEY,
         sendVerificationRequest: async (params) => {
             const { identifier: to, provider, url, request } = params
@@ -62,8 +63,18 @@ export default {
                 }),
             })
 
-            if (!res.ok)
-                throw new Error("Resend error: " + JSON.stringify(await res.json()))
+            if (!res.ok) {
+                const body = await res.json().catch(() => ({})) as { name?: string; message?: string }
+                const errorMessage = buildResendAuthErrorMessage({
+                    status: res.status,
+                    from: provider.from,
+                    to: emailTo,
+                    body,
+                    statusText: res.statusText,
+                })
+                console.error(`[Auth][Resend] Failed to send sign-in email${body.name ? ` (${body.name})` : ''}: ${errorMessage}`)
+                throw new Error(errorMessage)
+            }
         }
     })],
 } satisfies NextAuthConfig
--- a/src/env.mjs
+++ b/src/env.mjs
@@ -17,6 +17,10 @@ export const env = createEnv({
 
     // Auth
     RESEND_API_KEY: z.string().min(1),
+    // Sender ("from") address for sign-in magic-link emails. In local dev,
+    // set this to a Resend test sender (e.g. "OpenCouncil <onboarding@resend.dev>")
+    // since the production domain isn't verified on a fork's Resend account.
+    AUTH_EMAIL_FROM: z.string().min(1).default('OpenCouncil <auth@opencouncil.gr>'),
     NEXTAUTH_SECRET: z.string().min(1),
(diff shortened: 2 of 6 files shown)


src/env.mjs
```diff
@@ -46,6 +46,10 @@ export const env = createEnv({
 
     // Auth
     RESEND_API_KEY: z.string().min(1),
+    // Sender ("from") address for sign-in magic-link emails. In local dev,
+    // set this to a Resend test sender (e.g. "OpenCouncil <onboarding@resend.dev>")
+    // since the production domain isn't verified on a fork's Resend account.
+    AUTH_EMAIL_FROM: z.string().min(1).default('OpenCouncil <auth@opencouncil.gr>'),
```
kouloumos: This default is what makes the change safe to merge without touching any deployment. It is byte-identical to the string it replaces in `auth.config.ts`, so an environment that sets nothing keeps the current sender.

`.min(1)` also keeps the failure loud if somebody sets the variable to an empty string, instead of silently sending from an empty address.

src/auth.config.ts
```diff
@@ -62,8 +63,18 @@ export default {
                 }),
             })
 
-            if (!res.ok)
-                throw new Error("Resend error: " + JSON.stringify(await res.json()))
+            if (!res.ok) {
+                const body = await res.json().catch(() => ({})) as { name?: string; message?: string }
+                const errorMessage = buildResendAuthErrorMessage({
+                    status: res.status,
+                    from: provider.from,
```
kouloumos: I checked this at runtime because it reads like a bug and is not one.

`provider.from` on the provider object is the Auth.js default, and your configured value is at `provider.options.from`. Auth.js merges `options` over the defaults before it invokes this handler, so by the time this line runs, `provider.from` is the value from `AUTH_EMAIL_FROM`. I confirmed it: with `AUTH_EMAIL_FROM` set to a test value, the captured POST body to `https://api.resend.com/emails` carried that value, and the 403 hint quoted it back correctly.

No change needed.

PR #471, opened 2026-06-15
fix(summarize): run stale-data cleanup inside the repopulation transaction

### LINKED ISSUE #248: Re-summarize leaves stale topic labels and utterance discussion statuses
## Problem

When `handleSummarizeResult` runs a second time on the same meeting, summaries and subjects are properly replaced (upsert/update semantics), but **topic labels** and **utterance discussion statuses** accumulate stale data from previous runs.

### Topic labels accumulate

`TopicLabel` records are upserted with `update: {}` (no-op if exists), and old labels are never deleted before creating new ones.

```
Run 1: segment A → [Environment, Budget]     → 2 TopicLabel records created
Run 2: segment A → [Budget, Transport]       → Transport added, Budget unchanged
Result: segment A has [Environment, Budget, Transport]
                       ^^^^^^^^^^^ stale from run 1
```

**Relevant code:** `src/lib/tasks/summarize.ts` lines 86-107 — topic label upserts with no preceding cleanup.

### Utterance discussion statuses are partial

Only utterances present in the new response are updated. Utterances from the previous run that are not mentioned in the new response keep their old `discussionStatus` and `discussionSubjectId`.

```
Run 1: utterance A → SUBJECT_DISCUSSION, utterance B → VOTE
Run 2: only mentions utterance A → ATTENDANCE
Result: utterance A = ATTENDANCE (correct), utterance B = VOTE (stale from run 1)
```

**Relevant code:** `src/lib/tasks/summarize.ts` lines 124-191 — utterance updates only for IDs in response.

## Expected behavior

Re-summarize should produce the same result as if the meeting had been summarized for the first time with the new response. No stale data from previous runs should remain.

## Suggested fix

Before processing the new response, clear old data for the meeting's speaker segments:

1. **Topic labels:** Delete all `TopicLabel` records for the meeting's speaker segments before upserting new ones.
2. **Utterance discussion statuses:** Reset `discussionStatus` and `discussionSubjectId` to `null` for all utterances in the meeting's speaker segments before applying the new statuses.

Both cleanups should target only segments belonging to the specific meeting (scoped by `meetingId` + `cityId`), not all segments globally.

## Additional context

- Summaries and subjects already handle re-runs correctly (summary upsert, subject upsert by `agendaItemIndex`).
- Re-summarize is not a common operation, but when it happens the data should be consistent.
- Discovered while working on #211.

angelospk:
## What & why

Re-summarizing a meeting could leave it with no data. A force re-summarize already moved its stale-data cleanup out of `requestSummarize` and into `handleSummarizeResult` (so a failed task no longer wipes a meeting at request time). But the cleanup still ran as two separate awaited calls — `topicLabel.deleteMany` and `utterance.updateMany` — *before* the `$transaction` that repopulates summaries and topic labels.

So if the repopulation transaction failed, the cleanup had already committed: the meeting kept neither its old topic labels / discussion statuses nor any new ones.

## Changes

- `src/lib/tasks/summarize.ts`: prepend the two cleanup operations to the `operations` array, so cleanup and repopulation run in a single `$transaction`. If repopulation fails, the cleanup rolls back with it and the meeting keeps its previous data. Both cleanup ops now use the scalar `speakerSegmentId: { in: availableSpeakerSegmentIds }` form, consistent with how repopulation is scoped (equivalent coverage — that helper returns all of the meeting's segment ids).
- `src/lib/tasks/__tests__/summarizeForceCleanup.test.ts`: assert the scalar scoping and that cleanup is built into the same transaction as the save.

## Testing

- `npx jest summarizeForceCleanup` — 7 passing.
- `npx tsc --noEmit` — no errors in `summarize.ts`.

## Risks / notes

- Cleanup still runs on every successful callback (not only force), which keeps a normal re-run idempotent. The behavior change is purely the atomicity: same queries, now inside the transaction.
- Utterance discussion-status repopulation remains a separate best-effort pass (unchanged); this PR makes the topic-label/summary path atomic with its cleanup, which is where the "left empty on failure" risk was.

Closes #248

<h3>Greptile Summary</h3>

This PR fixes a data-loss race condition in `handleSummarizeResult`: stale-data cleanup (`topicLabel.deleteMany`, `utterance.updateMany`) previously ran as two sequential awaited calls *before* the repopulation `$transaction`, so a failed repopulation would leave the meeting with neither old nor new data. The fix prepends the cleanup operations to the `operations` array so all writes (cleanup + repopulation) execute atomically in one `$transaction`.

- **`summarize.ts`**: Cleanup ops are now array elements in `operations`, guaranteeing rollback on repopulation failure. The scope changes from a relational `speakerSegment: { meetingId, cityId }` filter to `speakerSegmentId: { in: availableSpeakerSegmentIds }` — confirmed equivalent because `getAvailableSpeakerSegmentIds` returns all segment IDs for the meeting with no additional filtering.
- **`summarizeForceCleanup.test.ts`**: Adds a Symbol-sentinel atomicity test that verifies the cleanup return values are actual *elements* of the `$transaction` array (not separately awaited), directly guarding the invariant this PR introduces.

- This PR is safe to merge — it tightens atomicity on an existing write path without changing observable behavior on the success case.
- The change is narrow and well-understood: two cleanup operations are prepended to an already-used $transaction array. The scope equivalence of speakerSegmentId: { in: availableSpeakerSegmentIds } vs the old relational filter is confirmed by reading getAvailableSpeakerSegmentIds, which performs a plain findMany with no extra predicates. The new Symbol-sentinel test actually guards the atomicity invariant rather than just asserting call counts. No new failure modes are introduced.
- No files require special attention.

| Filename | Overview |
|----------|----------|
| src/lib/tasks/summarize.ts | Cleanup ops (topicLabel.deleteMany, utterance.updateMany) moved from sequential pre-transaction awaits into the $transaction operations array; scope equivalence confirmed — getAvailableSpeakerSegmentIds returns all segment IDs for the meeting with no extra filtering. |
| src/lib/tasks/__tests__/summarizeForceCleanup.test.ts | Tests updated to assert scalar speakerSegmentId scoping; new Symbol-sentinel atomicity test properly verifies cleanup return values are elements of the $transaction array, not separately awaited. |

<sub>Reviews (4): Last reviewed commit: ["fix(summarize): run stale-data cleanup i..."](https://github.com/schemalabz/opencouncil/commit/7f5dd1a1e2c8f6c6e19a3d757f059ce4b0ddb1d2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37570800)</sub>

### DIFF
--- a/src/lib/tasks/__tests__/summarizeForceCleanup.test.ts
+++ b/src/lib/tasks/__tests__/summarizeForceCleanup.test.ts
@@ -124,22 +124,22 @@ describe('handleSummarizeResult — always cleans up stale data on success', () 
     jest.clearAllMocks();
   });
 
-  it('deletes topic labels scoped to the meeting', async () => {
+  it('deletes topic labels scoped to the meeting segments', async () => {
     await handleSummarizeResult(TASK_ID, EMPTY_RESPONSE);
 
     expect(mockTopicLabelDeleteMany).toHaveBeenCalledWith({
       where: {
-        speakerSegment: { meetingId: MEETING_ID, cityId: CITY_ID },
+        speakerSegmentId: { in: [] },
       },
     });
   });
 
-  it('resets utterance discussion statuses scoped to the meeting', async () => {
+  it('resets utterance discussion statuses scoped to the meeting segments', async () => {
     await handleSummarizeResult(TASK_ID, EMPTY_RESPONSE);
 
     expect(mockUtteranceUpdateMany).toHaveBeenCalledWith({
       where: {
-        speakerSegment: { meetingId: MEETING_ID, cityId: CITY_ID },
+        speakerSegmentId: { in: [] },
       },
       data: {
         discussionStatus: null,
@@ -148,7 +148,25 @@ describe('handleSummarizeResult — always cleans up stale data on success', () 
     });
   });
 
-  it('cleans up before saving new data', async () => {
+  it('passes the cleanup ops into the same $transaction array (atomicity)', async () => {
+    // Sentinel return values so we can assert by identity that the cleanup ops are
+    // ELEMENTS of the array handed to $transaction — not awaited separately before it.
+    // A sequential-await version would still call all three mocks, so asserting only
+    // "was called" wouldn't actually guard the atomicity invariant this PR introduces.
+    const deleteOp = Symbol('deleteTopicLabels');
+    const resetOp = Symbol('resetUtterances');
+    mockTopicLabelDeleteMany.mockReturnValueOnce(deleteOp);
+    mockUtteranceUpdateMany.mockReturnValueOnce(resetOp);
+
+    await handleSummarizeResult(TASK_ID, EMPTY_RESPONSE);
+
+    expect(mockTransaction).toHaveBeenCalled();
+    const [transactionArg] = mockTransaction.mock.calls[0];
+    expect(transactionArg).toContain(deleteOp);
+    expect(transactionArg).toContain(resetOp);
+  });
+
+  it('builds cleanup before saving new data', async () => {
     const callOrder: string[] = [];
     mockTopicLabelDeleteMany.mockImplementation(() => { callOrder.push('deleteTopicLabels'); return Promise.resolve({ count: 0 }); });
     mockUtteranceUpdateMany.mockImplementation(() => { callOrder.push('resetUtterances'); return Promise.resolve({ count: 0 }); });
--- a/src/lib/tasks/summarize.ts
+++ b/src/lib/tasks/summarize.ts
@@ -44,34 +44,6 @@ export async function handleSummarizeResult(taskId: string, response: SummarizeR
 
     const { councilMeeting } = task;
 
-    // Clean up stale data before saving new results.
-    // This runs on every successful callback (not just force), ensuring the DB
-    // is consistent even if a previous run left partial data.
-    // Summaries use upsert (keyed on speakerSegmentId) so they're overwritten naturally.
-    // TopicLabels and utterance discussion statuses need explicit cleanup because:
-    // - TopicLabels: upsert only creates/updates matching pairs; old labels for topics
-    //   no longer assigned to a segment remain as stale data.
-    // - Utterance statuses: only utterances present in the new response get updated;
-    //   utterances that had statuses from the previous run but aren't in the new
-    //   response keep their old (now stale) values.
-    console.log(`Cleaning up stale summarize data for meeting ${councilMeeting.id}`);
-
-    await prisma.topicLabel.deleteMany({
-        where: {
-            speakerSegment: { meetingId: councilMeeting.id, cityId: councilMeeting.cityId }
-        }
-    });
-
-    await prisma.utterance.updateMany({
-        where: {
-            speakerSegment: { meetingId: councilMeeting.id, cityId: councilMeeting.cityId }
-        },
-        data: {
-            discussionStatus: null,
-            discussionSubjectId: null
-        }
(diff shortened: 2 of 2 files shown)


src/lib/tasks/summarize.ts
```diff
@@ -87,8 +59,22 @@ export async function handleSummarizeResult(taskId: string, response: SummarizeR
     });
     const topicByName = new Map(topics.map(t => [t.name, t]));
 
-    // Prepare all operations for batch execution
-    const operations: any[] = [];
+    // Prepare all operations for batch execution.
+    // Stale-data cleanup is PREPENDED so it runs in the SAME transaction as the
+    // repopulation below: if repopulation fails, the cleanup rolls back too and the
```
kouloumos: This sentence is true for this transaction, but `handleSummarizeResult` commits two times.

- **Transaction 1** is this one. It deletes the topic labels, nulls the utterance discussion statuses, and upserts the summaries and labels.
- **Transaction 2** is inside `saveSubjectsForMeeting` (`src/lib/db/utils.ts:37`). It calls `applyUtteranceDiscussionStatusesInTx` (`src/lib/db/utils.ts:590`), which is the only place the discussion statuses are written back.

So the labels are now safe, because this transaction both deletes and re-creates them. The discussion statuses are not: they are nulled here and only restored in the later commit. When transaction 2 fails, the nulls stand.

I measured it. A forced failure in `saveSubjectsForMeeting` takes `athens/feb11_2026` from 4,823 utterances with a status to 0, on this branch and on main equally.

Two options, either is fine:

1. Narrow the comment to the transaction it describes, and open a follow-up issue for the status half.
2. Move the `utterance.updateMany` reset out of this array and into `saveSubjectsForMeeting`, next to the restore, so the reset and the restore share one commit.

Option 2 closes the gap. Option 1 is honest and keeps this PR at its current size. I would accept either.

src/lib/tasks/summarize.ts
```diff
@@ -87,8 +59,22 @@ export async function handleSummarizeResult(taskId: string, response: SummarizeR
...
+    // meeting keeps its old data instead of being left empty. This runs on every
+    // successful callback (not just force) — summaries upsert by speakerSegmentId, but
+    // TopicLabels and utterance discussion statuses need explicit cleanup because old
+    // rows for segments/topics no longer in the new response would otherwise persist.
+    const operations: any[] = [
+        prisma.topicLabel.deleteMany({
+            where: { speakerSegmentId: { in: availableSpeakerSegmentIds } }
+        }),
+        prisma.utterance.updateMany({
+            where: { speakerSegmentId: { in: availableSpeakerSegmentIds } },
```
kouloumos: This is the half that the later commit restores. See my comment on the block comment above.

If you take option 2 there, this operation is the one that moves.

src/lib/tasks/__tests__/summarizeForceCleanup.test.ts
```diff
@@ -148,7 +148,25 @@ describe('handleSummarizeResult — always cleans up stale data on success', ()
     });
   });
 
-  it('cleans up before saving new data', async () => {
+  it('passes the cleanup ops into the same $transaction array (atomicity)', async () => {
```
kouloumos: This test is the reason I trust the change. Asserting the sentinel values are **elements of the array** is what makes it a real guard — the sequential version would satisfy a "was called" assertion just as well.

One limit to record, not a request: `jest.mock('../../db/prisma')` replaces the whole client, so no test here runs SQL. Nothing in this file would notice if the transaction did not really roll back. That is why I ran the rollback against a real database.

PR #476, opened 2026-06-16
fix(api): scope task-status callback lookups to the path tenant

### LINKED ISSUE #445: Task-status callback endpoint: add tenant scoping and optional callback auth
## Summary

The external task backend posts results back to
`POST/PUT/DELETE /api/cities/{cityId}/meetings/{meetingId}/taskStatuses/{taskStatusId}`.
The handler resolves the task by its **UUID alone** — it ignores the `cityId`/`meetingId`
in the path and performs no authentication — then runs result processing
(`handleTaskUpdate` → `processResult`) that writes transcripts/summaries and other
derived data into the DB.

**Severity: moderate** (not critical). Task IDs are unguessable UUIDs that only the
app and the task backend hold, so this is effectively a capability-secret today —
which is "kind of okay". This issue is about closing it as defense-in-depth and
fixing the tenant-isolation gap, not patching an actively-exploited hole.

## Two distinct problems

1. **No tenant scoping (IDOR-shaped).** `getTaskStatus(taskStatusId)` looks the task up
   by `id` only and the route never checks it belongs to the path's
   `(cityId, councilMeetingId)`. Everything else in the codebase isolates tenants by the
   composite `(cityId, …)` key; this endpoint doesn't.
2. **No authentication on state-changing methods.** POST/PUT/DELETE have no bearer,
   signature, or session check. The frontend already calls the backend with
   `Authorization: Bearer ${TASK_API_KEY}` (`src/lib/tasks/tasks.ts`), but that shared
   secret is **not** required on the callback.

## Evidence

- `src/app/api/cities/[cityId]/meetings/[meetingId]/taskStatuses/[taskStatusId]/route.ts:56`
  — `handleUpdateRequest` calls `getTaskStatus(taskStatusId)` with no scope, no auth.
- `src/lib/db/tasks.ts:59` — `getTaskStatus` queries `findUnique({ where: { id } })`.
- `src/lib/tasks/tasks.ts:108` — outbound call uses `Bearer ${TASK_API_KEY}`; the
  callback URL is handed to the backend but the route enforces nothing.

## Suggested fix

1. **Tenant scope (unconditionally safe).** Extend `getTaskStatus` to take an optional
   `{ cityId, councilMeetingId }` and return `null` on mismatch; pass the path params
   from every handler. Returning `null` (→ existing 404) avoids leaking cross-tenant
   existence.
2. **Optional callback secret (env-gated rollout).** Add an optional
   `TASK_CALLBACK_SECRET`; when set, require `Authorization: Bearer <secret>` on
   POST/PUT/DELETE (constant-time compare), warn-and-allow when unset so it can ship
   before the backend is reconfigured. Closing the hole then needs two ops steps:
   set the env var **and** configure the task backend to send the header.

A full step-by-step implementation plan (with verification gates and tests) was drafted
by an audit pass and is available at `plans/001-secure-task-callback-endpoint.md` on the
`claude/elegant-dirac-35wvtv` branch.

## Out of scope

- `handleTaskUpdate` marks a task `succeeded` *before* `processResult` runs
  (`src/lib/tasks/tasks.ts:198-205`) — a separate correctness issue, tracked elsewhere.

_Found during a read-only codebase audit._

angelospk:
Closes #445

## Why

The task-status callback route resolved tasks by UUID alone. It ignored the `cityId`/`meetingId` in the path, so a task UUID from one tenant could be read, updated or deleted through another tenant's path.

## Fix

`getTaskStatusDirect` now takes a required `{cityId, councilMeetingId}` scope and looks the task up with `findFirst`. Every verb passes the path params through. A task belonging to another tenant resolves to null and returns 404, so there is no existence leak. DELETE goes through a scoped `deleteMany` and returns 404 when it removes 0 rows.

## Notes

- Callback authentication is no longer part of this PR. `main` now requires a per-task HMAC token (`src/lib/tasks/callbackToken.ts`, a1d265d7), which is strictly stronger than the optional `TASK_CALLBACK_SECRET` bearer this branch used to carry. That layer was dropped during the rebase.
- GET stays unauthenticated (read-only) but is now tenant-scoped, and keeps the field redaction from `main`.
- `deleteTaskStatus` gained an optional scope and returns a row count instead of void. Admin callers that ignore the return value are unaffected.

## Testing

Route tests cover tenant-mismatch 404 on GET/POST/DELETE, the scoped delete removing 0 rows, and the existing HMAC token paths.

### DIFF
--- a/src/lib/db/tasks.ts
+++ b/src/lib/db/tasks.ts
@@ -56,24 +56,53 @@ export async function getTasksForMeeting(cityId: string, meetingId: string): Pro
     }
 }
 
-export async function getTaskStatus(taskStatusId: string): Promise<TaskStatus | null> {
+/**
+ * Fetch a task status by id, optionally scoped to a (cityId, councilMeetingId) tenant.
+ *
+ * When `scope` is provided the lookup also matches on cityId/councilMeetingId, so a task
+ * that exists but belongs to a different tenant resolves to `null` (callers map that to a
+ * 404 — no cross-tenant existence leak). When omitted, behaves as a plain id lookup.
+ */
+export async function getTaskStatus(
+    taskStatusId: string,
+    scope?: { cityId: string; councilMeetingId: string }
+): Promise<TaskStatus | null> {
     try {
-        const taskStatus = await prisma.taskStatus.findUnique({
+        if (scope) {
+            return await prisma.taskStatus.findFirst({
+                where: {
+                    id: taskStatusId,
+                    cityId: scope.cityId,
+                    councilMeetingId: scope.councilMeetingId,
+                },
+            });
+        }
+
+        return await prisma.taskStatus.findUnique({
             where: { id: taskStatusId },
         });
-
-        return taskStatus;
     } catch (error) {
         console.error('Error fetching task status:', error);
         throw new Error('Failed to fetch task status');
     }
 }
 
-export async function deleteTaskStatus(taskStatusId: string): Promise<void> {
+/**
+ * Delete a task status by id, optionally scoped to a (cityId, councilMeetingId) tenant.
+ * Returns the number of rows deleted so callers can treat 0 as a tenant mismatch / 404.
+ */
+export async function deleteTaskStatus(
+    taskStatusId: string,
+    scope?: { cityId: string; councilMeetingId: string }
+): Promise<number> {
     try {
-        await prisma.taskStatus.delete({
-            where: { id: taskStatusId },
+        const result = await prisma.taskStatus.deleteMany({
+            where: {
+                id: taskStatusId,
+                ...(scope ? { cityId: scope.cityId, councilMeetingId: scope.councilMeetingId } : {}),
+            },
         });
+        return result.count;
     } catch (error) {
         console.error('Error deleting task status:', error);
         throw new Error('Failed to delete task status');
--- a/src/lib/db/tasks.ts
+++ b/src/lib/db/tasks.ts
@@ -61,7 +61,14 @@ export async function getTasksForMeeting(cityId: string, meetingId: string): Pro
 // possession of the unguessable id, so the read cannot carry a user gate and
 // must stay off the Server Action surface instead.
 
-export async function deleteTaskStatus(taskStatusId: string): Promise<void> {
+/**
+ * Delete a task status by id, optionally scoped to a (cityId, councilMeetingId) tenant.
+ * Returns the number of rows deleted so callers can treat 0 as a tenant mismatch / 404.
+ */
+export async function deleteTaskStatus(
+    taskStatusId: string,
+    scope?: { cityId: string; councilMeetingId: string }
+): Promise<number> {
     // Called directly from client components (admin tasks + voiceprint actions),
     // so it must gate itself. Scope to the task's own city; a missing task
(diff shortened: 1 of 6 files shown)


angelospk:
Addressed both review notes in `c2876dd1`:

- `TASK_CALLBACK_SECRET` is now validated with `z.string().trim().min(1)`. A whitespace-only value used to pass `min(1)`, then trim to `""` at request time and silently fall through to the unauthenticated path — now it fails fast at startup instead.
- Split the DELETE 404 test into two: a genuine tenant mismatch (scoped `getTaskStatus` resolves null → early 404, no delete attempted) and the scoped delete removing 0 rows (already gone / race). The old single test was named "tenant mismatch" but only covered the deleted-nothing path.

Tests pass (11/11).

src/lib/db/tasks.ts
```diff
@@ -71,9 +78,13 @@ export async function deleteTaskStatus(taskStatusId: string): Promise<void> {
     });
     await withUserAuthorizedToEdit(task ? { cityId: task.cityId } : {});
     try {
-        await prisma.taskStatus.delete({
-            where: { id: taskStatusId },
+        const result = await prisma.taskStatus.deleteMany({
```
kouloumos: This changes the function's failure contract, and two of its three callers still assume the old one.

`prisma.taskStatus.delete()` throws `P2025` when the row is absent, which your `catch` re-wrapped as `Failed to delete task status`. `deleteMany()` does not throw — it returns `{ count: 0 }`. The return type also moved from `Promise<void>` to `Promise<number>`.

The route was updated to read the count. The other two were not:

- **`src/components/admin/people/voiceprint-actions.tsx:122`** wraps the call in `try/catch` and shows an error toast on throw. It now shows **"Task deleted — successfully deleted"** for a row that did not exist.
- **`src/components/admin/tasks/ActiveTasks.tsx:73`** has no `try/catch` at all, so it silently drops the row from local state. Arguably an improvement, but it changed without anyone deciding it should.

Neither is serious. Both are one line now that the count comes back, for example in `voiceprint-actions.tsx`:

```
const deleted = await deleteTaskStatus(taskId);
if (deleted === 0) { /* toast: nothing to delete */ return; }
```

Worth doing while the contract is fresh — a silent success on a delete that removed nothing is the kind of thing that gets reported as a ghost bug months later.

src/lib/db/tasksInternal.ts
```diff
@@ -1,15 +1,31 @@
...
+ * Read a task status by id, scoped to a (cityId, councilMeetingId) tenant.
+ *
+ * The lookup also matches on cityId/councilMeetingId, so a task that exists but
+ * belongs to another tenant resolves to `null` — the callback route maps that to a
+ * 404, with no cross-tenant existence leak. The scope is required: the route is the
+ * only caller and always knows its path tenant, so there is no unscoped escape hatch.
+ */
+export async function getTaskStatusDirect(
+    taskStatusId: string,
+    scope: { cityId: string; councilMeetingId: string }
```
kouloumos: This is the part of the PR I would point at as well judged.

Making the scope **required** rather than optional means the unscoped call can no longer be written at all, so the old behaviour is not reachable by a future caller who forgets. That is stronger than a test that catches the mistake afterwards, and the comment explains why it is safe — the route is the only caller and always knows its path tenant.

No change requested.

PR #500, opened 2026-06-23
fix(auth): point magic-link emails at the sign-in host

christosporios:
## Problem
On the single multi-domain deployment (`opencouncil.gr` + `opencouncil.fr`), Auth.js builds the magic-link URL from the build-time `NEXTAUTH_URL`, so **every sign-in email points at the Greek domain**. A user signing in on `opencouncil.fr` can't actually complete login there — the callback lands on `opencouncil.gr` and sets the session cookie on the wrong domain.

(Sessions remain per-domain by design — different registrable domains can't share a cookie. This PR just makes each domain's login *work*; it is **not** a cross-domain SSO bridge.)

## Fix
`sendVerificationRequest` receives the `request` (Auth.js v5 passes it). New helper `signInUrlForRequest(url, request)` rewrites the magic-link **origin** to the host the request arrived on (via `x-forwarded-host`/`host` + `x-forwarded-proto`), preserving path + query (the verification token and `callbackUrl`).

- **Edge-safe:** reads headers off the passed `request` — no `next/headers`, so `auth.config.ts` (reachable from `proxy.ts`) stays out of the middleware bundle's way.
- **Graceful:** falls back to the original URL when no host header is present, so local dev (`localhost`/http) and the existing `.gr` flow are byte-identical. Worst case = today's behavior; it can't make login worse.

## Tests
`src/lib/auth/__tests__/signInUrl.test.ts` — rewrites `.gr → .fr`, preserves token/callbackUrl, `x-forwarded-host` precedence, dev no-op, same-host no-op, missing-host fallback. `tsc` clean; full suite **854 pass**.

## Out of scope / follow-up
`src/lib/auth/invite.ts` (admin invite links) still builds URLs from `NEXTAUTH_URL`. It has no request context (the invitee isn't on a domain yet) and the "right" domain depends on the invited realm — a separate decision.

---

> [!NOTE]
> **Medium Risk**
> Touches authentication email URLs and trusts forwarded Host headers, but rewrites are gated by `isKnownRealmHost` with a safe fallback to the original link.
> 
> **Overview**
> Fixes French-domain sign-in by rewriting magic-link **origins** to match the host the user actually used (`opencouncil.fr` vs `opencouncil.gr`), while keeping Auth.js path/query (token, `callbackUrl`) unchanged.
> 
> `sendVerificationRequest` now builds the email HTML and plain-text link via **`signInUrlForRequest`**, using `x-forwarded-host` / `host` and optional `x-forwarded-proto` from the Auth.js **`request`**. Rewrites only run when **`isKnownRealmHost`** accepts the host (realm apex/subdomains, including preview); unknown or spoofed hosts (e.g. `evil.com`, `localhost`) leave the original `NEXTAUTH_URL` link untouched.
> 
> Shared realm helpers extract **`hostMatchesDomain`** and add unit tests for cross-domain rewrite, proxy headers, same-host no-op, and rejection of untrusted hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 258b56459f90ee83d3a94c51542b234e8ba3ec08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR fixes magic-link sign-in emails on the multi-domain deployment (`opencouncil.gr` + `opencouncil.fr`) by rewriting the Auth.js-generated URL origin to match the host the user actually signed in on, so the callback cookie lands on the correct domain.

- **`signInUrl.ts`**: New `signInUrlForRequest` helper reads `x-forwarded-host`/`host` and `x-forwarded-proto` from the passed `Request`, validates the host against an explicit allowlist (`isKnownRealmHost`), and rewrites only the URL origin — path and query (token, `callbackUrl`) are preserved. Falls back to the original URL when headers are absent or the host is unrecognised, keeping local dev and the `.gr` flow byte-identical.
- **`realm.ts`**: Extracts `hostMatchesDomain` as a shared predicate and adds `isKnownRealmHost` — a strict membership check (no default-to-greece fallback) for use before trusting request headers in security-sensitive contexts.
- **`auth.config.ts`**: Applies the rewritten URL to both the HTML and plain-text email bodies in `sendVerificationRequest`.

Safe to merge — the change is well-scoped, the fallback preserves today's behaviour on failure, and the host-rewrite path is gated behind an explicit domain allowlist.

The host-allowlist concern raised in a prior review has been directly addressed: isKnownRealmHost rejects any host that isn't an apex or subdomain of a known realm domain, and there is an explicit test confirming evil.com is rejected. The helper is edge-safe (no next/headers), falls back to the original URL whenever it can't determine a known host, and is covered by nine focused unit tests spanning every meaningful branch including the security path.

| Filename | Overview |
|----------|----------|
| src/lib/auth/signInUrl.ts | New helper that rewrites a magic-link URL's origin to the request host, with an explicit allowlist guard via isKnownRealmHost |
| src/lib/realm.ts | Extracts shared hostMatchesDomain predicate and adds isKnownRealmHost for strict membership checks before trusting request host headers |
| src/auth.config.ts | Integrates signInUrlForRequest into sendVerificationRequest; both HTML and plain-text email bodies now use the rewritten URL |
| src/lib/auth/__tests__/signInUrl.test.ts | Comprehensive test suite covering multi-domain rewrite, proxy header precedence, protocol rewriting, dev no-op, same-host no-op, missing-host fallback, evil-host rejection, preview subdomain allowance, and comma-separated header handling |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(auth): point magic-link emails at th..."](https://github.com/schemalabz/opencouncil/commit/258b56459f90ee83d3a94c51542b234e8ba3ec08) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39263275)</sub>

### DIFF
--- a/src/auth.config.ts
+++ b/src/auth.config.ts
@@ -4,6 +4,7 @@ import { AuthEmail } from "./lib/email/templates/AuthEmail"
 import { renderReactEmailToHtml } from "./lib/email/render"
 import { env } from "./env.mjs"
 import { isTestUserEmail } from "./lib/dev/test-users"
+import { signInUrlForRequest } from "./lib/auth/signInUrl"
 
 // In development, use port-specific session cookie names to allow multiple
 // instances on different ports to have independent sessions. Without this,
@@ -25,8 +26,12 @@ export default {
         from: 'OpenCouncil <auth@opencouncil.gr>',
         apiKey: env.RESEND_API_KEY,
         sendVerificationRequest: async (params) => {
-            const { identifier: to, provider, url, theme } = params
-            const html = await renderReactEmailToHtml(AuthEmail({ url }))
+            const { identifier: to, provider, url, request } = params
+            // Point the magic link at the domain the user is signing in from
+            // (opencouncil.gr vs opencouncil.fr) instead of the single build-time
+            // NEXTAUTH_URL host, so the callback sets a cookie on the right domain.
+            const signInUrl = signInUrlForRequest(url, request)
+            const html = await renderReactEmailToHtml(AuthEmail({ url: signInUrl }))
 
             // Redirect test user emails to DEV_EMAIL_OVERRIDE if set
             // This allows testing different admin roles with a single real inbox
@@ -47,7 +52,7 @@ export default {
                     to: emailTo,
                     subject: `Συνδεθείτε στο OpenCouncil`,
                     html,
-                    text: `Συνδεθείτε στο OpenCouncil: ${url}`,
+                    text: `Συνδεθείτε στο OpenCouncil: ${signInUrl}`,
                 }),
             })
 
--- a/src/lib/auth/__tests__/signInUrl.test.ts
+++ b/src/lib/auth/__tests__/signInUrl.test.ts
@@ -1,0 +1,81 @@
+import { signInUrlForRequest } from '../signInUrl';
+
+const reqWith = (headers: Record<string, string>) =>
+    new Request('https://internal.local/api/auth/signin', { headers });
+
+describe('signInUrlForRequest', () => {
+    const greekUrl =
+        'https://opencouncil.gr/api/auth/callback/resend?callbackUrl=%2Fprofile&token=abc123&email=user%40example.com';
+
+    it('repoints the magic link to the French host when signing in on .fr', () => {
+        const result = signInUrlForRequest(greekUrl, reqWith({ host: 'opencouncil.fr' }));
+        const u = new URL(result);
+        expect(u.host).toBe('opencouncil.fr');
+        expect(u.protocol).toBe('https:');
+    });
+
+    it('preserves the path and query (token + callbackUrl) when rewriting', () => {
+        const result = signInUrlForRequest(greekUrl, reqWith({ host: 'opencouncil.fr' }));
+        const u = new URL(result);
+        expect(u.pathname).toBe('/api/auth/callback/resend');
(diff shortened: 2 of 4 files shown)


christosporios:
Addressed both review findings ([force-pushed](https://github.com/schemalabz/opencouncil/compare/af59a39dad7133eb0edff41ca4bd5d1ec3aefcf9..258b56459f90ee83d3a94c51542b234e8ba3ec08)):

- 🔒 **Greptile (security) — host-header injection.** Good catch. The rewrite now only trusts a host that matches one of our own domains: added `isKnownRealmHost(host)` to `@/lib/realm` (strict apex/subdomain membership against `REALMS`, unlike `realmForHost` which defaults to greece). An attacker-supplied `Host: evil.com` no longer produces a magic link to their domain — unknown hosts (incl. `localhost` dev) fall back to the original `NEXTAUTH_URL` link. Preview subdomains (`pr-N.preview.opencouncil.fr`) still match.
- 🟡 **Bugbot (medium) — multi-value forwarded headers.** `x-forwarded-host`/`-proto` are now parsed as comma-separated lists, taking the first (client-facing) value, so chained-proxy deployments don't get a malformed host/scheme.

Added tests for both (attacker host rejected, preview subdomain allowed, comma-separated header). `tsc` clean; full suite **857 pass**.

PR #522, opened 2026-07-03
fix(tasks): exclude scheduled and live streams from livestream matching

VasiaKoum:
## Problem

`pollLivestreamsForRecentMeetings` runs on a ±12h window, so a meeting scheduled up to 12h in the future can already have an **upcoming premiere** or an **in-progress live broadcast** on the administrative body's YouTube channel. `listRecentChannelVideos` returned those alongside finished uploads, so the LLM matcher could match one and trigger transcription of a stream that hasn't finished — a video with no complete recording.

Observed in production: the poll tried to transcribe [`Wlg3BQctCMk`](https://www.youtube.com/watch?v=Wlg3BQctCMk) — a scheduled special session ("Ειδική συνεδρίαση… Δ. Αργιθέας | Παρασκευή 3/7/2026") set to start at 16:00 UTC that day. YouTube reported it as `isUpcoming: true`, `isLiveNow: false`, no end timestamp.

## Fix

`search.list` already returns `snippet.liveBroadcastContent` per video: `"none"` for a finished/regular upload, `"live"` while broadcasting, `"upcoming"` for a scheduled stream that hasn't started. `listRecentChannelVideos` now filters out `"upcoming"` and `"live"`, keeping only finished videos.

The filter lives at the source (the YouTube client) rather than in the poller, so it's a single choke point no consumer can bypass. Unfinished streams reappear as ordinary uploads once they end, so the affected meeting is simply retried on a later run.

## Tests

New `src/lib/__tests__/youtube.test.ts` covers:
- excluding `"upcoming"` and `"live"` streams, keeping only finished videos
- treating a missing `liveBroadcastContent` field as finished
- the cache short-circuit path

All poll-livestreams tests continue to pass; no type errors.

## Note / possible follow-up

The filter trusts `search.list`'s `liveBroadcastContent`, which is an eventually-consistent index. In the rare case it lags and reports `"none"` for a still-upcoming stream, a video could slip through. A bulletproof follow-up would add a cheap `videos.list` confirmation (1 quota unit for up to 50 ids) checking `liveStreamingDetails` — a stream with a start time but no `actualEndTime` is not finished. Not included here to keep the fix minimal; happy to add if preferred.

<h3>Greptile Summary</h3>

This PR fixes a production bug where `pollLivestreamsForRecentMeetings` could match an upcoming or in-progress YouTube livestream and attempt to transcribe a video with no complete recording. The fix adds a filter in `listRecentChannelVideos` to drop items whose `liveBroadcastContent` is `"live"` or `"upcoming"`, keeping only finished uploads.

- `src/lib/youtube.ts`: Extends the `SearchListResponse` interface with `liveBroadcastContent` and inserts a `.filter()` step before `.map()` to exclude unfinished streams; the cache layer is left intact so filtered-out streams reappear as normal uploads once they end.
- `src/lib/__tests__/youtube.test.ts`: New unit tests cover the three core scenarios — filtering live/upcoming, treating a missing field as finished, and the cache short-circuit path.

The change is minimal and targeted: a single filter step is inserted before the mapping stage, with correct handling of undefined (passes through), and the Valkey cache is unaffected.

The fix surgically addresses the root cause by filtering at the data-source level before anything is cached or forwarded to the LLM matcher. The filter correctly treats undefined as finished, matching the test and the YouTube API contract. The 5-minute cache TTL means previously cached live/upcoming entries expire quickly, and live streams that complete reappear naturally on the next cache miss.

| Filename | Overview |
|----------|----------|
| src/lib/youtube.ts | Adds liveBroadcastContent to SearchListResponse and filters out "upcoming"/"live" items in listRecentChannelVideos; logic and types are correct, filter handles undefined gracefully. |
| src/lib/__tests__/youtube.test.ts | New test file covering filtering of upcoming/live streams, missing-field fallback, and cache short-circuit; setup/teardown is clean and assertions are complete. |

<sub>Reviews (1): Last reviewed commit: ["fix(tasks): exclude scheduled and live s..."](https://github.com/schemalabz/opencouncil/commit/0e1d093616ad33fd022fb7ec8b8de4461fbe6079) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41626131)</sub>

### DIFF
--- a/src/lib/__tests__/youtube.test.ts
+++ b/src/lib/__tests__/youtube.test.ts
@@ -1,0 +1,68 @@
+/** @jest-environment node */
+
+const mockEnv: { YOUTUBE_API_KEY?: string } = { YOUTUBE_API_KEY: 'test-key' };
+jest.mock('@/env.mjs', () => ({ env: mockEnv }));
+
+const mockCacheGetJSON = jest.fn();
+const mockCacheSetJSON = jest.fn();
+jest.mock('@/lib/cache/valkey', () => ({
+    cacheGetJSON: (...args: unknown[]) => mockCacheGetJSON(...args),
+    cacheSetJSON: (...args: unknown[]) => mockCacheSetJSON(...args),
+}));
+
+import { listRecentChannelVideos } from '../youtube';
+
+function searchItem(videoId: string, liveBroadcastContent: string, title = videoId) {
+    return {
+        id: { videoId },
+        snippet: { title, publishedAt: '2026-06-10T20:00:00Z', liveBroadcastContent },
+    };
+}
+
+function mockFetchOnce(items: unknown[]) {
+    (global.fetch as jest.Mock).mockResolvedValueOnce({
+        ok: true,
+        json: async () => ({ items }),
+    });
+}
+
+beforeEach(() => {
+    jest.clearAllMocks();
+    mockEnv.YOUTUBE_API_KEY = 'test-key';
+    mockCacheGetJSON.mockResolvedValue(null);
+    mockCacheSetJSON.mockResolvedValue(true);
+    global.fetch = jest.fn();
+});
+
+describe('listRecentChannelVideos', () => {
+    it('excludes scheduled ("upcoming") and live streams, keeping only finished videos', async () => {
+        mockFetchOnce([
+            searchItem('finished', 'none'),
+            searchItem('scheduled', 'upcoming'),
+            searchItem('broadcasting', 'live'),
+        ]);
+
+        const videos = await listRecentChannelVideos('UCchannel');
+
+        expect(videos.map(v => v.videoId)).toEqual(['finished']);
+    });
+
+    it('keeps videos with no liveBroadcastContent field (treated as finished)', async () => {
+        mockFetchOnce([
+            { id: { videoId: 'plain' }, snippet: { title: 'plain', publishedAt: '2026-06-10T20:00:00Z' } },
+        ]);
+
+        const videos = await listRecentChannelVideos('UCchannel');
+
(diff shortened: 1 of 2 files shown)


VasiaKoum:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/0e1d093616ad33fd022fb7ec8b8de4461fbe6079..f46b45442e28256e0a41b5a66d6805dc14138bd0) to rework the fix after reproducing the failure locally against the Athens meeting.

**What changed:** the previous commit only filtered `search.list` results by `liveBroadcastContent`. Local reproduction revealed a deeper root cause — `search.list` **intermittently 403s** (`"cannot act on behalf of the specified Google account"`) on our API key, measured at **~40% of calls (9/15 succeeded)**, while `channels.list` / `playlistItems.list` / `videos.list` were **15/15**. That flakiness — plus the search index's lag for just-finished livestreams — is why the Athens meeting (`jul3_2026`) never got its video `-jgt8v_lOeg` matched: the listing either errored or lagged during the ±12h window.

**New approach:** `listRecentChannelVideos` now reads the channel's **uploads playlist** (`playlistItems.list`) and fetches details + authoritative live status via **`videos.list`** — dropping `search.list` from the hot path entirely. This:

- eliminates the intermittent 403s (uses only the reliable endpoints),
- reflects finished livestreams immediately (no search-index lag),
- still excludes scheduled/in-progress streams (via `liveStreamingDetails.actualEndTime`),
- cuts quota from 100 → ~2 units per channel per run.

Verified end-to-end locally: the uploads-playlist path returns `-jgt8v_lOeg` every time, and the real LLM matcher then matches it to the meeting at 0.98 confidence.

Tests rewritten for the two-call flow (17 passing); `docs/task-architecture.md` updated.

PR #523, opened 2026-07-03
fix(tasks): list channel videos via uploads playlist, not search.list

VasiaKoum:
Follow-up to #522.

## Why

#522 filtered `search.list` results by `liveBroadcastContent`, but reproducing the Athens miss locally surfaced the real root cause: **`search.list` itself is the problem**, not just what it returns.

`search.list` queries YouTube's search index, which:
1. **Lags minutes-to-hours** and is unreliable for just-finished livestreams, and
2. **Intermittently returns `403 "cannot act on behalf of the specified Google account"`** — measured at **~40% of calls on our key (9/15 succeeded)**, while `channels.list` / `playlistItems.list` / `videos.list` were **15/15**.

Either failure leaves a meeting unmatched within its ±12h polling window — the correct finished video is absent from the lagging index, or the whole listing errors out. This is what happened to Athens `jul3_2026`: video [`-jgt8v_lOeg`](https://www.youtube.com/watch?v=-jgt8v_lOeg) was never even presented to the matcher. It also explains the earlier preview run where meetings came back as `"error"` rather than `"no_match"`.

## What

`listRecentChannelVideos` now reads the channel's **uploads playlist** instead of searching:

- `playlistItems.list` → the channel's recent uploads directly (immediate, reliable)
- `videos.list` → titles + authoritative `liveStreamingDetails` for those ids

Scheduled (`upcoming`) and in-progress (`live`) streams are still excluded — a stream with no `actualEndTime` has no complete recording to transcribe. `search.list` is dropped from the hot path entirely (it remains only in `resolveChannelId`'s `/c/` vanity fallback).

**Result:** no dependence on the flaky/lagging search index, and **~2 quota units per channel per run instead of 100**.

## Verification

Reproduced end-to-end locally against the real Athens channel + live Claude matcher:
- current `search.list` path: 403'd on all attempts
- uploads-playlist path: returned `-jgt8v_lOeg` every time, newest-first
- real LLM matcher then matched it to the meeting at **0.98 confidence** ✅

Tests rewritten for the two-call flow (17 passing); `docs/task-architecture.md` updated.

<h3>Greptile Summary</h3>

Replaces the flaky, high-quota `search.list` hot path in `listRecentChannelVideos` with `playlistItems.list` + `videos.list`, eliminating both the search-index lag and the intermittent 403 that caused the Athens meeting to go unmatched.

- **`youtube.ts`**: New `resolveUploadsPlaylistId` derives the uploads playlist ID via a fast UC→UU string swap (no API call) or a single `channels.list` fallback for non-standard IDs; `isUnfinishedStream` adds a defensive `liveStreamingDetails` check so a broadcast with a lagging `liveBroadcastContent: \"none\"` is still excluded if its `actualEndTime` is absent; `MAX_RECENT_VIDEOS` gains a comment documenting the ≤50 constraint for both downstream calls.
- **`youtube.test.ts`**: Tests rewritten around an `installFetch` router that dispatches by endpoint; five scenarios cover playlist ordering, live/upcoming/lagging exclusion, UC→UU fast path, non-UC channels.list fallback, and cache hit.
- **`docs/task-architecture.md`**: Updated to reflect the new two-call lookup and its reliability/quota advantages.

Safe to merge — the core logic is correct, the new API path is well-tested, and the change eliminates the root cause of the production miss.

The two-call flow (playlistItems → videos) is straightforward; the UC→UU transformation is a well-known YouTube convention; isUnfinishedStream correctly handles both the liveBroadcastContent field and the lagging-state edge case; all five new test scenarios pass; quota footprint drops from ~100 units to 2 units per channel per run.

| Filename | Overview |
|----------|----------|
| src/lib/youtube.ts | Core change: replaces search.list with playlistItems.list + videos.list; adds resolveUploadsPlaylistId and isUnfinishedStream helpers with correct UC→UU fast-path and defensive lagging-liveBroadcastContent handling. |
| src/lib/__tests__/youtube.test.ts | Tests rewritten for the two-call flow; installFetch helper routes by endpoint; covers ordering, live/upcoming exclusion, lagging liveBroadcastContent, UC→UU fast path, non-UC channels.list fallback, and cache hit. |
| docs/task-architecture.md | Documentation updated to reflect the switch to playlistItems.list + videos.list and the rationale for dropping search.list from the hot path. |

<sub>Reviews (2): Last reviewed commit: ["fix(tasks): list channel videos via uplo..."](https://github.com/schemalabz/opencouncil/commit/a66ac8dc5130fe39030876bb24e699b177dac742) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41636070)</sub>

### DIFF
--- a/docs/task-architecture.md
+++ b/docs/task-architecture.md
@@ -171,8 +171,10 @@ automatically.
 1. Finds candidate meetings: a `processAgenda` task has `succeeded`, `dateTime` is within
    ±12h of now, `youtubeUrl` is still null, the administrative body has a `youtubeChannelUrl`,
    and no `transcribe` task is already in flight.
-2. Lists recent videos for each distinct channel via the YouTube Data API v3 (resolved channel
-   id + listing are cached in Valkey, so meetings sharing a channel cost one API call per run).
+2. Lists recent videos for each distinct channel from the channel's uploads playlist
+   (`playlistItems.list` + `videos.list`) — reliable and immediate, unlike `search.list`,
+   whose index lags for just-finished livestreams and intermittently 403s. Resolved channel
+   id + listing are cached in Valkey, so meetings sharing a channel cost one lookup per run.
 3. Asks Claude to match each meeting to a single video. The matcher **refuses to match a video
    that appears to cover more than one council meeting** (e.g. a combined Λογοδοσία + Τακτική
    Συνεδρίαση stream) — those are handled manually.
--- a/src/lib/__tests__/youtube.test.ts
+++ b/src/lib/__tests__/youtube.test.ts
@@ -12,18 +12,61 @@ jest.mock('@/lib/cache/valkey', () => ({
 
 import { listRecentChannelVideos } from '../youtube';
 
-function searchItem(videoId: string, liveBroadcastContent: string, title = videoId) {
-    return {
-        id: { videoId },
-        snippet: { title, publishedAt: '2026-06-10T20:00:00Z', liveBroadcastContent },
-    };
+const CHANNEL_ID = 'UCX5CxaBSCrAJxQawnE1sKHw';
+
+// Route each mocked YouTube API call by endpoint. playlistItems returns the
+// ordered ids; videos returns per-id details keyed off `videosById`. When
+// `channelsUploads` is set, /channels resolves to that uploads-playlist id
+// (the non-UC fallback path).
+function installFetch(orderedIds: string[], videosById: Record<string, unknown>, channelsUploads?: string) {
+    global.fetch = jest.fn(async (input: unknown) => {
+        const url = String(input);
+        if (url.includes('/channels')) {
+            return {
+                ok: true,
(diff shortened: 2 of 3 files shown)


VasiaKoum:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/0bbacca6ef0db343830f4f4acc3d7c0fb991984c..a66ac8dc5130fe39030876bb24e699b177dac742) to address both Greptile review comments:

- **`videos.list` ≤50 coupling:** added a note on `MAX_RECENT_VIDEOS` documenting that it must stay ≤ 50 — both `playlistItems.list` (`maxResults`) and `videos.list` (comma-separated ids) cap at 50 per request, and we issue one call to each. Keeps the coupling visible if anyone bumps the constant.
- **Untested non-UC fallback:** added a test that resolves a non-`UC…` channel id through the `channels.list` path (`resolveUploadsPlaylistId`'s fallback), asserting it calls `/channels` and threads the returned uploads playlist id into `playlistItems.list`.

6 tests in the suite now pass.

PR #408, opened 2026-06-02
feat: redesign landing page with mock data

VasiaKoum:
<h3>Greptile Summary</h3>

This PR delivers a full landing page redesign (`LandingV2`) backed by new server-side data helpers, a set of React hooks, and API routes. The data layer is well-structured: realm scoping, caching, and error handling are consistently applied across nearly all new routes and DB helpers. The previous review cycle already resolved the cancelled-fetch stuck-loading issue (now using a correct derived-loading pattern in `useKeyedList`), the unreleased-meeting exposure, and the subject-count visibility mismatches.

- Adds `getMapSubjectsCached`, `getGeneralSubjectsCached`, `getSubjectCountsByCityCached`, and `getUpcomingMeetingsCached` — all realm-scoped with proper TTLs and cache-tag invalidation wired to meeting release toggles.
- Introduces `/api/geocode`, `/api/cities/at`, `/api/map/general-subjects`, and `/api/meetings/upcoming`; the `/api/map/subjects` route is refactored to a thin wrapper over the shared helper.
- `getPopularSearchQueries` in `searchQueries.ts` queries the shared `SearchQuery` table without a realm filter, and `logSearchQuery` does not store `realm` — so on the French deployment the popular-searches chips will surface Greek-language queries.

<h3>Confidence Score: 4/5</h3>

Safe to merge for the Greek realm; the popular-searches chips on the French deployment will show Greek queries until the SearchQuery table is realm-tagged.

The core landing data pipeline is realm-scoped and correct. The popular-searches feature lacks realm scoping, causing cross-language contamination on the French deployment. All other previously flagged issues appear resolved.

src/lib/db/searchQueries.ts and src/app/api/landing/popular-searches/route.ts — the popular-searches feature needs realm scoping before the French deployment goes live.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| src/lib/db/searchQueries.ts | New file implementing popular search queries; missing realm scoping means cross-realm query contamination |
| src/lib/db/subject.ts | Significantly refactored; adds getSubjectCountsByCityCached, getMapSubjectsCached, getGeneralSubjectsCached — all properly realm-scoped with correct visibility filters and caching |
| src/lib/db/meetings.ts | Adds getUpcomingMeetings(realm) and a cached variant; properly scoped to realm, listed cities, released meetings, and future date |
[...]

PR #533, opened 2026-07-10
feat: redesign landing page (iteration 2)

### LINKED ISSUE #554: Redesign landing page - iteration 2
## Iteration 2 — review comments, problems and solutions

Tracks which review comments from iteration 2 are addressed in this PR, and which
are deferred.

### P1 · A new visitor can't tell what they're looking at on this map

| # | Comment |
|---|---|
| 3 | The subject stats donut now sits below the design threshold of the rest of the page, but mainly it may confuse. Perhaps if we dropped the icons and the donut colours became a very thin ring around the counter (with no explanation that colour maps to a topic label), it would become simpler — the coloured ring wouldn't be clear to anyone but power users, yet it would be more minimal and, oddly, confuse less. |
| 5 | What I think we must optimise above all is the new visitor's experience: letting them quickly find something interesting discussed in their municipality. To do that we first have to learn where they live and what interests them. |
| 7 | They went looking for the "?" in the burger menu and didn't find it. |
| 8 | Asked "how do you find the most-discussed subjects?", they went to sign in — they hadn't realised subjects have a location, nor understood the order they appear in. |

**Addressed by:** dots instead of donuts (→ 3) · "?" made more visible and added to the burger menu (→ 7) · *(TBD)* new-visitor municipality dialog (→ 5)

### P2 · It's hard to find the location you care about

| # | Comment |
|---|---|
| 5 | (as above — we first need to learn where they live) |
| 10 | *(PostHog)* the first action is panning the map — users struggle to reach the location they care about. |

**Addressed by:** city donuts as the default Greece view · last map centre + zoom saved to local storage · *(TBD)* new-visitor municipality dialog

### P3 · There's no easy way to sign up for notifications

| # | Comment |
|---|---|
| 1 | It would help to show the administrative body holding the session — here I can't tell who is meeting, so I might consider subscribing to notifications. |
| 2 | The push notification is very useful; we could try other variants, not only "spent a long time in a municipality". |
| 9 | When the notification pop-up appeared, they tapped "status" in the burger menu instead. |

**Addressed by:** notifications surfaced in the burger menu for logged-out visitors (→ 9) · *(TBD)* turn the pop-up into a button above the preview subjects (→ 9)

### Standalone fixes

| # | Comment | Status |
|---|---|---|
| 4 | The date range picker has a different size and border radius from everything else. Consistency across the text input, filter button, date range picker and topic pills would help. | done |
| 6 | On the landing page, with a subject selected, they clicked outside the subject to dismiss it — which did nothing. | done — clicking the map background now dismisses the selected subject and the OpenCouncil preview |

### Solution status

| Solution | Addresses |
|---|---|
| Save in local storage the last lat, long and level size from the map. This will be used for location initialisation (for next revisit) | P2 (5, 10) |
| Dots for all subjects instead of donuts; icons return on zoom in. All topic pin/badge colours changed with it | P1 (3) |
| City donuts (default Greece view) for better map navigation | P2 (10) |
| Make the "?" more visible | P1 (7) |
| Add the "?" to the burger menu | P1 (7) |
| Clicking the map dismisses the selected subject | 6 |
| UI consistency: text input, filter button, date range picker, topic pills | 4 |
| Surface notifications in the burger menu for logged-out visitors | P3 (9) |
| Better notification pop up - municipality logo instead of bell | P3 (2) |

### Next round (good to have)
- *(TBD)* Show a new visitor a dialog asking which municipality they care about -> P1, P2 (5, 10)
- *(TBD)* Convert the notification pop-up into a button above the preview subjects -> P3 (2, 9)
- Municipalities with more than 10 petitions should be coloured somehow in the Δήμοι view.
- Images or the video of the discussion, per subject.
- Comment 8, second half — make the ordering of subjects (the ranking) understandable; nothing currently explains why subjects appear in the order they do.

VasiaKoum:
Iteration 2 of the landing redesign — follows #408 (now merged to `main`). Focus: **UI/UX polish** and **perf/correctness**.

### In this PR so far
- **Donut clusters split by municipality** above the regional-overview zoom — *different δήμος → different donut*. Previously a purely geographic grid cell could merge subjects from several δήμοι into one donut. Now the cell key includes `cityId` once zoomed in past a tunable `DONUT_CROSS_CITY_MAX_ZOOM` (default `7`); only the zoomed-out overview may still mix municipalities.

More to come (further UI/UX polish + correctness follow-ups). Early WIP — commits will be tidied before merge.

<h3>Greptile Summary</h3>

This PR delivers the second iteration of the landing page redesign, focusing on per-municipality donut markers and mobile UX polish. It replaces the proximity-cell donut clusters with one fixed-size donut per cooperating δήμος at zoomed-out levels, adds a greedy screen-space declutter pass (`markerDeclutter.ts`) to keep neighbouring donuts readable, and persists the map camera to localStorage so return visitors open where they left off.

- **Per-municipality overview**: `aggregateMunicipalityCounts` aggregates located + non-located subjects by city; `useMunicipalityCountMarkers` renders them as topic-arc donuts that zoom into the δήμος on click; subject pins yield at `MUNICIPALITY_COUNT_MAX_ZOOM = 9`.
- **Mobile preview mode**: tapping a map pin now highlights the pin and scrolls the bottom strip to it without opening the full subject card; a second tap (or selecting from the strip) opens the card.
- **Topic styling unified**: `topicStyle.ts` + `TopicIcon` replace ad-hoc alpha-suffix colour formulas spread across components, with a `soft`/`solid` variant model and a luminance-based contrast calculation.

Safe to merge; all changes are additive UI/logic work with no database, auth, or data-integrity side-effects.

The core new paths (donut rendering, declutter, saved-view persistence, mobile preview) are well-tested and well-commented. No auth, data-mutation, or API contract changes. The two flagged items are a stale type field and a question about reduced mobile map padding, neither of which blocks correctness.

**Files Needing Attention:** src/lib/landing/landingCore.ts — mobile fitBounds padding reduced from 120 px to 40 px; worth confirming it clears the bottom tab bar.

| Filename | Overview |
|----------|----------|
| src/lib/landing/donut.ts | Replaced the old proximity-cluster donut with a new per-municipality fixed-size arc. SVG is now emitted as one unbroken template literal to avoid a Turbopack minifier bug. |
| src/lib/landing/markerDeclutter.ts | New file: greedy screen-space decluttering for municipality donuts via zoom-invariant inverse-distance repulsion bearings. |
| src/lib/landing/savedView.ts | New file: localStorage-backed camera persistence with 30-day TTL, coordinate validation, and zoom clamping. |
| src/lib/landing/landingCore.ts | Added MUNICIPALITY_COUNT_MAX_ZOOM, SUBJECT_DOT_THRESHOLD, InfoSurface type. stylePin extended with intense/dot modes. Mobile fitBounds padding reduced from 120 px to 40 px — see comment. |
| src/lib/landing/landingData.ts | Added MunicipalitySubjectCount type and aggregateMunicipalityCounts. Located and non-located subject sets are mutually exclusive pools — no double-counting risk. |
| src/lib/topicStyle.ts | New file: single source of truth for topic colour derivation (soft/solid variants, contrastText). |
| src/lib/geo.ts | Added isPointInGeometry (ray-casting) and isInSupportedMunicipality for saved-view center validation. |
| src/components/landing/v2/LandingV2.tsx | Major wiring: saved-view restoration, per-municipality count bubbles, mobile preview mode, subjectDots threshold, and moveend camera persistence. |
| src/components/landing/v2/mapMarkers.tsx | Replaced createDonutMarker/createMunicipalityMarker with createMunicipalityCountMarker. Subject pins support dot mode and intense mode. |
| src/components/landing/v2/MobileLayout.tsx | Major redesign: full-screen list-over-map removed, replaced with bottom-sheet strip. Added overviewActive-driven collapse and preview-snapping. |
| src/components/landing/v2/hooks/useFilteredSubjects.ts | Removed NONLOCATED_MAX_ZOOM gate. Added previewId. mapZoom remains in Args type but is no longer read by the hook body. |
| src/components/landing/v2/hooks/useMapMarkers.tsx | useMunicipalityMarkers replaced by useMunicipalityCountMarkers with declutter. useSubjectMarkers now uses dots/previewId/active. |

<sub>Reviews (23): Last reviewed commit: ["feat(landing): show the δήμος logo in th..."](https://github.com/schemalabz/opencouncil/commit/4a98bac6ad3412a15fbdc07bfa2c28216c7207c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43273096)</sub>


VasiaKoum:
Pushed `ea8df50b` — mobile pass refinements:

- **Two-mode subject interaction** (preview vs select): tap a pin/strip card to preview (highlighted pin + map recentre), tap again to expand the selected card; selected/preview pins use the intense fill.
- **Filters sheet** (`MobileFilterBar`): added administrative-body and exact date-range filters, a full-width "Καθαρισμός φίλτρων" action (shown only when a filter is active) plus per-section clears.
- **Expanded subject card**: pins only the title + municipality logo; meta and description scroll together with an ombré fade above the link. Strip cards now show discussion duration.
- **Fixes**: general subjects stay listed when zoomed into a δήμος; loading pill centered; `?`/satellite controls hidden while an office/subject card covers the bottom.
- **Copy**: subject link → "Δες τη συζήτηση"; search placeholder → "Αναζήτηση θέματος" (el/en/fr).

`npx tsc --noEmit` clean; verified in the mobile preview (375×812).

VasiaKoum:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/ea8df50b5d56aca3e9d7dd9b8d8878d5a3fafdda..a4a200a963df59d767f9b8916032198531d6632d) to fold a round of mobile refinements into the `feat(landing): mobile immersive v2` commit:

- **Donut clustering**: `splitByCity` now floors the zoom (`Math.floor(zoom) > DONUT_CROSS_CITY_MAX_ZOOM`) so the split-by-city decision stays consistent with `clusterCellDegrees` within a zoom band (review comment).
- **Expanded subject**: added an `X` at the right of the category bar (same behaviour as the back `<`); tapping the title/logo also returns to the preview; the `>` chevron marks the active preview card.
- **City filter**: now a desktop-style single-select radio list (squared selector + avatar + full name) instead of chips.
- **Filters sheet**: "Καθαρισμός φίλτρων" only shows when a filter is active; date-range pickers stack full-width on iPhone; idle category chip labels take the category colour.
- **Tabs/misc**: collapse chevrons flipped (up = open); loading pill centered; subject link → "Δες τη συζήτηση"; search placeholder → "Αναζήτηση θέματος".

`npx tsc --noEmit` clean; verified in the mobile preview.

VasiaKoum:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/a4a200a963df59d767f9b8916032198531d6632d..5533ca4fdbd5109b07478fff49e169e810fbb26f) another round of mobile-search refinements, folded into the `feat(landing): mobile immersive v2` commit:

- **Search badge**: when a keyword search is active, the preview filter bar shows a leading badge (search icon + typed text, truncated) styled as the intense/filled orange chip. Tapping it now **reopens** the search panel (pre-filled) instead of clearing.
- **"Αναζήτηση" panel**: added an up-arrow at the right of the header to collapse/close the search.
- **Category chips**: labels are now bold; the search badge takes the first slot so one fewer category chip shows while a search is active.

`npx tsc --noEmit` clean; verified in the mobile preview.

VasiaKoum:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/5533ca4fdbd5109b07478fff49e169e810fbb26f..2cab3875349cda2bcb09bb5d44e455f0fab31b4f) two small mobile fixes, folded into the `feat(landing): mobile immersive v2` commit:

- **Selected-subject ×**: now fully dismisses — unselects the subject, drops the preview, and collapses the subjects list back to the bare map. The back `<` arrow / title-tap still return to the preview.
- **Date-range inputs**: constrained to ~80% width on iPhones (stacked) so the native date picker stays legible; side-by-side once there's room.

`npx tsc --noEmit` clean; verified in the mobile preview.

christosporios:
On donuts: I'm thinking that we should simplify this. How about defining a certain zoom level, above which you only see donuts, and under which no donuts are shown. Additionally, we should show exactly one donut per municipality -- and somehow also put the municipality round logo in the donut, along with the number of subjects? perhaps the topic category split could be omitted -- or even better, it could be a round thin border without the icons.

For example, this feels messy and inconsistent. There's 7 municipalities in this screen -- we should show one for each. Noting that this isn't trivial, because in the attica region there's lots of geographically small municipalities. Ideally, they should "float" and all be visible, kind of like bubbles -- even if they're not exactly at the center of the municipality (because they wouldn't fit).
 
 Screenshot 2026-07-15 at 19 08 03

VasiaKoum:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/2cab3875349cda2bcb09bb5d44e455f0fab31b4f..f22c9f8bd6b9362e14721537e678b97e0422ac16) to rebase onto `main` and re-scope this branch to a clean two-commit history:

- `b3fc04d7` — split donut clusters by municipality above the overview zoom
- `f22c9f8b` — mobile map-preview strips for subjects and δήμοι

**What changed vs the previous version of this branch:**
- Rebased onto `main`, which already contains the merged mobile-header refresh (#538). The branch's own header changes were **dropped** — `MobileHeader.tsx` is now byte-identical to `main`.
- Re-scoped to **only** the horizontal bottom strips (Θέματα + Δήμοι) with the tap-to-preview / tap-again-to-open interaction. The mobile filter bar, the ▽ filter sheet, and the header address/search dropdowns were removed; filters and keyword search use `main`'s existing search overlay + filter icon + date pill.
- Restored `main`'s locate control + geolocation-error tooltip (the branch's only geolocation entry was the dropped address panel), and reverted the i18n and `SearchBody` to `main`.

`npx tsc --noEmit` and `eslint` are clean; verified the layout renders on mobile.

VasiaKoum:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/f22c9f8bd6b9362e14721537e678b97e0422ac16..edcd6c481a8edf1afcdc1f92ccca365659dd7bb1) — the branch is now a **single commit** on top of `main`:

- **Dropped** `b3fc04d7` (*split donut clusters by municipality above the overview zoom*). Its cluster keying is back to `main`'s lng/lat-only cells and the `DONUT_CROSS_CITY_MAX_ZOOM` constant is gone; the strips' preview-pin styling was kept.
- **Strip cards now have a fixed height (150px)** instead of growing with the title: the title clamps to two lines with an ellipsis, and the meta (discussion time · date · address) is pinned to the bottom so it lines up across cards.

Verified all 861 cards report a single height of 150px, and the longest title (71 chars) clamps. `npx tsc --noEmit` and `eslint` clean.

VasiaKoum:
Rebased onto latest `main` (`775a16dd`), picking up #548 and #550. No code changes to this PR's own commits — pure rebase, 5 commits replayed (1a1484298408c70dd8117622d5241f23e32512a5 → 937c3448d74cad99ce97e407be1b209318955702).

Note: this does **not** resolve the failing `Deploy PR Preview` check. That failure is external — `actions/checkout` published a security patch across all major versions today (2026-07-20 ~15:23 UTC) that adds a guard refusing fork-PR checkouts under `pull_request_target`. The floating `@v6` tag now resolves to v6.1.0, which enforces it. Every fork PR preview is affected, not just this one; it needs a fix in the workflow itself (opt-in flag or a build/deploy split).

src/components/landing/v2/mapMarkers.tsx
```diff
@@ -337,79 +325,129 @@ export function createSubjectPin(
...
+ * includes off-screen ones and doesn't change as you zoom).
  */
 export function buildSubjectPins(
     map: MapboxMap,
     locationGroups: LandingSubject[][],
     opts: {
         selectedId: string | null;
+        /** mobile only: the previewed subject + whether we're on mobile (intense selection) */
+        previewId: string | null;
+        isMobile: boolean;
```
kouloumos: `buildSubjectPins` takes `isMobile` here but never reads it — `intense` is computed purely from `selected`/`previewId`. The whole chain feeding it is dead: `useSubjectMarkers` receives the prop, mirrors it into `isMobileRef`, then passes it in for nothing. Worth dropping from `buildSubjectPins`' opts, from `useSubjectMarkers` (param + ref + call), and from the call site in `LandingV2`. `useGeneralCityMarkers` genuinely uses its own `isMobile` for the ease offset, so only the subject-pin path is dead.

src/components/landing/v2/hooks/useMapMarkers.tsx
```diff
@@ -252,14 +268,18 @@ export function useSubjectMarkers({
...
     useEffect(() => {
         pinsRef.current.forEach((pin) => {
-            if (pin.subject) stylePin(pin, pin.subject, pin.subject.id === selectedId);
+            if (!pin.subject) return;
+            const selected = pin.subject.id === selectedId;
+            const intense = selected || pin.subject.id === previewId;
+            stylePin(pin, pin.subject, selected, intense, pin.dot);
         });
-    }, [selectedId]);
+    }, [selectedId, previewId, isMobile]);
```
kouloumos: Related: `isMobile` is in this effect's deps but the body only reads `selectedId`/`previewId`, so it re-runs on a viewport-class flip for no effect. The comment above it is also a touch misleading — `intense` is `selected || previewId` regardless of viewport; it's just that `previewId` is only ever set on mobile. Deps can drop to `[selectedId, previewId]`.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -374,24 +475,51 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
...
+            zoom: Math.max(mapInstance.getZoom(), SUBJECT_FOCUS_ZOOM),
+            duration: 500,
+        });
+    };
+
+    // A marker tap: mobile previews it (highlight + navigate, no box); desktop selects it (opens the
+    // tooltip/preview) and reports the funnel source.
+    const onMarkerClick = (id: string, source: 'map_pin' | 'cluster' | 'city_hall') => {
+        if (isMobile) previewSubject(id);
+        else trackedSelectSubject(id, source);
```
kouloumos, line 483: On desktop, if I'm on the Δήμοι tab (`view === 'municipalities'`) and click a subject pin, the map selects it but the left sidebar stays on the municipalities list — `selectSubject` never touches `view`. I'd expect picking a subject to drop me into the subjects view so I get the details, filters and search on the left. Could the desktop selection path `setView('subjects')` (probably cleanest in `trackedSelectSubject`, so co-located/general-box selections behave the same)? Mobile already handles this via the strip.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -404,14 +532,38 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
...
+                        [bounds.minLng, bounds.minLat],
+                        [bounds.maxLng, bounds.maxLat],
+                    ],
+                    { padding: isMobile ? 40 : 50 },
+                );
+            if (camera?.center != null && camera.zoom != null) {
+                mapInstance.easeTo({
+                    center: camera.center,
+                    // A touch past the exact fit so the δήμος sits a bit closer than edge-to-edge.
+                    zoom: isMobile ? Math.max(camera.zoom + 1, MUNICIPALITY_COUNT_MAX_ZOOM_SINGLE + 1) : Math.max(camera.zoom, MUNICIPALITY_COUNT_MAX_ZOOM_SINGLE + 1),
```
kouloumos: You already solved this exact problem here for the count-donut click: `cameraForBounds` + a floor of `MUNICIPALITY_COUNT_MAX_ZOOM_SINGLE + 1`, so even a large δήμος lands past the count threshold and shows its subjects. But the Cities-tab selection takes a different route — `setFilters({ cityIds: [id] })` → `flyToMunicipality`, a plain `fitBounds({ maxZoom: 17 })` with no floor. So on mobile, picking a geographically large δήμος lands below zoom 10 and leaves you on the city circle needing another tap, while a small δήμος lands zoomed into its subjects. The two "frame a δήμος" paths should share the same floor — ideally `flyToMunicipality` adopts the `cameraForBounds` + `MUNICIPALITY_COUNT_MAX_ZOOM_SINGLE + 1` logic you have here, or both call one small helper.

src/lib/landing/donut.ts
```diff
@@ -142,35 +41,131 @@ function donutSegmentPath(
...
-    const countText =
-        `<text x="${r}" y="${r}" dominant-baseline="central" text-anchor="middle" ` +
-        `font-family="'Relative Book Pro', Inter, sans-serif" font-weight="600" ` +
-        `font-size="${fontSize}" fill="#0c0a09">${totalCount}</text>`;
+    // The opening tag is deliberately one unbroken template literal. Split across `+` it is silently
+    // corrupted by the production minifier — Turbopack dropped the whole `xmlns`/`style` chunk, so
+    // the shipped markup read `viewBox="0 0 68 68<g transform="…`, the <g> was swallowed into the
+    // viewBox attribute and every arc rendered 11px off-centre from its logo. Dev builds don't
+    // minify, so it only ever showed up on deployed previews. Keep it in one piece.
     return (
```
kouloumos, line 166: `municipalityDonutSvg` is untested, and it carries the one invariant that already shipped a bug: the opening `<svg>` tag has to stay a single unbroken template literal — splitting it across `+` let Turbopack drop the `xmlns`/`style` chunk, folding `<g>` into the `viewBox` and offsetting every arc, and only on minified previews (dev builds don't minify, so it passed locally). It's a pure `(DonutSegment[]) => string`, so a cheap guard like `expect(svg).toMatch(/^<svg[^>]*viewBox="0 0 68 68"[^>]*><g transform=/)` — plus a segment-count and empty-fallback assertion — would stop a future refactor from silently reshipping it.

src/components/filters/TopicFilter.tsx
```diff
@@ -7,6 +7,8 @@ import { cn } from '@/lib/utils';
 import { motion } from 'framer-motion';
 import { Topic } from '@prisma/client';
 import Icon from '@/components/icon';
```
kouloumos: Unused now that `<TopicIcon>` replaced the inline `<Icon>` badge — safe to drop.

```suggestion
```

src/components/layout/Header.tsx
```diff
@@ -17,6 +17,7 @@ import { useSubjectHeaderOptional, SubjectHeaderInfo } from "@/contexts/SubjectH
 import { AutoScrollText } from "@/components/ui/auto-scroll-text"
 import Icon from "@/components/icon"
```
kouloumos: Same here — `Icon` is no longer referenced after the `<TopicIcon>` switch.

```suggestion
```

kouloumos (commented):
This is really solid. The progressive-zoom display (count donuts → pins → dots) is cohesive and well-documented — one block in `LandingV2` makes the decision, the thresholds are all named constants in `landingCore`, and the marker hooks are just consumers gated by `active`, so there's a single place to understand the rule. Performance holds up: no refetch on pan/zoom, and the count markers are repositioned rather than torn down and rebuilt. The `topicStyle`/`TopicIcon` consolidation is a genuine cleanup.

A couple of UX gaps came up while clicking around, both inline: on desktop, selecting a subject from the Δήμοι tab doesn't switch into the subject view; and on mobile, picking a city from the Cities tab lands at an inconsistent zoom because the filter path doesn't use the zoom floor the donut-click path already has. Otherwise just a dead `isMobile` path, two now-unused imports, and a missing regression test on the donut SVG. Details inline.

One optional thing while you're already in `landingData.ts` — and to be clear this is pre-existing, not something your change introduced: `toLandingSubjects` and `toGeneralCities` build a `LandingSubject` from an ~18-field object literal that's near-identical between them (they really differ only in `lat`/`lng` — point vs city centroid — and `where`). Since `GeneralSubjectRow` is just `Omit<MapSubjectRow, 'geometry' | 'locationText' | 'locationType'>`, both could collapse into one `rowToLandingSubject(row, { lng, lat, where })` helper, which also removes the drift risk when a `LandingSubject` field is added and only one mapper gets updated. Entirely fine to leave for a follow-up — flagging it here so it's on the record.

src/components/landing/v2/LandingV2.tsx
VasiaKoum: The desktop selection effect now flips to the subjects view when a subject is selected. I put it on the `selectedId` change (in `DesktopLayout`) rather than `trackedSelectSubject`, so co-located and general-box picks funnel through the same path and behave identically.

src/components/landing/v2/MunicipalitiesList.tsx
```diff
@@ -28,81 +32,93 @@ export function MunicipalitiesList({
...
             ))}
             <PetitionCta big source="municipalities_list" />
         </>
     );
 }
 
-/* δήμος card — stats + next meeting, links to the municipality page */
+/* δήμος card — stats + next meeting. Clicking the card filters to that δήμος (orange border while
+   selected, clicking again clears it); the bell opens its notifications and the arrow its page. */
 function MuniPanelCard({
```
kouloumos, line 46: Naming consistency: everything in the landing uses the full `Municipality*` (MunicipalitiesList, MunicipalitySubjectCount, MunicipalityPageButton…) — this `MuniPanelCard` and `MuniStat` (below) are the only two that abbreviate. Rename to match so there's one convention (grep-able, no "Muni or Municipality?" guessing).

src/components/filters/TopicFilter.tsx
```diff
@@ -101,6 +102,7 @@ export function TopicFilter({
             )}>
                 {topics.map(topic => {
                     const isSelected = isTopicSelected(topic);
+                    const s = topicStyle(topic.colorHex);
```
kouloumos, line 105: `s` doesn't say what it holds — it's the topic's resolved style. Rename `s` → `style` (there are a couple of `s.` reads just below to update too). Single-letter names for a domain value cost the next reader a lookup.

src/components/landing/v2/NotifyMunicipalityDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+                            onClick={() => {
+                                captureLandingAction('notify_cta', { surface: 'notify_dialog', city_id: city.id });
+                                onOpenChange(false);
+                            }}
+                            className="flex flex-col items-center gap-1.5 rounded-xl bg-white p-2.5 text-center no-underline shadow-sm transition-transform hover:scale-[1.03] hover:no-underline"
+                        >
+                            <CityAvatar city={city} />
+                            {/* the genitive form ("Δήμος Αθηναίων"), as everywhere else a δήμος is named */}
+                            <span className="text-md sm:text-lg font-semibold leading-tight text-foreground">
+                                {city.name_municipality}
```
kouloumos, line 72: Two small things: (1) the copy (`notify.pickMunicipality`) repeats "Επέλεξε τον Δήμο" when the δήμος options are right below it — we have wording on the notifications page to reuse; something like "Λάβε ενημερώσεις για τα θέματα και τις περιοχές που σε ενδιαφέρουν" reads better and drops the redundancy. (2) drop the comment on the line above — `{city.name_municipality}` rendering the genitive form is self-evident, and that gloss is repeated on ~5 fields (keep it once at the type root in `landingData.ts`).

src/components/landing/v2/NotifyPrompt.tsx
```diff
@@ -40,11 +44,25 @@ export function NotifyPrompt({
...
+                            known && logoImage
+                                ? 'border border-border bg-card p-1.5'
+                                : known
+                                  ? 'bg-primary/10 text-primary'
+                                  : 'bg-muted text-muted-foreground',
                         )}
                     >
-                        <Bell className="h-6 w-6 sm:h-7 sm:w-7" />
+                        {known && logoImage ? (
+                            <Image
```
kouloumos, line 56: This hand-rolls the logo `<Image>` that `CityAvatar` (`controls.tsx`) already does — and it's the shared avatar in SearchBody / NotifyMunicipalityDialog / MunicipalitiesList / MobileLayout. Reuse it for the logo case; the only difference is your no-logo fallback is a `Bell` (CityAvatar's is the city initial), so keep the Bell branch and use `CityAvatar` when there *is* a logo (or give it a fallback prop).

src/lib/landing/landingData.ts
```diff
@@ -81,6 +81,104 @@ export type LandingGeneralCity = {
...
+    const acc = new Map<
+        string,
+        {
+            cityId: string;
+            cityName: string;
+            nameMunicipality: string;
+            logoImage: string | null;
+            count: number;
+            members: LandingSubject[];
+            lngSum: number;
```
kouloumos, line 125: Field-name drift: the same concept is `cityLogo` on `bump`'s param but `logoImage` on the accumulator and on `MunicipalitySubjectCount` — worth picking one name so it reads as one thing.

[follow-up] the inline `acc` / `bump` types re-spell `MunicipalitySubjectCount` (9 and 7 fields) — derive them (`Pick`/`Omit` + the `lngSum/latSum/n` sum fields) rather than repeat the shape.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -262,9 +300,13 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
         setInterested,
     });
 
-    // Selecting a subject (map pin, search) closes the info drawer so its preview is visible.
+    // Selecting a subject (map pin, search) closes the info drawer so its preview is visible, and
+    // drops any strip-preview outline (the selected pin's own highlight takes over).
     useEffect(() => {
```
kouloumos, line 305: [follow-up] State written by effects instead of events: `interested` is set from four places (two effects here + imperative writes at `:317` and `useSubjectSelection.ts:62`), so precedence rides React's effect-flush order. A single `noteInterest(source)` called from each event handler is simpler and unambiguous. Same shape one screen up: `layoutSyncedRef` + the breakpoint effect (`:121-128`) re-derives `view` that render-time `desktopView(view)` already handles — deletable if the URL-sync writes `desktopView(view)` on desktop.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -404,14 +516,34 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
...
-            router.push(`/${cityId}`);
+        active: showMunicipalityCounts,
+        municipalityCounts,
+        onZoomToCity: (muni) => {
+            if (!mapInstance) return;
+            // cameraForBounds gives the fit zoom for the boundary (large δήμος → lower, small → higher);
+            // floor it one past MUNICIPALITY_COUNT_MAX_ZOOM so the δήμος's subjects always show (mobile
+            // goes a touch closer still). Desktop offsets the δήμος to the right ~2/3 of the screen, so
+            // it sits clear of the floating list on the left rather than under it.
+            const offset: [number, number] = isMobile ? [0, 0] : [mapInstance.getContainer().clientWidth / 6, 0];
```
kouloumos, line 531: [follow-up] A helper that should exist: the desktop-vs-mobile camera offset is re-encoded at 6 sites with drifting constants — `useMapActions.ts:88,120` (`[210,0]`), `useSubjectSelection.ts:52-55`, `useMapMarkers.tsx:344` (`[210,100]`), here (`clientWidth/6`), and `landingCore.ts:178` `flyToMunicipality` (`[120,0]`). One `landingCameraOffset(isMobile, kind: 'subjectFocus'|'municipalityFit'|'generalBox')` centralizes the panel-clearance geometry — and this drift is part of why mobile framing is inconsistent, so it pairs with the framing fix.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -357,14 +401,56 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
...
+            const municipalityInView = mapCitiesRef.current.some(
+                (city) =>
+                    city.lng >= b.getWest() &&
+                    city.lng <= b.getEast() &&
+                    city.lat >= b.getSouth() &&
+                    city.lat <= b.getNorth(),
+            );
+            if (!municipalityInView) return;
+            const c = mapInstance.getCenter();
+            writeSavedView({ center: [c.lng, c.lat], zoom: mapInstance.getZoom() });
```
kouloumos, line 439: [follow-up] Reuse the existing helper: this moveend hand-rolls a viewport test that `subjectInViewport()` already does. It also uses **centroid-in-bounding-box** here while the *read* (`initialView`, `isInSupportedMunicipality`) uses **point-in-polygon** — two definitions of "in a δήμος", worth reconciling. And it's a 2nd moveend handler, separate from the capture one in `useMapMarkers` — the two could fold together.

src/components/landing/v2/LandingV2.tsx
```diff
@@ -531,14 +677,15 @@ export function LandingV2({ defaultView, initial }: LandingV2Props) {
         count: listSubjects.length,
```
kouloumos, line 677: [follow-up] `ordered` and `trending` (the two lines just above) are the **same value** (`listSubjects`) under two names — one consumed only by desktop, one only by mobile. That's a symptom of `LayoutProps` (`landingCore.ts:274`) being a god-object: ~11 of its fields are dead for one layout or the other. Splitting it (Shared / Desktop / Mobile) and collapsing this duplicate would make each layout's real dependencies legible.

src/components/landing/v2/hooks/useMapMarkers.tsx
```diff
@@ -55,8 +57,6 @@ export function useMapViewCapture({
     useEffect(() => {
```
kouloumos, line 57: [follow-up] This moveend handler carries three cross-hook refs (`suppressViewCaptureRef`, `pendingCoLocatedRef`, `pendingGeneralRef`) so each "ease the camera, then open a box" case bolts its payload onto the shared handler. `map.once('moveend', () => openBox(...))` at each call site collapses that — Mapbox already gives you one-shot completion, so the parked-ref channel isn't needed.

src/lib/landing/savedView.ts
```diff
@@ -0,0 +1,66 @@
...
+/** Narrow untrusted localStorage JSON to a fresh, well-formed view. Anything off → null. */
+function parseSavedView(raw: string): SavedView | null {
+    const data: unknown = JSON.parse(raw);
+    if (typeof data !== 'object' || data === null) return null;
+    const { center, zoom, at } = data as Record<string, unknown>;
+    if (typeof at !== 'number' || !Number.isFinite(at) || Date.now() - at > MAX_AGE_MS) return null;
+    // The ceiling here (24, Mapbox's absolute max) is deliberately looser than write's MAX_SAVED_ZOOM
+    // cap: read only rejects impossible values, so an entry written before the cap changed still
+    // restores rather than being thrown away. write() is what enforces the product-level limit.
+    if (typeof zoom !== 'number' || !Number.isFinite(zoom) || zoom < 0 || zoom > 24) return null;
```
kouloumos, line 32: [follow-up] Hand-rolled type-narrowing where the house standard is Zod (`src/lib/zod-schemas/*`) — a small schema would be more idiomatic and less to read. Related: the read (≤24) vs write (≤14 `MAX_SAVED_ZOOM`) asymmetry guards a case that can't occur (write always caps), and its rationale is restated 3× — collapse to one bound + one comment. (The `MAX_SAVED_ZOOM` idea itself is a fine feature — context on return.)

src/components/landing/v2/MobileLayout.tsx
```diff
@@ -327,20 +382,326 @@ function MobileViewSwitch({
...
+            </div>
         </div>
     );
 }
 
-/* OpenCouncil preview (mobile) — same card treatment, anchored above the switcher */
+/* One δήμος card in the strip — the same content as the desktop Δήμοι card (logo · name · bell ·
+   stats · next meeting). Tapping the card filters to that δήμος (orange border when selected); the
+   bell opens its notifications and the arrow opens its page. */
+function MunicipalityCard({
```
kouloumos, line 446: [follow-up] This `MunicipalityCard` is ~90% identical to the desktop `MuniPanelCard` (`MunicipalitiesList.tsx:46`) — same data shape, same bell/arrow links + analytics, same stat grid + next-meeting row — and `MuniStat` is copy-pasted verbatim. Only card width, a divider, and font tokens differ. One shared `MunicipalityCard` with a `variant: 'panel' | 'strip'` (like the existing `SubjectCard`) would de-dupe it.

src/components/landing/v2/MobileLayout.tsx
```diff
@@ -70,114 +73,76 @@ export function MobileLayout({
...
+    // it only swaps the list, never the map. 'home' collapses to the Θέματα tab.
+    const tab: 'subjects' | 'municipalities' = view === 'municipalities' ? 'municipalities' : 'subjects';
+    // The list above the tabs can be collapsed to reveal the map; re-tapping the active tab toggles it.
+    // Starts collapsed in the zoomed-out overview (Greece / cluster view), where there's no δήμος to
+    // list subjects for.
+    const [listCollapsed, setListCollapsed] = useState(overviewActive);
+    // Follow the overview state on each crossing: collapse when zooming out to the country/cluster
+    // view, open when drilling into a δήμος. Only on the transition, so a manual re-tap within either
+    // band isn't overridden.
+    const prevOverviewRef = useRef(overviewActive);
```
kouloumos, line 86: [follow-up] `prevOverviewRef` here is the "adjust state when a prop changes, via a previous-value ref" pattern React's docs steer away from. Since `listCollapsed` both mirrors `overviewActive` and is user-togglable, do it *during render* with a `prevOverview` state (`if (overviewActive !== prevOverview) { setPrevOverview(...); setListCollapsed(overviewActive) }`) — no effect, no ref, React bails out of the extra commit.

src/components/landing/v2/hooks/useFilteredSubjects.ts
```diff
@@ -167,19 +167,23 @@ export function useFilteredSubjects({
...
             if (sel) list = [sel, ...list];
         }
+        // A previewed subject (e.g. tapped on the map, or from a co-located / general box) may be
+        // outside the current viewport — keep it in the list so the strip can highlight it.
+        if (previewId && previewId !== selectedId && !list.some((s) => s.id === previewId)) {
+            const prev = findSubject(previewId);
+            if (prev) list = [...list, prev];
+        }
         return list;
         // eslint-disable-next-line react-hooks/exhaustive-deps
```
kouloumos, line 185: [follow-up] These eslint-disables exist because `findSubject` (defined just above) is a fresh closure every render, so the linter can't verify the memo deps. Wrap `findSubject` in `useCallback([allSubjects, allGeneralSubjects])` and the memos can depend on it honestly — both disables here and at `:200` go away.

src/components/landing/v2/SearchBody.tsx
```diff
@@ -68,6 +69,7 @@ export function SearchBody({
     loading,
     onPickResult,
     onLocateAddress,
+    forceFilters = false,
```
kouloumos, line 72: [follow-up, minor] The `forceFilters` flag adds a third display mode to the panel (filters / query-results / forced-filters). Looked at it per the review — it's defensible for the "filters icon should show filters even with an active query" interaction, so flagging as minor, not asking to change it. Noting in case the panel's mode logic gets revisited.

src/components/landing/v2/LandingAside.tsx
```diff
@@ -107,6 +119,15 @@ export function LandingAside({
...
+                        {/* the "?" guide — parity with the mobile menu's info action */}
+                        <DropdownMenuItem
+                            onSelect={() => onToggleInfo('menu')}
+                            className="flex items-center gap-2 rounded-lg text-muted-foreground focus:bg-muted focus:text-foreground"
+                        >
+                            <HelpCircle className="h-4 w-4" />
+                            {t('info.title')}
+                        </DropdownMenuItem>
+                        <DropdownMenuSeparator className="bg-muted" />
                         {FOOTER_GROUPS.map((group, gi) => (
```
kouloumos, line 131: [follow-up, low] The overflow menu (this `FOOTER_GROUPS` link ladder + the account/session block) is reimplemented in `MobileHeader` too. The containers legitimately differ (dropdown vs full-screen sheet), but the per-link-type rendering and the account block are pure duplication — a shared `renderFooterLink(link)` + `<LandingAccountControls/>` would remove it. Opportunistic, lowest priority.

kouloumos (approved):
Strong iteration — the count-circle fix works, and I validated selection lands on subjects across all **11 δήμοι × 4 devices** (Laptop / Desktop / iPhone / Android). Nothing here blocks — I'm approving. Comments are inline; ones tagged **[follow-up]** are post-merge (captured as a reference, not asks for this PR), the rest are small suggestions for this PR at your discretion.

**Mobile framing [follow-up].** Selection lands on subjects everywhere ✓, but the mobile *framing* doesn't hug the δήμος border the way desktop does (desktop is the reference). Three patterns — each affects several δήμοι, the screenshots are representative:
- **Compact → too zoomed out** (the δήμος shrinks to a blob among its neighbours): **Αθήνα, Ζωγράφου, Χαλάνδρι, Βριλήσσια**
- **Large → cropped tighter than desktop** (the border falls outside the frame): **Σπάρτη, Ορεστιάδα, Άργος-Μυκήνες, Ξυλόκαστρο**
- **Off-center**: **Σαμοθράκη**

(Χανιά and Αργιθέα look right on both.) Reads like the mobile `cameraForBounds` padding + the fixed zoom floor over a narrow viewport — see the camera-offset comment on `LandingV2`.

<details><summary>📸 <b>Click to expand</b> — desktop vs mobile framing (Αθήνα · Ζωγράφου · Σπάρτη)</summary>

**Αθήνα** — compact, too zoomed out on mobile _(same on Ζωγράφου, Χαλάνδρι, Βριλήσσια)_


**Ζωγράφου** — same compact pattern


**Σπάρτη** — large, cropped tighter on mobile _(same on Ορεστιάδα, Άργος-Μυκήνες, Ξυλόκαστρο)_

</details>

**Comment hygiene.** A few rationales are copy-pasted across sites — keep each once at its root and drop the restatements: the "genitive form (Δήμος X)" gloss (root `landingData.ts` `nameMunicipality`, restated ~5×), currentColor (`landingCore` `stylePin` root, restated `:163`), `?view=help` (×3 in `landingCore`), the savedView read/write note (×3), and `TopicIcon.tsx:19-20` re-telling topicStyle's origin (→ pointer). The prose is genuinely good — this is just trimming repetition. Also `isPointInGeometry` is exported but only used internally — could unexport.

kouloumos:
Took a pass at the mobile-framing follow-up and pushed the fix — [force-pushed](https://github.com/schemalabz/opencouncil/compare/1d8a1de334b054eb167155662de803d76cf6e738..4a98bac6ad3412a15fbdc07bfa2c28216c7207c2) folding it into `863c1a85` (the overview-map commit).

**The change:** `flyToMunicipality`'s mobile `cameraForBounds` padding was `120` — a big fraction of a phone's width, which shrank compact δήμοι to a tiny blob among their neighbours on selection. Dropped it to `40`, which also **unifies** it with the value the donut-click path (`onZoomToCity`) already uses, so the two "frame a δήμος" paths behave the same now.

**Before → after** (Cities-tab selection, mobile — Αθήνα):
athens

<details><summary>📸 <b>More validation</b> (click to expand)</summary>

Other compact δήμοι — Ζωγράφου, Βριλήσσια, Χαλάνδρι:
zografou
vrilissia
chalandri

Consistent across mobile widths (320 → 430px):
sizes

Desktop unchanged (regression check — identical before/after):
desktop
</details>

Verified in the running app: 4 compact δήμοι all hug the border now, consistent across 4 screen widths, and no desktop regression (the change is mobile-only).

Left alone on purpose: **large δήμοι** (Σπάρτη etc.) are floor-clamped and unchanged — hugging a big rural δήμος's border would just show empty countryside; framing the subjects is better. **Σαμοθράκη**'s off-center framing is a separate centering bug (the geometry midpoint), noted as a follow-up. The broader `landingCameraOffset`/padding consolidation stays the post-merge follow-up from the review.

If you want to compare or roll back, the pre-fix state is preserved on my fork at [`backup/pr-533-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-533-original) (`1d8a1de3`).

PR #537, opened 2026-07-14
chore: upgrade Docker images to Node 24

christosporios:
## Why

Production has been emitting recurring `App Router render` error alerts:

```
TypeError: controller[kState].transformAlgorithm is not a function
    at transformStreamDefaultControllerPerformTransform (node:internal/webstreams/transformstream:527:37)
```

This is not an application bug — it's a known race condition inside Node's `TransformStream` implementation ([nodejs/node#62036](https://github.com/nodejs/node/issues/62036)), triggered when a client disconnects (or an LB times out) while Next.js is still streaming a server-rendered page. The stack trace contains no application frames; it also underlies [vercel/next.js#75994](https://github.com/vercel/next.js/issues/75994).

The upstream fix ([nodejs/node#62040](https://github.com/nodejs/node/pull/62040)) shipped in Node 24.15.0 (LTS) and 25.8.1. Node 20 — our current Docker base — went EOL in April 2026 and will never receive the backport, so these alerts recur for as long as we stay on it.

## Changes

- `Dockerfile` and `Dockerfile.dev`: `node:20-alpine` → `node:24-alpine` (active LTS, contains the fix)
- `package.json`: widen `engines.node` from `>=20.9.0 <23` to `>=20.9.0 <25` so a Node 24 runtime is allowed (Nix dev shells and preview deployments stay on the nixpkgs default, currently Node 22, and remain within range)

## Notes for review

- Production and staging deploy via DO App Platform from this Dockerfile, so the runtime bump takes effect on the next deploy of `main` (staging) — worth watching staging after merge before promoting to `production`.
- PR previews build with Nix, not Docker, so the preview for this PR does **not** exercise Node 24. If we want previews/dev shells on 24 too, that's a follow-up in `flake.nix` (`pkgs.nodejs` → `pkgs.nodejs_24`).
- Next.js 16 requires Node ≥ 20.9 and supports Node 24, so no framework constraint blocks this.

https://claude.ai/code/session_01HgSfUQJo3ifvUQr8mNxsgj

---
_Generated by [Claude Code](https://claude.ai/code/session_01HgSfUQJo3ifvUQr8mNxsgj)_

---

> [!NOTE]
> **Medium Risk**
> This changes the production/staging container runtime on deploy with no application code changes; validate staging after merge before promoting to production.
> 
> **Overview**
> Moves **production and development Docker** bases from `node:20-alpine` to **`node:24-alpine`** so deployed runtimes pick up the Node **TransformStream** fix tied to recurring Next.js App Router streaming errors on client disconnect.
> 
> Updates **`package.json` `engines.node`** to keep **Node 20.9–22.x** while also allowing **`>=24.15.0 <25`**, so Node 24 is permitted only at patch levels that include that fix (and unfixed Node 23 / early Node 24 are not declared compatible).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 236831655611855679df2eedaeb6964e1d574446. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR updates the application’s supported Node.js runtimes.

- Moves production and development Docker images from Node 20 to Node 24.
- Allows Node 24.15.0 and newer while excluding earlier Node 24 releases.
- Keeps Node 20 through Node 22 compatible with existing Nix environments.

This looks safe to merge.

- The updated engine range excludes the Node 24 versions that lack the required fix.
- No blocking issues remain in the changed code.

| Filename | Overview |
|----------|----------|
| Dockerfile | Updates the production container base image from Node 20 Alpine to Node 24 Alpine. |
| Dockerfile.dev | Updates the development container base image from Node 20 Alpine to Node 24 Alpine. |
| package.json | Allows Node 24.15.0 through Node 24 while retaining support for Node 20 through Node 22. |

<sub>Reviews (2): Last reviewed commit: ["Update package.json"](https://github.com/schemalabz/opencouncil/commit/236831655611855679df2eedaeb6964e1d574446) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44077152)</sub>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Base image
-FROM node:20-alpine AS base
+FROM node:24-alpine AS base
 
 # Set the working directory in the container
 WORKDIR /app
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,5 +1,5 @@
 # Stage 1: Base image
-FROM node:20-alpine AS base
+FROM node:24-alpine AS base
 
 # Set the working directory in the container
 WORKDIR /app
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "seed": "tsx prisma/seed.ts"
   },
   "engines": {
-    "node": ">=20.9.0 <25"
+    "node": ">=20.9.0 <23 || >=24.15.0 <25"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.56.0",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "seed": "tsx prisma/seed.ts"
   },
   "engines": {
-    "node": ">=20.9.0 <23"
+    "node": ">=20.9.0 <25"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.56.0",


kouloumos (approved):
Verified this locally since CI never builds the Docker image: built the image from this branch and ran the full container boot that DO performs at deploy (npm install → migrate deploy + generate → patch-package → next build → next start), all on `node:24-alpine` — everything passes, pages render, and `/_next/image` serves resized images so sharp's native musl binary is fine. Also confirmed `node:24-alpine` currently resolves to 24.18.0, comfortably past the 24.15.0 fix.

One behavioral change worth knowing about: Node 24 bundles npm 11, which skips dependency install scripts by default — the image build logs `npm warn allow-scripts` for prisma, sharp, ssh2, tree-sitter and protobufjs. I traced all five and they're harmless here (sharp ships prebuilt binaries via optionalDependencies, prisma's preinstall is just a version check and `prisma generate` runs in the entrypoint anyway, the rest are test-only or client-side deps). Note though that our nix dev shells (npm 10) still run these scripts, so a future dependency that needs one would work locally but arrive broken in the image — the `allow-scripts` warn list in the deploy build logs is the tell.

Two corrections for the PR body: the engines range it describes is the pre-suggestion one (`>=20.9.0 <25`), and the Nix dev shells/previews are actually on Node 20.19.1, not 22 — the flake pins nixos-24.11, whose default `nodejs` is 20.19.1. Still within the widened range, so nothing breaks on merge; it does mean dev and previews stay on the EOL runtime with the bug, which is fine in practice since nothing alerts there.

On the flake follow-up in the notes: it's slightly more involved than `pkgs.nodejs` → `pkgs.nodejs_24` — nixos-24.11 has no `nodejs_24` at all, and our locked nixpkgs-unstable only has 24.13.0, which predates the fix. The path is bumping the unstable input and taking `nodejs_24` from there (unstable only supplies ngrok today, so low blast radius). Node 22 isn't an alternative either: the fix hasn't been backported as of 22.23.1.

Approving — will watch staging after the merge as you suggest.

kouloumos:
Verified against the live App Platform config after merge — this unfortunately doesn't change the deployed runtime, and the alerts will keep recurring. Two findings:

1. Prod and staging build with the **Node buildpack**, not the Dockerfile (`dockerfile_path: null` in both app specs) — so the image bumps only affect local dev, and the runtime was already Node 22 (not 20).
2. The `engines.node` widening also doesn't get us to 24: when the buildpack sees a wide range spanning multiple release lines, it doesn't pick the newest match — it deliberately caps the choice at what its metadata considers the Active LTS line, and installs 22.22.2 again. The post-merge staging build log says this explicitly and points to the `NODEJS_ALLOW_WIDE_RANGE=true` opt-out.

The diagnosis itself checks out (the fix shipped in 24.15.0/25.8.1, with no V22 backport) — the runtime just needs to actually reach ≥ 24.15.0. Full evidence, build logs, and fix options in #547.

PR #548, opened 2026-07-20
feat(realm): resolve hosts to the most specific realm domain

kouloumos:
Realm resolution (#494) matches hosts against realm domains in declaration order, with each domain claiming its apex and all subdomains. That makes a realm hosted on a subdomain of another realm's domain unreachable — the parent's suffix match always wins, so e.g. a realm on `sub.opencouncil.gr` could never resolve. This PR resolves against domains **most-specific (longest) first**, so such a realm claims its own hosts while the parent realm keeps all others, including preview hosts like `pr-N.preview.opencouncil.gr`.

## What changed

- Resolution logic is built by `createRealmResolver(realms)`, which sorts the realm→domain map by domain length once (a subdomain is always longer than its parent, so length ordering is exactly specificity ordering) and returns the resolver; `realmForHost` binds it to `REALMS` with the existing greece fallback. The ordering is computed once at module load, not per request — resolution stays free in the proxy hot path, which is the point of keeping realm config in code.
- New `realm.test.ts` pins the contract against a synthetic nested domain pair — subdomain realm wins, parent keeps its other subdomains, a label that merely ends with the subdomain label doesn't false-match — plus the production apex/preview/fallback behavior and `isKnownRealmHost` trust boundaries.

The current realm domains don't nest, so production behavior is unchanged. Actually adding a subdomain realm later is the usual realm addition (enum value + `REALMS` entry + compiler-forced map entries) plus its DNS/TLS/routing.

## Verification

`npx tsc --noEmit` clean, `eslint .` clean, 1003 tests passing (15 new), production build passes. Also live-tested end-to-end with a spoofed-Host pass against a dev server with a temporarily nested realm config: robots/sitemap/metadata emit the nested realm's base URL on its host, the parent apex and preview hosts keep the parent realm's, and cross-realm data isolation holds.

<h3>Greptile Summary</h3>

This PR fixes a bug in realm resolution where a realm hosted on a subdomain of another realm's domain was unreachable because the parent realm's suffix match always won. The fix sorts the realm→domain map by domain length (descending) once at module load inside the new `createRealmResolver` factory, so the most-specific domain always claims its hosts first.

- `createRealmResolver(realms)` replaces the previous inline loop in `realmForHost` with a factory that pre-sorts entries and returns a closure — zero per-request overhead.
- `realmForHost` becomes a thin wrapper around the pre-built resolver with the existing `greece` fallback.
- 15 new tests pin the nested-domain contract (child wins, dot-separator guard, port stripping, case folding) plus production apex/preview/fallback and `isKnownRealmHost` trust boundaries.

Safe to merge — the change is purely additive, production behavior is unchanged (no current domain nesting), and the fix is gated behind a well-tested sort.

The sort-by-length approach correctly establishes specificity ordering because a proper subdomain always adds at least one label plus a dot, making it strictly longer than its parent. All edge cases (port stripping, case normalization, null/empty input) are handled correctly and covered by the new tests. No existing callers change behavior since current realm domains do not nest.

| Filename | Overview |
|----------|----------|
| src/lib/realm.ts | Introduces `createRealmResolver` factory that sorts realms by domain length (descending) once at module load and returns a per-host lookup closure; `realmForHost` wraps it with the existing `greece` fallback. Logic is correct, hot path is free. |
| src/lib/__tests__/realm.test.ts | Adds 15 new tests covering nested-domain resolution (child wins over parent, dot-separator guard, port stripping, case normalization) and expands production-domain and isKnownRealmHost coverage; contract is well-pinned. |

<sub>Reviews (3): Last reviewed commit: ["feat(realm): resolve hosts to the most s..."](https://github.com/schemalabz/opencouncil/commit/e499ea446c40388bdb80f002c7e760953d193ba6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45604683)</sub>

### DIFF
--- a/src/lib/__tests__/realm.test.ts
+++ b/src/lib/__tests__/realm.test.ts
@@ -1,17 +1,74 @@
-import { realmForHost, isKnownRealmHost } from '../realm';
+import { realmForHost, createRealmResolver, isKnownRealmHost } from '../realm';
 
-describe('realmForHost', () => {
-    it('resolves the cypriot apex domain', () => {
-        expect(realmForHost('opencouncil.cy')).toBe('cyprus');
+describe('createRealmResolver', () => {
+    // Parent declared first so a naive first-match would wrongly return it for
+    // hosts on the child's domain — the ordering under test must not depend on
+    // declaration order.
+    const NESTED = {
+        parent: { domain: 'example.org' },
+        child: { domain: 'sub.example.org' },
+    };
+    const resolve = createRealmResolver(NESTED);
+
+    it('resolves a host on a subdomain realm to that realm, not its parent', () => {
+        expect(resolve('sub.example.org')).toBe('child');
     });
 
-    it('resolves cypriot subdomains, like preview hosts, to cyprus', () => {
-        expect(realmForHost('pr-7.preview.opencouncil.cy')).toBe('cyprus');
+    it('resolves subdomains of a subdomain realm to that realm', () => {
+        expect(resolve('pr-7.sub.example.org')).toBe('child');
+    });
+
+    it('resolves other subdomains of the parent domain to the parent realm', () => {
+        expect(resolve('pr-7.preview.example.org')).toBe('parent');
+    });
+
+    it('does not match a label that merely ends with a subdomain realm label', () => {
+        expect(resolve('notsub.example.org')).toBe('parent');
+    });
+
+    it('strips the port and lowercases the host', () => {
+        expect(resolve('SUB.Example.ORG:443')).toBe('child');
+    });
+
+    it('returns null when no realm domain matches', () => {
+        expect(resolve('unknown.com')).toBeNull();
+        expect(resolve(null)).toBeNull();
     });
 });
 
+describe('realmForHost', () => {
+    it('resolves the production apex domains', () => {
+        expect(realmForHost('opencouncil.gr')).toBe('greece');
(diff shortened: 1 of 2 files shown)


kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/0954ad12256a13dae318661c11fcbad3aa514ae7..e9b2885dc6e4ce66683cce8778877340601ffb97) to eliminate per-request work from the resolver:

- `resolveRealmForHost(host, realms)` became a factory, `createRealmResolver(realms)`, which computes the domain-specificity ordering once per realm map instead of sorting on every call (~122ns vs ~344ns per resolution). The realm config being static in code is what makes resolution free in the proxy hot path — deriving the ordering per request contradicted that.
- Same contract, same tests (adapted to the factory API).

christosporios (approved):

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/e9b2885dc6e4ce66683cce8778877340601ffb97..e499ea446c40388bdb80f002c7e760953d193ba6) to rebase onto main after #550 merged:

- Resolved the anticipated `realm.test.ts` add/add conflict by unioning the two files: this PR's resolver-contract suites plus #550's cyprus pins, folded into the shared production sections (all three realms now covered for apex/preview/`isKnownRealmHost`).
- `realm.ts` merged cleanly — no code changes to this PR's diff.

Issue #242, opened 2026-03-01
optimization suggestions


angelospk:
I’ve done a review using Claude Code and the Vercel React Best Practices skill on the project. It found some findings; I’ll share them in a gist, and when you have time, I’d like us to go over them to see whether you agree that these are optimizations that could be made.
(Summary has some extra suggestions from another similar review)

https://gist.github.com/angelospk/91ec6ba091e4e3c751fcc3231815c721

angelospk:
Picked four more items off the audit list and split them into separate PRs, since they touch unrelated parts of the app and are easier to judge one at a time:

- #562 replaces a `find()` nested inside a `map()` in `getSubjectsForMeeting` with a Map lookup.
- #563 moves `getTimestampBounds` off `Math.min(...arr)`, which throws outright on very long meetings, and deletes a per-render speaker scan that turned out to be unused.
- #564 memoizes the filtering in `List`, stabilizes the `filter` callbacks its two callers pass in, and stops `CityCard` from requesting a `100vw` logo.
- #565 keeps `mapbox-gl` out of the initial bundle on the pages that show a map conditionally or not at all, and stops shipping `mapbox-gl-draw` to people who never open edit mode.

One thing worth flagging from #565: the landing page still loads mapbox eagerly. The marker and popup hooks import `mapboxgl` as a value for `new Marker()` and `new Popup()`, so a dynamic import of the map component alone does not move the library out of that chunk. Getting it out means moving markers and popups behind the same lazy boundary, which I would rather do as its own PR than bolt onto this one.

#377 covered the parallel-awaits item earlier.

PR #563, opened 2026-07-23
perf(transcript): memoize hot paths in the meeting transcript UI

angelospk:
Part of #242.

Two hot paths on the meeting transcript page.

`getTimestampBounds` used `Math.min(...timestamps)` / `Math.max(...)`. On a long meeting that spreads thousands of arguments onto the stack every time it runs, and past roughly 100k entries it throws instead of returning a number. It moved to `src/lib/utils/timestamps.ts` as a plain loop. Tests cover the empty array, a single utterance, unsorted input, NaN propagation (so it keeps matching `Math.min`/`Math.max` semantics), and a 250k-utterance array that the spread version could not handle.

`TranscriptControls` rescanned every speaker segment on each render to compute `currentSpeaker`. That value turned out to be dead: nothing renders it, and the hover tooltip reads `hoveredSpeaker`, which is maintained separately by the mouse and touch handlers. So it is deleted rather than memoized. Thanks to Greptile for catching that the first version of this PR memoized a computation that nobody consumed.

### Testing

`npm test` passes (1028 tests, 5 new). Typecheck clean.

Unrelated note: `utils.test.ts` has a flaky case, `klitiki > should handle mixed casing`, that failed once across my runs on both this branch and `main` and passes on re-run. It is untouched by this PR.

<h3>Greptile Summary</h3>

This PR removes an unused transcript speaker scan and adds a stack-safe timestamp bounds utility.
- Replaces timestamp array expansion with a single-pass bounds calculation.
- Adds timestamp utility tests for empty, unsorted, invalid, and large inputs.
- Removes the unused `currentSpeaker` calculation from `TranscriptControls`.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/meetings/CouncilMeetingDataContext.tsx | Replaces the inline timestamp bounds calculation with the shared utility. |
| src/components/meetings/TranscriptControls.tsx | Removes the unused current-speaker calculation without affecting the separate hover and touch tooltip path. |
| src/lib/utils/timestamps.ts | Adds a single-pass timestamp bounds calculation that avoids argument-count limits. |
| src/lib/utils/__tests__/timestamps.test.ts | Tests the timestamp utility across empty, unsorted, invalid, and large inputs. |

<sub>Reviews (5): Last reviewed commit: ["refactor(transcript): drop unused curren..."](https://github.com/schemalabz/opencouncil/commit/aceff47b734f6c4a1574eaf6b0889db88cfb71f5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46688183)</sub>

### DIFF
--- a/src/components/meetings/CouncilMeetingDataContext.tsx
+++ b/src/components/meetings/CouncilMeetingDataContext.tsx
@@ -8,6 +8,7 @@ import { Transcript } from '@/lib/db/transcript';
 import { MeetingData } from '@/lib/getMeetingData';
 import { PersonWithRelations } from '@/lib/db/people';
 import { getPartyFromRoles } from "@/lib/utils";
+import { getTimestampBounds } from '@/lib/utils/timestamps';
 import type { HighlightWithUtterances } from '@/lib/db/highlights';
 
 // Actions are mutations + getters that don't depend on transcript identity.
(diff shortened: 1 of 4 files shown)


angelospk:
Pushed cdf88d2.

Greptile was right that the memoized `currentSpeaker` never gets rendered. I traced it and the value is unused everywhere: the tooltip reads `hoveredSpeaker`, which the mouse and touch handlers set directly. So I removed the scan instead of memoizing it, which is the better version of the change. The `getTimestampBounds` part is unchanged.

PR #564, opened 2026-07-23
perf(list): memoize filtering and add sizes to the CityCard image

angelospk:
Part of #242.

`List` recomputed `selectedFilters` and `filteredItems` on every render, including renders that had nothing to do with the filters. Both are memoized now. `filterAvailableValues` also gets a stable default object, otherwise the fresh `{}` literal on each render would invalidate the memo immediately.

That is only half of it: `CityPeople` and `CityMeetings` passed inline arrow functions as `List`'s `filter` prop, so the prop identity changed on every parent render and the new memo would have been invalidated anyway. Both callbacks are `useCallback`ed on `resolvedBodyId`, which is the only value they close over.

Separately, the `CityCard` image uses `fill` without a `sizes` prop, so Next.js assumes `100vw` and picks a source much wider than the card is ever rendered. `sizes` is now a prop that defaults to the `CitiesList` grid geometry. `PilotPage` passes its own, because cards stack to the full container width there below `lg` and the shared default would have understated the slot by half.

### Testing

`npm test` passes (1023 tests). Typecheck clean.

<h3>Greptile Summary</h3>

Memoizes list filtering and selected-filter derivation, stabilizes city meeting and people filter callbacks, and adds layout-aware responsive sizing metadata to city-card images.
- Uses `useMemo` to avoid unnecessary filter recomputation.
- Uses `useCallback` for stable meeting and person filter identities.
- Adds a configurable `logoSizes` prop with defaults for the cities grid and a PilotPage override.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/List.tsx | Memoizes selected filters and filtered items while providing a stable default filter-values array. |
| src/components/cities/CityMeetings.tsx | Stabilizes the existing meeting filter callback without changing its filtering behavior. |
| src/components/cities/CityPeople.tsx | Stabilizes the existing people filter callback without changing its filtering behavior. |
| src/components/cities/CityCard.tsx | Adds responsive image sizing metadata configurable by each card layout. |
| src/components/PilotPage.tsx | Supplies CityCard image sizing metadata tailored to the page’s responsive layout. |

<sub>Reviews (5): Last reviewed commit: ["perf(list): stabilize filter callbacks a..."](https://github.com/schemalabz/opencouncil/commit/db972e8ffacb2e0c94b965346d10ab392014a954) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46688201)</sub>

### DIFF
--- a/src/components/List.tsx
+++ b/src/components/List.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState, useEffect, useRef } from 'react';
+import React, { useCallback, useMemo, useState, useEffect, useRef } from 'react';
 import { useSearchParams } from 'next/navigation';
 import FormSheet from './FormSheet';
 import { Search, ChevronLeft, ChevronRight } from "lucide-react";
@@ -39,6 +39,10 @@ interface ListProps<T, P = {}, F = string | undefined> extends BaseListProps {
     renderAfterFilters?: React.ReactNode | ((selectedValues: F[]) => React.ReactNode);
 }
 
+// Stable default so an absent `filterAvailableValues` prop doesn't produce a
+// fresh array identity on every render (which would defeat memoization below).
+const EMPTY_FILTER_VALUES: never[] = [];
+
 export default function List<T extends { id: string }, P = {}, F = string | undefined>({
     items,
     editable,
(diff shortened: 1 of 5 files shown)


angelospk:
Pushed 1db04b3, covering both review comments.

The inline `filter` callbacks in `CityPeople` and `CityMeetings` would have invalidated the new memo on every render, so the original change did nothing for those two callers. Both are `useCallback`ed on `resolvedBodyId` now.

On `sizes`: the `PilotPage` layout really is different, cards stack to the full container width below `lg` there. One string cannot serve both layouts, so `sizes` became a prop defaulting to the `CitiesList` geometry, with `PilotPage` passing its own.

PR #573, opened 2026-07-27
feat(landing): iteration-2 follow-ups — guide hint, ranked-list label, bubble clusters, pin morphs, Δήμοι petition layer + leaderboard

christosporios:
Six independent-ish improvements to the landing redesign (iteration 2), from user testing and design review. **Each commit is self-contained and written to be cherry-picked** — see the dependency notes below.

## The commits

1. **`feat(landing): point first-time visitors to the "?" guide`** — user testing showed the guide goes unnoticed. Until the first open (localStorage) and for signed-out visitors only, the "?" gets a solid orange fill and a "Τι είναι αuτό;" label — in the rail flow on desktop, a bubble above the button on mobile, gated so it never covers other UI.
2. **`feat(landing): name the subject list's ordering and explain it`** — a "Πρόσφατα πολυσυζητημένα" label + on-demand popover (tap-friendly) explaining the ranking; subtitle on desktop, floating pill over the map on mobile (zero layout cost). Hidden during text search. Fires `ranking_explain_opened`.
3. **`feat(landing): pack overview donuts into live bubble clusters`** — each pile of overlapping municipality donuts becomes a bubble cluster: busiest δήμος anchored full-size on its real spot, the rest at 75% packed *touching* it on their true azimuths, second ring only when the first is full. Markers are modelled as their real ink ellipse (exact Minkowski contact). Layout recomputes every frame of a zoom (translation-invariant, so pans are free) with per-frame hysteresis + CSS transitions so nothing dithers. 20 unit tests.
4. **`feat(landing): morph subject pins dot↔badge in place, live during moves`** — the dense-viewport dot mode now morphs each pin in place (size/colour transition, icon opacity fade, "+N" badge scales away) instead of tearing the layer down on moveend, evaluated per frame with an exit threshold below entry so it can't flap. Icon React roots still mount lazily (all-dots viewports keep costing zero roots).
5. **`feat(landing): Δήμοι-view map with logo bubbles and a petition layer`** — the Δήμοι view swaps the map to plain logo bubbles (same packing engine), plus out-of-network municipalities with ≥10 petitions: boundary shaded in brand blue by (log-scaled, distribution-relative) intensity, logo marker with a matching ring, and the "request it" preview saying "N+ κάτοικοι το έχουν ζητήσει" (coarse buckets 10+/25+/50+/100+).
6. **`feat(landing): petitioned-δήμοι leaderboard in the Δήμοι lists`** — thin ranked rows (desktop) / cards (mobile strip) under the δήμος cards, plus the dynamic tail "και N ακόμα δήμοι με λιγότερα από 10 αιτήματα" (ICU plural, el/en/fr).

## Cherry-picking notes

- Commits **1, 2, 3, 4** are mutually independent (3 and 4 touch different parts of `useMapMarkers.tsx`; picking one without the other may need a trivial conflict resolution there).
- Commit **5 requires 3** (packing engine + marker shell).
- Commit **6 requires 5** (petition data + click handler).

## Privacy (commits 5–6)

Enforced invariants, documented in `src/lib/landing/petitions.ts`:
- Petitioner data (user ids, names, emails, timestamps) never leaves the server — the SQL aggregates with `COUNT` only.
- Which municipalities have <10 petitions is never revealed — they ship as one aggregate integer.
- Exact counts are not recoverable client-side: coarse buckets + colour intensity quantised to 0.1 steps (a precise intensity would be invertible).
- Verified empirically by grepping the full server-rendered payload for seeded below-threshold city ids and petitioner artifacts (zero hits).

## Testing & verification

- `tsc` clean, 1,106 unit tests green (includes new suites for the packing engine and petition helpers).
- Behaviour verified in-browser on desktop + mobile viewports, including frame-by-frame instrumentation: tangency held within 1px across zoom gestures, dot↔badge morphs animate mid-gesture, no layout dithering (max 1.1px/frame movement).
- Perf/DRY review pass applied: shared repack listener, shared leaderboard ordering, early-exit viewport counting, parallelised petition queries.

## Notes for reviewers

- **Staging has seeded demo data**: fake users `petition-test-N@test.invalid` + petitions (`ptest-*` ids) giving Αγία Παρασκευή 25, Αιγάλεω 13, and 23 municipalities with 1–9 petitions. Remove with `DELETE FROM "Petition" WHERE id LIKE 'ptest-%'; DELETE FROM "User" WHERE id LIKE 'ptest-%';`
- The petition payload is cached with a 1h TTL (`cities:petitioned-map`), so petition changes show up within the hour.

---

> [!NOTE]
> **Medium Risk**
> Touches core landing map behavior, new cached petition aggregation, and privacy-sensitive public counts; well-tested but regressions in focus/tab switching or marker performance are possible.
> 
> **Overview**
> **Landing iteration 2** bundles several map and Δήμοι-tab improvements: first-time **guide** emphasis, a **ranked list** label with explainer, and richer **el/en/fr** copy (including petition strings).
> 
> On the **Δήμοι** view, the map switches to logo bubbles (with the same **live bubble-cluster packing** as zoomed-out count donuts) plus **petitioned** municipalities (≥10 aggregated petitions): blue boundary shading, markers, and a preview showing coarse **N+** buckets—not exact counts. The home page loads this via **`getPetitionedMapCitiesCached`**; **`clearMapFocus`** keeps petition previews from stacking when switching tabs or selecting subjects. Desktop and mobile **leaderboards** list ranked petitioned δήμοι with overflow and below-threshold aggregate lines.
> 
> **Subject markers** now **morph** between dots and badges during pan/zoom (hysteresis via **`nextDotMode`**), instead of rebuilding the layer on `moveend`. Municipality donuts use **ellipse-aware packing** with live repack on zoom and **`.oc-packed-marker`** CSS for click hit-testing.
> 
> **`messages/fr.json`** is largely reformatted (indentation); **`EmbedConfigurator` / `EmbedWidget`** gain location and hot-subjects strings in that pass.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21e227b1e9df92fe2bb26f28ebc2bcc0c885635c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

Adds landing-page guidance and ranking explanations, live marker clustering and pin morphing, plus a privacy-preserving municipality petition map and leaderboard.
- Introduces packed municipality marker layouts with live zoom repacking.
- Morphs subject markers between dots and badges without rebuilding the layer.
- Adds server-aggregated petition buckets, map shading, markers, previews, and municipality leaderboards.
- Extends Greek, English, and French landing translations and translation consistency tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/landing/v2/LandingV2.tsx | Coordinates municipality-map mode, petition focus state, marker layers, and layout data without an eligible follow-up issue. |
| src/components/landing/v2/hooks/useMapMarkers.tsx | Adds live dot-mode transitions and shared packed-marker repacking for overview and municipality layers. |
| src/components/landing/v2/mapMarkers.tsx | Adds morphable subject pins and animated packed marker shells; the generated donut SVG contains fixed geometry and topic colors rather than user-facing text. |
| src/lib/landing/markerDeclutter.ts | Replaces overlap nudging with deterministic, hysteresis-aware bubble-cluster packing. |
| src/lib/db/cities.ts | Aggregates petition demand using count-only SQL and returns only thresholded cities plus one anonymous below-threshold total. |
| src/lib/landing/petitions.ts | Converts petition totals into coarse display buckets and map intensities while excluding raw counts and petitioner fields. |
| src/components/landing/v2/hooks/useMapFeatures.ts | Adds petition-intensity municipality shading and focused petition boundary styling. |
| src/components/landing/v2/MunicipalitiesList.tsx | Adds desktop and mobile petition leaderboards using bucketed demand data. |

<sub>Reviews (8): Last reviewed commit: ["feat(landing): petitioned-δήμοι leaderbo..."](https://github.com/schemalabz/opencouncil/commit/21e227b1e9df92fe2bb26f28ebc2bcc0c885635c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47571622)</sub>

**Context used:**


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/2a621145699a859440588588fd6a649736a9e8d4..200a6eff679058ccd8e46fa764ecc2e8286a92d7) to fold a mobile-leaderboard revision into the leaderboard commit (kept as 6 clean cherry-pickable commits):

- The mobile strip now shows the petitioned-δήμοι leaderboard as **one card** (header · ranked rows · tail line) instead of one thin card per δήμος.
- The card caps at the top 5 rows; longer rankings fold into an honest overflow line ("και N ακόμα δήμοι με 10+ αιτήματα") — no nested vertical scroll inside the horizontal strip.
- Removed the now-dead `municipality.petitionsShort` message key.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/200a6eff679058ccd8e46fa764ecc2e8286a92d7..81eed702bc32d6e4f09f50bb561c26586f52319b) two design revisions, folded into their commits (still 6 clean cherry-pickable commits):

**Δήμοι-view markers** (into the petition-layer commit):
- Cooperating municipalities' logo bubbles now carry a brand-orange ring (this view only).
- Petitioned bubbles are visually subordinate: rendered at 0.8× and well faded (0.65 opacity) under the cooperating ones — the map reads "these are in OpenCouncil, those are asked-for" at a glance.

**Mobile leaderboard card** (into the leaderboard commit):
- The strip now bottom-aligns cards at their natural heights (`items-end`, matching the subjects strip), so the leaderboard card is the only one that grows — the δήμος cards no longer stretch to match it.

kouloumos (dismissed):
Update on this one, because it took a turn.

**Commits 1–2 are on `main`.** **Commits 3–4 landed, broke donut clicking in production, and have been reverted off both `main` and `production`.** This PR should now carry 3–6.

## What's on main

Rebased onto 52913128, so new SHAs — and **neither is patch-identical to what's on this branch**:

| here | on main | folded in |
|---|---|---|
| ca476f06 | c0d928da | `localStorage` moved behind guarded `readInfoHintSeen`/`markInfoHintSeen` in `savedView.ts`; key → `oc:landing:info-hint-seen`; guide button's `aria-label` dropped while the hint shows |
| 47dbc9de | 09bdc234 | desktop ranking label gated on `!loading && count > 0` |

The `localStorage` one: bare access throws on iOS Safari with cookies blocked, and it sat in a mount effect, so it took the whole landing to an error boundary. `savedView.ts` already guards its own access and documents why, so the two flag helpers went next to it.

## Why 3–4 came back off

Zoomed out, some δήμος donuts stopped being clickable — the click fell through to the map and opened the out-of-network municipality underneath instead of zooming in.

**Mechanism.** Commit 3 moved the cluster displacement off Mapbox's `Marker({ offset })` onto a CSS transform on an inner wrapper, so it could transition and be re-set per frame:

```js
// before
new mapboxgl.Marker({ element: rootEl, offset })
// after
new mapboxgl.Marker({ element: rootEl })   // + offsetEl.style.transform = translate(...)
```

Mapbox's `offset` moves the marker's own element — box and hit area together. An inner transform moves only what's *painted*. `rootEl`'s 68px box stays anchored at the δήμος centroid, invisible, where the donut is no longer drawn. Markers are appended into the map's canvas container (`getCanvasContainer().appendChild`) and `.mapboxgl-marker` sets no `pointer-events`, so a click landing on that box finds no handler to call `stopPropagation()`, bubbles to the map's own click handler, and runs the `/api/cities/at` lookup.

Confirmed live before the revert — every line here is a click-swallowing box, with the donut drawn 15–52px away:

```
dead zone at 659 397 → donut actually at 696 378
dead zone at 658 398 → donut actually at 663 359
dead zone at 598 397 → donut actually at 579 420
dead zone at 604 419 → donut actually at 617 412
dead zone at 657 401 → donut actually at 703 425
```

The boxes cluster at the satellites' *true* centroids — which is exactly where the anchor sits — so the busiest δήμος of each cluster was the least clickable one.

**Status of the fix.** The obvious patch is `pointer-events: none` on the root and `auto` on the drawn button. We tried it and it did **not** resolve the symptom in local testing — though we never established whether the built bundle actually contained it, so treat that as inconclusive rather than as evidence the analysis is wrong. Rather than ship a second unverified patch to production, we reverted.

**Suggested direction.** The pre-release code was structurally immune because Mapbox moved the whole element. The repack already runs per frame, so keeping `Marker.setOffset()` and interpolating the offset in the existing rAF loop would give both the smooth motion *and* a hit area that's correct by construction — rather than a CSS transition that has to be chased with `pointer-events` patches. Whatever the approach, the invariant to hold is: **anything Mapbox positions must have its hit area coincide with what's drawn.**

Two things worth knowing before you rebuild it:

- **Commit 4 was collateral, not faulty.** It only came off because it shares `useMapMarkers.tsx` with 3 and the revert wouldn't separate cleanly. I checked the pins specifically: `stylePin` sets the button's own `h-3`/`h-7`/`h-9`, so the box tracks the visual, and the co-located `+N` badge already gets an explicit `pointer-events-none` when scaled away. No equivalent problem.
- **`packedMarkerShell` in 18d577c4 is this same structure**, applied to two more marker types. Since it's already the shared extraction, it's the right single place to get this right.

## The anchor bug in commit 3

Separate from the above, and still real. `HOME_RETURN_MARGIN` is applied to every marker, but the component cut only clears anchors of each other at exactly `s`. An anchor in the band `[2, 2.25)` that was a satellite the previous frame — whichever marker's pile just split as you zoom in — failed the margin check, fell into the ring branch, and got placed on a ring around its own centre (`a === i`, so `relX = relY = 0`):

```
pts [{0,0},{2.1·rx,0}], priorities [9,5], memory[1] = {home:false}
→ placements[1] = { offset: {x:-25.18, y:-49.38}, scale: 1 }
```

A full-scale donut ~55px off its δήμος, latching, because it kept reporting `home: false`. A 60-frame zoom sweep over an 8-δήμος Attica pile hit it on 6 frames.

The margin is a satellite rule and should never reach an anchor — `markerDeclutter.ts`:

```diff
-        const wasOnRing = memory?.[i]?.home === false;
+        const wasOnRing = !anchors.has(i) && memory?.[i]?.home === false;
```

and a test that fails against the current code and passes with it:

```js
it('never applies that margin to an anchor, so a pile that splits does not fling one', () => {
    // Two δήμοι at normalised distance 2.1: far enough apart to be separate piles (the cut is
    // at 2), so both anchor at full size — but inside the 2.25 the return-home margin would
    // ask for. Applying it here drops the second δήμος into the ring branch, where its own
    // anchor is itself, flinging it ~55px off its real position at full scale — and latching,
    // because it keeps reporting home:false.
    const pts: Point[] = [
        { x: 0, y: 0 },
        { x: 2.1 * EXTENT.rx, y: 0 },
    ];
    const split = packMunicipalityMarkers(pts, [9, 5], EXTENT, SCALE, [
        { home: true, side: 0 },
        { home: false, side: 1 },
    ]);
    expect(split.placements[1]).toEqual({ offset: ZERO, scale: 1 });
    expect(split.memory[1].home).toBe(true);
});
```

This is pure geometry, independent of how the markers end up positioned on screen — so it holds whichever way the hit-area question gets resolved.

## Rebasing

Commits 1–2 aren't patch-identical to what landed, so a plain rebase will replay them and conflict. Replay 3–6 onto the new main:

```
git rebase --onto upstream/main 47dbc9de feat/landing-v2-tweaks
```

## Scope for round 2

**Commit 3 — packing**
- The hit-area problem above. Foundational; everything else in this list is secondary to it.
- The anchor fling, fix above.
- The ±π ring sweep dropped the old `MAX_BEARING_DEVIATION = π/6` invariant without replacement, so a satellite can settle on the *opposite* side of its own δήμος. It's also what makes `PackMemory.side` necessary at all — a bounded deviation would make that half of the hysteresis near-vestigial. Worth making that trade deliberately rather than silently.
- Repack is ~O(n²): measured 0.6ms at n=10, 9.2ms at n=60, 18.1ms at n=80 — past ~70 markers a single repack exceeds the frame budget, and it fires on every zoom tick.
- The packing reserves a 48×58 ink ellipse but the button box is 68×68, so two correctly-tangent satellites still have clickable boxes overlapping by ~15px.
- Hysteresis is only pinned at a single frame with hand-built memory, never as a gesture feeding memory forward — which is the only way the dithering property can actually fail.
- `markerDeclutter.test.ts` hardcodes `EXTENT = {rx:24, ry:29, cy:6}` rather than deriving it from the donut constants, so changing `MUNICIPALITY_DONUT_LOGO_SIZE` silently desyncs the whole suite from what ships.

**Commit 4 — pins.** Nothing outstanding beyond `countInView` fabricating `{clng: 0, clat: 0}`, and the dot↔badge thresholds having no test (lifting the decision into `landingCore` as `nextDotMode(n, dot)` would make it a five-line one).

**Commits 5–6 — petition layer.** Blocking:

- **The leaderboard's ordinals don't match the counts.** `rankPetitionedCities` sorts on `intensity`, quantised to 0.1 before it leaves the server — so ties spanning a 1.585× range in true count fall through to alphabetical order, under explicit `1.` `2.` `3.` numbering. Executed: counts `Ωρωπός 120, Αθήνα 100, Βόλος 1000` render as `1. Βόλος · 2. Αθήνα (100) · 3. Ωρωπός (120)`. The query also has no `ORDER BY`, so map anchor tie-breaks follow arbitrary row order — a third ordering for the same data.

  Fix is `ORDER BY COUNT(p.id) DESC, c.name` and ranking on server order. Worth saying why the quantisation is there at all: it serves the third privacy invariant in `petitions.ts` ("exact counts are not recoverable"). But we already publish a coarse bucket against a *named* δήμος — "Δήμος X: 25+" — so exact counts aren't something we're keeping secret, and that invariant is buying nothing while costing a visibly wrong ranking. I'd drop invariant 3 from the docstring along with the fix, so nobody restores the quantisation later believing it's load-bearing.

- **`petitionFill`'s white-text switch fails WCAG AA.** `pct >= 55` isn't the crossover. Measured contrast against white is 2.91 / 3.29 / 3.79 / 4.40 at intensity 0.4 / 0.5 / 0.6 / 0.7, and the badge is 12px so 4.5:1 applies, not 3:1. No switch point rescues the white ↔ `#1c2a3a` pair — both sit at ~3.8:1 around pct 70. `pct >= 85 ? '#ffffff' : '#000000'` passes all 11 steps, worst case 4.78:1.

- **The copy says "residents"; the query counts everyone.** `PetitionFormStep` accepts citizen-only petitions (`!isResident && !isCitizen` is the only rejection), and a δημότης may live elsewhere — but neither query filters `is_resident`. Three shipped strings assert it: `municipalityTooltip.petitions`, `marker.petitionedAria`, `municipality.petitionedTitle`. Either filter both queries or drop the qualifier — the tail lines already use the neutral "αιτήματα".

- **The three open Bugbot findings are all still live** — the mobile strip's `return null` landing before the leaderboard and CTA, and the two focus-exclusivity ones. There's a fourth of the same class: `trackedSetView` doesn't clear `clickedMunicipality` on tab switch, so the petition-blue fill and the "request it" popup leak into the Θέματα map with no petition layer around them. Probably one `focusMunicipality(kind, id)` owning exclusivity rather than four separate patches.

<details>
<summary><b>Worth fixing (5–6)</b></summary>

- The two petition queries disagree on `geometry IS NOT NULL` — query 1 requires it, query 2 doesn't but is capped at `< threshold`. A δήμος with ≥10 petitions and no geometry lands in neither the leaderboard nor the tail.
- The below-threshold tail renders inside a block gated on `cities.length > 0`, so it's invisible in exactly the young-deployment case where it's the only signal available.
- Petition geometries ship uncapped in every landing first-paint payload, though only the Δήμοι view reads them. Grows with adoption.
- `MUNICIPALITY_VIEW_EXTENT` doesn't account for the 4px ring spread or `baseScale: 0.8` — cooperating rings overlap by ~5px, petitioned bubbles are held ~3px apart and never actually touch.
- `captureLandingAction('petitioned_municipality_opened')` fires before the `!city.geometry` guard, so dead clicks count as opens.
- Leaderboard rows' `aria-label` uses `nameMunicipality` while the visible label is `name` — WCAG 2.5.3, breaks speech activation.

</details>

<details>
<summary><b>Structural, if you want them</b></summary>

- The leaderboard is two implementations (`MunicipalitiesList` / `MobileLayout`), ~35 near-identical lines. `rankPetitionedCities` was centralised precisely so both surfaces rank identically — then the rendering that depends on it was forked.
- Same shape in `useMapMarkers`: `attachLiveRepack` was extracted, but the `packMemory` feedback protocol is still hand-written in both packed layers.
- `petitionFill` reimplements what `topicStyle.ts` exists for (including its `contrastText`), and introduces a second near-black constant. "Petition blue" is five literals in three spellings across `petitions.ts`, `mapMarkers.tsx` and `useMapFeatures.ts`.
- `PetitionedMapCityRow` re-spells `MapCityRow` field-for-field, and the centroid + `ST_SimplifyPreserveTopology(…, 0.001)` projection is copy-pasted between the two readers — in a file that already has `cityCoversPoint` for exactly this reason.

</details>

<details>
<summary><b>Tests (5–6)</b></summary>

Untested and worth pinning: `rankPetitionedCities` (which carries the "both surfaces rank identically" claim), `petitionFill` (which carries the contrast decision), and the row → payload mapping in `getPetitionedMapCitiesCached` — a key-set assertion there is a direct encoding of the "petitioner data never leaves the server" invariant, and it fires the day someone widens the `SELECT`.

Separately: `translations.test.ts` only walks `messages/en` and `messages/el`, so nothing guards `fr` at all — a missing key or broken ICU plural there would still report green. Pre-existing, but this PR's fr keys ride on it.

</details>

---

For the record, the two privacy invariants that carry real weight hold end-to-end: no petitioner column — user id, name, email, per-petition timestamp — is selected in either query, and below-threshold δήμοι leave the server as a single aggregate integer, never as a list. I traced the full server → client payload and nothing beyond the nine declared fields crosses.

The clustering itself is worth getting back — it reads far better than the old spread. It just needs a positioning mechanism that doesn't put the hit area and the pixels in different places.

christosporios:
Rebased as prescribed: `git rebase --onto origin/main 47dbc9de` — the branch now carries exactly commits 3–6 on top of main (which has the fixed-up 1–2). Prisma client regenerated for `allowFeedbackCalls`; tsc clean, 1114 tests green. Round-2 rework of the review findings starts from here — plan in the works, the hit-area invariant first.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/4926020145a0aab2bc5c29441401fbe466b71be8..81a4e207deb33dad5ff6ea17fc772511db6365c5) to fold the round-2 review fixes **into the four commits themselves** — each commit now carries its own fixes, so cherry-picking any prefix stays safe (notably: the hit-area architecture lives in the pack commit, not in a fixup at the tip). Commit messages were updated to match. Point-by-point against the review:

**Pack commit (`feat(landing): pack overview donuts into live bubble clusters`)**
- **Hit-area architecture**: the packing displacement now goes through `Marker.setOffset()` (box + paint move together), animated by a self-terminating rAF easing loop — never a CSS translate on an inner wrapper. The invariant is documented at `packedMarkerShell`. A second click-swallower found while verifying: scaled satellites' wrappers keep their unscaled layout boxes, so all wrappers are now pointer-transparent (via an `!important` class, because mapbox-gl re-sets inline `pointer-events` during occlusion updates) and only the drawn button captures clicks. The marker element is now exactly the donut's ink box, so neighbouring tangent donuts' boxes can't overlap. Verified with an `elementFromPoint` grid probe over every marker — zero dead zones, both settled and mid-animation.
- **Anchor fling**: fixed — `HOME_RETURN_MARGIN` hysteresis now applies to satellites only (`!anchors.has(i)`), so a freshly-split anchor can't latch onto a ring around itself. Your failing-first test is in `markerDeclutter.test.ts` verbatim, plus a gesture-level sweep (zoom 1→3→1 in 120 steps, memory threaded forward): per-marker home transitions ≤ 2, side flips = 0. The suite's `EXTENT` is now derived from the real donut constants instead of hardcoded numbers. 22 packing tests total.
- **±π sweep**: kept, per Christos — the full sweep is what guarantees one-ring packing; the side-hysteresis exists *because* the sweep is unbounded. Now documented as a deliberate trade right at the sweep code.
- **Repack cost**: the ~O(n²) repack (~9 ms at n=60) is time-gated to 100 ms during gestures, with an ungated `moveend` pass so every gesture settles on an exact layout; the shells ease toward the latest targets at frame rate in between.

**Pins commit** — the dot↔badge decision is extracted as pure `nextDotMode()` with unit tests, and `countInView` takes a bounds-only viewport type (no fabricated centre) and early-exits at the entry threshold.

**Δήμοι-view commit** — `PETITION_BLUE` defined once for every surface (map paints get comma-syntax hsl for mapbox's parser); the view-marker extent now includes the ring shadows (`RING_SPREAD`), so "touching" bubbles no longer overlap rings; petition copy is neutral in el/en/fr (petitions aren't verified residents): "Αιτήματα προσθήκης" / "{count}+ αιτήματα για να προστεθεί". Focus exclusivity: a single `clearMapFocus()` owner runs before every focus-opening path and on view switches, so petition previews can't stack or survive tab changes (verified live in the browser).

**Leaderboard commit**
- **Ranking**: moved into the query (`ORDER BY COUNT(p.id) DESC, name`) — the server order *is* the ranking; client-side `rankPetitionedCities` is gone.
- **Quantisation**: dropped, per Christos — exact-count secrecy is now explicitly documented as **not** an invariant (we already publish a bucket against a named δήμος; quantised intensities made leaderboard ordinals contradict the badges). The two real invariants (no petitioner data ever; below-threshold δήμοι only as one aggregate integer) are documented in `petitions.ts` and enforced by `buildPetitionedCities`, a pure field-by-field mapping whose payload key set a test pins — a widened SELECT or a stray spread can't silently ship more.
- **WCAG**: the ramp's text flip now sits at 85% where white actually clears 4.5:1 (worst case 4.78:1) — asserted by a test that computes contrast across the whole ramp.
- **Punch list**: boundary-less ≥10 δήμοι now reach the leaderboard as disabled rows (no map bubble, no dead click — the guard runs before analytics); the tail line renders even when no δήμος clears the threshold; row aria-labels use the visible name; shared `PetitionedRow` for desktop/mobile; `translations.test.ts` is now en-reference × {el, fr} including modular files (it caught 16 missing `fr.json` keys, now filled).

`npx tsc --noEmit` clean and 1,135 tests green on the tip, and each intermediate commit type-checks on its own.

src/components/landing/v2/mapMarkers.tsx
```diff
@@ -275,95 +287,230 @@ export function createSubjectPin(
...
+    const shell = packedMarkerShell(offset, scale);
     const el = document.createElement('button');
     el.type = 'button';
     el.setAttribute('aria-label', t('marker.municipalityZoom', { name: muni.nameMunicipality, count: muni.count }));
     el.className =
         'relative block cursor-pointer border-0 bg-transparent p-0 leading-none transition-transform hover:scale-110';
-    el.style.width = `${MUNICIPALITY_DONUT_DIAMETER}px`;
-    el.style.height = `${MUNICIPALITY_DONUT_DIAMETER}px`;
+    el.style.pointerEvents = 'auto'; // the drawn box is the click target (wrappers are transparent)
+    el.style.width = `${DONUT_INK_WIDTH}px`;
```
kouloumos, line 501: Follow-up, not a merge condition.

The clickable element is a rectangle. The packer calculates the space between two markers as ellipses. Two markers that touch as ellipses still have rectangles that overlap. The overlap area is on top of the logo of the other marker.

I tested this on the preview. A click on the logo of Δήμος Αθηναίων goes to the handler for Δήμος Βριλησσίων. In the image, the dashed lines are the two clickable areas. The magenta points are the test points that go to the wrong marker.



https://github.com/user-attachments/assets/34c435a6-b0b1-485e-a40b-d0361dc30f05

The affected area is small:

- It is a band of 2 px to 7 px at the edge of the logo.
- It is approximately 0.1% of the test points on the logo.
- It occurs only when the markers form a cluster.

This is much smaller than the markers that did not respond before the rollback. The user must click the extreme edge of the logo.

`border-radius: 50%` on `el` corrects it. The browser applies border-radius to hit-testing. 50% of 48×58 gives `rx: 24, ry: 29`, and that is the same as `MUNICIPALITY_MARKER_EXTENT`. The clickable area then becomes the same shape that the packer uses. `createMunicipalityViewMarker` already uses `rounded-full`, and it does not have this problem.

If you make this change, please check the count text. The ellipse gives ±16.6 px at that height, and three digits need approximately 12.5 px. The text must stay clickable.

src/lib/landing/petitions.ts
```diff
@@ -0,0 +1,127 @@
...
+ *  Shared by the map bubbles and the Δήμοι-list leaderboard so "how blue" means one thing. */
+export function petitionFill(intensity: number): { background: string; text: string } {
+    const pct = Math.round(25 + 75 * intensity);
+    return {
+        background: `color-mix(in srgb, ${PETITION_BLUE.deep} ${pct}%, ${PETITION_BLUE.pale})`,
+        // The badges are 12px text, so WCAG AA wants 4.5:1. Neither white nor near-black clears
+        // that across the whole ramp with a mid-ramp switch — white only reaches 4.5:1 from
+        // pct ≈ 85, so that's where the flip lives (worst case 4.78:1, measured; see the
+        // contrast test).
+        text: pct >= 85 ? '#ffffff' : '#000000',
```
kouloumos: Follow-up. This is a future problem, not a current one.

The value `85` is four steps too high. The text stays black at pct 82, 83, and 84. The contrast at those three steps is 4.429, 4.346, and 4.264 to 1. A 12 px badge needs 4.5 to 1.

The image shows the full ramp. I rendered it in the browser with the real `color-mix`. The failed steps have a red outline.



The value `81` gives a minimum contrast of 4.600 to 1 across the full ramp.

The current data does not reach these steps. There are two petitioned δήμοι, and their badges are at pct 100 and pct 46. The problem starts when more δήμοι have counts that spread across the range.

One correction for the comment above: the value `4.78:1` is the contrast at pct 78. That is the nearest step the test measures. It is not the minimum contrast of the ramp.

src/lib/landing/petitions.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        const out = buildPetitionedCities([
+            { ...row, id: 'first', petitions: 1000 },
+            { ...row, id: 'second', petitions: 12 },
+        ]);
+        expect(out.map((c) => c.id)).toEqual(['first', 'second']);
+    });
+});
+
+describe('petitionFill', () => {
+    it('meets WCAG AA (4.5:1 — the badges are 12px text) across the whole ramp', () => {
```
kouloumos: Follow-up, paired with the comment on `petitions.ts`.

`intensity += 0.1` gives pct 25, 33, 40, 48, 55, 63, 70, 78, 85, 93, and 100. These are 11 steps of the 76 possible steps. The failed steps are 82, 83, and 84, and the test does not measure them. `intensity` is continuous, so a loop over the integer values of `pct` measures all of them.

A second point: lines 22-23 repeat the two colours of the ramp instead of reading `PETITION_BLUE`. The docstring of `PETITION_BLUE` says "change it here or nowhere". If someone changes that colour, this test continues to measure the old colour, and it continues to pass. This is the same problem that the `EXTENT` derivation corrected in `markerDeclutter.test.ts`.

src/components/landing/v2/hooks/useMapMarkers.tsx
```diff
@@ -244,16 +266,68 @@ export function useSubjectMarkers({
...
             t: tRef.current,
         });
         pinsRef.current = pins;
+
+        const syncDotMode = () => {
+            const n = countInView();
+            const next = nextDotMode(n, dotMode);
+            if (next === dotMode) return;
+            dotMode = next;
+            pins.forEach((pin) =>
```
kouloumos: This is the smaller of the two regressions in this review, and it is cosmetic.

This code reads `selected` and `intense` from `pin.subject`. But `createSubjectPin` sets `subject: coLocated ? null : rep` (`mapMarkers.tsx:327`). For a "+N" pin, `pin.subject` is always `null`. Therefore both values are always `false`, and the morph gives a selected group the style of an unselected group.

The restyle effect below cannot correct this, because it starts with `if (!pin.subject) return`. The dependencies of this layer are `[mapInstance, visibleSubjects, active]`. There is no viewport dependency, so a pan or a zoom does not rebuild the layer. The highlight stays incorrect until a filter changes.

`buildSubjectPins` calculates the highlight for the full group (`mapMarkers.tsx:661`), so only the morph code path loses it. At `e77ca11d` the dependency array was `[mapInstance, visibleSubjects, dots, active]`. A change of dot mode rebuilt the layer and calculated the highlight again. That rebuild is what this commit removes.

To see it:

1. Open the Θέματα tab and find a pin with a "+N" badge.
2. Tap the pin and select one θέμα from the box. The pin becomes solid and slightly larger.
3. Zoom out until the pins become plain dots.
4. Zoom back in until they become badges again.
5. The pin is now drawn as an unselected pin, and the θέμα card is still open.

This is reasoned from the code and not yet seen in a browser. Athens stays in dot mode at every zoom level I could reach, so a "+N" pin never returned to badge mode there. A δήμος with a subject count near the threshold, such as Χαλάνδρι, should cross it in both directions.

src/components/landing/v2/hooks/useMapMarkers.tsx
```diff
@@ -399,35 +485,232 @@ export function useMunicipalityCountMarkers({
...
+
+/**
+ * Live layout for a packed marker layer: the packing only ever changes with zoom (it's
+ * translation-invariant, and pitch/rotation are disabled on the landing map), so 'zoom' — which
+ * fires every animation frame of a wheel/pinch/easeTo — is the one signal that matters. Repacks
+ * are rAF-coalesced AND time-gated to REPACK_MIN_INTERVAL_MS — the shells' easing carries the
+ * motion between targets — with an ungated 'moveend' pass so every gesture settles on an exact
+ * final layout. Returns the detach function.
+ */
+function attachLiveRepack(map: MapboxMap, reposition: () => void): () => void {
```
kouloumos, line 553: Follow-up.

The docstring says the packing is translation-invariant. That is the reason to use the `zoom` event and not the `move` event. But it is not fully correct for the values that `map.project()` returns.

With integer coordinates, the layout is identical after a pan. With fractional coordinates, a pan moves approximately 60% of the placements by more than 1 px. `clear()` compares the distance against exact tangency and uses no epsilon, and the binary refinement stops at that limit. A difference of one ULP changes the result of the comparison, and the marker moves to a different position.

On the preview the movement is 2 px to 9 px, and the shell animates it over approximately 120 ms. That is small, and I do not recommend a behaviour change for it.

Two smaller effects follow from the same cause:

- `onMoveEnd` has no gate. Every pan runs the full O(n²) calculation for a result that the docstring says is identical. A gate on "the zoom level changed" removes that cost.
- `markerDeclutter.test.ts:245` tests the invariance with integer inputs only, so the test cannot fail. Use fractional inputs with a tolerance, or add an epsilon to `clear()` and keep the exact comparison.

src/lib/__tests__/translations.test.ts
```diff
@@ -14,46 +12,43 @@ function getAllKeys(obj: object, prefix = ''): string[] {
...
+const REFERENCE = 'en';
+const LOCALES = ['el', 'fr'];
+
 function getModularFiles(): string[] {
-    const enDir = path.join(messagesDir, 'en');
-    const elDir = path.join(messagesDir, 'el');
-    const enFiles = fs.existsSync(enDir) ? fs.readdirSync(enDir).filter(f => f.endsWith('.json')) : [];
-    const elFiles = fs.existsSync(elDir) ? fs.readdirSync(elDir).filter(f => f.endsWith('.json')) : [];
-    return [...new Set([...enFiles, ...elFiles])].sort();
+    const files = [REFERENCE, ...LOCALES].flatMap((locale) => {
```
kouloumos, line 20: Follow-up.

The en-reference × {el, fr} change and the modular-directory walk are both correct. A file that exists in one locale and not in `en` makes the test fail, as intended.

The test cannot find an invalid message. It compares key sets, and it does not parse any message. This PR adds ICU plurals in three locales. If a translator removes the `other` arm, this test passes and the page fails at render. I tested that with the new `fr` strings and got `MISSING_OTHER_CLAUSE`.

next-intl already installs `@formatjs/icu-messageformat-parser`. A third test case that gives every message to `parse()` finds this class of error. Please run it against the current messages first, to confirm that they all pass.

Minor: `LOCALES` has the fixed value `['el','fr']`, and `routing.locales` is `['en','el','fr']`. If you read the value from `routing`, a fourth locale cannot stay untested.

src/lib/db/cities.ts
```diff
@@ -213,6 +225,69 @@ export async function getMapCitiesCached(realm: Realm): Promise<MapCityRow[]> {
...
+                prisma.$queryRaw<PetitionedCityQueryRow[]>`
+                    SELECT c.id, c.name, c.name_municipality, c."logoImage", ${CITY_MAP_PROJECTION},
+                           COUNT(p.id)::int AS petitions
+                    FROM "City" c
+                    JOIN "Petition" p ON p."cityId" = c.id
+                    WHERE c."officialSupport" = false
+                      AND c.realm = ${realm}::"Realm"
+                    GROUP BY c.id
+                    HAVING COUNT(p.id) >= ${PETITION_DISPLAY_THRESHOLD}
+                    ORDER BY COUNT(p.id) DESC, c.name ASC
```
kouloumos: Follow-up. Three small items for this query.

**1. The order is not total.** Staging has δήμοι with the same name in the out-of-network set: `Άγιος Νικόλαος`, `Φιλοθέη - Ψυχικό`, `Μακρονήσι`, and `Πρασονήσι`. Each name occurs twice. Two δήμοι with the same count and the same name can change order when the cache refreshes. That changes their position numbers in the list. It also changes which δήμος becomes the anchor of a cluster, because the last tie-break in the packer is the array index. `, c.id` makes the order stable.

**2. The query has no `LIMIT`, and it selects `geometry`.** The landing page sends this data on every first paint, but only the Δήμοι view reads it. The cost is small now, with two rows. The out-of-network boundaries are 4.7 KB on average, and there are 358 of them. If 100 δήμοι go over the threshold, the payload increases by approximately 978 KB. All the city geometry in the current payload is 31 KB. A `LIMIT`, or a separate request when the user opens the Δήμοι tab, solves this.

**3. A question, not a defect.** This query uses `officialSupport = false` to select out-of-network δήμοι. The δήμος cards use `status = 'listed'`. These are two independent columns. A city with `status = 'listed'` and `officialSupport = false` shows as a δήμος card and also as an out-of-network "request it" row. On staging the two columns agree, but no constraint enforces that.

kouloumos (commented):
Reviewed at `81a4e207`. One item needs a decision before this merges. Everything else is a follow-up.

## On the Δήμοι tab, selecting a δήμος no longer shows its subjects



The steps are the same on both sides: open the Δήμοι tab, then click the **Βριλήσσια** card. The view stays on Δήμοι, which is intended — the call site says "filter to it, stay on Δήμοι". On `main` the map then shows the boundary of the δήμος with its subject pins inside it. On this branch the map shows the boundary and one logo bubble. Βριλήσσια has 144 θέματα, and none of them are on the map.

The cause is the new `municipalitiesView` gate on two layers, at `LandingV2.tsx:497` and `LandingV2.tsx:512`:

```
// e77ca11d
active: !showMunicipalityCounts,
// 81a4e207
active: !showMunicipalityCounts && !municipalitiesView,
```

`useSubjectMarkers` and `useGeneralCityMarkers` are both off for the whole Δήμοι tab, at every zoom level. The city-hall marker that carries the θέματα without a location disappears with the subject pins.

The layer swap itself is deliberate, and the description explains it. The part that looks unintended is the effect on the δήμος card. That action was useful because it showed the subjects of the δήμος, and now it shows an empty boundary.

Two more results from comparing the same steps against `main`:

- **There is a way out.** If the user switches to the Θέματα tab, the city filter stays, and the subjects appear. So the data is still reachable. It is not reachable from the tab where the user selected the δήμος.
- **The Δήμοι map also stopped showing the θέματα counts.** On `main` that map used the donut markers, and each δήμος showed its number of θέματα. The logo bubbles do not. The description states this, so I record it as intended. But together with the point above, the Δήμοι map now carries less information than before at every zoom level.

There are at least three ways to resolve it, and all of them are product decisions:

- change the view to Θέματα when the user selects a δήμος card
- keep the subject pins in the Δήμοι view while a city filter is active
- show the subject pins in the Δήμοι view past a zoom level

## The rest of the branch

The round-2 work does what the description says:

- `Marker.setOffset()` runs on every animation frame, so the element box and the painted marker cannot separate. The mechanism that caused the rollback is gone. There are no dead zones at four zoom levels, or after eight pans.
- The anchor-fling guard holds. Reverting `!anchors.has(i)` makes the test fail with the ~55 px offset.
- Server-side ranking, `CITY_MAP_PROJECTION`, the shared `PetitionedRow`, and the payload key-set test are all correct.
- The first-paint payload contains no `ptest-`, `userId`, `@test.invalid`, or `is_resident`. The privacy invariants hold from the server to the client.

The inline comments are follow-up items, not merge conditions. I will open a follow-up issue that collects them, with the screenshots and the older map-highlight behaviour that this PR partly corrects.

Two smaller points.

**1. One cosmetic regression.** A "+N" pin loses its highlight after the map enters dot mode and returns to badge mode. `dots` left the dependency array at `useMapMarkers.tsx:339`, and the rebuild that recalculated the group highlight went with it. The θέμα stays selected and its card stays open. Only the highlight on the map is wrong. The detail, and the steps to see it, are inline at `useMapMarkers.tsx:305`.

**2. The contrast item becomes live without a signal.** The three failed steps of the ramp are unreachable with two petitioned δήμοι. They become reachable when more δήμοι have counts across the range. The test that should find them measures 11 of the 76 steps, and does not measure those three. The detail is inline at `petitions.ts:110`.

One warning about the tests: only the dedicated test finds the anchor fling. The 120-frame gesture sweep passes when the guard is reverted, because it checks `memory` and never checks `placements`. Please do not treat it as second coverage for that bug.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/81a4e207deb33dad5ff6ea17fc772511db6365c5..21e227b1e9df92fe2bb26f28ebc2bcc0c885635c) with the round-3 fixes (rebased on today's `main` first — the Next 16.2.12 / Node 24 bump). As before, each fix is folded into the commit that introduced the problem.

**The merge blocker — decision: keep the subject pins while a city filter is active.** The gates now key on `municipalitiesMapMode = municipalitiesView && filters.cityIds.length === 0`: picking a δήμος (card or logo bubble) hands the map back to the subject layers — pins, city-hall markers, and the zoomed-out count donuts — while the filter lasts, and clearing it returns the bubble map. The tab stays on Δήμοι. Verified on this preview with your exact steps: the Βριλήσσια card now shows its boundary with 17 pins (including a "+N" group) instead of one bubble; toggling the card off returns the 13 bubbles; zooming far out while filtered brings up the count donut. In the Δήμοι-view commit, with its message updated to say so.

**The "+N" highlight regression** — fixed at the root rather than at the call site: `setDot` now takes `(dot, selectedId, previewId)` and the pin resolves its own highlight from its members (it closes over the whole group; `pin.subject` is null for "+N" pins precisely because there is no single subject). The morph now applies the same group-aware rule as `buildSubjectPins`' initial styling, so the two paths can't diverge again. In the pins commit.

**The contrast flip** — moved to pct ≥ 81, and the test now scans every integer step of the ramp (25..100) instead of the 0.1-intensity sample that missed 82–84; measured minimum is now 4.60:1, matching your number. The test also reads the ramp endpoints from `PETITION_BLUE` itself, per your second point — a hue change now re-runs the measurement instead of silently testing the old colour. Thanks for the correction on my "4.78:1 worst case" — that was indeed just the nearest sampled step. In the leaderboard commit.

**The ranking tiebreak** — `ORDER BY COUNT(p.id) DESC, c.name ASC, c.id ASC`, with a comment naming the duplicate-δήμος-names reason. In the leaderboard commit.

On the rest of the inline items — the rectangle-vs-ellipse hit band (nice catch, and the `border-radius: 50%` observation is elegant), the fractional-projection invariance + ungated `moveend`, the ICU `parse()` test case, the payload `LIMIT`, and the `officialSupport`/`status` consistency question — I've left them for your follow-up issue as proposed. On the last one: you're right that nothing constrains the two columns; that wants a decision about which column is authoritative for "in the network" rather than a point fix here.

Also noted your warning that the gesture sweep only checks `memory`, not `placements` — agreed, the dedicated anchor-fling test is the only real coverage for that bug, and I won't count the sweep as a second line of defence.

`tsc` clean and the full suite green (1,158 tests) on the tip; the intermediate commits still type-check individually.

kouloumos (approved):
Approved at `21e227b1`.

**The blocker is fixed and verified.** Selecting a δήμος on the Δήμοι tab returns the subject layers: 17 pins for Βριλήσσια against `main`'s 19 — that gap is data (prod has 151 θέματα, staging 144), not behaviour. Clearing the filter returns the 13 bubbles. Driving the same steps against `main` and this preview now shows only the intended bubble-map difference.

`municipalitiesMapMode` is the right shape for it: one derived value at all five gate sites, so the layers stay mutually exclusive instead of four booleans drifting apart.

Also confirmed:

- **Contrast** — `pct >= 81`, minimum 4.600:1, matching the number I measured independently. The test now scans all 76 integer steps and reads the ramp endpoints from `PETITION_BLUE`, so a hue change re-runs the measurement instead of silently testing the old colour.
- **Ranking** — `ORDER BY COUNT(p.id) DESC, c.name ASC, c.id ASC`.
- **"+N" highlight** — `setDot` resolves the highlight from the group, using the same expression `buildSubjectPins` does, so the two paths cannot diverge again. This one is verified by reading only: across three attempts I could not get a selection to land inside a "+N" box in a browser, so treat the runtime behaviour as unconfirmed rather than checked.

One correction to my previous review: I wrote that a stale petition popup would sit over a δήμος with no shading. That was wrong — the focused-δήμος fill comes from `__clicked-city__`, which sits outside the `showPetitioned` branch, so the shading is there. Checked on the preview.

## Follow-ups

Not blocking, and not in an issue yet — recording them here so they aren't lost.

1. **Map highlights aren't exclusive on two code paths** (pre-dates this PR). `toggleInfo` (`LandingV2.tsx`) and the δήμος-card `onSelect` (`DesktopLayout.tsx`, `MobileLayout.tsx`) don't call `clearMapFocus()`. Reproduce: Δήμοι tab → click **Αγία Παρασκευή** in the petition list → click **Βριλήσσια** in the δήμος list; the Αγία Παρασκευή popup stays. Clicking Βριλήσσια's map bubble instead clears it, because `onOpenCity` calls `clearMapFocus()`.
2. **The selection-restyle effect still skips "+N" pins** — `useMapMarkers.tsx:339` returns early on `!pin.subject`. The morph path is now group-aware and this one isn't, so selecting a subject inside a group still doesn't highlight it without a layer rebuild. Pre-existing, but the correct expression is now one line away.
3. **Donut hit area** — the rectangle-vs-ellipse band; `border-radius: 50%` on the marker element makes the clickable shape the ellipse the packer already proves disjoint.
4. **Packing isn't translation-invariant on fractional coordinates** — the docstring overstates it, `onMoveEnd` is ungated, and the test uses integer inputs only.
5. **`translations.test.ts` doesn't validate ICU** — key parity only; a plural missing its `other` arm passes and throws at render.
6. **Petition query** — no `LIMIT`, and `geometry` ships on every first paint though only the Δήμοι view reads it; separately, `officialSupport = false` and `status = 'listed'` are independent columns.

The inline threads for the fixed items are resolved.

src/lib/landing/petitions.ts
kouloumos: Fixed — `pct >= 81`, minimum 4.600:1 across the ramp, matching the number I measured independently. Resolving.

src/components/landing/v2/hooks/useMapMarkers.tsx
kouloumos: Fixed at the root: `setDot` now resolves the highlight from the group, using the same expression `buildSubjectPins` does, so the two paths can't diverge again. Verified by reading — I couldn't get a selection to land inside a "+N" box in a browser, so the runtime behaviour is unconfirmed either way. The restyle effect at line 339 still returns early on `!pin.subject`; that's pre-existing and listed as a follow-up in the review. Resolving this thread.

src/components/landing/v2/mapMarkers.tsx
kouloumos: Follow-up in the review, not blocking. Resolving.

src/components/landing/v2/hooks/useMapMarkers.tsx
kouloumos: Follow-up in the review, not blocking. Resolving.

src/lib/__tests__/translations.test.ts
kouloumos: Follow-up in the review, not blocking. Resolving.

src/lib/landing/petitions.test.ts
kouloumos: Fixed — the test now walks all 76 integer steps and reads the ramp endpoints from `PETITION_BLUE`, so a hue change re-runs the measurement. Resolving.

src/lib/db/cities.ts
kouloumos: `ORDER BY COUNT(p.id) DESC, c.name ASC, c.id ASC` fixed, with the reason in a comment. The `LIMIT` and the `officialSupport`/`status` question are follow-ups in the review. Resolving.

PR #575, opened 2026-07-28
fix(analytics): never identify against the cookieless sentinel

christosporios:
## Problem

PostHog production data shows at least **five different signed-up users merged into a single PostHog person** (`13452e1d-b5bf-5f8b-805e-5f47dc9fa755`, 2,470 events). Their `$identify` events all share `anon_distinct_id: "$posthog_cookieless"` — the sentinel distinct id every cookieless visitor shares.

The race: consent status lives in localStorage (cross-tab), but cookieless mode is per-tab runtime state. A tab still running cookieless can observe granted consent (set by another tab, with the sync effect re-triggered by a `useSession` refetch) and call `identify()` while its distinct id is still the sentinel. Each such identify aliases another account to the same shared value, and PostHog merges them all into one person. Merges are irreversible.

## Fix

- Guard `identify()` on `posthog.get_distinct_id() !== "$posthog_cookieless"`. The affected tab keeps capturing anonymously; identify happens on the next cookied page load.

## Also included

- Team members now get `$internal_or_test_user: true` set as a person property (once per device, localStorage-guarded to avoid a billable `$set` per page load). The per-device `internal_user` event stamp misses events sent from a fresh device before the first team sign-in (~1,900 events leaked past the project filters since June 12); the person property puts them in the *Internal / Test users* cohort (152880), which the project filter excludes retroactively at query time.
- Unit tests for `identifyPostHogUser` covering the sentinel guard, the consent gate, device marking, and the once-per-device person property.

## Verification

- `npx tsc --noEmit` clean, 12/12 tests pass.
- After deploy, the frankenperson can be deleted in PostHog so future events re-create clean per-user persons.

---

> [!NOTE]
> **Medium Risk**
> Changes identity aliasing and internal-user filtering in production analytics; wrong guards could leave users unidentified or mis-tag cohorts, but scope is limited to client-side PostHog helpers with new tests.
> 
> **Overview**
> Fixes a cross-tab race where **PostHog `identify()`** could run while a tab still used the shared **`$posthog_cookieless`** distinct id, merging multiple real users into one person. **`identifyPostHogUser`** now bails out when `get_distinct_id()` is that sentinel so identification waits for a cookied load.
> 
> Team/internal traffic also gets a **person-level backstop**: after consent, team members (including superadmins) receive **`$internal_or_test_user: true`** via `setPersonProperties`, deduped per account with new **`INTERNAL_PERSON_KEY`** in localStorage (still one device stamp via `internal_user` even when consent is pending or cookieless). **`identifyUser.test.ts`** adds jsdom coverage for consent, sentinel, device marking, and per-account person properties.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 736daba9a05c3368795fb523d5863284ebf2b16c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR prevents PostHog identification against the shared cookieless sentinel and improves internal-user attribution.
- Skips identification and person-property updates while the current distinct ID is `$posthog_cookieless`.
- Records the last team account stamped on each device so shared browsers stamp each distinct team member once.
- Adds tests for consent, cookieless mode, device marking, superadmins, and per-account person properties.

The PR appears safe to merge.

The previously reported shared-device attribution failure is fixed by storing and comparing the team user ID, and no blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/analytics/identifyUser.ts | Adds the cookieless sentinel guard and correctly changes internal-person deduplication from device-wide to per-account. |
| src/components/analytics/__tests__/identifyUser.test.ts | Covers identification gates and verifies that different team accounts sharing a device each receive the internal person property. |
| src/lib/utils/analyticsConsent.ts | Defines and documents the localStorage key that tracks the most recently stamped team account. |

<sub>Reviews (3): Last reviewed commit: ["fix(analytics): never identify against t..."](https://github.com/schemalabz/opencouncil/commit/736daba9a05c3368795fb523d5863284ebf2b16c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47935844)</sub>

### DIFF
--- a/src/components/analytics/__tests__/identifyUser.test.ts
+++ b/src/components/analytics/__tests__/identifyUser.test.ts
@@ -1,0 +1,143 @@
+/**
+ * @jest-environment jsdom
+ *
+ * identifyPostHogUser reads localStorage, so it needs jsdom despite the
+ * .test.ts → node split in jest.config.js.
+ */
+import posthog from 'posthog-js';
+import type { Session } from 'next-auth';
+import { identifyPostHogUser } from '../identifyUser';
+import { INTERNAL_PERSON_KEY, INTERNAL_USER_KEY } from '@/lib/utils/analyticsConsent';
+
+jest.mock('posthog-js', () => ({
+  __esModule: true,
+  default: {
+    __loaded: true,
+    get_explicit_consent_status: jest.fn(),
+    get_distinct_id: jest.fn(),
+    identify: jest.fn(),
+    register: jest.fn(),
+    setPersonProperties: jest.fn(),
+  },
+}));
+
+const mocked = posthog as jest.Mocked<typeof posthog>;
+
+function session(user: Partial<Session['user']>): Session {
+  return { user, expires: '' } as Session;
+}
+
+const visitor = session({ id: 'user-1', email: 'someone@example.com' });
+const teamMember = session({ id: 'team-1', email: 'someone@opencouncil.gr' });
+
+describe('identifyPostHogUser', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    localStorage.clear();
+    mocked.__loaded = true;
+    mocked.get_explicit_consent_status.mockReturnValue('granted');
+    mocked.get_distinct_id.mockReturnValue('0197-anon-device-uuid');
+  });
+
+  it('identifies a consenting signed-in user on their user id', () => {
+    identifyPostHogUser(visitor);
+
+    expect(mocked.identify).toHaveBeenCalledWith('user-1');
+  });
+
+  it('does nothing when posthog is not initialized', () => {
+    mocked.__loaded = false;
+
+    identifyPostHogUser(teamMember);
+
+    expect(mocked.register).not.toHaveBeenCalled();
+    expect(mocked.identify).not.toHaveBeenCalled();
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+    expect(localStorage.getItem(INTERNAL_USER_KEY)).toBeNull();
+  });
+
+  it.each(['pending', 'denied'] as const)('does not identify while consent is %s', (status) => {
+    mocked.get_explicit_consent_status.mockReturnValue(status);
+
+    identifyPostHogUser(visitor);
+
+    expect(mocked.identify).not.toHaveBeenCalled();
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+  });
+
+  it('still marks the team device while consent is pending', () => {
+    // The device stamp must not sit behind the consent gate: it exists
+    // precisely so logged-out and cookieless visits from a team device are
+    // filtered too.
+    mocked.get_explicit_consent_status.mockReturnValue('pending');
+
+    identifyPostHogUser(teamMember);
+
+    expect(mocked.register).toHaveBeenCalledWith({ internal_user: true });
+    expect(localStorage.getItem(INTERNAL_USER_KEY)).toBe('1');
+    expect(mocked.identify).not.toHaveBeenCalled();
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+  });
+
+  it('does not identify while the tab still runs cookieless', () => {
+    // The cookieless sentinel is shared by every visitor: identifying
+    // against it would merge all such users into one PostHog person.
+    mocked.get_distinct_id.mockReturnValue('$posthog_cookieless');
+
+    identifyPostHogUser(visitor);
+
+    expect(mocked.identify).not.toHaveBeenCalled();
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+  });
+
+  it('marks a team device even while cookieless', () => {
+    mocked.get_distinct_id.mockReturnValue('$posthog_cookieless');
+
+    identifyPostHogUser(teamMember);
+
+    expect(localStorage.getItem(INTERNAL_USER_KEY)).toBe('1');
+    expect(mocked.register).toHaveBeenCalledWith({ internal_user: true });
+    expect(mocked.identify).not.toHaveBeenCalled();
+    // Writing the person property onto the shared sentinel person would flag
+    // every anonymous visitor as internal — the same class of bug as the
+    // sentinel identify.
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+  });
+
+  it('treats superadmins on external emails as team members', () => {
+    const superAdmin = session({ id: 'admin-1', email: 'someone@example.com', isSuperAdmin: true });
+
+    identifyPostHogUser(superAdmin);
+
+    expect(mocked.register).toHaveBeenCalledWith({ internal_user: true });
+    expect(mocked.setPersonProperties).toHaveBeenCalledWith({ $internal_or_test_user: true });
+  });
+
+  it('sends the internal person property once per account for team members', () => {
+    identifyPostHogUser(teamMember);
+    identifyPostHogUser(teamMember);
+
+    expect(mocked.setPersonProperties).toHaveBeenCalledTimes(1);
+    expect(mocked.setPersonProperties).toHaveBeenCalledWith({ $internal_or_test_user: true });
+    expect(localStorage.getItem(INTERNAL_PERSON_KEY)).toBe('team-1');
+  });
+
+  it('re-sends the person property when a different team member uses the same device', () => {
+    const secondTeamMember = session({ id: 'team-2', email: 'other@opencouncil.gr' });
+
+    identifyPostHogUser(teamMember);
+    identifyPostHogUser(secondTeamMember);
+
+    expect(mocked.setPersonProperties).toHaveBeenCalledTimes(2);
+    expect(localStorage.getItem(INTERNAL_PERSON_KEY)).toBe('team-2');
+  });
+
+  it('does not mark regular users as internal in any way', () => {
+    identifyPostHogUser(visitor);
+
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+    expect(mocked.register).not.toHaveBeenCalled();
+    expect(localStorage.getItem(INTERNAL_USER_KEY)).toBeNull();
+    expect(localStorage.getItem(INTERNAL_PERSON_KEY)).toBeNull();
+  });
+});
--- a/src/components/analytics/identifyUser.ts
+++ b/src/components/analytics/identifyUser.ts
@@ -1,6 +1,11 @@
 import posthog from "posthog-js";
 import type { Session } from "next-auth";
-import { INTERNAL_USER_KEY } from "@/lib/utils/analyticsConsent";
+import { INTERNAL_PERSON_KEY, INTERNAL_USER_KEY } from "@/lib/utils/analyticsConsent";
+
+// posthog-js's cookieless sentinel distinct id (not exported publicly).
+// While in cookieless mode, EVERY visitor shares this value as their
+// distinct id.
+const COOKIELESS_SENTINEL = "$posthog_cookieless";
 
 // Team traffic on production can't be excluded by host filters, and person
 // filters can't see logged-out or cookieless visits. Instead, any device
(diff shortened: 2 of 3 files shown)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/c84615f02ee688927298e97fefc8ea6c3b39a245..c3034bd433ddbd6cb79cb855af1ca046ba46e68b) to address the review comment about the device-wide person-property marker:

- `INTERNAL_PERSON_KEY` now stores the **user id** that was stamped instead of a boolean, so when a second team member signs in on the same browser profile, their person also gets `$internal_or_test_user: true` (previously the first sign-in permanently suppressed it for everyone else on that device).
- Added a test covering the two-team-members-one-device case (7 tests total, all passing).

src/components/analytics/identifyUser.ts
```diff
@@ -26,11 +31,29 @@ export function identifyPostHogUser(session: Session | null) {
...
+    // keeps capturing anonymously, and identify happens on the next cookied
+    // page load.
+    if (posthog.get_distinct_id() === COOKIELESS_SENTINEL) return;
     if (session?.user?.id) {
         // Deliberately no person properties: the id alone is enough to link
         // events to the account, keeps direct PII (email) out of PostHog, and
         // avoids the billable $set that identify() emits on every page load
         // when properties are passed for an already-identified person.
         posthog.identify(session.user.id);
+        if (isTeamMember(session) && localStorage.getItem(INTERNAL_PERSON_KEY) !== "1") {
```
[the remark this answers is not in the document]
christosporios: Valid catch — fixed in c3034bd: the marker now stores the stamped user id rather than a boolean, so each team account on a shared device gets the person property once. Covered by a new test.

src/components/analytics/__tests__/identifyUser.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    expect(mocked.setPersonProperties).toHaveBeenCalledTimes(2);
+    expect(localStorage.getItem(INTERNAL_PERSON_KEY)).toBe('team-2');
+  });
+
+  it('does not send person properties for regular users', () => {
+    identifyPostHogUser(visitor);
+
+    expect(mocked.setPersonProperties).not.toHaveBeenCalled();
+  });
```
kouloumos: All 7 pass, but they don't defend the guards. I broke `identifyUser.ts` one way at a time and re-ran the suite — **6 of 7 mutations survive**; only the consent gate is actually pinned. The three that matter:

- **drop `isTeamMember(session) &&` from `identifyUser.ts:29`** → every visitor's device gets stamped `internal_user: true`, permanently, and re-applied on each load by `instrumentation-client.ts:54`. The project's "filter internal and test users" setting then excludes all production traffic — no error, no exception, dashboards just quietly go to zero. Survives.
- **move the consent gate above the device stamp** → reverts the stamp for exactly the logged-out/cookieless visitors that `identifyUser.ts:10-15` says it exists for. Survives.
- **make the sentinel guard skip `identify()` but not the person property** (e.g. `const cookieless = …; if (!cookieless) posthog.identify(…)`) → writes `$internal_or_test_user: true` onto the shared sentinel person, flagging every anonymous visitor as internal. Same class as the bug this PR fixes. Survives.

The common cause is that 5 of the 7 tests assert a single mock, where `src/lib/__tests__/analyticsConsent.test.ts` asserts both sides in every test. The function has five observable outputs (`register`, `identify`, `setPersonProperties`, and the two localStorage keys) and each test pins one or two.

Four lines close all three:

```ts
// 'does not send person properties for regular users'
expect(mocked.register).not.toHaveBeenCalled();
expect(localStorage.getItem(INTERNAL_USER_KEY)).toBeNull();

// 'does not identify while the tab still runs cookieless'
expect(mocked.setPersonProperties).not.toHaveBeenCalled();
```

Plus one new test with `get_explicit_consent_status` → `'pending'` asserting the device stamp *still* fires. Every current test mocks `'granted'` in `beforeEach`, so nothing pins the stamp-above-consent ordering — which is the whole point of the feature.

Smaller, while you're in here: `beforeEach` resets `mocked.__loaded = true` but no test ever sets it false, so the `!posthog.__loaded` early return is untested and that setup line is inert (the sibling file has exactly this test). Both team fixtures use `@opencouncil.gr`, so the `isSuperAdmin` arm of `isTeamMember` never runs. And consent is only tested as `'pending'`, never `'denied'` — the sibling uses `it.each` for this.

kouloumos (commented):
The diagnosis is right and the guard is correct — I verified `"$posthog_cookieless"` against `COOKIELESS_SENTINEL_VALUE` (`posthog-js/lib/src/constants.js:96`) on the pinned 1.386.5, and `identifyUser.ts` is the only `identify()`/`alias()` call site in `src/`. Two things about the *residual*, though, and one about the tests.

### The comment at `identifyUser.ts:34-42` describes the opposite of what happens

Being cookieless is maintained by two mechanisms with different lifetimes:

- the **identity** (`distinct_id = $posthog_cookieless`) is written once, imperatively, into that tab's memory — at init (`posthog-core.js:528`) or inside `opt_out_capturing()` (`:3277`)
- the **flag** (`$cookieless_mode: true`) is computed per-event from `_inCookielessMode()` (`:1139`), which reads consent live from localStorage on every access (`consent.js:83`)

So when another tab grants consent, the live half updates instantly and the write-once half never does:

| | before grant | after grant |
|---|---|---|
| `get_distinct_id()` | `$posthog_cookieless` | `$posthog_cookieless` |
| `_inCookielessMode()` | true | **false** |
| `$cookieless_mode` sent | yes | **no** |
| capturing | yes | yes |

The tab now sends the sentinel distinct id *without* the flag that tells ingestion to substitute the daily hash. With `person_profiles: "always"` that creates and then keeps feeding a real person whose distinct id is the literal string. I think that's the frankenperson — the five accounts are what `identify()` *named* onto it, not what created it.

Which means "this tab keeps capturing anonymously" is precisely what it stops doing. Concretely:

- the guard stops accounts being aliased on — the irreversible harm, worth having
- it doesn't stop the person being fed
- **logged-out visitors never reach it at all** (`PostHogAuthSync.tsx:22` only calls it when authenticated), and since `person_profiles: "always"` was chosen because most visitors never sign in, they're probably most of the volume. 2,470 events against ~5 identifies fits that.

That last part is inference, not measurement — worth confirming by breaking that person's events down by the `logged_in` super property.

Not asking you to redesign this here; you have context on the PostHog side I don't. But the comment should say what actually happens, because as written it closes the question for the next reader.

### The tab can get stuck on the sentinel permanently

"identify happens on the next cookied page load" doesn't hold on this path:

1. Two tabs open, neither has answered the chip. `ConsentChip` reads the key once on mount and has no `storage` listener, so tab A's chip stays up after B accepts.
2. Accept in A → `opt_in_capturing()` → `_inCookielessMode()` is already false → **reset block skipped** (`:3197`)
3. falls through to `_sync_opt_out_with_persistence()` (`:3220`) → `set_disabled(false)` → `save()` → **sentinel written to localStorage**
4. guard fires, no identify — correct
5. next load: consent granted → init reaches `else if (!this.get_distinct_id())` (`:534`) → persisted sentinel is truthy → no device id minted

From there the guard returns on every load forever; only sign-out or clearing site data recovers. Note `opt_out_capturing()` avoids exactly this by disabling persistence *before* registering the sentinel (`:3266-3277`) — `opt_in_capturing()` inverts that order.

Pre-PR this path self-corrected, because `identify()` did `register({distinct_id: userId})`. So it's a trade rather than a regression, but a silent one. Cheapest fix is to heal instead of returning, reusing the pair already on the sign-out path:

```ts
if (posthog.get_distinct_id() === COOKIELESS_SENTINEL) {
    posthog.reset();
    applyStoredAnalyticsConsent();
    return; // next pass identifies against a real device id
}
```

Separately: the new tests all pass but don't defend the guards — 6 of 7 mutations I tried survive the suite. Details inline on the test file.

### Smaller

- `:49-50` — the marker is written before the `$set` it guards, so a dropped request (adblocker on `/ingest`, offline) records success permanently with no retry. Swap the two lines; both are synchronous.
- `:49` — `INTERNAL_PERSON_KEY` is a single slot, so two team members alternating on a shared device re-fire on every switch. A per-user key removes it. (posthog's own dedupe at `:2169` is memory-only, so the localStorage guard is doing real work — just not for alternating accounts.)
- `:20-26` now contradicts `:34-42`: it says granted consent implies out-of-cookieless-mode, which is the thing `:34-42` exists because of. Worth merging.
- HEAD is a `fixup!` — needs autosquash before merge.

### One thing I can't check from here

`:10-15` says the exclusion works off the **event** property `internal_user`; `:51-55` says the cohort filter works off the **person** property `$internal_or_test_user`. Those are different filter conditions — is the person-property one actually configured on the project filter? If not, that half of the PR ships a billable `$set` and nothing else.

Related, and pre-existing: on the accept path `opt_in_capturing()` internally calls `reset(true)` → `persistence.clear()`, which drops the `internal_user` super property. `:29`'s localStorage gate then skips re-registering it on an already-marked device, so the `$pageview` two lines later in `ConsentChip` goes out unstamped. Plausibly part of the leak you're trying to plug — gating only the `setItem` and calling `register` unconditionally would make it self-healing.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/c3034bd433ddbd6cb79cb855af1ca046ba46e68b..736daba9a05c3368795fb523d5863284ebf2b16c) to harden the test suite per @kouloumos's review (and to squash the stray fixup commits into a single clean commit — apologies for the noise):

- All five observable outputs (`register`, `identify`, `setPersonProperties`, both localStorage keys) are now asserted in the tests where their absence matters, not just the one under test
- New test: the device stamp still fires while consent is `pending` — pins the stamp-above-consent ordering
- New test: `posthog.__loaded === false` does nothing (the previously inert `beforeEach` reset now has a reason to exist)
- New test: superadmin on a non-opencouncil.gr email is treated as team — covers the `isSuperAdmin` arm of `isTeamMember`
- Consent gate now tested with `it.each(['pending', 'denied'])` like the sibling suite
- 11 tests total, all passing

Verified against the three P1 mutations from the review by re-applying each one — every mutation now fails exactly one targeted test (previously 6 of 7 mutations survived).

src/components/analytics/__tests__/identifyUser.test.ts
christosporios: Great catch, and thanks for actually running the mutations. Fixed in 736daba9: every test now asserts all the outputs whose absence matters (register, identify, setPersonProperties, both localStorage keys), plus the four missing cases — stamp-fires-while-consent-pending, __loaded=false, the isSuperAdmin arm, and it.each(['pending','denied']) for the consent gate. Re-ran your three P1 mutations against the new suite: each one now fails exactly one targeted test.

kouloumos:
Merged as ac17aaba and going out to production. Three checks to confirm it actually did what it should, and the order matters.

### 1. Does the internal-traffic half actually do anything?

The open question from review: `identifyUser.ts:10-15` filters on the *event* property `internal_user`, while `:51-55` relies on the *person* property `$internal_or_test_user` via cohort 152880. Those are different filter conditions, and nothing in the code confirms the project filter carries the second one.

On a fresh browser profile (or after clearing site data): accept the cookie chip, sign in with an `@opencouncil.gr` account, load a page. Then check:

- the person carries `$internal_or_test_user: true`
- the person appears in cohort 152880
- with "filter internal and test users" enabled, that person's events drop out of a query that otherwise shows them

If the third fails, the code is fine and the console config isn't — the project filter needs the person-property condition added. Nothing to revert either way: the mechanism excludes at query time, so fixing the setting later also fixes the historical events.

### 2. Measure the frankenperson before deleting it

On `13452e1d-b5bf-5f8b-805e-5f47dc9fa755`, break the events down by the `logged_in` super property and check whether `$cookieless_mode` is present on them. Record the event count and date range.

This settles what was only inferred during review. If the mechanism is what we think, `$cookieless_mode` should be absent on essentially all of them (that's what makes them land on a literal-sentinel person rather than a daily-hash one), and `logged_in: false` should dominate.

Do this before step 3 — deleting first loses the evidence.

### 3. Delete it, then watch whether it regrows

This is the actual test of whether the fix was sufficient, and it answers the open design question with data instead of argument.

- **Doesn't regrow** → the identify path was the whole problem, the residual is theoretical, and we close it out.
- **Regrows from anonymous traffic** → confirms the residual: stale tabs, and any browser stuck via the persistence path, are still feeding it. The guard can't reach either, since neither goes through `identifyPostHogUser`.

The regrowth *rate* is the useful number — it prices the healing work. A handful of events a week is a footnote; back into the thousands means it's worth doing properly. Worth looking at ~1 week and ~1 month, and the `logged_in` split on the regrown events distinguishes the logged-out gap from the stuck-browser path.

We'll post the numbers here once these have run.

PR #589, opened 2026-08-04
refactor(subjects): remove deprecated SubjectSpeakerSegment

christosporios:
## Summary

`SubjectSpeakerSegment` has been superseded by `SpeakerContribution` and utterance-level subject tagging (`Utterance.discussionSubjectId`) for a while now. This PR removes the model and all remaining code paths that referenced it, and drops the table.

## Changes

**Schema & migration**
- Remove the `SubjectSpeakerSegment` model and its relations on `Subject` and `SpeakerSegment`
- New migration (`20260804000000_drop_subject_speaker_segment`, created but not applied) drops the table — it drops the dependent `SubjectSpeakerSegmentSearchView` first, which a Prisma-generated migration wouldn't know about

**Legacy fallbacks removed**
- `getStatisticsFor` / `getBatchStatisticsForSubjects` / `getDiscussionSecondsForSubjects` no longer fall back to the join table; subjects without tagged utterances get empty statistics
- City reset route, `prisma/seed.ts`, `generate_seed_dump.ts`, and `copy_db.sh` no longer create/delete/copy the table
- Deleted `src/components/meetings/Subjects.tsx` (unused — nothing imported it) and the subject page's legacy speaker-segment fallback rendering; the subject page now renders contributions only

**Behavior rewired to the new system**
- Detailed search results (chat context) derive a subject's speaker segments from utterances tagged with `discussionSubjectId`, keeping the same output shape for the chat API
- ES person "spoke in subject" filter and nested full-text queries now use `speaker_contributions`; the `speaker_segments` nested field is removed from `schema.json` / `views.sql` / `validate-views.sql`
- `matchedSpeakerSegmentIds` (driven by ES inner hits on the removed field) is gone, along with the chat "key segment" 🔹 highlighting built on it
- Landing map speaker count comes from contribution rows (one per speaker per subject); podcast spec generation resolves segment IDs via tagged utterances

## Deploy notes

- The migration must run before/with deploy: it drops both the table and `SubjectSpeakerSegmentSearchView`
- This is a **PGSync schema change** — the Elasticsearch index needs re-indexing per `elasticsearch/README.md`

## Testing

- `npx tsc --noEmit` clean
- `npm test`: 89 suites, 1253 tests pass (statistics and search-query tests updated)
- `npm run build` succeeds after `prisma generate`
- `npm run test:integration` could not run in the dev container (no Docker daemon) — relying on CI

https://claude.ai/code/session_01Y4GioPiosYYAFuDzNsReR6

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y4GioPiosYYAFuDzNsReR6)_

---

> [!NOTE]
> **High Risk**
> Drops a production table and ES nested field (re-index required), and changes subject stats/search/chat behavior for data that never migrated to utterance tagging or contributions.
> 
> **Overview**
> Removes the deprecated **`SubjectSpeakerSegment`** join table and **`SubjectSpeakerSegmentSearchView`**, and deletes all app paths that still depended on them. Subject discussion is modeled only through **`SpeakerContribution`** and utterances tagged with **`discussionSubjectId`**.
> 
> **Database & sync:** Prisma drops the model; migration drops the view then the table. PGSync/Elasticsearch **`speaker_segments`** nested data and the segment search view are removed; search docs and examples now use **`speaker_contributions`** only. Deploy needs the migration and a **full ES re-index**.
> 
> **Statistics & data tooling:** Subject speaking time and batch stats no longer fall back to the join table—subjects without **`SUBJECT_DISCUSSION`** utterances get empty statistics. Seeds, DB copy, seed dumps, and city reset skip the old table; podcast spec generation resolves segment IDs from tagged utterances.
> 
> **Product & search:** The subject page shows **contributions only** (legacy segment UI and unused **`Subjects.tsx`** removed). Detailed search/chat context loads speaker segments from utterances tagged to the subject, not from ES inner hits; **`matchedSpeakerSegmentIds`** and chat “key segment” highlighting are removed. Person “spoke in subject” filters and nested text search target **`speaker_contributions`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ad25940dfec90e6eec74c5bb41eff37ddc6f49c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR removes the deprecated SubjectSpeakerSegment model, database table, Elasticsearch view and nested field, then rewires subject statistics, search, rendering, and tooling to use contributions and utterance-level tags.
- Drops the legacy schema and migration artifacts.
- Replaces legacy subject statistics and UI fallbacks with contribution and tagged-utterance data.
- Moves detailed search and Elasticsearch subject content to SpeakerContribution and discussionSubjectId-based paths.

The PR is not yet safe to merge because detailed search still contaminates subject context with unrelated utterances, and contribution rows still inflate the reported speaker count.

Whole speaker segments are assigned to every subject represented by any contained utterance and then concatenated without subject or status filtering; separately, the map and MCP speaker count remains the number of contribution rows rather than distinct speakers.

**Files Needing Attention:** src/lib/search/index.ts, src/lib/db/subject.ts, src/lib/utils.ts

| Filename | Overview |
|----------|----------|
| src/lib/search/index.ts | Rebuilds detailed subject segments from utterance tags, but emits complete mixed-subject segments and therefore retains the reported context-contamination defect. |
| src/lib/db/subject.ts | Replaces legacy segment-based map data with contribution counts; the previously reported row-count-versus-speaker-count issue remains outstanding. |
| prisma/schema.prisma | Removes SubjectSpeakerSegment and its relations while retaining SpeakerContribution as the replacement subject-speaker model. |
| prisma/migrations/20260815000000_drop_subject_speaker_segment/migration.sql | Drops the dependent search view before removing the deprecated join table. |
| elasticsearch/schema.json | Removes the speaker_segments nested mapping and PGSync child while retaining speaker_contributions. |

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `src/lib/search/index.ts`, line 320-327 ([link](https://github.com/schemalabz/opencouncil/blob/7f3d2901c83425c0a3c241739147486fac549df1/src/lib/search/index.ts#L320-L327)) 

   <a href="#">P1</a> **Mixed segments contaminate subject context**

   When a speaker segment contains utterances assigned to multiple discussion subjects or statuses, this mapping concatenates every utterance in the segment for each represented subject, causing detailed search and chat context to present unrelated speech as evidence for the returned subject.

   **Knowledge Base Used:** [MCP assistant integration](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/mcp-assistant.md)
   
   `````markdown
   Path: src/lib/search/index.ts
   Line: 320-327
   
   Comment:
   **Mixed segments contaminate subject context**
   
   When a speaker segment contains utterances assigned to multiple discussion subjects or statuses, this mapping concatenates every utterance in the segment for each represented subject, causing detailed search and chat context to present unrelated speech as evidence for the returned subject.
   
   **Knowledge Base Used:** [MCP assistant integration](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/mcp-assistant.md)
   
   ---
   
   `````
   </details>
   
   <a href="https://app.greptile.com/ide/claude-code?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20src%2Flib%2Fsearch%2Findex.ts%0ALine%3A%20320-327%0A%0AComment%3A%0A**Mixed%20segments%20contaminate%20subject%20context**%0A%0AWhen%20a%20speaker%20segment%20contains%20utterances%20assigned%20to%20multiple%20discussion%20subjects%20or%20statuses%2C%20this%20mapping%20concatenates%20every%20utterance%20in%20the%20segment%20for%20each%20represented%20subject%2C%20causing%20detailed%20search%20and%20chat%20context%20to%20present%20unrelated%20speech%20as%20evidence%20for%20the%20returned%20subject.%0A%0A**Knowledge%20Base%20Used%3A**%20%5BMCP%20assistant%20integration%5D%28https%3A%2F%2Fapp.greptile.com%2Fschema-labs%2F-%2Fcustom-context%2Fknowledge-base%2Fschemalabz%2Fopencouncil%2F-%2Fdocs%2Fmcp-assistant.md%29%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=589&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaude.svg?v=6">Fix in Claude Code</picture></a>
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Greploop%20schemalabz%2Fopencouncil%20PR%20%23589%3A%20work%20through%20Greptile's%20open%20review%20comments%2C%20then%20keep%20reviewing%20and%20fixing%20until%20it%20comes%20back%20clean%20at%205%2F5%20with%20zero%20unresolved%20comments.%0AStart%20by%20reading%20the%20comments%20off%20the%20PR%20itself.%20On%20GitHub%2C%20use%20paginated%20%60gh%20api%20graphql%60%20to%20query%20%60pullRequest.reviewThreads%60%20with%20each%20thread's%20%60isResolved%60%20value%20and%20inline%20%60comments%60.%20%60gh%20pr%20view%20--comments%60%20only%20includes%20conversation%20comments%2C%20so%20do%20not%20use%20it%20as%20the%20findings%20source.%20They%20are%20not%20listed%20in%20this%20prompt%20on%20purpose%3A%20the%20PR%20is%20current%2C%20a%20pasted%20copy%20would%20not%20be.%20Skip%20anything%20already%20resolved%2C%20and%20if%20you%20judge%20a%20comment%20wrong%2C%20say%20so%20rather%20than%20changing%20code%20to%20satisfy%20it.%0A%0APrefer%20the%20Greptile%20CLI%2C%20which%20reviews%20the%20working%20tree%20with%20no%20push%20and%20no%20CI%20run.%20Fall%20back%20to%20PUSH%20LOOP%20only%20where%20a%20step%20below%20says%20to.%0A1.%20Run%20%60command%20-v%20greptile%60.%20Missing%3A%20go%20to%20PUSH%20LOOP%2C%20and%20pitch%20the%20CLI%20at%20the%20end.%0A2.%20Run%20%60greptile%20whoami%60.%20It%20exits%200%20even%20when%20signed%20out%2C%20so%20read%20the%20output%3A%20if%20it%20says%20%22Not%20signed%20in%22%2C%20ask%20the%20user%20to%20run%20%60greptile%20login%60%20%28it%20opens%20a%20browser%2C%20do%20not%20run%20it%20yourself%29%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%0A3.%20Look%20for%20%60greploop%2FSKILL.md%60%20under%20%60.agents%2Fskills%2F%60%20or%20%60.claude%2Fskills%2F%60%2C%20both%20in%20the%20repo%20root%20and%20in%20the%20user's%20home%20directory.%20Found%3A%20follow%20that%20skill%2C%20never%20pushing%20between%20rounds%2C%20and%20you%20are%20done.%0A4.%20Missing%3A%20run%20%60greptile%20skills%20list%60.%20If%20it%20fails%20as%20an%20unknown%20command%2C%20this%20CLI%20predates%20the%20skill%2C%20so%20tell%20the%20user%20to%20run%20%60greptile%20update%60%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%20Otherwise%20ask%20the%20user%20before%20running%20%60greptile%20skills%20install%20greploop%60%2C%20then%20follow%20the%20skill%20it%20writes.%0A%0APUSH%20LOOP%20%28only%20when%20a%20step%20above%20sends%20you%20here%29%3A%0A%0AGREPLOOP%20ENABLED%3A%20After%20fixing%20all%20issues%2C%20commit%20and%20push%20your%20changes.%20Greptile%20will%20automatically%20re-review%20the%20PR%20via%20webhook.%20Wait%20for%20the%20review%20to%20complete%2C%20then%20read%20the%20new%20review%20comments%20and%20confidence%20score.%20If%20the%20confidence%20score%20is%20below%205%2F5%20or%20there%20are%20unresolved%20comments%2C%20fix%20those%20issues%20and%20push%20again.%20Repeat%20until%20the%20confidence%20score%20is%205%2F5%20with%20zero%20unresolved%20comments.%20Maximum%205%20iterations.%20Do%20NOT%20open%20a%20new%20PR%20%E2%80%94%20keep%20pushing%20to%20the%20same%20branch.%0A%0AIF%20YOU%20ENDED%20UP%20ON%20PUSH%20LOOP%20BECAUSE%20THE%20CLI%20WAS%20MISSING%3A%20once%20the%20loop%20is%20done%2C%20print%20the%20block%20below%20and%20nothing%20else%20about%20the%20CLI.%20Reproduce%20it%20character%20for%20character%20inside%20a%20fenced%20code%20block%2C%20so%20the%20borders%20line%20up.%20Do%20not%20reword%20it%2C%20translate%20it%2C%20summarize%20it%2C%20add%20to%20it%2C%20or%20drop%20the%20box.%20Print%20it%20once%2C%20and%20do%20not%20bring%20the%20CLI%20up%20again%20later%20in%20the%20session.%0A%0A%E2%95%AD%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AE%0A%E2%94%82%20%20You%20can%20run%20greploops%20faster%20locally%20with%20our%20CLI.%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%E2%94%82%0A%E2%94%82%20%20Install%20it%20at%20https%3A%2F%2Fwww.greptile.com%2Fcli%2C%20or%20I%20can%20install%20it%20for%20you.%20%20%E2%94%82%0A%E2%95%B0%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AF%0A%0AIf%20they%20take%20you%20up%20on%20it%2C%20install%20with%20%60npm%20install%20-g%20greptile%60%20%28or%20%60brew%20install%20greptileai%2Ftap%2Fgreptile%60%29%2C%20then%20%60greptile%20skills%20install%20greploop%60.%20Leave%20%60greptile%20login%60%20to%20them%2C%20it%20opens%20a%20browser.&repo=schemalabz%2Fopencouncil&pr=589&platform=github">Fix all with Greploop</a>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Flib%2Fsearch%2Findex.ts%3A320-327%0A**Mixed%20segments%20contaminate%20subject%20context**%0A%0AWhen%20a%20speaker%20segment%20contains%20utterances%20assigned%20to%20multiple%20discussion%20subjects%20or%20statuses%2C%20this%20mapping%20concatenates%20every%20utterance%20in%20the%20segment%20for%20each%20represented%20subject%2C%20causing%20detailed%20search%20and%20chat%20context%20to%20present%20unrelated%20speech%20as%20evidence%20for%20the%20returned%20subject.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=589&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
src/lib/search/index.ts:320-327
**Mixed segments contaminate subject context**

When a speaker segment contains utterances assigned to multiple discussion subjects or statuses, this mapping concatenates every utterance in the segment for each represented subject, causing detailed search and chat context to present unrelated speech as evidence for the returned subject.

---

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["test(statistics): pin subject time to ta..."](https://github.com/schemalabz/opencouncil/commit/7f3d2901c83425c0a3c241739147486fac549df1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50074083)</sub>

### CODE REVIEW

**Comment by kouloumos** [comment]
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/1ad25940dfec90e6eec74c5bb41eff37ddc6f49c..04f5bd6cb59c04c34df38f6aa35a60237b285bcf) to rebase onto current main. The branch was 126 commits behind. Four files conflicted:

- `src/app/api/chat/route.ts` and `src/lib/db/chat-mock-data.ts`: main removed `/chat` in e143c3d2. The changes to these two files no longer apply, so I accepted the deletion.
- `src/lib/search/query.ts` and `src/lib/search/index.ts`: main only moved the surrounding lines. I removed the `speaker_segments` nested query and `matchedSpeakerSegmentIds`, as the commit intends.

Two changes are not part of the conflict resolution:

- The migration is now `20260815000000_drop_subject_speaker_segment`. The old timestamp `20260804000000` sorted before two migrations that main added after this branch (`20260811190506`, `20260814120000`). The three migrations are independent, so the apply order was safe. The filename order did not match it.
- 04f5bd6c exposes `speakerCount` on `list_hot_subjects`. The MCP surface hid the field because the count came from the join this PR removes. `speakerCountOf` now counts contributions, so the value is correct for every subject.

`npx tsc --noEmit` passes. All 104 test files pass.

The state before the rebase is at [`backup/pr-589-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-589-original).

A preview database that already applied `20260804000000_drop_subject_speaker_segment` needs a reset.

File: src/lib/search/index.ts [hunk 1]
```diff
@@ -214,6 +213,35 @@ export async function search(
...
         const subjectMap = new Map(subjects.map(subject => [subject.id, subject]));
 
+        // For detailed results, fetch the speaker segments discussing each subject
+        // (segments whose utterances are tagged with the subject via discussionSubjectId)
+        const segmentsBySubject = new Map<string, SubjectDiscussionSegment[]>();
+        if (request.config?.detailed && subjectIds.length > 0) {
+            const subjectIdSet = new Set(subjectIds);
+            const segments = await prisma.speakerSegment.findMany({
+                where: {
+                    utterances: { some: { discussionSubjectId: { in: subjectIds } } }
```
kouloumos: This matches any utterance tagged to the subject, whatever its `discussionStatus`. Every other reader pairs the two columns, and they use three different sets: `SUBJECT_DISCUSSION` alone in `statistics.ts` and `getDiscussionSecondsForSubjects`; `SUBJECT_DISCUSSION | PROCEDURAL_VOTE | VOTE` in `getUtterancesForSubject` and the MCP transcript tool; `VOTE` alone in the voting-utterances route. These two new queries add a fourth set, which is everything.

The effect: an agenda item that passes without debate has `VOTE` utterances tagged to it. The subject page reports no discussion and zero minutes, because `getStatisticsFor` filters on `SUBJECT_DISCUSSION`. This query returns the roll-call segment as that subject's discussion. The same subject answers two different ways.

The legacy join could not do this, because the summarize task wrote only discussion segments into it.

Suggest one shared helper that both call sites import, with the status set as a required argument so a later call site cannot leave it out. That also removes the status literal that is already duplicated in two files.

src/lib/search/index.ts
```diff
@@ -270,26 +298,19 @@ export async function search(
...
                     votes: [],
                     attendance: []
                 };
 
                 // If detailed results are requested, add speaker segment text
                 if (request.config?.detailed) {
-                    const speakerSegments = subject.speakerSegments
-                        .map(ss => ss.speakerSegment)
+                    const speakerSegments = (segmentsBySubject.get(subject.id) ?? [])
                         .filter(segment => {
```
kouloumos: A segment is one continuous block from one speaker, so its utterances can carry two different `discussionSubjectId` values when a speaker runs past the end of one agenda item into the next. This filter keeps the whole segment for each subject it touches, and the map below joins every utterance of the segment into one string. The same words then appear as the discussion of two different subjects.

The status filter above does not fix this. Only selecting the utterances tagged to the subject does.

src/lib/search/index.ts
```diff
@@ -214,6 +213,35 @@ export async function search(
         // Create a map of subjects by ID for efficient lookup
         const subjectMap = new Map(subjects.map(subject => [subject.id, subject]));
 
+        // For detailed results, fetch the speaker segments discussing each subject
+        // (segments whose utterances are tagged with the subject via discussionSubjectId)
+        const segmentsBySubject = new Map<string, SubjectDiscussionSegment[]>();
+        if (request.config?.detailed && subjectIds.length > 0) {
```
kouloumos: Is this path still live? No caller in the repo sets `detailed` to true. The search page and the MCP tool both pass false. The only route to it is an external `POST /api/search` with an explicit flag. Its original consumer was `/chat`, which main removed in e143c3d2.

If nothing external depends on it, deleting the detailed path removes this branch and the fan-out below it.

If it stays, it needs a bound. The query has no `take`, `pageSize` accepts up to 100, and the include pulls every utterance of every matched segment plus the speaker roles per segment. Before this PR the same include hung off the legacy join, which is empty for most subjects, so the cost was small. Now it scales with the tagged utterances, and the heaviest subjects on staging carry several thousand utterance rows each.

src/lib/search/index.ts
```diff
@@ -32,6 +33,34 @@ const logEssential = (message: string, data?: any) => {
     console.log(`[Search Analytics] ${message}`, data || '');
 };
 
+// Relations needed to turn a speaker segment into a detailed search result entry
+const subjectDiscussionSegmentInclude = {
```
kouloumos: This include repeats a shape that `roleWithRelationsInclude` in `src/lib/db/types/roles.ts` already defines. That helper has 22 call sites, three of them in `src/lib/db/subject.ts`, which this PR edits. `getLatestSegmentsForSpeaker` in `src/lib/db/speakerSegments.ts` builds the same include and uses the helper.

The PR moves an existing hand-written copy rather than inventing one, but hoisting it into a named const is the moment to point it at the helper. Better still, move this const next to the two matching includes in `speakerSegments.ts` so all three share one definition.

src/lib/search/index.ts
```diff
@@ -270,26 +298,19 @@ export async function search(
...
                     councilMeeting: subject.councilMeeting,
                     votes: [],
                     attendance: []
                 };
 
                 // If detailed results are requested, add speaker segment text
                 if (request.config?.detailed) {
-                    const speakerSegments = subject.speakerSegments
-                        .map(ss => ss.speakerSegment)
+                    const speakerSegments = (segmentsBySubject.get(subject.id) ?? [])
```
kouloumos: The guard here and the map below reproduce code in `src/lib/db/speakerSegments.ts`. The `hasPerson` / `hasRoles` / `text.length >= 100` triple and the seven-field object are the same in both files, and `speakerSegments.ts` already holds two copies. `SearchResultDetailed.speakerSegments` is typed `SegmentWithRelations`, so the type is shared but the producer is not.

Suggest exporting a mapper and a predicate from `speakerSegments.ts` and using them at all three sites.

src/lib/tasks/generatePodcastSpec.ts
```diff
@@ -38,10 +37,17 @@ export async function requestGeneratePodcastSpec(cityId: string, councilMeetingI
             throw new Error(`Subject with id ${subject.id} not found`);
         }
 
+        // Segments discussing the subject, via utterances tagged with discussionSubjectId
+        const taggedUtterances = await prisma.utterance.findMany({
+            where: { discussionSubjectId: subject.id },
```
kouloumos: Same missing `discussionStatus` filter as the search query. A vote or attendance utterance tagged to the subject puts its segment into `speakerSegmentIds`, and the podcast backend receives it as source material.

The query also has no `orderBy`, so the ids arrive in index-scan order rather than chronological order. The backend consumes the list as a sequence.

Querying segments instead of utterances fixes both, and matches the shape the search path uses.

src/lib/__tests__/statistics.test.ts
```diff
@@ -118,8 +115,7 @@ describe('Statistics', () => {
       }));
     });
 
-    it('should call prisma with correct parameters for subject statistics (new system)', async () => {
-      // Mock utterances query (new system)
+    it('should call prisma with correct parameters for subject statistics', async () => {
```
kouloumos: This test mocks `speakerSegment.findMany` to return an empty array, so `getStatisticsForTranscript` iterates nothing. The duration map built above is never read, and the returned statistics are never inspected. The test asserts the call arguments and stops.

I checked what that leaves uncovered. I changed `getStatisticsForTranscript` to ignore `utteranceDurationsBySegment` and always use the full segment length, which reverts the central behaviour of this PR, and all 1400 tests still passed.

Suggest returning a real segment from the mock and asserting the value. A segment from 0 to 100 with tagged utterances of 15 and 5 seconds should produce 20 seconds, not 100. The batch function has no test at all.

src/lib/db/subject.ts
```diff
@@ -329,9 +296,9 @@ export function buildMapSubjectWhere(realm: Realm, f: MapSubjectFilters): Prisma
     };
 }
 
-/** Unique speaker count for a subject from its (legacy) speaker-segment join. */
+/** Speaker count for a subject: one contribution per speaker (unique on subjectId+speakerId). */
 function speakerCountOf(s: MapSubjectPayload): number {
```
kouloumos: This changes the count from distinct speakers to a row count. The version on main builds a `Set` of speaker tag ids, so it deduplicates. This one does not.

The unique constraint on `(subjectId, speakerId)` covers identified speakers, so the two agree for them. It does not cover rows where `speakerId` is null, because Postgres treats nulls as distinct.

I looked at the rows behind this on a copy of the data. The null-speaker rows hold separate statements, and the writer nulls `speakerId` when it cannot resolve the person, so they look like different speakers rather than duplicates of one. Counting them is defensible. The point is narrower: main deduplicated and this does not, so the change deserves a decision rather than arriving as a side effect.

This matters more because 04f5bd6c publishes the number over MCP. I wrote that commit, and its message claims the value is correct for every subject. That claim is too strong. Suggest softening the message, or dropping the commit until the semantics are settled.

prisma/migrations/20260815000000_drop_subject_speaker_segment/migration.sql
```diff
@@ -0,0 +1,17 @@
+/*
+  Warnings:
+
+  - You are about to drop the `SubjectSpeakerSegment` table. If the table is not empty, all the data it contains will be lost.
+
+*/
+-- Drop the PGSync helper view that depends on the table (see elasticsearch/views.sql)
+DROP VIEW IF EXISTS "SubjectSpeakerSegmentSearchView";
```
kouloumos: This drops a view that PGSync creates and reads. PGSync runs from opencouncil-tasks with its own copy of `schema.json`, and nothing in this repo orders the two. If the migration lands while PGSync still runs the old schema, PGSync reads a view that no longer exists and sync stops.

The removed nested field also needs a full re-index, which `elasticsearch/README.md` already requires for a `schema.json` change.

Suggest stating the order in the description: deploy PGSync with the new schema, run the migration, then re-index.

src/lib/search/index.ts
kouloumos: This matches any utterance tagged to the subject, whatever its `discussionStatus`. Every other reader pairs the two columns, and they use three different sets: `SUBJECT_DISCUSSION` alone in `statistics.ts` and `getDiscussionSecondsForSubjects`; `SUBJECT_DISCUSSION | PROCEDURAL_VOTE | VOTE` in `getUtterancesForSubject` and the MCP transcript tool; `VOTE` alone in the voting-utterances route. These two new queries add a fourth set, which is everything.

The effect: an agenda item that passes without debate has `VOTE` utterances tagged to it. The subject page reports no discussion and zero minutes, because `getStatisticsFor` filters on `SUBJECT_DISCUSSION`. This query returns the roll-call segment as that subject's discussion. The same subject answers two different ways.

The legacy join could not do this, because the summarize task wrote only discussion segments into it.

Suggest one shared helper that both call sites import, with the status set as a required argument so a later call site cannot leave it out. That also removes the status literal that is already duplicated in two files.

src/lib/search/index.ts
kouloumos: A segment is one continuous block from one speaker, so its utterances can carry two different `discussionSubjectId` values when a speaker runs past the end of one agenda item into the next. This filter keeps the whole segment for each subject it touches, and the map below joins every utterance of the segment into one string. The same words then appear as the discussion of two different subjects.

The status filter above does not fix this. Only selecting the utterances tagged to the subject does.

src/lib/search/index.ts
kouloumos: Is this path still live? No caller in the repo sets `detailed` to true. The search page and the MCP tool both pass false. The only route to it is an external `POST /api/search` with an explicit flag. Its original consumer was `/chat`, which main removed in e143c3d2.

If nothing external depends on it, deleting the detailed path removes this branch and the fan-out below it.

If it stays, it needs a bound. The query has no `take`, `pageSize` accepts up to 100, and the include pulls every utterance of every matched segment plus the speaker roles per segment. Before this PR the same include hung off the legacy join, which is empty for most subjects, so the cost was small. Now it scales with the tagged utterances, and the heaviest subjects on staging carry several thousand utterance rows each.

src/lib/search/index.ts
kouloumos: This include repeats a shape that `roleWithRelationsInclude` in `src/lib/db/types/roles.ts` already defines. That helper has 22 call sites, three of them in `src/lib/db/subject.ts`, which this PR edits. `getLatestSegmentsForSpeaker` in `src/lib/db/speakerSegments.ts` builds the same include and uses the helper.

The PR moves an existing hand-written copy rather than inventing one, but hoisting it into a named const is the moment to point it at the helper. Better still, move this const next to the two matching includes in `speakerSegments.ts` so all three share one definition.

src/lib/search/index.ts
kouloumos: The guard here and the map below reproduce code in `src/lib/db/speakerSegments.ts`. The `hasPerson` / `hasRoles` / `text.length >= 100` triple and the seven-field object are the same in both files, and `speakerSegments.ts` already holds two copies. `SearchResultDetailed.speakerSegments` is typed `SegmentWithRelations`, so the type is shared but the producer is not.

Suggest exporting a mapper and a predicate from `speakerSegments.ts` and using them at all three sites.

src/lib/tasks/generatePodcastSpec.ts
kouloumos: Same missing `discussionStatus` filter as the search query. A vote or attendance utterance tagged to the subject puts its segment into `speakerSegmentIds`, and the podcast backend receives it as source material.

The query also has no `orderBy`, so the ids arrive in index-scan order rather than chronological order. The backend consumes the list as a sequence.

Querying segments instead of utterances fixes both, and matches the shape the search path uses.

src/lib/__tests__/statistics.test.ts
kouloumos: This test mocks `speakerSegment.findMany` to return an empty array, so `getStatisticsForTranscript` iterates nothing. The duration map built above is never read, and the returned statistics are never inspected. The test asserts the call arguments and stops.

I checked what that leaves uncovered. I changed `getStatisticsForTranscript` to ignore `utteranceDurationsBySegment` and always use the full segment length, which reverts the central behaviour of this PR, and all 1400 tests still passed.

Suggest returning a real segment from the mock and asserting the value. A segment from 0 to 100 with tagged utterances of 15 and 5 seconds should produce 20 seconds, not 100. The batch function has no test at all.

src/lib/db/subject.ts
kouloumos: This changes the count from distinct speakers to a row count. The version on main builds a `Set` of speaker tag ids, so it deduplicates. This one does not.

The unique constraint on `(subjectId, speakerId)` covers identified speakers, so the two agree for them. It does not cover rows where `speakerId` is null, because Postgres treats nulls as distinct.

I looked at the rows behind this on a copy of the data. The null-speaker rows hold separate statements, and the writer nulls `speakerId` when it cannot resolve the person, so they look like different speakers rather than duplicates of one. Counting them is defensible. The point is narrower: main deduplicated and this does not, so the change deserves a decision rather than arriving as a side effect.

This matters more because 04f5bd6c publishes the number over MCP. I wrote that commit, and its message claims the value is correct for every subject. That claim is too strong. Suggest softening the message, or dropping the commit until the semantics are settled.

prisma/migrations/20260815000000_drop_subject_speaker_segment/migration.sql
kouloumos: This drops a view that PGSync creates and reads. PGSync runs from opencouncil-tasks with its own copy of `schema.json`, and nothing in this repo orders the two. If the migration lands while PGSync still runs the old schema, PGSync reads a view that no longer exists and sync stops.

The removed nested field also needs a full re-index, which `elasticsearch/README.md` already requires for a `schema.json` change.

Suggest stating the order in the description: deploy PGSync with the new schema, run the migration, then re-index.

src/lib/search/index.ts
kouloumos, line 223: This matches any utterance tagged to the subject, whatever its `discussionStatus`. Every other reader pairs the two columns, and they use three different sets: `SUBJECT_DISCUSSION` alone in `statistics.ts` and `getDiscussionSecondsForSubjects`; `SUBJECT_DISCUSSION | PROCEDURAL_VOTE | VOTE` in `getUtterancesForSubject` and the MCP transcript tool; `VOTE` alone in the voting-utterances route. These two new queries add a fourth set, which is everything.

The effect: an agenda item that passes without debate has `VOTE` utterances tagged to it. The subject page reports no discussion and zero minutes, because `getStatisticsFor` filters on `SUBJECT_DISCUSSION`. This query returns the roll-call segment as that subject's discussion. The same subject answers two different ways.

The legacy join could not do this, because the summarize task wrote only discussion segments into it.

Suggest one shared helper that both call sites import, with the status set as a required argument so a later call site cannot leave it out. That also removes the status literal that is already duplicated in two files.

src/lib/search/index.ts
kouloumos, line 314: A segment is one continuous block from one speaker, so its utterances can carry two different `discussionSubjectId` values when a speaker runs past the end of one agenda item into the next. This filter keeps the whole segment for each subject it touches, and the map below joins every utterance of the segment into one string. The same words then appear as the discussion of two different subjects.

The status filter above does not fix this. Only selecting the utterances tagged to the subject does.

src/lib/search/index.ts
kouloumos, line 219: Is this path still live? No caller in the repo sets `detailed` to true. The search page and the MCP tool both pass false. The only route to it is an external `POST /api/search` with an explicit flag. Its original consumer was `/chat`, which main removed in e143c3d2.

If nothing external depends on it, deleting the detailed path removes this branch and the fan-out below it.

If it stays, it needs a bound. The query has no `take`, `pageSize` accepts up to 100, and the include pulls every utterance of every matched segment plus the speaker roles per segment. Before this PR the same include hung off the legacy join, which is empty for most subjects, so the cost was small. Now it scales with the tagged utterances, and the heaviest subjects on staging carry several thousand utterance rows each.

src/lib/search/index.ts
kouloumos, line 37: This include repeats a shape that `roleWithRelationsInclude` in `src/lib/db/types/roles.ts` already defines. That helper has 22 call sites, three of them in `src/lib/db/subject.ts`, which this PR edits. `getLatestSegmentsForSpeaker` in `src/lib/db/speakerSegments.ts` builds the same include and uses the helper.

The PR moves an existing hand-written copy rather than inventing one, but hoisting it into a named const is the moment to point it at the helper. Better still, move this const next to the two matching includes in `speakerSegments.ts` so all three share one definition.

src/lib/search/index.ts
kouloumos, line 313: The guard here and the map below reproduce code in `src/lib/db/speakerSegments.ts`. The `hasPerson` / `hasRoles` / `text.length >= 100` triple and the seven-field object are the same in both files, and `speakerSegments.ts` already holds two copies. `SearchResultDetailed.speakerSegments` is typed `SegmentWithRelations`, so the type is shared but the producer is not.

Suggest exporting a mapper and a predicate from `speakerSegments.ts` and using them at all three sites.

src/lib/tasks/generatePodcastSpec.ts
kouloumos, line 42: Same missing `discussionStatus` filter as the search query. A vote or attendance utterance tagged to the subject puts its segment into `speakerSegmentIds`, and the podcast backend receives it as source material.

The query also has no `orderBy`, so the ids arrive in index-scan order rather than chronological order. The backend consumes the list as a sequence.

Querying segments instead of utterances fixes both, and matches the shape the search path uses.

src/lib/__tests__/statistics.test.ts
kouloumos, line 119: This test mocks `speakerSegment.findMany` to return an empty array, so `getStatisticsForTranscript` iterates nothing. The duration map built above is never read, and the returned statistics are never inspected. The test asserts the call arguments and stops.

I checked what that leaves uncovered. I changed `getStatisticsForTranscript` to ignore `utteranceDurationsBySegment` and always use the full segment length, which reverts the central behaviour of this PR, and all 1400 tests still passed.

Suggest returning a real segment from the mock and asserting the value. A segment from 0 to 100 with tagged utterances of 15 and 5 seconds should produce 20 seconds, not 100. The batch function has no test at all.

src/lib/db/subject.ts
kouloumos: This changes the count from distinct speakers to a row count. The version on main builds a `Set` of speaker tag ids, so it deduplicates. This one does not.

The unique constraint on `(subjectId, speakerId)` covers identified speakers, so the two agree for them. It does not cover rows where `speakerId` is null, because Postgres treats nulls as distinct.

I looked at the rows behind this on a copy of the data. The null-speaker rows hold separate statements, and the writer nulls `speakerId` when it cannot resolve the person, so they look like different speakers rather than duplicates of one. Counting them is defensible. The point is narrower: main deduplicated and this does not, so the change deserves a decision rather than arriving as a side effect.

This matters more because 04f5bd6c publishes the number over MCP. I wrote that commit, and its message claims the value is correct for every subject. That claim is too strong. Suggest softening the message, or dropping the commit until the semantics are settled.

prisma/migrations/20260815000000_drop_subject_speaker_segment/migration.sql
kouloumos, line 18: This drops a view that PGSync creates and reads. This PR already removes the view definition from `views.sql`, so `views.sql` is where the removal belongs as well — re-running that file should bring a database to the intended state, and today it leaves the orphaned view in place.

Keep a `DROP VIEW IF EXISTS` here too. The table cannot drop while the view exists, and this migration is the only step that always runs, because `views.sql` is applied by hand. `IF EXISTS` makes both copies idempotent, so carrying it in both costs nothing.

That combination also allows a sequence with no sync gap: deploy PGSync on the new `schema.json`, run `views.sql`, then migrate. Without it, a running PGSync breaks the moment the view disappears and stays broken until it redeploys. Worth stating the intended order in the description.

kouloumos (commented):
## Search no longer indexes what speakers say

`SubjectSpeakerSegment` was the only path from a subject to its speaker segments in the index. `Subject` is the root entity, so we index subjects. The join gave each subject its segments, and the view aggregated those segments' utterances into one verbatim text field. This PR removes the join, so the path goes with it.

The removal is correct. The join is deprecated and this PR is right to drop it. But a capability leaves alongside it. The removed field held concatenated utterance text, which is what people actually said. The replacement holds an AI-written contribution per speaker per subject. Both are useful, and they are not the same content.

We need to think about how to represent this now. `Utterance.discussionStatus` tells us the type of each utterance, and `discussionSubjectId` tells us which subject it belongs to. The old join could carry neither. So the question is how to index that and how to use it.

## Notes

This PR changes the same four files under `elasticsearch/` as #346. This PR keeps `SubjectSearchView`, and #346 removes it to fix deletion propagation. Whichever merges second gets a conflict, and a careless resolution can restore the bug that #346 fixes. Worth agreeing the order.

`src/lib/utils.ts:140` still declares `speakerSegments` on `SortableSubject`, and the appearance branch at `src/lib/utils.ts:184` still reads it. No query can populate that field now. This PR removes the identical field from `CategorizableSubject` in `src/lib/utils/subjects.ts`, so this looks like a missed twin. Six tests keep it green by casting literals to `any`. No production caller passes `appearance`, so nothing breaks today.

src/lib/__tests__/statistics.test.ts
kouloumos: Added four tests in 7f3d2901. They assert values rather than call shape: a segment longer than its tagged part, a segment with no tagged utterance, several utterances of one segment summed, and two subjects sharing a segment.

Checked they have teeth. Removing the `utteranceDurationsBySegment` override, so the code reads whole segment lengths again, previously left all 1400 tests passing. All four now fail.

src/lib/db/subject.ts
kouloumos: Addressed in d64a828b. `speakerCountOf` is gone; the landing rows call `getContributionCount`, which the subject ranker and `contributor_count` in `SubjectMetricsView` already use, so the three agree by construction instead of by a comment asking them to. `mapSubjectInclude` now selects `_count` rather than a row per contribution.

The doc comment that claimed one row per speaker is corrected, and the commit message no longer claims the value is correct for every subject.

The semantics question stays open, and I raised it in the update comment: rows with a null `speakerId` escape the unique constraint, so several unidentified speakers stay several rows.

prisma/migrations/20260815000000_drop_subject_speaker_segment/migration.sql
kouloumos: Settled the other way round, which I think is the right split: `views.sql` owns the view definitions, and the migration owns the table plus the view that dies with it. The `DROP VIEW` stays here because the table cannot drop while that view exists. The tagged-only `SubjectMetricsView` lives only in `views.sql`.

The rebase made this concrete. #628 added `SubjectMetricsView`, which reads the table for a legacy fallback, so the migration as written fails on any database that carries the PGSync views. I reproduced both paths against a prod-shaped database:

- migration alone: `cannot drop table "SubjectSpeakerSegment" because other objects depend on it`
- `views.sql` first, then the migration: completes, table gone, view still selectable

So the deploy order is now a real prerequisite rather than a preference, and the migration header states it. Preview and staging carry no PGSync views and will pass either way, which is the gap #638 is about.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/04f5bd6cb59c04c34df38f6aa35a60237b285bcf..7f3d2901c83425c0a3c241739147486fac549df1) after rebasing onto main, which had moved 16 commits ahead and now includes #346 and #628.

**The rebase found a blocker.** #628 added `SubjectMetricsView`, which reads `SubjectSpeakerSegment` for a legacy fallback. Postgres refuses `DROP TABLE` while a view reads the table, so this migration would have failed on production. Preview and staging carry no PGSync views, so both would have passed first. I reproduced it against a prod-shaped database: without the fix, `cannot drop table "SubjectSpeakerSegment" because other objects depend on it`; with `views.sql` applied first, the migration completes.

`views.sql` now carries the tagged-only `SubjectMetricsView`, and the migration documents the order it depends on. The migration SQL is otherwise unchanged.

**The legacy branch is gone from the metric.** `discussion_speaking_seconds` comes only from tagged `SUBJECT_DISCUSSION` utterances. The two branches did not measure the same thing: the legacy branch summed whole segments, and the tagged branch sums the utterances that belong to the subject. A subject with no tagged utterance now reports 0 until the summarize task runs for its meeting.

**Also in this push:** `speakerCountOf` is gone; the landing rows call `getContributionCount`, which the ranker and `contributor_count` already use, and `mapSubjectInclude` selects `_count` instead of a row per contribution. Four tests pin subject time to tagged utterances — removing the duration override previously left all 1400 tests passing, and now fails all four.

**Verified with the pgsync harness** against a seeded database (455 subjects, 4,068 segments, 630 contributions, 30,521 tagged utterances): bootstrap indexes 455 documents, no document carries `speaker_segments`, REF links are stripped from descriptions and contributions, and `contributor_count` and `discussion_speaking_seconds` agree with the SQL for the subjects I checked. Deletion propagation is healthy — the daemon counters move together on insert and on delete.

**Not addressed here.** The missing `discussionStatus` filter, the whole-segment text attribution, the unbounded `detailed` fetch, and the two duplications all sit on `detailed: true` and `requestGeneratePodcastSpec`, and neither has a caller in the repo. Fixing dead paths in this PR did not seem right, especially while the open question is whether `detailed` should be removed. Those threads stay open.

**One decision left.** `contributor_count` and the landing count both count contribution rows. The unique constraint on `(subjectId, speakerId)` does not bind rows with a null `speakerId`, so several unidentified speakers on one subject stay several rows. The data says they are distinct statements rather than duplicates, so counting them is defensible, but it is worth a deliberate answer.

src/lib/db/subject.ts
```diff
@@ -326,9 +293,9 @@ export function buildMapSubjectWhere(realm: Realm, f: MapSubjectFilters): Prisma
...
-    return new Set((s.speakerSegments ?? []).map((sss) => sss.speakerSegment.speakerTag.id)).size;
+    return s.contributions.length;
```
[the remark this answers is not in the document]
kouloumos: The mechanism is right, the conclusion does not hold. `speakerId` is nullable and Postgres treats nulls as distinct, so the unique constraint does not bind those rows, and several of them can sit on one subject. On staging, 196 subjects carry more than one, up to 16.

They are not repeats of one speaker though. I read the rows behind the worst case: five contributions, each with different text. The writer explains why — `src/lib/db/utils.ts` nulls `speakerId` when it cannot resolve the incoming speaker to a person:

```ts
speakerId: contrib.speakerId && validSpeakerIds.has(contrib.speakerId)
    ? contrib.speakerId
    : null,
```

So each row is a separate speaker that lost its person link, not one speaker counted several times. Counting them reports the number of people who spoke, which is what the field is for. Collapsing them to one would undercount.

Worth adding that the row-count semantics is not introduced here. `getContributionCount` in `src/lib/utils.ts` and `contributor_count` in `SubjectMetricsView` both counted rows before this branch. This PR moves the landing count onto the same helper, so the three now agree rather than diverge.

Leaving the thread open, because what a null-speaker row should count as is a product question and worth an explicit answer. I do not think it blocks this PR.

PR #591, opened 2026-08-06
fix(transcript): keep the viewport still when an utterance edit is saved

### LINKED ISSUE #367: Editing: Screen jumps to the end of a long segment after pressing Enter
Often, when an editor is working inside a long segment and makes a change, pressing Enter causes the screen to scroll/jump to the end of the segment, losing the editor's current position in the text. This forces the editor to manually scroll back to where they were, which is disruptive and time-consuming - especially in long transcriptions.

[Video](https://www.loom.com/share/711a94ece3f94496a0bd0e74210d9db7)

Steps to Reproduce (it doesn't happen every time!)

Open a transcription in editing mode
Navigate to a long speaker segment
Make a change anywhere in the middle of the segment
Press Enter
Observe that the screen scrolls down to the end of the segment instead of staying at the current cursor position

Expected Behavior:
After pressing Enter, the screen should stay at the editor's current cursor position within the text. The viewport should not move unless the cursor itself moves outside the visible area.
Actual Behavior:
The screen jumps to the end of the segment, losing the editor's current position and requiring manual scrolling to return to the editing point.

christosporios:
Closes #367

## The bug

Browsers keep your place through layout changes with a feature called **scroll anchoring**: when content resizes, the browser picks an element you're looking at and adjusts the scroll position so it stays put. Its preferred anchor is the **focused element**.

Editing an utterance replaces its inline `<span>` with a block-level editor box, so while you're typing, the focused textarea *is* the anchor. Pressing Enter does two things in the same frame:

1. destroys the textarea → the browser loses its anchor;
2. removes a block-shaped box from the middle of a paragraph → the text reflows.

The anchor is gone in exactly the frame it was needed. And the ground underneath isn't solid: speaker segments are virtualised with `content-visibility: auto` (`globals.css:226`), so off-screen segments aren't laid out and their heights are estimates that keep getting corrected. Scroll anchoring normally hides that; for one frame it can't, and the scroll lands wherever those estimates leave it — in a long segment, usually near its end. `VideoProvider.tsx:179` already re-scrolls three times after a deep link to work around the same instability.

That also explains the "doesn't happen every time" in the report: it depends on what the layout above the viewport happens to be doing at that moment.

### How this was pinned down

Frame-by-frame extraction of the screen recording in the issue shows the jump is a **single-frame** event at 30fps — 4.383s still has the open editor with the caret in it, 4.417s is already ~712px further down, and nothing moves for 0.27s before or 0.5s after. That rules out the app's own `scrollIntoView({behavior: 'smooth'})` calls, which animate over many frames, and places the jump exactly at the commit that closes the editor.

## The fix

Measure the edited node, commit the close synchronously, re-pin whatever node replaced it:

```ts
const closeEditor = (commit?: () => void) => {
    const anchor = captureScrollAnchor(rootRef.current);
    flushSync(() => {
        commit?.();
        setIsEditing(false);
    });
    restoreScrollAnchor(anchor, rootRef.current);
};
```

Both closing paths (save and cancel) go through it, passing whatever state should land in the same pass. Opening the editor is deliberately **not** anchored — the `EDIT_NEXT_UTTERANCE` shortcut relies on the browser scrolling to an off-screen utterance.

Keeping this in the two handlers rather than in a hook matters on this page: a hook would have added a ref and a layout effect to all ~9,000 utterances to serve the one being edited, on a component that is explicitly tuned for that scale.

`flushSync` is a deliberate deopt — it's React's documented escape hatch for measuring the DOM right after a state update, and it's called from event handlers here, so it's a sanctioned use. It forces a synchronous render of one memoized component.

### Also: the textarea's ref callback

It was written inline, so React re-attached it on every render and re-ran `focus()` on **every keystroke** — each call asking the browser to scroll the element into view. Wrapping the same callback in `useCallback([])` makes React invoke it only when the editor actually mounts, which is what "focus once" wants.

## Changes

| File | |
|---|---|
| `src/lib/utils/scrollAnchor.ts` | new — `captureScrollAnchor` / `restoreScrollAnchor`, pure DOM. Restore takes the replacement node explicitly, since React swaps the editor `<div>` for a fresh `<span>`. |
| `src/components/meetings/transcript/Utterance.tsx` | route both closing paths through `closeEditor`; make the textarea ref stable |
| `src/lib/utils/__tests__/scrollAnchor.test.tsx` | new — 5 tests (replacement node, sub-pixel drift, no scroll container, null guards) |

## Verification

- Full suite: 1270 passing. `tsc --noEmit` and `eslint` clean. `npm run build` compiles and type-checks (it then fails collecting page data for `/api/chat` on missing Elasticsearch env in the sandbox — unrelated).
- Reproduced the transcript's DOM in Chromium via Playwright (`content-visibility` segments, sticky headers, the inline↔block editor swap) and ran the exact capture → swap → jump → restore sequence: **−818px stranding without the anchor, +1px with it.**

## Follow-up worth considering

This compensates rather than cures. The root cause is that edit mode swaps *inline* text for a *block* box; an inline-sized editor wouldn't reflow the paragraph and would need no anchoring at all. That's a real rewrite of the editing UI (timestamp row, save/cancel buttons, auto-growing rows), so it isn't folded into this fix.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T7bs2JUZvryhquRy7tCRFU)_

---

> [!NOTE]
> **Low Risk**
> Localized transcript UX fix with guarded DOM helpers and tests; `flushSync` is limited to single-utterance edit close handlers.
> 
> **Overview**
> Fixes **#367**: saving or canceling an utterance edit no longer jumps the transcript scroll position when the block editor collapses back into inline text.
> 
> A new **`scrollAnchor`** helper measures the utterance’s position in the meeting **`data-scroll-container`**, then **`restoreScrollAnchor`** nudges **`scrollTop`** after layout reflow. **`Utterance`** routes save, cancel, and no-op close through **`closeEditor`**, which uses **`flushSync`** so the DOM swap completes before the second measurement; **`rootRef`** tracks whichever node is active (editor **`div`** or read-only **`span`**). **`Transcript`** reuses **`getScrollContainer`** for banner scroll tracking instead of duplicating the selector.
> 
> jsdom tests cover replacement nodes, sub-pixel drift, and missing-container guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 607ba4d5fc74aae9d7ab29045631f394c2eb2841. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds explicit scroll-position preservation when closing the utterance editor and centralizes transcript scroll-container lookup.
- Captures the edited utterance’s viewport position, synchronously commits save or cancel state, and restores any resulting scroll drift.
- Adds focused DOM-helper tests for replacement nodes, drift correction, container discovery, and null handling.
- Reuses the scroll-container helper in the transcript’s banner scroll listener.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/meetings/transcript/Utterance.tsx | Routes save, unchanged-submit, and cancel closures through synchronous anchor restoration while preserving the existing editor interaction behavior. |
| src/lib/utils/scrollAnchor.ts | Introduces small DOM utilities for locating the transcript scroller and correcting replacement-node position drift. |
| src/lib/utils/__tests__/scrollAnchor.test.tsx | Covers scroll correction, replacement nodes, sub-pixel drift, container discovery, and null guards in jsdom. |
| src/components/meetings/transcript/Transcript.tsx | Replaces a duplicated scroll-container lookup with the shared helper without changing listener behavior. |

<sub>Reviews (2): Last reviewed commit: ["fix(transcript): keep the viewport still..."](https://github.com/schemalabz/opencouncil/commit/607ba4d5fc74aae9d7ab29045631f394c2eb2841) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50798871)</sub>

### DIFF
--- a/src/components/meetings/transcript/Utterance.tsx
+++ b/src/components/meetings/transcript/Utterance.tsx
@@ -1,6 +1,7 @@
 "use client";
 import { Utterance } from "@prisma/client";
-import React, { useEffect, useRef, useState } from "react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
+import { flushSync } from "react-dom";
 import { useTranslations } from 'next-intl';
 import { useVideoActions } from "../VideoProvider";
 import { useTranscriptOptions } from "../options/OptionsContext";
@@ -19,6 +20,7 @@ import { cn } from "@/lib/utils";
 import { useEditing } from "../EditingContext";
 import { formatTimestamp } from "@/lib/formatters/time";
 import { useLocalizeText } from "@/hooks/useLocalizeText";
+import { captureScrollAnchor, restoreScrollAnchor } from "@/lib/utils/scrollAnchor";
 
 /**
  * Resolve the character offset of a click within a text span.
@@ -61,6 +63,11 @@ const UtteranceC: React.FC<{
 
     const clickOffsetRef = useRef<number | null>(null);
     const [isEditing, setIsEditing] = useState(false);
+    // Whichever node currently stands for this utterance — the read-only span
+    // or the editor box. Closing the editor swaps one for the other, which is
+    // exactly the moment the browser loses its own scroll anchor (#367).
+    const rootRef = useRef<HTMLElement | null>(null);
+    const setRootRef = useCallback((el: HTMLElement | null) => { rootRef.current = el; }, []);
     const [localUtterance, setLocalUtterance] = useState(utterance);
     const [editedText, setEditedText] = useState(utterance.text);
     const [editedStartTime, setEditedStartTime] = useState(utterance.startTimestamp);
@@ -79,6 +86,19 @@ const UtteranceC: React.FC<{
         setEditedStartTime(utterance.startTimestamp);
         setEditedEndTime(utterance.endTimestamp);
     }, [utterance]);
+
+    // Stable identity, so React runs this when the editor actually mounts. An
+    // inline callback is re-attached on every render, which re-ran `focus()` on
+    // every keystroke — and each call asks the browser to scroll the textarea
+    // into view.
+    const focusEditor = useCallback((el: HTMLTextAreaElement | null) => {
+        if (!el) return;
+        el.focus();
+        if (clickOffsetRef.current !== null) {
+            el.selectionStart = el.selectionEnd = clickOffsetRef.current;
+            clickOffsetRef.current = null;
+        }
+    }, []);
 
     // Check if this utterance is highlighted in the current editing highlight
     const isHighlighted = editingHighlight?.highlightedUtterances.some(hu => hu.utteranceId === localUtterance.id) || false;
@@ -141,6 +161,21 @@ const UtteranceC: React.FC<{
         }
     };
 
+    /**
+     * Closing the editor collapses a block-level box back into the running text
+     * of the segment. Commit that synchronously, between two measurements, so
+     * the reader keeps their place through the reflow (#367). `commit` carries
+     * any state the caller wants applied in the same pass.
+     */
+    const closeEditor = (commit?: () => void) => {
+        const anchor = captureScrollAnchor(rootRef.current);
+        flushSync(() => {
+            commit?.();
+            setIsEditing(false);
+        });
+        restoreScrollAnchor(anchor, rootRef.current);
+    };
+
     const handleEdit = async (e: React.FormEvent | React.MouseEvent) => {
         e.preventDefault();
         const originalText = localUtterance.text;
@@ -153,19 +188,18 @@ const UtteranceC: React.FC<{
 
         // Nothing changed - just close the editor
         if (!textChanged && !timestampsChanged) {
-            setIsEditing(false);
+            closeEditor();
             return;
         }
 
         // Optimistic update - immediate
-        setLocalUtterance({ 
-            ...localUtterance, 
+        closeEditor(() => setLocalUtterance({
+            ...localUtterance,
             text: editedText,
             startTimestamp: editedStartTime,
             endTimestamp: editedEndTime
-        });
-        setIsEditing(false);
-        
+        }));
+
         // Background save
         try {
             // Update text first
@@ -221,10 +255,11 @@ const UtteranceC: React.FC<{
     };
 
     const handleCancel = () => {
-        setIsEditing(false);
-        setEditedText(localUtterance.text);
-        setEditedStartTime(localUtterance.startTimestamp);
-        setEditedEndTime(localUtterance.endTimestamp);
+        closeEditor(() => {
+            setEditedText(localUtterance.text);
+            setEditedStartTime(localUtterance.startTimestamp);
+            setEditedEndTime(localUtterance.endTimestamp);
+        });
     };
 
     const setStartTimeToCurrentVideo = () => {
@@ -259,7 +294,7 @@ const UtteranceC: React.FC<{
 
     if (isEditing) {
         return (
-            <div className="relative w-full py-1 border border-blue-300 rounded-md p-2 bg-blue-50/30">
+            <div ref={setRootRef} className="relative w-full py-1 border border-blue-300 rounded-md p-2 bg-blue-50/30">
                 {/* Text Editor */}
                 <form onSubmit={handleEdit} className="relative">
                     <textarea
@@ -292,15 +327,7 @@ const UtteranceC: React.FC<{
                             }
                         }}
                         className="w-full resize-none border border-gray-300 rounded px-2 py-1 pr-16 text-sm min-h-[2.5em] transcript-text"
-                        ref={(el) => {
-                            if (el) {
-                                el.focus();
-                                if (clickOffsetRef.current !== null) {
-                                    el.selectionStart = el.selectionEnd = clickOffsetRef.current;
-                                    clickOffsetRef.current = null;
-                                }
-                            }
-                        }}
+                        ref={focusEditor}
                         rows={Math.max(1, editedText.split('\n').length)}
                         style={{
                             height: 'auto',
@@ -417,6 +444,7 @@ const UtteranceC: React.FC<{
     // and renders one set of items instead of one Radix Menu per utterance.
     const utteranceSpan = (
         <span
+            ref={setRootRef}
             className={cn(className, emptyUtteranceClass)}
             id={localUtterance.id}
             data-utterance-id={localUtterance.id}
--- a/src/components/meetings/transcript/Utterance.tsx
+++ b/src/components/meetings/transcript/Utterance.tsx
@@ -1,6 +1,7 @@
 "use client";
 import { Utterance } from "@prisma/client";
-import React, { useEffect, useRef, useState } from "react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
+import { flushSync } from "react-dom";
 import { useTranslations } from 'next-intl';
 import { useVideoActions } from "../VideoProvider";
 import { useTranscriptOptions } from "../options/OptionsContext";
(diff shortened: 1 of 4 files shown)


src/components/meetings/transcript/Utterance.tsx
```diff
@@ -80,6 +87,19 @@ const UtteranceC: React.FC<{
         setEditedEndTime(utterance.endTimestamp);
     }, [utterance]);
 
+    // Stable identity, so React runs this when the editor actually mounts. An
+    // inline callback is re-attached on every render, which re-ran `focus()` on
+    // every keystroke — and each call asks the browser to scroll the textarea
+    // into view.
+    const focusEditor = useCallback((el: HTMLTextAreaElement | null) => {
```
kouloumos: This is the one blocker.

Stabilising the ref removes a focus restoration the editor was leaning on. Before, React re-attached this callback on every render and re-ran `focus()`, so any click that moved focus onto a toolbar button was silently undone on the next render. Now focus stays where the click put it, and the shortcuts bound to the textarea's `onKeyDown` — Enter, Escape, Shift+[/], Shift+arrows — are unreachable until you click back into the text.

Same steps on both builds, base on the left. Watch the `focus:` and `editors open:` readout: after the timestamp button is clicked, base returns focus to the textarea and Escape closes the editor; here focus stays on the button, and neither Escape nor Enter does anything.

https://github.com/user-attachments/assets/da7d2cd7-b9c7-41dc-838e-ec477c85071c

| | base | this branch |
|---|---|---|
| focus after clicking the timestamp button | `TEXTAREA` (1 `focus()` call) | `BUTTON` (0) |
| Escape then closes the editor | yes | no |
| Enter then saves | yes | no — editor stays open |

The skip-interval dropdown does the same thing, since it also writes to `options` and so re-renders this component. The speed control doesn't — it goes through `useVideo()` rather than the options context, so it never re-rendered the utterance and there was no restoration to lose. Keyboard `Shift+[` is fine throughout, because focus never leaves the textarea.

Three ways out, in increasing order of coverage:

- `onMouseDown={e => e.preventDefault()}` on the two timestamp buttons — smallest, but leaves the skip-interval case.
- Keep the element in `focusEditor` and re-focus it after the timestamp setters.
- Drop the `useCallback`. The comment above overstates what it buys: `focus()` on the already-focused element returns early per the focusing steps, and I counted zero `scrollIntoView` calls on either build across ten keystrokes — so those per-keystroke calls were wasted work, not a source of scrolling.

src/lib/utils/scrollAnchor.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Measure with `captureScrollAnchor`, commit the DOM change synchronously
+ * (`flushSync`), then hand the replacement node to `restoreScrollAnchor`.
+ */
+export interface ScrollAnchor {
+    container: HTMLElement;
+    top: number;
+}
+
+export function captureScrollAnchor(element: HTMLElement | null): ScrollAnchor | null {
+    const container = element?.closest<HTMLElement>('[data-scroll-container]');
```
kouloumos: Third place `[data-scroll-container]` is spelled out — the meeting layout sets it, and `Transcript.tsx:52` already reads it with the same `closest()` call. Worth exporting the lookup from here and having Transcript import it, so this module owns the convention instead of being its second independent reader of a magic string.

kouloumos (changes requested):
Verified against a base build running on the same data: saving an utterance mid-segment moved the scroll container 625px before, 23px here. It does what it says. Same steps on both, base on the left:

https://github.com/user-attachments/assets/0d19d594-cbfa-4b72-a227-4fcf2fc41b8d

One thing to fix first. Making the textarea's ref stable removes a focus restoration the rest of the editor was leaning on. The unstable inline callback re-ran `focus()` on every render, so whenever a click moved focus onto a toolbar control the next render silently pulled it back into the textarea. That no longer happens — and since every editor shortcut is bound to the textarea's `onKeyDown`, they all go dead until you click back into the text. Enter stops saving, Escape stops cancelling, and nothing indicates why. Reachable through the timestamp buttons and through the skip-interval dropdown.

I didn't exercise the failed-save path, and I only tested at 1440px.

src/components/meetings/transcript/Utterance.tsx
christosporios: Good catch, and thanks for the side-by-side — the focus readout makes it unarguable.

Took your third option: the `useCallback` is gone and the inline callback is back verbatim. Your measurement is the deciding factor. My comment claimed each `focus()` asked the browser to scroll the element into view; you counted zero `scrollIntoView` calls on either build, and `focus()` on the already-focused element returns early per the focusing steps. So the stabilisation bought nothing measurable and cost working shortcuts — not a trade worth making, least of all inside a scroll fix.

I've left a comment on the ref explaining what the changing identity is actually load-bearing for, so the next person doesn't tidy it away the same way I did:

```tsx
// Deliberately inline: the changing identity makes React
// re-attach this on every render, which re-focuses the
// textarea. Every editor shortcut below hangs off its
// onKeyDown, so anything that moves focus onto a control
// in here — the timestamp buttons, the skip-interval
// dropdown — would otherwise leave Enter and Escape dead
// until the user clicked back into the text.
```

That restores the previous behaviour exactly rather than replacing the accident with something deliberate. The restoration is still incidental to a re-render, so it holds only for controls that happen to re-render this component — which is why the speed control never had it. Worth making explicit at some point, but as its own change rather than folded in here.

On the failed-save path you didn't exercise: `closeEditor` has already run by then, so the `catch` only reverts local state and toasts into an editor that's closed. It can reflow the paragraph when the reverted text differs in length, and nothing re-pins that — but it's an async failure well after the interaction, not the frame this PR is about.

---

src/lib/utils/scrollAnchor.ts
christosporios: Agreed. Exported the lookup and pointed `Transcript.tsx:52` at it:

```ts
export function getScrollContainer(element: Element | null | undefined): HTMLElement | null {
    return element?.closest<HTMLElement>('[data-scroll-container]') ?? null;
}
```

`captureScrollAnchor` now goes through it too, so the selector appears once outside the layout that sets it. Covered by a test for both the hit and the miss.

---

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/e9e4270c7e575cedc22092b91dfa39ebd295f4fa..607ba4d5fc74aae9d7ab29045631f394c2eb2841) to address both review comments:

- Reverted the textarea ref stabilisation — the inline callback is back verbatim, with a comment recording that its changing identity is what re-focuses the editor after a click lands on a control inside it. This was the blocker: without it, Enter and Escape went dead after using the timestamp buttons or the skip-interval dropdown.
- Added `getScrollContainer()` to `scrollAnchor.ts` and pointed `Transcript.tsx:52` at it, so `[data-scroll-container]` is spelled out once outside the layout that sets it.

The PR description's "Also: the textarea's ref callback" section no longer describes the change — the measurement in review showed that `focus()` on an already-focused element returns early and issues no `scrollIntoView`, so the premise was wrong. The scroll fix itself is untouched.

Suite is 1272 passing (one new test covering `getScrollContainer` hit and miss), `tsc --noEmit` and `eslint` clean, build compiles and type-checks.

---

kouloumos (approved):
Re-measured rather than read: after the timestamp button focus returns to the textarea and Escape closes the editor, matching main; skip-interval too. The save still holds the viewport to 23px where main moves 625px, so the revert didn't cost the fix anything.

The comment on the inline callback is what I'd most have wanted here. It's what stops this being tidied up again in six months.

PR #595, opened 2026-08-07
Send the city's country on summarize and processAgenda

christosporios:
The tasks backend now accepts an optional `country` on `/summarize` and `/processAgenda` — an ISO 3166-1 alpha-2 code that restricts geocoding of subject locations to that country. Without it, everything is geocoded as if it were in Greece, which is why locations for French and Serbian cities come back wrong.

## Where the country comes from

Derived from the city's **realm**, not from a new column and not from `cityLanguage`:

- A realm is exactly one country (`greece→GR`, `france→FR`, `cyprus→CY`, `serbia→RS`), and `City.realm` already carries it. A separate `City.country` column could only ever drift out of sync with the realm the city is actually served under.
- `cityLanguage` can't stand in for it: `fr` spans France/Belgium/Switzerland and `sr` spans Serbia/Bosnia/Montenegro.

`REALMS` already held the ISO code for Google Places geocoding, so this is one lookup away. It now stores the code uppercase — the form the tasks backend wants — and `getRealmGeocoding` lowercases it for Google's ccTLD-style `region`/`components=country:` params. Both existing geocoding call sites behave exactly as before.

## Changes

- `src/lib/apiTypes.ts` — new `Country` union, self-contained like the neighboring `CityLanguage`; `country` field on `ProcessAgendaRequest` and on `RequestOnTranscript` (which `SummarizeRequest` extends).
- `src/lib/realm.ts` — new `getRealmCountry(realm)`.
- `src/lib/db/utils.ts` and `src/lib/tasks/processAgendaInternal.ts` — send it. These two builders are the only paths into the two endpoints (batch rerun and the admin actions both funnel through them), so every request now carries a country.

No schema change, no migration.

## Testing

`npm test` (1269 tests) and `npx tsc --noEmit` are clean. New coverage:

- `getRealmCountry` per realm, plus a loop over `ALL_REALMS` so a newly added realm can't silently geocode as Greece.
- `getRealmGeocoding` still returns the lowercase ccTLD form.
- A Serbian city in `db-utils.test.ts` asserting `country: 'RS'` alongside `cityLanguage: 'sr'` — the pair the language can't derive. (That fixture mocked a city without a `realm`; completed it.)

## Note for review

`fixTranscript` and `generatePodcastSpec` build on the same shared body types, so they now include `country` too. Harmless if the backend ignores unknown fields — which it must already do, since `generatePodcastSpec` posts the full summarize payload — but worth a glance if either endpoint validates strictly.

---

> [!NOTE]
> **Low Risk**
> Additive contract fields on existing task request builders with no schema migration; main caveat is downstream tasks endpoints accepting the new field (and any Google geocoding callers if they relied on lowercase realm country codes).
> 
> **Overview**
> Task payloads for **summarize** (via `getRequestOnTranscriptRequestBody`) and **processAgenda** now include a **`country`** ISO 3166-1 alpha-2 code so the tasks backend can geocode subject locations in the correct country instead of defaulting to Greece.
> 
> The code is derived from **`City.realm`** through new **`getRealmCountry`**, not from `cityLanguage`. **`REALMS`** country codes are stored **uppercase** and typed as the new **`Country`** union on **`ProcessAgendaRequest`** and **`RequestOnTranscript`** (and thus summarize / fixTranscript / podcast builders that extend it). Tests cover per-realm mapping, an **`ALL_REALMS`** loop, and a Serbian city asserting **`RS`** with **`sr`** language.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd72f9e67a1458c48705141f7684d69702636a48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds an uppercase realm-derived country code to transcript and agenda task requests so backend geocoding is scoped to the correct country.
- Adds the shared `Country` request type and country fields.
- Derives task countries from the city realm configuration.
- Preserves the existing realm geocoding representation and adds mapping/request tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/realm.ts | Stores uppercase ISO country codes, exposes `getRealmCountry`, and keeps realm display and geocoding behavior consistent with the updated representation. |
| src/lib/apiTypes.ts | Adds the supported country-code union to agenda and transcript-derived task request contracts. |
| src/lib/db/utils.ts | Adds the city realm’s country to transcript-related task request bodies using the required alias import convention. |
| src/lib/tasks/processAgendaInternal.ts | Adds the city realm’s country to process-agenda requests using the required alias import convention. |
| src/lib/__tests__/db-utils.test.ts | Replaces the previously flagged `as any` city fixture and verifies Serbian realm and language values are transmitted independently. |
| src/lib/__tests__/realm.test.ts | Covers every configured realm’s uppercase country mapping and validates representative geocoding outputs. |

<sub>Reviews (3): Last reviewed commit: ["feat(tasks): send the city&#39;s country on ..."](https://github.com/schemalabz/opencouncil/commit/dd72f9e67a1458c48705141f7684d69702636a48) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51160083)</sub>

### DIFF
--- a/src/lib/apiTypes.ts
+++ b/src/lib/apiTypes.ts
@@ -20,6 +20,12 @@ export interface TaskRequest {
 // the Prisma `CityLanguage` enum.
 export type CityLanguage = 'el' | 'fr' | 'sr';
 
+// ISO 3166-1 alpha-2 code (uppercase) of the country a task's city is in. The
+// backend restricts geocoding of subject locations to it; without it everything
+// is geocoded as if it were in Greece. Comes from the city's realm, not its
+// language — see `getRealmCountry`.
+export type Country = 'GR' | 'FR' | 'CY' | 'RS';
+
 /*
  * System endpoints
  */
@@ -125,6 +131,7 @@ export interface ProcessAgendaRequest extends TaskRequest {
     topicLabels: TopicLabelInfo[];
     cityName: string;
     cityLanguage: CityLanguage;
+    country: Country;
     date: string;
 }
 
@@ -258,6 +265,7 @@ export interface RequestOnTranscript extends TaskRequest {
     topicLabels: TopicLabelInfo[];
     cityName: string;
     cityLanguage: CityLanguage;
+    country: Country;
     administrativeBodyName: string | null;
     partiesWithPeople: {
         name: string;
--- a/src/lib/realm.ts
+++ b/src/lib/realm.ts
@@ -1,4 +1,5 @@
 import { Realm } from '@prisma/client';
+import type { Country } from '@/lib/apiTypes';
 
 /**
  * Realm (tenant) configuration. A single deployment serves all domains off one
@@ -15,15 +16,15 @@ import { Realm } from '@prisma/client';
  * (`proxy.ts`). The request-scoped resolver lives in `realm.server.ts`.
  */
 export const REALMS = {
-    greece: { domain: 'opencouncil.gr', defaultLocale: 'el', country: 'gr' },
-    france: { domain: 'opencouncil.fr', defaultLocale: 'fr', country: 'fr' },
-    cyprus: { domain: 'opencouncil.cy', defaultLocale: 'el', country: 'cy' },
+    greece: { domain: 'opencouncil.gr', defaultLocale: 'el', country: 'GR' },
+    france: { domain: 'opencouncil.fr', defaultLocale: 'fr', country: 'FR' },
+    cyprus: { domain: 'opencouncil.cy', defaultLocale: 'el', country: 'CY' },
     // Serbian is digraphic: `sr` (Cyrillic) is the default, `sr-Latn` is the
     // realm-exclusive Latin variant reachable via the script switcher.
-    serbia: { domain: 'opencouncil.rs', defaultLocale: 'sr', extraLocales: ['sr-Latn'], country: 'rs' },
+    serbia: { domain: 'opencouncil.rs', defaultLocale: 'sr', extraLocales: ['sr-Latn'], country: 'RS' },
 } as const satisfies Record<
     Realm,
-    { domain: string; defaultLocale: 'el' | 'fr' | 'sr'; country: string; extraLocales?: readonly string[] }
+    { domain: string; defaultLocale: 'el' | 'fr' | 'sr'; country: Country; extraLocales?: readonly string[] }
 >;
 
 /**
@@ -39,7 +40,7 @@ export const ALL_REALMS = Object.keys(REALMS) as Realm[];
  * `REALMS` — no per-realm labels or translation keys to maintain anywhere.
  */
 export function getRealmDisplayName(realm: Realm, locale: string): string {
-    const country = REALMS[realm].country.toUpperCase();
+    const country = REALMS[realm].country;
     return new Intl.DisplayNames([locale], { type: 'region' }).of(country) ?? country;
 }
 
@@ -218,6 +219,17 @@ export function getRealmDomain(realm: Realm): string {
  */
 export function getRealmGeocoding(realm: Realm): { country: string; language: string } {
     return { country: REALMS[realm].country, language: REALMS[realm].defaultLocale };
+}
+
+/**
+ * ISO 3166-1 alpha-2 country code of a realm (uppercase), as the tasks backend
+ * expects it on /summarize and /processAgenda to restrict geocoding of subject
+ * locations. Derived from the realm rather than the city's language: `fr` spans
+ * France/Belgium/Switzerland and `sr` spans Serbia/Bosnia/Montenegro, while a
+ * realm is exactly one country.
+ */
+export function getRealmCountry(realm: Realm): Country {
+    return REALMS[realm].country;
 }
 
 /** Fallback map center/zoom for a realm, used when a city has no stored geometry. */
--- a/src/lib/__tests__/db-utils.test.ts
+++ b/src/lib/__tests__/db-utils.test.ts
@@ -39,6 +39,12 @@ const MEETING_ID = 'meeting-1';
 const ADMIN_BODY_ID = 'admin-body-1';
 const MEETING_DATE = new Date('2024-06-15T10:00:00Z');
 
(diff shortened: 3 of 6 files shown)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/7228eb6d1562230576c314d7da6308245690705a..f87268254bcd392ad8fa7bd9669149117bb35884) to address @greptile-apps' review:

- **`@/` import convention** — the three new imports (`realm.ts`, `db/utils.ts`, `processAgendaInternal.ts`) now use the alias. I had originally matched each file's local style, since both files import exclusively via relative paths today, but CLAUDE.md states the convention explicitly and new code should follow it.
- **Untyped city fixture** — `db-utils.test.ts` no longer casts its city mocks to `any`. A small `mockCity()` helper types them as `Partial<...>` of `getCity`'s actual return, so `realm` and `language` are checked against the real enums. This one had teeth: the fixture that broke while writing this PR broke precisely because a missing `realm` sailed through `as any` and only surfaced as an `undefined` at runtime. With the helper, `realm: 'serbiaa'` is now a compile error (`TS2820: Did you mean "serbia"?`). The remaining `as any` casts in that file are pre-existing and out of scope here.

No production behavior changed. Verified in a clean checkout of this branch alone (no other working-tree changes present): `npx tsc --noEmit` clean, `npm test` 1269 passing.

src/lib/realm.ts
```diff
@@ -217,7 +218,19 @@ export function getRealmDomain(realm: Realm): string {
  * right country (e.g. France on opencouncil.fr) instead of hardcoding Greece.
  */
 export function getRealmGeocoding(realm: Realm): { country: string; language: string } {
-    return { country: REALMS[realm].country, language: REALMS[realm].defaultLocale };
+    // Google's `region`/`components=country:` want the ccTLD form (lowercase).
+    return { country: REALMS[realm].country.toLowerCase(), language: REALMS[realm].defaultLocale };
+}
```
kouloumos, line 222: Both halves of this are wrong. `region` isn't used anywhere in the repo (only S3 config), and `components=country:` is case-insensitive — I ran the actual Places Autocomplete call site both ways against our key:

```
components=country:RS → "Kneza Mihaila, Београд, Србија"  place_id EitLbmV6YSBNaWhhaWxh…
components=country:rs → "Kneza Mihaila, Београд, Србија"  place_id EitLbmV6YSBNaWhhaWxh…
```

Identical predictions, identical place_id. So the `.toLowerCase()` is a no-op that exists only to undo the flip in `REALMS`, and the comment invents a Google requirement to justify it.

Uppercase is the right canonical form — it's ISO 3166-1, and it's what the tasks backend uses (`DEFAULT_COUNTRY = 'GR'`, its tests assert `country:FR`). So keep the flip, drop the lowercasing:

```suggestion
export function getRealmGeocoding(realm: Realm): { country: string; language: string } {
    return { country: REALMS[realm].country, language: REALMS[realm].defaultLocale };
}
```

The `getRealmGeocoding` test below asserts the lowercase form, so it needs updating alongside this.

src/lib/apiTypes.ts
```diff
@@ -20,6 +20,13 @@ export interface TaskRequest {
 // the Prisma `CityLanguage` enum.
 export type CityLanguage = 'el' | 'fr' | 'sr';
 
+// ISO 3166-1 alpha-2 code (uppercase) of the country a task's city is in. The
+// backend restricts geocoding of subject locations to it; without it everything
+// is geocoded as if it were in Greece. NOT derivable from `CityLanguage` — `fr`
```
kouloumos: The `fr`/`sr` argument is made in full three times: here, in `getRealmCountry`'s JSDoc, and again at `db-utils.test.ts:103`. Three copies drift — when the derivation changes, one gets updated and the others quietly start contradicting the code.

`getRealmCountry` is where the "realm, not language" decision is actually implemented, so it should keep the argument. This comment already points there (`see getRealmCountry`) — it just restates the whole case first. And the test comment is covered by the test's own name.

Two deletions, no rewrites.

kouloumos (approved):
Two inline comments — the first is worth acting on, the second is hygiene. Good to go from me once the first one is fixed.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/f87268254bcd392ad8fa7bd9669149117bb35884..dd72f9e67a1458c48705141f7684d69702636a48) with both review points applied.

**1. Dropped the `.toLowerCase()`** — agreed, uppercase is the right canonical form. Updated the `getRealmGeocoding` test alongside it.

One correction on the reasoning, since it changes *why* this is safe: `region` **is** used, at [`src/app/api/geocode/route.ts:17`](https://github.com/schemalabz/opencouncil/blob/main/src/app/api/geocode/route.ts#L17), and it's fed by `getRealmGeocoding` — so this call site does now receive `RS` instead of `rs`. Your evidence covered `components=country:` (Places Autocomplete), not `region` (Geocoding API), so I ran that endpoint against our key too:

```
Toledo  region=es -> Toledo, Spain     ChIJkwyrlqwLag0RiQIn2fdIshM
Toledo  region=us -> Toledo, OH, USA   ChIJeU4e_C2HO4gRRcM6RZ_IPHw
Toledo  region=ES -> Toledo, Spain     ChIJkwyrlqwLag0RiQIn2fdIshM
Toledo  region=US -> Toledo, OH, USA   ChIJeU4e_C2HO4gRRcM6RZ_IPHw
```

The first two differ, so `region` is live and does bias results; the second pair matches the first exactly, so it's case-insensitive. Same result for `rs`/`RS`. The change is inert either way — just not for the stated reason.

**2. Removed the two duplicate copies** of the `fr`/`sr` argument, in `apiTypes.ts` and `db-utils.test.ts`. `getRealmCountry`'s JSDoc keeps it, and the `apiTypes.ts` comment now just points there.

`npx tsc --noEmit` clean, `npm test` 1269 passing.

Issue #594, opened 2026-08-07
Serbian assembly groups are not parties


christosporios:
In Serbia, councillors do not sit in parties. They sit in **assembly groups**
(одборничка група), and one group can be a coalition of several parties.

We model this with the `Party` table, so a coalition group becomes one `Party`
row and the parties inside it disappear. There is nowhere to record that a
councillor belongs to party X while sitting in group Y.

For the Niš demo we only renamed the labels in the Serbian translation files
("Party" -> "Assembly Group", "Party Leader" -> "Assembly Group Leader"). The
screen now reads correctly, but the data underneath is still wrong.

Two things to decide:

1. **Data model.** Do we add a real "group" concept that can contain parties,
   or keep one flat table and accept that coalition members lose their party?
   This matters before a Serbian city becomes a paying customer.

2. **Naming per client.** Zvezdana also asked whether terminology can follow
   the client: a *city* and a *municipality* use different words, and Greece
   and Serbia use different words again. Right now the label is fixed per
   locale. If we make it configurable, that decision should cover the
   party/group naming too, so we do not do it twice.

Context: email thread "Niš Demo | OpenCouncil", 6 Aug 2026.

PR #596, opened 2026-08-07
fix(i18n): correct Serbian terminology and replace hardcoded Greek strings

christosporios:
Two fixes to the Serbian UI, both reported by our Serbian partner while
preparing the Niš demo.

## 1. Assembly-group terminology

Serbian councillors do not sit in parties. They sit in **assembly groups**
(одборничка група), and a group can be a coalition of several parties. The
catalogue called them `странка` throughout, which names the wrong entity.

`partyLeader` was a mistranslation rather than a synonym. The source strings
are `Επικεφαλής` ("Head") and `Party Leader`, but the Serbian asserted
**"Председник странке"** — *President of the Party*, an office nobody on the
page holds, since a party president is usually not a councillor. The short
form was worse: bare **"Председник"**, identical to the assembly president's
title two keys above it in the same file.

Every form is declined rather than substituted, including the ICU plural
branches:

| case | before | after |
|---|---|---|
| nom. sg. | Странка | Одборничка група |
| gen. sg. | странке | одборничке групе |
| acc. sg. | странку | одборничку групу |
| instr. sg. | странком | одборничком групом |
| gen. pl. | странака | одборничких група |

## 2. Hardcoded Greek strings

Serbian pages rendered Greek in places. Key parity was already complete across
all four locales, so nothing was falling back — these strings were **literals
in the components**, bypassing i18n entirely. Among them the exact one she
reported (`Πρόσωπα εκτός παρατάξεων` on the city parties page) and the whole
sign-in form.

Missing keys added in el/en/sr/fr and the components routed through them:
sign-in, person and party pages, transcript, meeting cards, share dropdown,
playback options, and the admin controls used to assign speakers and set
subject categories.

Two call sites needed a translator threaded in rather than a literal swap:

- `getWithdrawnLabel` now takes `t` first, matching `getAgendaLabel` in the
  same module.
- `SubjectCardFooter` takes `minutesText` as a prop, matching the existing
  `withdrawnLabel`, because it is a hook-free presentational shell shared with
  the embed widget — the same contract `SubjectCardContent` documents. (An
  earlier revision of this description claimed the embed renders outside the
  i18n provider. That is wrong: `(embed)` is a route group under `[locale]`
  and inherits `NextIntlClientProvider`.)

## Deliberately not included

- **The minutes docx and its preview keep their Greek.** That feature is Greek
  by construction — Greek headings, Greek date locale, ΑΔΑΜ/ΚΗΜΔΗΣ/ΑΔΑ
  identifiers — and never renders outside the Greek realm. The label is
  inlined there with a comment rather than pretending it is localised.
- **The magic-link email is still entirely Greek** — subject, body and the
  `AuthEmail` template. This is not a string swap: `sendVerificationRequest`
  has no locale at that point, so it needs the locale derived from the request
  host, the way `signInUrlForRequest` already derives the domain. A Serbian
  user signing in today receives a Greek email. Worth its own PR.
- **`generateMetadata` on the person and party pages is still Greek**, so a
  Serbian user gets a Serbian party name inside a Greek `<title>`. Left out
  deliberately: it is server-side metadata across several routes and needs
  `getTranslations` threaded through each, which is a different shape of change
  from the client components here. Its own PR.
- **Renaming the label is not a model fix.** A coalition group still collapses
  into one `Party` row and its constituent parties are unrepresentable. The
  screen now reads correctly while the data underneath does not. Tracked in
  #594, along with her related request for per-client terminology (city vs.
  municipality), since making labels configurable and renaming party → group
  are the same decision.

## Default-locale (Greek) copy changes

Two Greek strings change, because routing them through existing keys was
preferable to adding near-duplicates:

| where | before | after |
|---|---|---|
| `SubjectAdminControls` category select | Ημερησίας | Ημερησίας διάταξης |
| `DecisionsPanel` separator | Εκτός Ημ. Διάταξης | Εκτός ημερησίας |

The select trigger moved from `w-[160px]` to `min-w-[160px]`, since the longer
label lands on the old fixed width.

## Notes for review

- `Header.orChatWithAI` was added and then dropped during rebase: #587 removed
  that chat button, so the key would have been dead on arrival.
- The three remaining Greek strings in the Serbian catalogue (ΑΔΑΜ, ΚΗΜΔΗΣ,
  ΑΔΑ) are Greek procurement identifiers, not untranslated UI.

## Verification

`npx tsc --noEmit` clean; 96 suites / 1314 tests pass. Catalogue parity
verified programmatically at 2096 keys across all four locales, with no
missing or extra keys. Serbian Latin is derived from Cyrillic at load time and
is unaffected — `transliterateCatalog` only maps Cyrillic code points, so the
ICU syntax in the new plural strings is untouched.

Not verified in a browser: no dev server was running locally.

Refs #594

---

> [!NOTE]
> **Low Risk**
> Changes are mostly copy and i18n wiring across locales; minutes export stays on the existing Greek-only helper, so runtime risk is low aside from missing translation keys or ICU plural mistakes.
> 
> **Overview**
> **Serbian terminology** replaces `странка` / party-leader wording with **одборничка група** and **шеф одборничке групе** across `messages/sr.json` and related modular catalogs, with glossary and `sr-glossary.test.ts` bans/requirements updated so the wrong entity and “председник” leader titles cannot creep back in.
> 
> **Internationalization** adds matching keys in el/en/fr/sr and wires components that previously showed Greek literals: sign-in (`SignIn`), share menu (`ShareDropdown` with per-context full sentences), meeting overview/upcoming cards (`CouncilMeeting`), transcript empty/procedural states, person/party pages, meeting options, list/search errors, and subject withdrawn labels. **`getWithdrawnLabel`** now takes a `Subject` translator like other helpers; **`SubjectCardFooter`** receives a pre-pluralized **`minutesText`** for embed and app cards.
> 
> **Minutes docx/preview** keep Greek withdrawn TOC text via **`getWithdrawnLabelGreek`** in `lib/minutes/builders`, separate from the localized UI path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ff3f33d1836f5b98c80a1aded9ada9c55910945. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR corrects Serbian assembly-group terminology and routes previously hardcoded Greek interface text through locale catalogs.
- Adds corresponding translations across Greek, English, French, and Serbian catalogs.
- Localizes meeting, transcript, person, party, sign-in, sharing, playback, and administrative UI.
- Threads translated withdrawn and count-sensitive duration labels through shared subject-card components.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/subject/SubjectCardFooter.tsx | The shared footer now receives a fully localized, count-sensitive speaking-duration string. |
| src/components/subject-card.tsx | The application subject card supplies localized withdrawn and ICU-formatted duration labels. |
| src/components/embed/EmbedSubjectCard.tsx | The embed card supplies the same count-sensitive localized duration label as the main application. |
| messages/sr.json | Serbian terminology and new localized UI messages are added with count-sensitive ICU forms. |
| src/components/user/sign-in.tsx | The sign-in interface now obtains its user-facing copy from the SignIn translation namespace. |

<sub>Reviews (3): Last reviewed commit: ["fix(i18n): replace hardcoded Greek strin..."](https://github.com/schemalabz/opencouncil/commit/9ff3f33d1836f5b98c80a1aded9ada9c55910945) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51161668)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/afe4dcc37b314d5e8f0ebfd80442b3ddc1c9c8d9..7adb011f4f0fc92c761330fd92d1d1fc5af072b0) to address the review. All three findings were valid; each is squashed into its own commit.

**Bugbot — `Party.leaderLabel` still said "Председник: "** — correct, and it was the same defect the PR set out to fix. I missed it because my replacements keyed on `странка`, and this string is a standalone "Председник: " with no group noun attached. It renders directly beside the group head's name on the party page, so the page said *President:* next to someone the rest of the UI now calls group leader.

Now "Шеф групе: ", matching `partyLeaderShort`. I audited every remaining `Председник` in the Serbian catalogue: the only other one is `Person.president`, which is the actual assembly president (`Πρόεδρος` / `President`) and is correct as-is.

**Greptile — count-insensitive minutes label** — valid, though pre-existing rather than introduced here: `main` had `{stats.minutes} λεπτά`, equally invariant. Since this PR is specifically about Serbian being grammatically right, it was worth fixing rather than carrying forward.

Added `Subject.minutesCount` as an ICU message. I did **not** convert `Subject.minutes`, because three other call sites (`PilotPage`, `subject.tsx`, `SubjectDemo`) use it as a bare standalone noun under a separately-rendered number, and making it ICU would break them. `SubjectCardFooter` now takes the fully-formatted `minutesText` instead of a unit noun.

Verified rendering through `intl-messageformat`:

| n | el | en | sr |
|---|---|---|---|
| 1 | 1 λεπτό | 1 minute | 1 минут |
| 2 | 2 λεπτά | 2 minutes | 2 минута |
| 5 | 5 λεπτά | 5 minutes | 5 минута |
| 21 | 21 λεπτά | 21 minutes | 21 минут |

Serbian's `one` category correctly catches 21, so "21 минут" rather than "21 минута".

**Greptile — misgrouped `next-intl` imports** — valid per CLAUDE.md. I moved only the four lines I added up into the framework group. I deliberately did not reorder the surrounding imports: several of these files already interleave local and third-party imports (`List.tsx` has a local import above `lucide-react`), and CONTRIBUTING.md asks that formatting fixes not ride along with code changes. A tidy-up pass would be its own commit.

`tsc` clean; 96 suites / 1314 tests pass, confirmed stable across three consecutive runs.

src/components/persons/PersonBadge.tsx
```diff
@@ -292,13 +294,13 @@ function PersonBadge({
...
                                     <CommandItem
-                                        onSelect={() => handleSetLabel(nextUnknownLabel || "Άγνωστος Ομιλητής")}
+                                        onSelect={() => handleSetLabel(nextUnknownLabel || tEditing('unknownSpeaker'))}
                                         className="flex items-center gap-2"
                                     >
                                         <div className="w-10 h-10 relative shrink-0 flex items-center justify-center bg-muted rounded-full">
                                             <span className="text-xs font-medium">?</span>
                                         </div>
-                                        <span className="font-medium">{nextUnknownLabel || "Άγνωστος Ομιλητής"}</span>
+                                        <span className="font-medium">{nextUnknownLabel || tEditing('unknownSpeaker')}</span>
```
kouloumos: This fallback can't be reached, so the Serbian editor still sees `Άγνωστος Ομιλητής 1`.

The popover only renders under `editable && availablePeople`, and the one production call site that satisfies that — `SpeakerSegment.tsx:315-323` — always passes `nextUnknownLabel`, which comes from a `useMemo` that always returns `buildUnknownSpeakerLabel(n)`. So `editing.actions.unknownSpeaker` is dead in all four locales.

Worth not just flipping the operands, though. `UNKNOWN_SPEAKER_LABEL` (`src/lib/utils.ts:47`) is a persisted value rather than copy: it's written server-side by `transcribe.ts:99` and prefix-matched in three places — `SpeakerSegment.tsx:217`, `EditingModeBar.tsx:70`, `SpeakersOverviewSheet.tsx:111`. `handleSetLabel` persists through `onLabelChange`, so if this branch ever did fire, a Serbian write would drop that segment out of unknown-speaker navigation, mis-group it in the overview sheet, and reset the auto-numbering.

So either revert these two lines and treat it like the docx exclusion, or split display from the stored value. As it stands it's the awkward middle — no user-visible fix, plus a latent trap.

src/components/meetings/subject/SubjectAdminControls.tsx
```diff
@@ -74,14 +76,14 @@ export function SubjectAdminControls({ subject, cityId, meetingId }: SubjectAdmi
                                     <SelectValue />
                                 </SelectTrigger>
                                 <SelectContent>
-                                    <SelectItem value="agenda">Ημερησίας</SelectItem>
-                                    <SelectItem value="beforeAgenda">Προ ημερησίας</SelectItem>
-                                    <SelectItem value="outOfAgenda">Εκτός ημερησίας</SelectItem>
+                                    <SelectItem value="agenda">{t('categories.agenda.shortLabel')}</SelectItem>
```
kouloumos, line 79: This changes the Greek label, not just where it comes from: the literal was `Ημερησίας`, and `Subject.categories.agenda.shortLabel` is `Ημερησίας διάταξης`. The other two options are byte-identical, so it's only this one.

It also lands exactly on the trigger's limit. `h-7 w-[160px]` minus 24px padding, the 16px chevron and the 4px gap leaves 116px; measured in the app's own stylesheet, `Ημερησίας διάταξης` is 116px (`Ημερησίας` was 61px). Any font fallback tips it into the base `line-clamp-1`. `min-w-[160px]` would settle it.

src/components/meetings/admin/DecisionsPanel.tsx
```diff
@@ -587,7 +588,7 @@ export function DecisionsPanel({ open, onOpenChange }: DecisionsPanelProps) {
                                     <Fragment key={subject.id}>
                                         {showOutOfAgendaSeparator && (
                                             <div className="pt-3 pb-1 text-xs font-medium text-muted-foreground uppercase tracking-wide">
-                                                Εκτός Ημ. Διάταξης
+                                                {tSubject('categories.outOfAgenda.shortLabel')}
```
kouloumos, line 591: Same class as `SubjectAdminControls.tsx:79` — the Greek copy changes here too, `Εκτός Ημ. Διάταξης` → `Εκτός ημερησίας` (rendered uppercase). No objection, but it's a default-locale UI change riding on an i18n PR, so worth a line in the description.

messages/en.json
```diff
@@ -911,7 +926,9 @@
         "breakdownCouncil": "{count} on the city council",
         "breakdownCommittees": "{count} on committees",
         "breakdownCommunities": "{count} in communities",
-        "breakdownOthers": "{count} others"
+        "breakdownOthers": "{count} others",
+        "peopleWithoutParties": "People not in a group:",
+        "independents": "Independents"
```
kouloumos: `Party.independents` has no consumer — the only `independents` hit in `src/` is an unrelated LLM prompt string in `cityCreatorAI.ts:154`. It's added in all four locales, so the parity test will pin it indefinitely.

`peopleWithoutParties` right above and `Person.independentCouncilMember` are both used, so this looks like a leftover from an earlier shape of the fix. Drop it, or wire it into `src/lib/sorting/people.ts:98`, which still has `|| "Ανεξάρτητοι"` hardcoded — though that branch is currently unreachable, since the only caller passes `'id'`.

src/components/subject/SubjectListContainer.tsx
```diff
@@ -144,7 +145,7 @@ export function SubjectListContainer({
     return (
       <div className="flex justify-center items-center min-h-[400px]">
         <div className="text-center space-y-2">
-          <div className="text-destructive text-lg font-medium">Σφάλμα φόρτωσης</div>
+          <div className="text-destructive text-lg font-medium">{tCommon('loadingError')}</div>
           <span className="text-muted-foreground">
             {dataError?.message || statsError?.message || 'An error occurred'}
```
kouloumos: The heading above is translated, this line isn't — a Serbian user gets "Грешка при учитавању" over "Failed to fetch data".

Note both `dataError.message` and `statsError.message` are themselves hardcoded English (lines 71 and 102), so adding a `Common.genericError` for the tail only fixes the case where neither error has a message.

src/components/meetings/docx/MinutesDocx.ts
```diff
@@ -541,8 +540,10 @@ function createTOCTable(subjects: MinutesSubject[], useSequentialNumbers: boolea
 
     const dataRows = subjects.map((subject, index) => {
         const seqNum = useSequentialNumbers ? `${index + 1}` : `${subject.agendaItemIndex ?? ''}`;
+        // Greek-only document (see the Greek headings and date locale above), so
+        // the label is inlined rather than routed through the i18n catalogue.
         const decisionText = subject.withdrawn
-            ? getWithdrawnLabel(subject)
+            ? (subject.nonAgendaReason === 'outOfAgenda' ? 'Δεν εγκρίθηκε' : 'Αποσύρθηκε')
```
kouloumos: No argument with keeping the minutes Greek. But dropping the shared helper means this ternary and its two literals are now duplicated verbatim in `MinutesPreviewContent.tsx:423`, and the rationale comment is duplicated with it. These two render the same TOC table, so they'll drift.

The repo already has the idiom for this: `getAbsentLabel` in `src/lib/formatters/name.ts`, which both files already import. A `getWithdrawnLabelGreek(subject)` sibling keeps the "Greek by construction" rationale in one place and the two call sites in lockstep.

src/components/subject/SubjectCardFooter.tsx
```diff
@@ -9,6 +9,8 @@ interface SubjectCardFooterProps {
     speakers: PersonWithRelations[];
     withdrawn?: boolean;
     withdrawnLabel?: string;
+    /** Localized speaking-time text, already pluralized (e.g. "12 minutes"). */
+    minutesText: string;
```
kouloumos, line 13: The prop itself is fine — it matches `SubjectCardContent`, which documents itself as a hook-free presentational shell. But the reason given in the description isn't right, and it'll mislead whoever touches this next: the embed does render inside the provider. `(embed)` is a route group under `[locale]`, its layout says so in its own docblock (`(embed)/layout.tsx:11`), and `EmbedSubjectCard` already called `useTranslations('Subject')` before this PR — that's how `t('noLocation')` and `getAgendaLabel(t, …)` work there today.

Same note on `Subject.minutes`: of the three call sites cited for not converting it, only `subject.tsx:235` actually reads that key. `PilotPage` reads `PilotPage.minutes` and `SubjectDemo` reads `about.demos.subject.minutes`, both different namespaces. The conclusion still holds — both surviving layouts render the number separately above a caption noun — but the constraint is narrower than stated.

kouloumos (commented):
Ran this against the preview with staging as the A/B baseline. Two things worth reporting: `/athens/parties` in Greek is byte-identical to `main` apart from the dev login bar, and `/en/embed/subjects` went from `4 λεπτά` to `4 minutes` — the embed widget was serving Greek to every locale, so `minutesCount` fixes a live externally-embedded surface, not just the app.

What I couldn't do: there's no Serbian-realm city in the staging DB, and Greek cities pin the locale (`/sr/...` redirects), so the preview can't render any of the pages this PR is actually about. The Serbian side is verified from the catalogue and by running `transliterateCatalog` over the new strings (ICU intact, `Party.leaderLabel`'s trailing space survives, `Одборничка група` → `Odbornička grupa`) — not in a browser. Might be worth seeding a Serbian fixture city on staging before the Niš demo, otherwise every Serbian PR has this hole.

The one thing I'd fix before merge is outside the diff, so it can't go inline:

**The glossary guard wasn't extended.** `docs/guides/serbian-localization.md:50` still reads `| political party | странка | |` — it now documents the banned term as the approved one — and `BANNED` in `sr-glossary.test.ts` still holds only the `консултациј` entry. I reinjected `странка` into `Party.item`, `PersonCard.party` and `Party.items` and ran the three guard suites: 43 tests, all green. Nothing stops the next machine-translation pass from undoing all 46 strings. Both the guide and the test's own docblock say to extend them when a term is settled. This patch passes at HEAD and fails at BASE:

```ts
// BANNED
{ pattern: /[Сс]транк|[Сс]транак/, use: 'одборничка група' },
// REQUIRED
{ key: 'sr.json.Party.item',         stem: /[Оо]дборничк/ },
{ key: 'sr.json.Person.partyLeader', stem: /Шеф/ },
```

A blanket ban on `Председник` would be wrong — `Person.president` is the actual assembly president — hence pinning the leader keys in `REQUIRED` instead. `SignIn` and `ShareDropdown` are also new top-level namespaces missing from the review checklist at :104-111, so they currently fall outside the native-review gate.

Two scope questions rather than findings. The meeting page (`[meetingId]/page.tsx`) still has eight Greek literals including `"Χωρίς θέματα"` — the exact string this PR just keyed as `MeetingCard.noSubjects` — and it's already `"use client"` with a `Subject` translator in scope. And `generateMetadata` on the person and party pages is still Greek, so a Serbian user gets a Serbian party name inside a Greek `<title>`. Both look like deliberate deferrals; worth saying so explicitly if they are.

The terminology commit I'd take as-is — the declension is right case by case, and `Party.items` declines correctly inside the plural branches.

Longer term, nothing validates ICU plural *categories* per locale. I checked: a Serbian plural missing `few`, one with a bogus category, and a Greek plural with a spurious `few` are all accepted by `validate-sr-catalogs.ts`, which only compares argument sets. `minutesCount` is correct in all four locales, but `serbian-localization.md:121` gates launch on this and currently hands it to a human eye. Happy to open an issue.

christosporios:
Thanks — all of it landed. [Force-pushed](https://github.com/schemalabz/opencouncil/compare/7adb011f4f0fc92c761330fd92d1d1fc5af072b0..9ff3f33d1836f5b98c80a1aded9ada9c55910945), rebased onto current `main` (picks up #595), squashed into the same two commits.

**Glossary guard.** Extended, and I confirmed it bites: reinjecting `странка` into `Party.item`, `PersonCard.party` and `Party.items` fails both `contains no banned terms` and `keeps required stems`; restored, all three suites green. Guide row replaced — it had been documenting the banned term as approved — plus a second row for the leader title, and `SignIn`/`ShareDropdown` added to the checklist.

I pinned six leader keys in `REQUIRED` rather than two. Your reasoning about `Person.president` applies to `Party.leaderLabel` as well, which is exactly the key that regressed here, so pinning only `Person.partyLeader` would have left the one that actually broke unguarded.

**PersonBadge.** Reverted — you're right that it was the awkward middle. One deviation: rather than restoring the bare literal I used the `UNKNOWN_SPEAKER_LABEL` constant already exported from `@/lib/utils`, since it makes the coupling to `transcribe.ts` and the three prefix-matchers explicit at the call site. Same value, same behaviour, and the next person to reach for a translator there has a reason not to. Test mock reverted with it; `editing.actions.unknownSpeaker` stays unused, as it was before this PR.

**Minutes duplication.** Extracted, but to `@/lib/minutes/builders` rather than `formatters/name`. `formatSubjectLabel` is already there doing exactly this job — a Greek-only minutes label — and both files already import from it, so `getWithdrawnLabelGreek` sits beside its sibling and the rationale comment lives in one place.

**Also fixed:** `Party.independents` dropped (no consumer); the untranslated error tail, via `Common.genericError` — noted in the commit that both `.message` values are themselves hardcoded English, so this only covers the neither-has-a-message case; `min-w-[160px]` on the select trigger.

**Scope questions.** Meeting page: fixed, you were right that it was arbitrary to key `MeetingCard.noSubjects` and leave the identical literal one file away. Eight literals gone, and `subjectsCount` folded the `> 0` ternary into one ICU message with an `=0` branch, so "Χωρίς θέματα" and "N θέματα" are now a single key. `generateMetadata`: deliberate, now stated in the description — server-side across several routes, needs `getTranslations` threaded through each, different shape of change. `sorting/people.ts:98` I left alone too, since as you note the branch is unreachable and it feeds AI prompt context rather than UI.

**Description corrected.** The embed/provider claim was wrong and I've said so inline rather than quietly editing it — `(embed)` is under `[locale]` and its layout docblock says so. Your `Subject.minutes` correction is right too: only `subject.tsx` reads that key. Also added a table of the two Greek copy changes, which should have been there from the start.

**Plural categories:** #597.

**Staging fixture:** agreed, and it's the gap that matters most — every Serbian claim in this PR is catalogue-level, including yours. Worth seeding before the demo rather than after.

`tsc` clean; 96 suites / 1318 tests.

PR #598, opened 2026-08-08
feat: Notis playground — agent core, simulator, and admin panel (PR 1)

christosporios:
PR 1 of the Notis plan (PRD §9): the full iteration loop for ο Νότης — the real agent core running against real city data in a playground — before any Bird/WhatsApp infrastructure.

## What's in here

**npm workspaces + shared UI**
- Repo converts to npm workspaces: root app + `packages/ui` + `services/notis`, one root lockfile.
- 46 shadcn primitives (+ `use-mobile`, `use-toast`, `phone`, `pagination`, the tailwind preset and CSS tokens) move to `@opencouncil/ui`. The main app keeps working through re-export stubs at the old paths — no import site changed.
- Nix build, CI, Dockerfiles and the preview workflow updated for the workspace layout. `nix build .#opencouncil-prod` passes.

**`services/notis` — new Next.js app** (deploys as a second DO component, `notis.opencouncil.gr`)
- Static Greek landing page; admin area behind a shared-secret HMAC cookie (interim auth for PR 1).
- Agent core (`src/agent/`): pure `runWake` loop over injected deps — client tools (`send_message`, `update_taste_profile`, `schedule_wakeup`, `unsubscribe_user`) are recorded into the outcome, MCP research runs server-side via the Anthropic MCP connector against opencouncil.gr/mcp. Editorial pass scores each meeting's subjects once; per-user wakes read the brief instead of raw subjects. Wire shapes have one source of truth (zod schemas → inferred types).
- Model: claude-sonnet-5 at low effort. Prompt caching with 1h TTL; a cache breakpoint on the user turn keeps multi-pass research wakes cheap (measured: $0.37 → $0.09 for a researching send; silences ~$0.01–0.03).
- WhatsApp rules enforced in the sim: outside the 24h service window messages ride the approved Bird template shells (`demos_*`, copied verbatim from the Bird dashboard); inside it they go free-form. ΣΤΟΠ triggers `unsubscribe_user` and is surfaced in the UI.
- A repair pass in `runWake` catches a real failure mode: on a direct user question the model sometimes writes its answer into the rationale and never calls `send_message` — one bounded nudge makes it send or own the silence.

**Admin panel** (`/admin`, mock-first — no DB yet)
- Sidebar with six sections: dashboard (all metrics honest zeros), conversations (list + read-only detail), wakes feed, playground, templates (rendered live from `agent/templates.ts`), settings (config, env checks, outbound kill switch).
- Each section reads from a small `_lib` module that returns zeros/empties; PR 2 swaps those internals for DB queries with no UI work.
- The inspection UI (timeline, WhatsApp-style chat with OG link previews, wake inspector) is shared between the playground and the future real-conversation viewer — simulator controls are optional props.

**Playground** (the point of the PR)
- Build a simulated resident (cities, topics, geocoded addresses on a map, enrollment origin), derive a queue of real released meetings, and step through wakes with the real model: per-wake traces (context blocks, tool calls, rationale, cost), prompt override for A/B, rewind via snapshots, golden-scenario export, fast-forward until the next send.
- Golden replay tests run recorded scenarios offline with full-content assertions; the notis CI job runs on every PR (notis depends on packages/ui, so path-filtering would under-run it). A separate live lane for prompt-regression checks is future work (see README).

## Not in this PR
Database, Bird integration, real sends, the poller, real auth, and the rollout flag (`User.notisEnabledAt` + release panel in the main app's admin — PR 2, per PRD §9). The kill switch and panel metrics are wired to mocks. See `services/notis/README.md` for DO component settings and local dev.

## Testing
- `npm test` (root + notis projects), `npx tsc --noEmit`, `nix build --impure .#opencouncil-prod` all green.
- Live end-to-end sessions against production data: verified silence-by-default, grounded sends with correct links, template vs free-form delivery, ΣΤΟΠ, capability boundaries, follow-up scheduling, and cost per wake in the inspector.

<h3>Greptile Summary</h3>

The PR introduces the Notis workspace service, shared UI package, agent wake loop, authenticated administration surface, and real-data playground, while adapting repository builds and CI for npm workspaces.

- Moves shared UI primitives into `@opencouncil/ui` with compatibility re-exports for the main application.
- Adds the Notis agent core, editorial processing, prompt assembly, templates, schemas, and offline replay tests.
- Adds the Notis admin panel and stateful playground backed by authenticated server routes and public OpenCouncil data.
- Updates Docker, Nix-adjacent build inputs, TypeScript, Jest, Tailwind, and GitHub Actions for the workspace layout.

The PR should not merge until the public login throttle can no longer let one unauthenticated caller lock out every administrator.

The login endpoint counts all clients in one process-wide quota before validating their requests, making sustained remote denial of the newly added admin and playground surface straightforward; the mutable CI action references are additional non-blocking supply-chain hardening work.

**Files Needing Attention:** services/notis/src/app/api/admin/login/route.ts, .github/workflows/ci.yml

<details open><summary><h3>Security Review</h3></summary>

The public shared-secret login route has a process-global throttle that lets any unauthenticated caller lock every administrator out. The new CI job also uses mutable action tags rather than immutable commit references.
</details>

| Filename | Overview |
|----------|----------|
| services/notis/src/app/api/admin/login/route.ts | Adds shared-secret cookie authentication, but its process-global pre-validation throttle permits trivial unauthenticated denial of admin login. |
| services/notis/src/proxy.ts | Protects admin and API routes with the HMAC cookie while intentionally exposing health and login endpoints. |
| services/notis/src/agent/runWake.ts | Implements the bounded model/tool loop, repair pass, outcome recording, usage accounting, and immutable state application. |
| services/notis/src/lib/mcp-client.ts | Implements a reusable Streamable HTTP MCP client; no concrete incompatibility with the configured external server was established. |
| services/notis/src/app/admin/(panel)/playground/page.tsx | Coordinates real-data simulation, lazy editorial briefs, wake execution, WhatsApp delivery modes, rewind, and scenario export. |
| services/notis/src/app/admin/(panel)/playground/store.ts | Provides persisted reducer state, bounded traces and snapshots, queue updates, and rewind behavior for the playground. |
| packages/ui/package.json | Defines the shared UI workspace package and its exported component surface. |
| .github/workflows/ci.yml | Adds Notis type-check, test, and build coverage, but the new job executes actions referenced through mutable tags. |

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Aservices%2Fnotis%2Fsrc%2Fapp%2Fapi%2Fadmin%2Flogin%2Froute.ts%3A15-18%0A**Global%20login%20quota%20enables%20lockout**%0A%0AWhen%20any%20unauthenticated%20client%20submits%20five%20login%20requests%20within%20sixty%20seconds%2C%20the%20process-wide%20quota%20rejects%20every%20subsequent%20request%20before%20parsing%20or%20checking%20its%20secret%2C%20causing%20legitimate%20administrators%20to%20receive%20HTTP%20429%20and%20lose%20access%20to%20the%20admin%20panel%20and%20playground.%20Sustained%20requests%20can%20keep%20the%20shared%20quota%20exhausted%20indefinitely.%0A%0A**How%20this%20was%20verified%3A**%20The%20public%20login%20allowlist%20was%20traced%20to%20the%20module-level%20counter%2C%20which%20records%20every%20request%20before%20credential%20validation.%0A%0A%23%23%23%20Issue%202%0A.github%2Fworkflows%2Fci.yml%3A34-38%0A**CI%20actions%20use%20mutable%20tags**%0A%0AThe%20new%20Notis%20job%20executes%20%60actions%2Fcheckout%40v6%60%20and%20%60actions%2Fsetup-node%40v4%60%20through%20mutable%20tags%2C%20so%20a%20moved%20or%20compromised%20tag%20can%20silently%20alter%20CI%20behavior%20or%20run%20untrusted%20code%20with%20the%20workflow's%20available%20credentials.%20Pin%20these%20actions%20to%20full%20commit%20SHAs%20to%20make%20the%20executed%20code%20immutable.%0A%0A**How%20this%20was%20verified%3A**%20Both%20newly%20added%20%60uses%60%20entries%20reference%20version%20tags%20rather%20than%20full%20commit%20SHAs.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=598&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
services/notis/src/app/api/admin/login/route.ts:15-18
**Global login quota enables lockout**

When any unauthenticated client submits five login requests within sixty seconds, the process-wide quota rejects every subsequent request before parsing or checking its secret, causing legitimate administrators to receive HTTP 429 and lose access to the admin panel and playground. Sustained requests can keep the shared quota exhausted indefinitely.

**How this was verified:** The public login allowlist was traced to the module-level counter, which records every request before credential validation.

### Issue 2
.github/workflows/ci.yml:34-38
**CI actions use mutable tags**

The new Notis job executes `actions/checkout@v6` and `actions/setup-node@v4` through mutable tags, so a moved or compromised tag can silently alter CI behavior or run untrusted code with the workflow's available credentials. Pin these actions to full commit SHAs to make the executed code immutable.

**How this was verified:** Both newly added `uses` entries reference version tags rather than full commit SHAs.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(notis): simplify pass over rece..."](https://github.com/schemalabz/opencouncil/commit/d8bd00b8a630336b58f0f0bc4673c591d994b98a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51458922)</sub>

> Greptile also left **2 inline comments** on this PR.

---

> [!NOTE]
> **Medium Risk**
> Large structural change (shared UI package + workspace installs) affects every build and deploy path; mis-linked imports or lockfile drift could break the main app or Docker builds until CI catches it.
> 
> **Overview**
> Turns the repo into an **npm workspace** (`packages/*`, `services/*`) with a single root lockfile. Shared shadcn-style primitives, hooks, tokens, and small libs move into **`@opencouncil/ui`** (`packages/ui`) with package exports; the root app depends on it and **`next.config`** now transpiles `@opencouncil/ui` instead of the old internal alias.
> 
> **Tooling** is aligned with the layout: Docker images copy workspace `package.json` files before install; **preview deploy** lockfile checks any workspace manifest (including peer deps); root **ESLint** ignores `services/**`; **Jest** picks up tests under `packages/*/src`; a **`check:ui`** script type-checks the UI package. **CI** adds a **Notis** job (`npm ci`, `tsc`, tests, build for the `notis` workspace). Ignore rules expand for nested `node_modules` and `services/*/.next`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a8327b8176fa715df881443310d9be12966d44d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### CODE REVIEW

**Comment by christosporios** [comment]
@greptileai review

File: services/notis/src/app/api/admin/login/route.ts [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+// for a single-instance interim gate; goes away with PR 2's cookie auth.
+let attempts: number[] = [];
+
+export async function POST(request: NextRequest) {
+  const now = Date.now();
+  attempts = attempts.filter((t) => now - t < 60_000);
+  if (attempts.length >= 5) {
+    return NextResponse.json({ error: "too many attempts, wait a minute" }, { status: 429 });
+  }
+  attempts.push(now);
```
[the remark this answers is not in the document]
christosporios: Known and accepted for this PR: this shared-secret gate is explicitly interim — PR 2 of the Notis plan replaces it entirely with validation of the main app's session cookie against a read-only view (see the PRD's shipping plan). The throttle's job here is only to slow secret guessing on a draft-stage panel with no user data behind it; the worst case of the global counter is a 60-second lockout of the playground, which self-heals and also clears on any instance restart. Not worth per-IP bookkeeping for code with a planned deletion date.

.github/workflows/ci.yml
```diff
@@ -26,3 +26,24 @@ jobs:
...
+
+  notis:
+    name: Notis (services/notis)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '24'
```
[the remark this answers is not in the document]
christosporios: Declining in this PR for consistency: every workflow in this repo uses mutable version tags (actions/checkout@v6, actions/github-script@v9, cachix/*@v31/v17 — 17 uses: entries across ci.yml, preview-deploy.yml, preview-cleanup.yml). Pinning only the two lines this PR adds wouldn't change the repo's actual supply-chain posture and would leave mixed conventions. Happy to do a repo-wide SHA-pinning pass as a separate chore.

christosporios:
Pushed a fix for a wake-loop crash found while testing: when a turn ended with `stop_reason: tool_use` but contained only server-side MCP blocks, `runWake` sent an empty `tool_result` message and the API rejected it with a 400. Those turns now continue like `pause_turn`. One new regression test (37 total).

christosporios:
Prompt improvement from playground testing: «tell me everything about X» was answered with a referral to the site's notification settings instead of just granting it. The capability section of system.md now makes the hierarchy explicit — only the city list is outside the agent's power; topic categories are a seed the taste profile outranks. Replaying the same exchange now yields a plain yes + profile update, with the site mentioned only as an aside.

christosporios:
Fixed a second shape of the undelivered-message bug, found by reviewing an exported thread: on «τίποτα άλλο από το ΔΣ;» the model composed two messages but wrote the first as a plain text block beside the send_message call for the second — so the person got only message two (starting with «επίσης»…), while the rationale claimed both were sent. Now: a prompt rule (no prose in tool-calling turns) plus an extension of the runWake repair pass that quotes stranded prose back and makes the model either send it or explicitly own it as reasoning. Regression-tested (38 tests).

christosporios:
Three more playground findings, pushed: (1) rewinding a user-message step left its bubble in the chat — pending records now render only while actively running, so instant-send still works but rewind truly erases; (2) the agent was scheduling wakeups for a meeting's aftermath, which meeting_summarized already covers — the tool description and prompt now reserve schedule_wakeup for things no event covers; (3) two delivered messages carried vote counts cross-contaminated from other subjects — Grounding now requires counts to come from that subject's own votes array or be omitted.

christosporios:
Prompt: taught the agent the full opencouncil.gr link surface — subject/meeting/person/party pages plus the per-utterance ?t= moment deep-links that transcripts return. Quotes now link the exact second of the video instead of the subject page (verified live: «στείλε μου το λινκ να ακούσω τη στιγμή» → …?t=993), and the internal REF:UTTERANCE markers are explicitly banned from outgoing messages.

christosporios:
Speed batch: (1) new finish_wake tool carries the rationale, letting a wake end in the same turn as its sends — best case halves model passes and their tokens; repair guards and fallbacks preserved. (2) Two-tier replies: a short holding line goes out before genuinely-needed research. (3) The simulator now mirrors Bird/WhatsApp receipt mechanics (mark-as-read blue ticks at wake start, typing indicator with the 25s dismissal cap) — researched against Bird's Conversations API and WhatsApp Cloud API docs. 40 tests green, verified live.

christosporios:
Prompt: the agent now gauges and records each reader's civic-vocabulary level (a «τι είναι κανονιστική;» question is a durable profile fact) and matches their register from then on. Validation note: plain-language register matching verified live twice; the profile write itself is still unreliable at low effort — flagged a structural nudge as the follow-up if review-queue data confirms. Also tightened both repair nudges to demand the wake's own rationale after a trace recorded the model's answer-to-the-check as its journal rationale.

christosporios:
Ran a 17-wake API-driven test session (adversarial + proactive quality, ~$0.90). Three fixes pushed: (1) briefs now carry real meeting/subject URLs — a no-research send had fabricated a wrong-shaped link; (2) the stranded-prose repair can no longer replace the wake's genuine rationale with meta chatter about the check (was polluting the journal in 4/17 wakes); (3) main-app MCP get_subject now returns a precomputed voteSummary tally — two wakes miscounted the same subject's votes array two different ways (12-6 and 13-6 for a 13-6 vote). Full findings in the session notes: injections/neutrality/boundaries/ΣΤΟΠ/dedup all clean; finish_wake landed same-turn in 12/17 wakes (simple replies now ~$0.014 / 10-12s).

christosporios:
Title-attribution fix (the «Μαλτέζος as σύμβουλος» issue), both layers: (1) main-app MCP now attaches each speaker's role as of the meeting date to subject contributions, introducedBy, and both transcript surfaces — role name when stored, Δήμαρχος for city-head, party membership as the factual fallback; verified end-to-end locally (Δούκας → Δήμαρχος, Λιμνιωτάκη → Πρόεδρος). Zero extra queries on subjects, one widened batch select on transcripts, ~100 tokens per research wake. (2) Notis prompt: state an office only as the record gives it, name alone otherwise; vote counts now come from the new voteSummary. Also shipped earlier this hour: language accommodation — readers who write in another language get answered in it, recorded as a profile preference.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/78e5f0b8bda4f8ec59e164179583f9f480a64e77..b1909f4aa5a6ba795d14a2a17ef03f24aba43edb) to extract the main-app MCP changes into their own PR:

- The speaker-roles and voteSummary commits now live in #599 (independently useful to all MCP consumers and deployable ahead of Notis)
- This branch keeps only the Notis-side prompt guards that consume those fields once #599 deploys; until then the fields are simply absent and behavior degrades gracefully

christosporios:
Pushed two fixes out of a fresh 10-wake test round (run against the pr-599 preview MCP, so the new role labels and voteSummary were exercised end to end):

- `8b439322` **fix(notis): quote stranded prose when a user question got no send** — the worst finding: a wake could end with the reader getting nothing (or only a bare link) because the model saw its own undelivered prose above and concluded it had already replied. The zero-send nudge now quotes the stranded text back and states it was not delivered; both nudges also demand the wake's own rationale, so repair chatter stays out of the journal. Reproduced 2/2 before, full-answer delivery 2/2 after.
- `cef49b6d` **prompt(notis): fixed Greek vote vocabulary** — ABSTAIN was rendered «παρών»/«δήλωσαν παρών»; Grounding now pins FOR=υπέρ, AGAINST=κατά, ABSTAIN=λευκό, PRESENT=παρών, DID_NOT_VOTE=αποχή. Verified on a 14–4–2 vote and a «πόσοι απείχαν;» follow-up.

Also verified in the round: role titles from the record («ο δήμαρχος Ιωάννης Μαλτέζος»), neutrality under verdict-baiting, future-meeting boundary, English-language accommodation with profile write, plain-language register write, and `?t=` moment links.

christosporios:
Pushed the review-round cleanup — seven commits out of the two-reviewer DRY/perf pass:

**Bugs**
- `b8e450ee` MCP client recovers from session expiry (404 → reset + one retry) — the module-scope singleton could otherwise brick until process restart.
- `f40c5330` Playground store v2: snapshots hold references + an id→status queue list instead of 30 deep copies of the whole sim; saves debounced 500ms; the quota fallback's retry is guarded; `userMessage` refuses to race a wake already in flight; dead `Sim.cursor` removed. Bonus fix: a rewind no longer silently reverts the prompt override, so the edit-prompt-then-rewind A/B loop now works as intended.
- `05dfc00c` Only failed logins consume the throttle budget.

**Perf**
- `100a3d2f` Moving cache breakpoint in the wake loop — the newest tool_results message carries the ephemeral marker, so follow-up turns re-read history (incl. MCP transcript payloads) from cache. Verified live: a two-turn research wake now reports 8 uncached input tokens against 109k cache reads. Editorial top-8 detail fetches run concurrently.

**DRY** (`301fb53b`, `bf1781cc`)
- `NOTIS_MCP_URL` / `OPENCOUNCIL_BASE_URL` env vars with prod defaults replace hardcoded URLs; `lib/api.ts` (parseJsonBody / errorResponse / proxyOpencouncilGet) replaces the boilerplate copied across four routes — and error responses now carry the upstream status instead of flattening everything to 502.
- One home each for: the `Origin` union (was inline in 5 files), `introTemplateFor`, the WhatsApp palette, el-GR date/number formatting, `LocationPoint`, the trace-evicted fallback. Composer extracted so typing stops re-rendering the thread. Riding along: Ticks unfreeze, viewer quick-replies disable properly, AddressSearch spinner clears on network failure, tables use next/link.
- `6563ddd6` send_message tool description now says the reader's language, not 'in Greek'.

Verified: tsc clean, 40/40 tests, lint-silent, playground boots fresh on the v2 store (wizard renders, cities load, zero console errors), and a live research wake through the new caching path.

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,286 @@
...
+  }, []);
+
+  // Debounced: during run-until-send every step dispatches several actions,
+  // and serializing the whole store (traces included) on each would jank the
+  // main thread. A 500ms trailing save loses at most half a second of state.
+  useEffect(() => {
+    if (!mounted) return;
+    const t = setTimeout(() => saveStore(store), 500);
+    return () => clearTimeout(t);
+  }, [store, mounted]);
```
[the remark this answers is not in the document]
christosporios: Fixed in 42506650: an unmount/`pagehide` flush closes the debounce window (guarded so a pre-hydration flush can't overwrite storage with the initial empty store).

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,286 @@
...
+      const item: WakeRecord = {
+        id: `user:${Date.now()}`,
+        event: { type: "user_message", at: clock, text },
+        status: "pending",
+      };
+      dispatch({ type: "userMessage", item });
+      void runItem(item);
+    },
+    [runItem, busyItemId],
+  );
```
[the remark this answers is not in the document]
christosporios: Fixed in 42506650: the guard moved into `runItem` behind a ref (`busyRef`), so it's synchronous — two clicks in the same tick can't both pass. `step`, `runUntilSend` and `userMessage` all funnel through it.

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,286 @@
...
+        if (stopRef.current) break;
+        const next = nextPending();
+        if (!next) break;
+        const outcome = await runItem(next);
+        if (!outcome || outcome.messages.length > 0) break;
+      }
+    } finally {
+      setAutoRun(false);
+    }
+  }, [nextPending, runItem]);
```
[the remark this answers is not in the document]
christosporios: Fixed in 42506650 for the part that's the simulator's job: after `unsubscribedAt` is set, `nextPending` returns nothing, so step/run-until-send stop advancing proactive wakes — matching production, where the scheduler would never fire them. Injected user messages still run, since inbound survives a ΣΤΟΠ. Recording the unsubscribe on `WakeState`/journal is deliberately left to the real store in PR 2 — in this stateless core the outcome's `unsubscribe` field plus the journaled rationale of that wake carry the fact.

.github/workflows/ci.yml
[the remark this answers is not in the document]
christosporios: Yes — please remember that rule.

christosporios:
Pushed 42506650 addressing the open review threads: the debounced playground save now flushes on unmount/pagehide, the busy guard is a synchronous ref inside `runItem` (no same-tick double runs), and a ΣΤΟΠ halts proactive queue advancement while still allowing injected user messages. Details in the inline replies.

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,315 @@
...
+  }, []);
+
+  // After a ΣΤΟΠ (unsubscribe) the production scheduler would never fire
+  // another proactive wake, so the simulator stops advancing too. Injected
+  // user messages still run — inbound survives an unsubscribe.
+  const nextPending = useCallback(
+    () =>
+      storeRef.current.sim.unsubscribedAt
+        ? undefined
+        : storeRef.current.sim.queue.find((q) => q.status === "pending"),
```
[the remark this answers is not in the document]
christosporios: Fixed in f1fd72f7: the chat now derives the halt from the records themselves (any outcome carrying `unsubscribe`) instead of trusting its own queue scan, so step/skip/fast-forward disable the moment a wake unsubscribes the reader — and the derivation lives in the component, so the PR 2 read-only viewer inherits it. The composer stays enabled, since inbound survives a ΣΤΟΠ.

christosporios:
Pushed f1fd72f7: the chat's step/skip/fast-forward buttons now disable after a ΣΤΟΠ, deriving the halt from the records so it also holds in the PR 2 viewer. Closes the remaining Bugbot thread.

christosporios:
Merged `origin/main` into the branch (4ace663f) to clear the conflicts — 37 commits since the branch point. Conflicts were `package.json` (both sides added a dependency; union of `@opencouncil/ui` and `@posthog/mcp`) and `package-lock.json` (regenerated with `npm install` against the merged manifest). Re-ran the full gate battery on the merge result: root tsc, notis tsc, eslint, 1400 jest tests, and `nix build --impure .#opencouncil-prod` all green (the CityStatus schema change from main needed a `prisma:generate`, nothing else). Marking ready for review.

services/notis/src/lib/mcp-client.ts
```diff
@@ -0,0 +1,121 @@
...
+      res = await this.postCall(tool, args);
+    }
+    if (!res.ok) throw new Error(`MCP tools/call ${tool} failed: HTTP ${res.status}`);
+    const body = await parseBody(res);
+    if (body.error) throw new Error(`MCP ${tool}: ${body.error.message}`);
+    if (body.result?.isError) {
+      const msg = body.result.content?.map((c) => c.text).join("\n") ?? "unknown tool error";
+      throw new Error(`MCP ${tool}: ${msg}`);
+    }
+    const text = body.result?.content
```
kouloumos: **Blocks the playground at tip.** The main-app MCP server now appends an advisory text block to every tool response (`[SERVER]: Reuse conversation_id=…`, from the analytics wrapper in a13776b1 — inside this branch since the merge). This join concatenates the data block and the note, so `JSON.parse` fails and the method returns a raw string. Every consumer then hedges the string into an empty result: the wizard reports «καμία δημοσιευμένη συνεδρίαση» for every city, and `editorialPass` throws on `"meeting" in raw`. Verified against the tip through a running dev server.

Two-part fix: parse each text block individually and return the first block that parses as JSON; and pin the proxy response shapes with zod at the boundary, so the next contract drift fails loudly instead of returning an empty wizard.

services/notis/src/agent/runWake.ts
```diff
@@ -0,0 +1,322 @@
...
+            "if needed so the conversation still reads naturally. If it was only " +
+            "reasoning, do not send it — give the wake's own rationale, about the reader " +
+            "and this wake's decision, never about this check.",
+        });
+        strandedProse = undefined;
+        continue;
+      }
+    }
+
+    // end_turn, max_tokens, or anything else: stop.
```
kouloumos, line 294: **A cut turn records itself as a deliberate silence.** A turn that stops at the `max_tokens` ceiling falls through to this `break`. Tool calls are only read under the `tool_use` branch, so any `send_message` or `unsubscribe_user` calls in the cut turn are dropped, and the outcome derives `decision: "silence"` from `sent.length`. A truncated wake and a considered quiet wake produce identical records. The ceiling is real: no `thinking` parameter is sent, so Sonnet 5 thinks by default and thinking spends the same budget.

This inflates the silence rate — the PRD's headline metric — and the Batch API lane would hide a systematic truncation inside normal-looking silence. A natural-language opt-out lives in exactly the tool call this can drop (literal ΣΤΟΠ is safe — production handles it before the agent). Processing the cut turn's calls is also wrong, because a truncated `tool_use` block carries partial JSON.

Fix: an explicit truncation marker on `WakeOutcome`, so no caller can read a cut turn as a decision. Mechanism verified by two independent reads; the runtime scenario is inferred, not reproduced live.

services/notis/src/agent/runWake.ts
```diff
@@ -0,0 +1,322 @@
...
+      system,
+      messages,
+      tools: buildTools(),
+      mcp_servers: buildMcpServers(deps.config.mcpUrl),
+      output_config: { effort: deps.config.effort },
+    });
+
+    const usage = normalizeUsage(response.usage);
+    usageTotal = addUsage(usageTotal, usage);
+    turns.push({ content: response.content, stopReason: response.stop_reason ?? "unknown", usage });
```
kouloumos, line 119: **Repaired wakes leave no trace.** The repair nudges are injected as `role: "user"` messages, but the trace stores only assistant turns plus the first user turn, and `WakeOutcome` has no `repaired` field. A rescued wake is indistinguishable from a healthy one in the inspector and in every future record. How often the net fires is precisely the number that says whether the prompt works at low effort, and the PR 5 review queue reads these same records.

Fix: record injected turns in the trace with their role, and add a `repaired` flag (or the nudge kind) to the outcome.

services/notis/src/agent/runWake.ts
```diff
@@ -0,0 +1,322 @@
...
+      continue;
+    }
+
+    // end_turn: two observed low-effort failure modes get ONE bounded repair
+    // pass between them, so a broken wake self-corrects instead of shipping.
+    if (response.stop_reason === "end_turn" && !repaired) {
+      // (a) The person wrote to us and the model produced neither a send nor
+      // an unsubscribe: it wrote its answer into the final text instead of
+      // the tool. Never fires on proactive wakes — silence is their default.
+      if (event.type === "user_message" && sent.length === 0 && !unsubscribe) {
```
kouloumos, line 242: **One of four repair paths lacks the rationale protection.** The four nudge sites are one 2×2: {unanswered, stranded prose} × {finish_wake, end_turn}. Three of the four set `preNudgeRationale` so that post-nudge check-chatter cannot become the journal rationale. This branch — end_turn, zero-send — does not, so a post-nudge silence confirmation can ship text about the system check as the rationale that the next thirty wakes read as memory. The journal is also the only evidence adaptive frequency has (PRD §5).

Suggestion: extract one `checkTurnCompletion()` used from both exit points. That removes the four near-duplicate strings and makes this asymmetry impossible.

services/notis/src/agent/__tests__/golden.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        anthropic: replay,
+        now: () => FIXED_NOW,
+        prompts: { system: "SYSTEM", contextPack: "PACK", editorial: "ED" },
+        config: { model: "claude-sonnet-5", maxTurns: 8, mcpUrl: "https://opencouncil.gr/mcp", effort: "low" },
+        mcp: { call: async () => null },
+      };
+
+      const { outcome, trace } = await runWake(fixture.state, fixture.event, deps);
+
+      expect(outcome.decision).toBe(fixture.expected.decision);
```
kouloumos, line 55: **The fixtures are stale, and the assertions cannot see content.** Three facts, all verified. First: both fixtures were recorded at e27384c7 on Opus — the thinking signatures base64-decode to `claude-opus-5` — before the model change and before the agenda embargo, which the silence fixture now violates (a pre-meeting wake whose brief is full of outcomes). Second: `expected` holds counts, so mutation tests are blind — chopping every sent message to three characters replays green, and never appending tool results replays green. Third: both creation sites write this weak shape, so new recordings inherit it.

The lane's design is right: replay is deterministic, so full-text comparison is valid here and guards the code. Fixes: re-record both fixtures on the shipped pipeline; store full message texts, the `scheduledWakes` array, and an `unsubscribe` field in `expected` (about five lines in `scripts/record-scenario.ts`). The PRD's second job for golden scenarios — guarding prompt changes — needs a separate live-replay lane with behavioral assertions (decision, cadence caps, link presence); that is future work, worth a line in the README so the two jobs stay distinct.

services/notis/src/agent/runWake.ts
```diff
@@ -0,0 +1,322 @@
...
+        }
+      }
+
+      messages.push({ role: "assistant", content: response.content });
+      // A tool_use stop with no client tool calls happens when the turn holds
+      // only server-side (MCP) blocks. There is nothing for us to answer —
+      // an empty user message is an API error — so continue like pause_turn.
+      if (results.length > 0 || nudge) {
+        const content = nudge ? [...results, { type: "text", text: nudge }] : results;
+        if (!finished) markLatest(content);
```
kouloumos, line 229: **The moving cache breakpoint has zero test coverage.** Deleting the unmark step inside `markLatest` passes all 40 tests. The regression it would cause is production-only: stale markers accumulate, and any wake with three or more tool turns exceeds the API's four-breakpoint limit and fails with a 400. Neither fixture is deep enough to reach it. A mutation-verified test exists from the review (three mutations, all caught) — happy to hand it over.

services/notis/src/lib/mcp-client.ts
```diff
@@ -0,0 +1,121 @@
...
+      jsonrpc: "2.0",
+      id: this.nextId++,
+      method: "tools/call",
+      params: { name: tool, arguments: args },
+    });
+  }
+
+  async call(tool: string, args: Record<string, unknown>): Promise<unknown> {
+    let res = await this.postCall(tool, args);
+    if (res.status === 404) {
```
kouloumos, line 115: **Session recovery races under concurrency.** Two concurrent 404s both reset and re-initialize the shared singleton; a request in flight between one caller's reset and re-init posts without a session, burns its single retry, and fails although the client is healthy. Fine for one operator today; wrong for the PR 3/4 concurrent lanes. Fix: route recovery through one shared re-init promise, so the first detector resets and the rest await it.

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,315 @@
...
+  );
+
+  const step = useCallback(() => {
+    const next = nextPending();
+    if (next) void runItem(next);
+  }, [nextPending, runItem]);
+
+  // Fast-forward: keep running wakes until ο Νότης actually writes (or the
+  // queue runs dry / an error stops us / the user hits stop).
+  const runUntilSend = useCallback(async () => {
```
kouloumos, line 156: **Fast-forward runs every silent wake twice, and the guard does not cover it.** Evidence first: a reproduction probe on this workspace's React 19.2.8 shows every silent wake executing exactly twice, and re-running it **with the `busyRef` guard included** still shows ×2. The guard from 42506650 does fix the overlap case the bot described — two calls in one tick. This is a different mechanism: the loop is sequential, `finally` clears the ref before the next pick, and `nextPending` reads `storeRef`, which only updates on render — the continuation runs before the render, so it re-picks the wake it just finished and pays for a second identical `dryRun`.

The duplicate spend is invisible: traces are keyed by item id, so the second overwrites the first and `totalCost` does not double. The loop also has no iteration or cost cap, so a model that stays silent but keeps scheduling follow-ups runs indefinitely at real cost.

One fix covers all of it: a run-epoch ref plus a ledger of run item ids; drop any result from a previous epoch; cap the loop.

services/notis/src/app/admin/(panel)/playground/page.tsx
```diff
@@ -0,0 +1,315 @@
...
+    },
+    [runItem],
+  );
+
+  const selectedSnapshotId = useMemo(
+    () => store.snapshots.find((sn) => sn.itemId === selectedId)?.id,
+    [store.snapshots, selectedId],
+  );
+
+  const rewind = useCallback(() => {
```
kouloumos, line 213: **Rewind and Reset are unguarded against an in-flight wake.** `canRewind` only checks that a snapshot exists. Rewind during a running wake lands, then the late `stepDone` writes the post-wake state onto the rewound store — timeline shows the past, state holds the future — and the auto-save persists the corruption. The run-epoch ref from the fast-forward comment fixes this too; also disable Rewind and Reset while busy or auto-running. Verified by read, not reproduced.

services/notis/src/app/admin/(panel)/playground/store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    case "userMessage": {
+      const queue = insertChronological(store.sim.queue, action.item);
+      return { ...store, sim: { ...store.sim, queue } };
+    }
+
+    case "setPromptOverride":
+      return { ...store, sim: { ...store.sim, promptOverride: action.value } };
+
+    case "restoreSnapshot": {
```
kouloumos, line 203: **Rewind then re-run leaves two snapshots with the same item id.** `restoreSnapshot` never prunes snapshots taken at or after the restore point, and both the rewind lookup and `exportScenario` use `.find`, which returns the pre-rewind one. Rewind to before X, inject a message, run forward, then rewind X again: the restore silently erases the injected message's effects, and an export pairs the new outcome with the abandoned timeline's state. Fix: prune snapshots from the restored item onward, or match with `.findLast`.

services/notis/src/app/admin/(panel)/playground/store.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+export function loadStore(): PlaygroundStore {
+  if (typeof window === "undefined") return emptyStore();
+  try {
+    window.localStorage.removeItem("notis:playground:v1");
+    const raw = window.localStorage.getItem(STORAGE_KEY);
+    if (!raw) return emptyStore();
+    const parsed = JSON.parse(raw) as PlaygroundStore;
+    if (parsed.version !== 2) return emptyStore();
```
kouloumos: **Store durability: silent destruction in two paths.** An unknown `version` wipes the whole store on sight, so an older deploy in another tab erases a newer store. And the quota fallback drops **all** traces and snapshots silently — the operator's main artifact — while the inspector labels the loss "LRU" (`InspectorPane.tsx:31`), which points the operator at the wrong cause. Fix: leave unknown-version stores untouched, and surface the quota fallback ("traces dropped to fit storage") instead of mislabeling it.

services/notis/prompts/context-pack/95-notis-boundaries.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Τι μπορεί
+
+- Να απαντά σε ερωτήσεις για οποιονδήποτε δήμο καλύπτεται από το OpenCouncil — όχι μόνο για τους δήμους στους οποίους είναι εγγεγραμμένος ο χρήστης: τι συζητήθηκε, τι αποφασίστηκε, ποιος είπε τι, πότε είναι η επόμενη συνεδρίαση.
+- Να εξηγεί πώς λειτουργούν οι δήμοι, τα δημοτικά συμβούλια και οι επιτροπές.
+- Να προσαρμόζει τη συζήτηση στις προτιμήσεις του χρήστη: αν κάποιος πει π.χ. «λιγότερα για Χ» ή «με ενδιαφέρει περισσότερο το Υ», ο Νότης το σέβεται στη συνέχεια της συνομιλίας.
+
+## Τι δεν μπορεί
+
+- Δεν μπορεί να προσθέσει ή να αφαιρέσει δήμους από τις εγγραφές του χρήστη, ούτε να αλλάξει τις αποθηκευμένες προτιμήσεις ειδοποιήσεων (δήμοι, θέματα, τοποθεσίες). Αυτά γίνονται μόνο από τον ίδιο τον χρήστη, στο προφίλ του στο opencouncil.gr. Όταν ζητηθεί κάτι τέτοιο, ο Νότης παραπέμπει εκεί.
```
kouloumos: **This contradicts `system.md` and the PRD.** This line says Notis cannot change «θέματα, τοποθεσίες» and must refer to the website. `system.md:96-112` — the shipped fix for a live incident — says only cities are out of reach and "a capability you have is never a referral to the website". PRD §5 settles it: city is the one hard filter; topics and locations are a seed the agent overrides on request. Both files ship in the same system array on every wake, so the model receives a rule and its opposite and guesses — the already-fixed bug is half-alive. Fix: shrink «Τι δεν μπορεί» to cities only.

services/notis/src/agent/tools.ts
```diff
@@ -0,0 +1,98 @@
...
+    name: "send_message",
+    description:
+      "Send one WhatsApp message to the reader, in their language (Greek unless they have " +
+      "shown otherwise). One specific thing per message; " +
+      "two or three sentences is a whole message. Include an opencouncil.gr link. " +
+      "Call once per message, in sending order.",
+    input_schema: {
+      type: "object" as const,
+      properties: {
+        text: { type: "string", description: "The message text, in Greek." },
```
kouloumos: The parameter contradicts the tool it belongs to: the description one line up says "in their language (Greek unless they have shown otherwise)"; the parameter still says "The message text, in Greek." Delete "in Greek".

services/notis/src/agent/tools.ts
```diff
@@ -0,0 +1,98 @@
...
+ * the WakeOutcome and answers with synthetic tool_results — no side effect
+ * ever happens inside runWake; the caller applies the outcome.
+ */
+
+export const CLIENT_TOOLS = [
+  {
+    name: "finish_wake",
+    description:
+      "End the wake, carrying your operator rationale (2-4 honest sentences; 1-2 for " +
+      "silence) about why this was, or was not, worth the reader's attention. REQUIRED " +
```
kouloumos, line 12: **REQUIRED in the prompt, unenforced in the core.** `finish_wake` is REQUIRED here, but the loop falls back to the last text block when it never arrives, and the canonical send fixture itself ends `end_turn` with prose and no `finish_wake` — the repair paths cannot fire on that shape. Either enforce it (a truncation-style marker when it is missing) or soften REQUIRED to match reality. The recorded evidence violating its own contract is worth knowing before golden fixtures become required checks.

services/notis/src/agent/templates.ts
```diff
@@ -0,0 +1,148 @@
...
+    buttons: [
+      { label: "Περισσότερα", kind: "url" },
+      { label: "Ας γνωριστούμε", kind: "quick_reply" },
+    ],
+  },
+  demos_update_agenda: {
+    name: "demos_update_agenda",
+    category: "utility",
+    bodyPrefix: "Πριν την επόμενη συνεδρίαση, κάτι που σε αφορά:\n\n",
+    bodySuffix: "\n\nΘα σου πω τι συζητήθηκε.",
```
kouloumos: **A written promise the design breaks 85–95% of the time.** Every agenda send ends with «Θα σου πω τι συζητήθηκε.» — an unconditional commitment, while silence-by-default means the follow-up usually never comes. This text is expensive to change after Meta approval, which is why it outranks most code here. Same file, lines 46 and 59: both intro templates introduce «ο βοηθός του OpenCouncil», while `system.md` requires the reader-facing persona «ο Νότης». Decide the persona and the promise before submission, not after.

services/notis/src/agent/templates.ts
```diff
@@ -0,0 +1,148 @@
...
+  return { template: name, body, footer: def.footer, buttons: def.buttons };
+}
+
+/** Which shell opens the thread, by how the reader entered Notis. */
+export function introTemplateFor(origin: "transition" | "signup"): TemplateName {
+  return origin === "transition" ? "demos_transition" : "demos_intro";
+}
+
+/** Which shell a cold proactive send must use, by the event that woke the agent. */
+export function templateForEvent(eventType: WakeEvent["type"]): TemplateName {
```
kouloumos, line 153: **Two event mappings frame sends wrongly.** A wake the agent scheduled after a **proactive** send maps to `demos_followup` («Σχετικά με αυτό που με ρώτησες») — an answer to a question nobody asked. The jul14 fixture contains exactly such a self-scheduled check. And `heartbeat` maps to `demos_update_news` while nothing produces heartbeats in PR 1. Worth a comment or a guard now, because the mapping goes live in PR 4. (The window-closed case follows the PRD's table and is fine.)

services/notis/src/agent/editorialPass.ts
```diff
@@ -0,0 +1,203 @@
...
+    }))
+    .filter((s) => s.id)
+    .sort((a, b) => b.discussionSeconds - a.discussionSeconds);
+
+  // Enrich the most-discussed subjects with their full record — but only for
+  // the post-meeting brief. The agenda brief runs before the meeting: subject
+  // details carry outcomes and exchanges that must not leak into a preview.
+  const detailed = new Map<string, unknown>();
+  if (phase === "summary") {
+    const top = subjects.slice(0, TOP_SUBJECTS_FOR_DETAIL).filter((s) => s.discussionSeconds > 0);
```
kouloumos, line 110: **The silence evidence is thinnest exactly where the product cares most.** Mechanism, confirmed: only the top eight subjects by `discussionSeconds > 0` get enriched; everything else is scored from a truncated description; the prompt then forbids research before deciding. So a high-impact hyperlocal item with zero debate time is scored from its agenda title and silenced without the record ever being read — and `discussionSeconds` measures councillor interest, not resident impact. Honest status: the one recorded silence audits as **correct**, so the harmed case is so far hypothetical. The decisive experiment is cheap: record a summary-phase scenario on a meeting with a consequential zero-debate subject.

Two adjacent gaps in the same function: an omitted `subjectId` in the model's response yields all-zeros with an empty note, indistinguishable from "deliberately unremarkable" (line 172) — validate coverage; and every `get_subject` failure collapses to `null` (line 112), so a degraded MCP silently yields a detail-free brief — count and surface failures.

services/notis/src/agent/prompt.ts
```diff
@@ -0,0 +1,113 @@
...
+    case "meeting_summarized":
+      return (
+        `A meeting has concluded and its record is published.\n` +
+        `Meeting: ${event.meetingName} (${event.meetingDate})${
+          event.adminBody ? ` — ${event.adminBody}` : ""
+        }, city ${event.cityId}, id ${event.meetingId}.\n` +
+        `Editorial brief (a map, not a source — read the record before quoting):\n${renderBrief(event.brief)}`
+      );
+    case "user_message":
+      return `The reader wrote to you on WhatsApp:\n«${event.text}»`;
```
kouloumos: **Inbound text is interpolated unescaped, and the repair loop trained a dangerous precedent.** The reader's WhatsApp text goes into the prompt verbatim inside an XML-tagged structure, and the repair mechanism has taught the model that user-role text beginning "(system check)" carries operator authority — bytes any inbound user can type once PR 3 lands. The "transcripts are data" rule covers tool results, not user turns. Fix before PR 3: escape or fence inbound text, and move the nudges off the plain user role (or mark them so the model can distinguish them from inbound).

services/notis/src/agent/types.ts
```diff
@@ -0,0 +1,154 @@
...
+export const DEFAULT_CONFIG: DepsConfig = {
+  // Sonnet 5: near-Opus quality on agentic work at ~40% of the cost and
+  // noticeably lower latency — the right default for per-resident wakes.
+  model: "claude-sonnet-5",
+  maxTurns: 8,
+  mcpUrl: "https://opencouncil.gr/mcp",
+  effort: "low",
+};
+
+export const JOURNAL_WINDOW = 30;
```
kouloumos, line 174: **The journal window truncates silently.** After thirty entries the model is told this is "the record of what this person has already been told" while older entries are silently gone — it will confidently repeat itself. One line: render an "older entries omitted" marker when the window clips.

services/notis/src/agent/journal.ts
```diff
@@ -0,0 +1,27 @@
+import { JournalEntry, WakeEvent } from "./types";
+
+/**
+ * The journal entry is built deterministically in code, never by the model,
+ * so its shape and honesty are testable without a model in the loop.
+ */
+export function buildJournalEntry(
```
kouloumos, line 7: **The journal never records profile rewrites or unsubscribes.** The agent's own memory changes (taste-profile rewrites, an unsubscribe) leave no journal entry, so future wakes cannot see that they happened. Cheap now; structural after PR 2 fixes the record shapes.

services/notis/src/lib/anthropic.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/** Real Deps.anthropic implementation over the beta namespace (MCP connector). */
+export const realAnthropic: AnthropicLike = {
+  async create(params: ModelRequest): Promise<ModelResponse> {
+    // Streamed under the hood: the SDK rejects non-streaming requests whose
+    // max_tokens could exceed its 10-minute heuristic (bites at 32k).
+    const response = await getClient()
+      .beta.messages.stream({
+        betas: [MCP_BETA, CACHE_TTL_BETA],
```
kouloumos, line 23: **Thinking is invisible to the instrument built to show it.** No `thinking` parameter is sent, so display defaults to omitted and every trace shows an empty thinking field — the inspector can only show a post-hoc self-report. One line: `thinking: { type: "adaptive", display: "summarized" }`.

services/notis/src/agent/pricing.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * claude-sonnet-5 pricing, USD per million tokens (list price — an intro rate
+ * of $2/$10 applies through 2026-08-31, so real spend runs lower until then).
+ * Cache reads bill at 0.1× input; cache-write rates depend on TTL (below).
+ */
+export const SONNET_5_RATES = {
+  inputPerMTok: 3,
+  outputPerMTok: 15,
+  // 1h-TTL cache writes bill at 2× base input (5m would be 1.25× = $3.75).
+  cacheWritePerMTok: 6,
```
kouloumos: **One cache-write rate for two TTLs.** All cache writes bill at the 1h rate (2×), but only the system breakpoint uses `ttl: "1h"` — the user-turn and moving breakpoints write at the 5m rate (1.25×). The inspector therefore overstates write costs, and the PRD §8 figures inherit the overstatement — the fix makes measured costs cheaper. The SDK usage object splits the two; use the split.

services/notis/src/env.mjs
```diff
@@ -0,0 +1,30 @@
...
+export const env = createEnv({
+  server: {
+    ANTHROPIC_API_KEY: z.string().min(1),
+    // Gates /admin and the agent API routes until PR 2 replaces this with
+    // shared-cookie validation against the main app's sessions.
+    NOTIS_ADMIN_SECRET: z.string().min(16),
+    // The MCP endpoint the agent researches against. Must be publicly
+    // reachable (the MCP connector calls it from Anthropic's side); override
+    // to point wakes at a preview deployment.
+    NOTIS_MCP_URL: z.string().url().default("https://opencouncil.gr/mcp"),
```
kouloumos, line 13: **Structurally Greece-only while both prompts promise otherwise.** The default MCP endpoint is the Greek realm, and the main app's MCP gate scopes results by Host — so «any city OpenCouncil covers» in `system.md` and the context pack is not true for Cyprus, Serbia, or France. Either scope the prompt's promise or plan multi-realm reads; today the prompt writes a check the endpoint cannot cash.

services/notis/prompts/system.md
```diff
@@ -0,0 +1,253 @@
...
+this, and be glad it arrived? If the answer is "probably not" or "mildly
+interesting", say nothing. Only "they would want to know this" is a message.
+
+Good reasons to stay quiet: it is procedural or routine; it is far from where
+they live and does not affect the whole municipality; you wrote recently and
+this is not more important; it is the first step of a long process, so use
+schedule_wakeup and come back; you are not certain it is true; you would only be
+writing to seem useful.
+
+## Political neutrality
```
kouloumos: **The neutrality rules cover office-holders only.** The section addresses «the mayor, every councillor and the opposition», but κοινότητες meetings are where private citizens speak, and the send fixture already quotes a resident's intervention (correctly, in that case — the named «Γέφυρα Πουλόπουλου» is a footbridge, and the live record lists the speaker as anonymous). Add one rule for private participants: no names, no identifying micro-detail beyond what the story needs.

services/notis/src/app/admin/(panel)/playground/components/SetupWizard.tsx
```diff
@@ -0,0 +1,410 @@
...
+  const [cities, setCities] = useState<CityOption[]>([]);
+  const [topics, setTopics] = useState<TopicOption[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [busy, setBusy] = useState(false);
+
+  const [citySearch, setCitySearch] = useState("");
+  const [drafts, setDrafts] = useState<CityDraft[]>([]);
+  const [focus, setFocus] = useState<MapFocus | null>(null);
+  const [name, setName] = useState("Μαρία");
+  const [profile, setProfile] = useState(
```
kouloumos, line 46: **The wizard population is not the migration population.** The PRD (§7) says the simulated profile is "seeded from preferences exactly as migration would seed it". Here the profile is hand-typed free text. Tuning on hand-written profiles and launching on mechanically-seeded ones means the tuning population differs from the launch population. Either implement the seeding rule (it will exist for PR 6 anyway) or note the gap in the README.

services/notis/src/app/api/dry-run/route.ts
```diff
@@ -0,0 +1,43 @@
...
+const requestSchema = z.object({
+  state: wakeStateSchema,
+  event: wakeEventSchema,
+  options: z
+    .object({
+      promptOverride: z.string().optional(),
+      contextPackOverride: z.string().optional(),
+      model: z.string().optional(),
+      maxTurns: z.number().int().optional(),
+      effort: effortSchema.optional(),
```
kouloumos, line 21: **The API accepts levers the UI cannot pull.** `contextPackOverride` and `effort` are accepted here, and no client sends them — while `effort` is the lever most likely to remove the need for the repair machinery. Meanwhile dead `SimSettings` plumbing ships in the store. Either expose effort in the playground (an A/B on effort is exactly what this playground is for) or delete the dead plumbing.

.github/workflows/preview-deploy.yml
```diff
@@ -96,16 +96,22 @@ jobs:
...
+          # Dependencies changed in a manifest but lockfile wasn't updated.
           # (The Nix build reads dependency hashes straight from the lockfile
           # via importNpmLock, so no flake.nix hash check is needed.)
           if [ -n "$PKG_JSON_CHANGED" ] && [ -z "$LOCKFILE_CHANGED" ]; then
-            OLD_DEPS=$(git show origin/${{ github.base_ref }}:package.json | jq -S '{dep: .dependencies, dev: .devDependencies}')
-            NEW_DEPS=$(jq -S '{dep: .dependencies, dev: .devDependencies}' package.json)
-            if [ "$OLD_DEPS" != "$NEW_DEPS" ]; then
-              echo "::error::Dependencies in package.json were modified but package-lock.json was not updated. Run 'npm install' and commit the updated lockfile."
+            CHANGED=false
+            for f in $PKG_JSON_CHANGED; do
```
kouloumos, line 108: **The loop crashes on a deleted manifest.** `git diff --name-only` lists deleted files, and `jq` on a missing path kills the step with a raw error under `pipefail` before the friendly message. Add `[ -f "$f" ] || { CHANGED=true; continue; }`. Also worth one line to the team: the check now includes `peerDependencies`, so a root peerDeps-only change without a lockfile update starts failing — deliberate, but stricter than before.

.github/workflows/ci.yml
```diff
@@ -26,3 +26,24 @@ jobs:
           pushFilter: "(\\.tgz$|opencouncil-.*-sources$)"
 
       - run: nix flake check --print-build-logs
+
+  notis:
```
kouloumos, line 30: The PR description says "path-filtered CI job"; this job has no path filter and runs on every PR. Running always is arguably right — notis depends on `packages/ui` — so fix the description, not the job.

services/notis/src/app/api/admin/login/route.ts
```diff
@@ -0,0 +1,42 @@
...
+let attempts: number[] = [];
+
+export async function POST(request: NextRequest) {
+  const now = Date.now();
+  attempts = attempts.filter((t) => now - t < 60_000);
+  if (attempts.length >= 5) {
+    return NextResponse.json({ error: "too many attempts, wait a minute" }, { status: 429 });
+  }
+
+  const parsed = bodySchema.safeParse(await request.json().catch(() => null));
```
kouloumos: This inlines the body of `parseJsonBody` extracted in this same PR; use the helper, or add a comment saying why not.

services/notis/src/app/admin/(panel)/conversations/[id]/view.tsx
```diff
@@ -0,0 +1,38 @@
...
+ */
+export function ConversationDetailView({ detail }: { detail: ConversationDetail }) {
+  const [selectedId, setSelectedId] = useState<string | undefined>();
+  const s = detail.summary;
+
+  return (
+    <>
+      <PageHeader title={s.userName}>
+        <span className="truncate text-xs text-muted-foreground">
+          {s.phone} · {s.cityNames.join(", ")} · από {s.startedAt.slice(0, 10)}
```
Comment (kouloumos) [inline services/notis/src/app/admin/(panel)/conversations/[id]/view.tsx]: Renders a raw ISO date; every sibling surface renders el-GR via `fmtDate`.

services/notis/src/app/admin/(panel)/page.tsx
```diff
@@ -0,0 +1,107 @@
+import { Activity, Euro, MessagesSquare, Users } from "lucide-react";
+import { PageHeader } from "./_components/PageHeader";
+import { LIVE_DATA, getPanelMetrics } from "./_lib/metrics";
+
+export const metadata = { title: "Νότης · admin" };
+
+function fmt(n: number | null, unit = ""): string {
```
kouloumos, line 9: `fmt` duplicates `fmtInt` from `_lib/format.ts`, and `MetricCard` re-implements `StatsCard` from `@opencouncil/ui`, whose JSDoc names this exact use.

services/notis/src/app/admin/login/page.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const router = useRouter();
+  const [secret, setSecret] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [busy, setBusy] = useState(false);
+
+  async function submit(e: React.FormEvent) {
+    e.preventDefault();
+    setBusy(true);
+    setError(null);
+    const res = await fetch("/api/admin/login", {
```
kouloumos: No catch: a network failure here leaves the button stuck on "…". Wrap in try/finally.

services/notis/src/app/admin/(panel)/_components/Timeline.tsx
```diff
@@ -0,0 +1,322 @@
...
+    const strip = stripRef.current;
+    const node = nextRef.current;
+    if (!strip || !node) return;
+    const left = node.offsetLeft + node.offsetWidth / 2 - strip.clientWidth / 2;
+    strip.scrollTo({ left: Math.max(0, left), behavior: "smooth" });
+  }, [nextId]);
+
+  function beginHover(item: WakeRecord, el: HTMLElement) {
+    clearTimeout(hoverTimer.current);
+    hoverTimer.current = setTimeout(() => {
```
kouloumos, line 110: The hover timer has no unmount cleanup; a pending timer fires `setHover` after unmount.

services/notis/services/notis/fixtures/scenarios/athens-jul6-agenda-silence.json
```diff
@@ -0,0 +1,247 @@
+{
```
kouloumos: Accidentally nested duplicate of this file exists at `services/notis/services/notis/fixtures/scenarios/` — byte-identical, not loaded by the suite; delete the nested tree (`record-scenario.ts` resolves its output from cwd, which is how it appeared).

packages/ui/src/lib/pagination.ts
```diff
@@ -0,0 +1,36 @@
+export interface PaginationParams {
```
kouloumos, line 1: `PaginationParams` now exists here and in `src/lib/db/types`; re-export the type from the package so the copies cannot drift.

kouloumos (commented):
This is a strong PR 1, and the strongest parts verify rather than merely read well. The workspace extraction was checked byte by byte: all fifty moved components differ from their originals only in import paths, the one default export has an explicit stub, shared dependencies resolve to identical versions before and after, and a runtime A/B of the preview against a control preview showed zero differences that this PR causes. The agent core holds up the same way: `runWake` is genuinely pure over its injected dependencies, the repair machinery is a bounded state machine rather than patch accumulation, the store reducer never mutates, and the React layer carries zero lint suppressions. Two audits that go beyond the code also came back good: the recorded silence is correct against the live July 6 meeting record, and the recorded Greek passes the prompt's own neutrality and grounding tests. The recent fixes hold — the save flush, the ΣΤΟΠ queue stop, and the dead controls are verified fixed, and inbound-still-answers matches the PRD.

The review found one item that blocks using the playground at the current tip, four items I would fix before merge, and a set of forward-looking items that this slice makes cheap to fix now and expensive later. Specifics are inline; the themes are these.

**The playground is broken at the tip, and the branch now contains both halves of the break.** The main-app MCP server appends an advisory text block to every tool response since the analytics change (a13776b1, brought into this branch by the merge). The Notis MCP client cannot parse any response that carries it, and every consumer downgrades the failure to an empty result. Every city in the setup wizard reports «καμία δημοσιευμένη συνεδρίαση», and the editorial pass throws. The inline comment carries the two-part fix. One related point for the team, outside this PR: the analytics wrapper changed the de facto contract of the public MCP endpoint. Any programmatic consumer that parses the joined text content broke the same day. The PRD's own design says Notis consumes the MCP "exactly like any outside consumer" — and this is what happened to the first outside consumer.

**The wake record can misstate what happened, and the record is the product's instrument panel.** A turn cut at the token ceiling records itself as a deliberate silence. A repaired wake leaves no trace of the repair. The journal window truncates silently under a prompt that calls the journal a complete record. Each is small in code. Together they matter because the silence rate is the PRD's headline metric, the review queue in PR 5 reads these same records, and adaptive frequency reads the journal as its only evidence.

**The golden evidence is stale, and its assertions are count-only.** Both fixtures were recorded on Opus before the model change and before the agenda embargo, which one fixture now violates. The assertions cannot see message content, so mutation tests that corrupt every sent message replay green. The offline lane's design is sound — deterministic replay guards the code, and comparing full text is valid there. The gap is that the PRD assigns golden scenarios a second job, guarding prompt changes, and that requires a separate live lane with behavioral assertions. The comment inline separates the cheap fix from the future work.

Two one-sentence deltas against the PR description: the notis CI job has no path filter, so it runs on every PR — which is fine, and arguably better, but not what the description says; and the golden lane is not yet the prompt-regression check the description implies, per the inline comment.

christosporios:
Thank you for this review — the A/B verification and the reproduced fast-forward double-run especially. Pushed four commits addressing the blocker, all four pre-merge items, and most of the forward-looking list:

- `2b17ae37` **MCP client survives advisory text blocks** — parses text blocks individually, returns the first JSON block; playground fetchers pin shapes with zod so the next contract drift fails loudly. Session recovery is single-flight. Five regression tests.
- `4d2cb362` **Wake records tell the truth** — `truncated` on outcome+journal for max_tokens cuts (calls unprocessed); repair nudges recorded as injected trace turns + a `repairs` list on the outcome (replay skips them); the fourth repair path got the missing rationale protection; `finishWakeMissing` records the unenforced contract; journal records profile rewrites/unsubscribes and renders an 'older entries omitted' marker; editorial pass surfaces unscored subjects and failed detail fetches; inbound text fenced in `<reader_message>` + system.md declares system checks structural (pre-PR 3); context pack shrunk to cities-only; TTL-split cache pricing; `thinking: summarized`.
- `07c63e3f` **Playground** — tick-yield + ledger + 50-step cap kills the double-run; run-epoch guards rewind/reset against in-flight wakes (both disable while busy); `restoreSnapshot` prunes the abandoned timeline's snapshots; unreadable/newer stores are stashed aside instead of clobbered; the trace-gone message names both causes.
- `508a3f54` **Sweep** — golden `expected` now carries full message texts / scheduled wakes / unsubscribe / repairs and the test asserts them when present; record-scenario resolves paths from the package root (nested duplicate tree deleted); effort exposed in the prompt editor; StatsCard/fmtInt/fmtDate/parseJsonBody/try-finally/hover-cleanup/PaginationParams/preview-deploy guard. Mutation-verified test for the moving cache breakpoint (your three mutations welcome as well if they cover more).

PR description corrected on both deltas (CI runs on every PR; golden lane's second job is future work, now in the README).

**Left open, needing decisions or live runs**: the template texts («Θα σου πω τι συζητήθηκε» promise + «ο βοηθός του OpenCouncil» vs «ο Νότης») before Meta submission; the realm scoping of «any city»; the zero-debate editorial experiment; re-recording both fixtures on the shipped pipeline (script is ready; recording after the prompt changes settle); and the migration-seeded wizard profile (README-noted, needed for PR 6 anyway).

christosporios:
Pushed a2f6d1d4: template decisions applied in Bird and mirrored in `templates.ts` — the reader-facing persona is now **ο Δήμος** (intro/transition texts, system.md, context pack, simulator header; Notis stays the internal codename), the agenda shell closes with the keepable «Αν συζητηθεί κάτι που σε αφορά, θα σου πω.», the intro quick reply is «Τι είναι αυτό;» (with prompt guidance to answer it as a real question + explain link), and a new `demos_checkin` template exists for the low-frequency downgrade (unused until a PR 4+ policy adopts it). All revised Bird versions were submitted for Meta review; intro, transition and the first checkin version were approved within minutes.

christosporios:
Pushed 5f5b03b8: correction — the persona is **ο Νότης**; the previous commit's rename to «ο Δήμος» was my misread of a decision. Reverted in the template bodies (fixed in Bird too, corrected versions submitted to Meta), system.md, the context pack and the simulator header. The rest of the template round stands (conditional agenda closing, «Τι είναι αυτό;» intro button + prompt guidance, demos_checkin).

christosporios:
Pushed f7f6d89b: golden fixtures re-recorded on the shipped pipeline (sonnet-5, current prompts, briefs regenerated through the phase-aware editorial pass — a new `refresh-brief` script handles that, since the brief is wake input). The agenda fixture is still a silence and no longer violates the embargo. The send fixture drifted honestly: two grounded messages instead of one, and the self-scheduled aftermath follow-up is gone (the no-aftermath-scheduling rule at work — it was also the exact case behind the `demos_followup` mapping gap). Full-content assertions active on both. ~$0.35 total.

christosporios:
Pushed c8946be1: the coverage promise in system.md is scoped to the Greek archive — the agent reads the Greek-realm MCP, so «any city OpenCouncil covers» was a check the endpoint cannot cash for other realms. Realm design settled and recorded in the PRD: realm is a property of cities and meetings, never users; Notis launches Greece-only; PR 2's views will carry city realm+language, PR 4's poller picks the MCP origin from the meeting's realm, and a reader's default language derives from their preference cities. No speculative plumbing lands before those PRs.

PR #599, opened 2026-08-09
feat(mcp): speaker roles and vote tallies on subject records

christosporios:
Two data enrichments to the public MCP server, both motivated by agent testing on #598 but independently useful to every MCP consumer (ChatGPT, Claude, and later Notis):

## Speaker roles on quoted speech

`get_subject` (contributions, `introducedBy`), `get_subject_transcript` (utterances) and `get_transcript` (segments) now attach each speaker's most senior role **active on the meeting date**:

```json
{ "speaker": "Χάρης Δούκας", "role": "Δήμαρχος", "text": "…" }
```

Before this, records carried only names, so model consumers guessed offices from context — in testing, the mayor of Argos got labeled «δημοτικός σύμβουλος». Labels prefer the stored role name («Αντιδήμαρχος Παιδείας & Διαγενεακής Αλληλεγγύης»), fall back to «Δήμαρχος» for a city-head role, and otherwise state party membership («παράταξη «Αθήνα Ψηλά»») — a fact, never an invented title. Uses the existing `isRoleActiveAt` / `getRoleTypePriority` helpers.

Cost: zero extra queries on the subject path (roles were already fetched), one widened batch select on transcripts. `requireVisibleMeeting` now also returns the meeting date it already looks up.

## Precomputed vote tally

`get_subject` adds `voteSummary` (`{"FOR": 15, "AGAINST": 6}`) next to the `votes` array. Two test wakes miscounted the same subject's votes two different ways (12–6 and 13–6); consumers should copy a tally, not count an array.

## Testing

`npx tsc --noEmit` and the MCP test suite (20 tests) pass; verified end-to-end against the local MCP: Δούκας → «Δήμαρχος», Λιμνιωτάκη → «Πρόεδρος», deputy-mayor role name verbatim, party fallback for plain councillors.

---

> [!NOTE]
> **Low Risk**
> Additive MCP fields and shared UI labeling with tests; visibility/auth behavior unchanged aside from returning meeting date from the existing gate query.
> 
> **Overview**
> **MCP responses now include meeting-date speaker roles and a precomputed vote tally**, so agents and other consumers stop inferring offices from names alone.
> 
> Role badge text is centralized: `getRoleText` and `getRoleLabelAt` move from `RoleDisplay` into `src/lib/utils/roles`, with `filterActiveRoles` accepting an optional `at` date. `RoleDisplay` imports the shared helpers. MCP loads Greek `Person` translations and attaches a `role` field on `get_subject` (`introducedBy`, contributions), `get_subject_transcript` utterances, and `get_transcript` segments, using each person’s top active role on the meeting date. `requireVisibleMeeting` now returns `dateTime` (no extra query) for that resolution; transcript paths widen person selects to include roles where needed.
> 
> `get_subject` also adds **`voteSummary`** via `calculateVoteResult`, matching site rules (e.g. `PRESENT` / `DID_NOT_VOTE` are not counted as votes). MCP tool descriptions document role timing and vote semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a9d6b3ba9216c6aa20a2e1aac5641ee7d82e803. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR enriches public MCP subject and transcript responses with meeting-date speaker roles and adds a precomputed subject vote tally.
- Moves shared role-label rendering into `src/lib/utils/roles.ts`.
- Loads role relations on transcript speaker records and resolves labels against the meeting date.
- Returns `voteSummary` alongside individual subject votes.
- Extends meeting visibility lookup results with the meeting date and updates MCP descriptions and tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/mcp/data.ts | Adds role labels and vote summaries to MCP responses; the previously flagged role-helper import now uses the required source alias. |
| src/lib/utils/roles.ts | Introduces shared role-text and date-sensitive primary-role labeling helpers. |
| src/lib/mcp/gate.ts | Returns the selected meeting date with the existing visibility result for role resolution. |
| src/components/persons/RoleDisplay.tsx | Reuses the extracted shared role-text helper without changing badge behavior. |
| src/lib/mcp/server.ts | Documents the newly exposed speaker-role fields and vote semantics. |

<sub>Reviews (4): Last reviewed commit: ["feat(mcp): speaker roles on subjects and..."](https://github.com/schemalabz/opencouncil/commit/3a9d6b3ba9216c6aa20a2e1aac5641ee7d82e803) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51572313)</sub>

### DIFF
--- a/src/lib/mcp/data.ts
+++ b/src/lib/mcp/data.ts
@@ -12,6 +12,7 @@ import { upsertHighlightCore, canUserEditCity, canActorManageHighlight } from '@
 import { requestGenerateHighlightCore } from '@/lib/tasks/generateHighlight-core';
 import { NotFoundError, UnauthorizedError, BadRequestError, ForbiddenError } from '@/lib/api/errors';
 import { canSeeUnreleased, requireVisibleMeeting } from './gate';
+import { calculateVoteResult } from '@/lib/utils/votes';
 import {
     renderOptionsFromRequestBody,
     resolveRenderOptions,
@@ -294,6 +295,10 @@ export async function mcpGetSubject(subjectId: string, identity: McpIdentity) {
             }
             : null,
         votes: subject.votes.map(vote => ({ person: vote.person.name, vote: vote.voteType })),
+        // Precomputed tally so consumers (humans and agents alike) never have
+        // to count the array themselves — and with the same rules the site
+        // uses: PRESENT and DID_NOT_VOTE are declarations, not votes.
+        voteSummary: subject.votes.length > 0 ? calculateVoteResult(subject.votes) : null,
         url: urls.subject(subject.cityId, subject.councilMeetingId, subject.id),
     };
 }
--- a/src/lib/mcp/data.ts
+++ b/src/lib/mcp/data.ts
@@ -12,7 +12,10 @@ import { upsertHighlightCore, canUserEditCity, canActorManageHighlight } from '@
 import { requestGenerateHighlightCore } from '@/lib/tasks/generateHighlight-core';
 import { NotFoundError, UnauthorizedError, BadRequestError, ForbiddenError } from '@/lib/api/errors';
 import { canSeeUnreleased, requireVisibleMeeting } from './gate';
+import { getRoleLabelAt, RoleTextTranslator } from '@/lib/utils/roles';
 import { calculateVoteResult } from '@/lib/utils/votes';
+import { roleWithRelationsInclude } from '@/lib/db/types';
+import { getTranslations } from 'next-intl/server';
 import {
     renderOptionsFromRequestBody,
     resolveRenderOptions,
@@ -43,6 +46,16 @@ const urls = {
     moment: (cityId: string, meetingId: string, seconds: number) =>
         `${currentBaseUrl()}/${cityId}/${meetingId}?t=${Math.floor(seconds)}`,
 };
+
+
+/**
+ * Speaker role labels ride along with quoted speech so consumers copy titles
+ * instead of guessing them. The label is the site's own (getRoleLabelAt),
+ * resolved as of the meeting date; MCP serves the Greek public record, so it
+ * is always rendered from the Greek messages.
+ */
+const getRoleTranslations = (): Promise<RoleTextTranslator> =>
+    getTranslations({ locale: 'el', namespace: 'Person' });
 
 /** Turn contribution markdown ("[text](REF:UTTERANCE:id)") into plain text. */
 function stripRefLinks(text: string): string {
@@ -268,7 +281,12 @@ export async function mcpGetMeeting(cityId: string, meetingId: string, identity:
 export async function mcpGetSubject(subjectId: string, identity: McpIdentity) {
     const subject = await getSubject(subjectId);
     if (!subject) throw new NotFoundError('Subject not found');
-    await requireVisibleMeeting(subject.cityId, subject.councilMeetingId, identity);
+    const { dateTime: meetingDate } = await requireVisibleMeeting(
+        subject.cityId,
+        subject.councilMeetingId,
+        identity
+    );
+    const t = await getRoleTranslations();
 
     return {
         id: subject.id,
@@ -280,10 +298,15 @@ export async function mcpGetSubject(subjectId: string, identity: McpIdentity) {
         topic: subject.topic?.name ?? null,
         location: subject.location?.text ?? null,
         introducedBy: subject.introducedBy
-            ? { id: subject.introducedBy.id, name: subject.introducedBy.name }
+            ? {
+                id: subject.introducedBy.id,
+                name: subject.introducedBy.name,
+                role: getRoleLabelAt(subject.introducedBy.roles, t, meetingDate),
+            }
             : null,
         contributions: subject.contributions.map(contribution => ({
             speaker: contribution.speaker?.name ?? contribution.speakerName ?? 'Unknown',
+            role: getRoleLabelAt(contribution.speaker?.roles, t, meetingDate),
             text: stripRefLinks(contribution.text),
         })),
         decision: subject.decision
@@ -311,7 +334,12 @@ export async function mcpGetSubjectTranscript(subjectId: string, page: number, i
         select: { id: true, name: true, cityId: true, councilMeetingId: true },
     });
     if (!subject) throw new NotFoundError('Subject not found');
-    await requireVisibleMeeting(subject.cityId, subject.councilMeetingId, identity);
+    const { dateTime: meetingDate } = await requireVisibleMeeting(
+        subject.cityId,
+        subject.councilMeetingId,
+        identity
+    );
+    const t = await getRoleTranslations();
 
     const where: Prisma.UtteranceWhereInput = {
         discussionSubjectId: subjectId,
@@ -333,7 +361,16 @@ export async function mcpGetSubjectTranscript(subjectId: string, page: number, i
                 speakerSegment: {
                     select: {
                         speakerTag: {
-                            select: { label: true, person: { select: { id: true, name: true } } },
+                            select: {
+                                label: true,
+                                person: {
+                                    select: {
+                                        id: true,
+                                        name: true,
+                                        roles: roleWithRelationsInclude,
+                                    },
+                                },
+                            },
                         },
                     },
                 },
@@ -352,6 +389,7 @@ export async function mcpGetSubjectTranscript(subjectId: string, page: number, i
             utteranceId: utterance.id,
             speaker: utterance.speakerSegment.speakerTag.person?.name
                 ?? utterance.speakerSegment.speakerTag.label,
+            role: getRoleLabelAt(utterance.speakerSegment.speakerTag.person?.roles, t, meetingDate),
             startSec: utterance.startTimestamp,
             endSec: utterance.endTimestamp,
             text: utterance.text,
@@ -370,7 +408,8 @@ export async function mcpGetTranscript(
     options: { page: number; segmentsPerPage: number; includeUtteranceIds: boolean; personId?: string },
     identity: McpIdentity
 ) {
-    await requireVisibleMeeting(cityId, meetingId, identity);
+    const { dateTime: meetingDate } = await requireVisibleMeeting(cityId, meetingId, identity);
+    const t = await getRoleTranslations();
 
     const allSegments = await getTranscript(meetingId, cityId);
 
@@ -401,9 +440,10 @@ export async function mcpGetTranscript(
     const personIds = [...new Set(pageSegments.map(s => s.speakerTag.personId).filter((id): id is string => !!id))];
     const persons = await prisma.person.findMany({
         where: { id: { in: personIds } },
-        select: { id: true, name: true },
+        select: { id: true, name: true, roles: roleWithRelationsInclude },
     });
     const personNames = new Map(persons.map(person => [person.id, person.name]));
+    const personRoles = new Map(persons.map(person => [person.id, getRoleLabelAt(person.roles, t, meetingDate)]));
 
     return {
         cityId,
@@ -414,6 +454,7 @@ export async function mcpGetTranscript(
         segments: pageSegments.map(segment => ({
             speaker: (segment.speakerTag.personId && personNames.get(segment.speakerTag.personId))
                 || segment.speakerTag.label,
+            role: (segment.speakerTag.personId && personRoles.get(segment.speakerTag.personId)) || null,
             startSec: segment.startTimestamp,
             endSec: segment.endTimestamp,
             summary: segment.summary?.text ?? null,
--- a/src/lib/mcp/gate.ts
+++ b/src/lib/mcp/gate.ts
@@ -29,18 +29,19 @@ export async function requireVisibleMeeting(
     cityId: string,
     meetingId: string,
     identity: McpIdentity
-): Promise<{ released: boolean }> {
+): Promise<{ released: boolean; dateTime: Date }> {
     const meeting = await prisma.councilMeeting.findFirst({
         // Realm-scoped: a connector added on one domain must not reach another
         // realm's councils. Covers meetings, subjects and transcripts, which
         // all pass through here.
         where: { cityId, id: meetingId, city: { realm: currentRealm() } },
-        select: { released: true },
+        select: { released: true, dateTime: true },
     });
 
     if (!meeting || (!meeting.released && !(await canSeeUnreleased(identity, cityId)))) {
         throw new NotFoundError('Meeting not found');
     }
+
 
     return meeting;
 }
--- a/src/components/persons/RoleDisplay.tsx
+++ b/src/components/persons/RoleDisplay.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { useTranslations } from 'next-intl';
 import { Link } from '@/i18n/routing';
 import { Badge } from '../ui/badge';
-import { cn, sortRolesByPriority, getPrimaryRole } from '@/lib/utils';
+import { cn, sortRolesByPriority, getPrimaryRole, getRoleText } from '@/lib/utils';
 import { Star, Building, Users } from 'lucide-react';
 import { RoleWithRelations } from '@/lib/db/types';
 const BADGE_MAX_CHARS_SMALL = 40;
@@ -39,38 +39,6 @@ const getRoleIconColor = (role: RoleWithRelations) => {
     if (role.cityId) return "text-[#a4c0e1]";
     if (role.administrativeBodyId) return "text-muted-foreground/70";
     return "text-muted-foreground/70";
-};
-
-const getRoleText = (role: RoleWithRelations, t: ReturnType<typeof useTranslations>) => {
-    // Party roles (has partyId)
-    if (role.partyId && role.party) {
-        const text = role.party.name;
-        if (role.isHead) return `${text} (${t('partyLeader')})`;
-        if (role.name) return `${text} - ${role.name}`;
-        return text;
-    }
-
-    // Administrative body roles (has administrativeBodyId)
-    if (role.administrativeBodyId && role.administrativeBody) {
-        // For administrative body roles, show role name first (e.g., "President"), then body name
-        if (role.name) {
-            return `${role.name} - ${role.administrativeBody.name}`;
-        }
-        // If no role name but is head, use the localized "president" label
-        if (role.isHead) {
-            return `${t('president')} - ${role.administrativeBody.name}`;
-        }
-        // Otherwise just show the administrative body name
-        return role.administrativeBody.name;
-    }
-
-    // City-level roles (has cityId but no partyId or administrativeBodyId, e.g., mayor)
-    if (role.cityId && !role.partyId && !role.administrativeBodyId) {
-        if (role.isHead) return t('mayor');
-        return role.name || t('member');
-    }
-
-    return role.name || t('member');
 };
 
 const getRoleVariant = (role: RoleWithRelations) => {
--- a/src/lib/mcp/__tests__/gate.test.ts
+++ b/src/lib/mcp/__tests__/gate.test.ts
@@ -28,31 +28,32 @@ describe('requireVisibleMeeting', () => {
     });
 
     it('passes released meetings for everyone', async () => {
-        mockMeetingFindFirst.mockResolvedValue({ released: true });
-        await expect(requireVisibleMeeting('athens', 'm1', null)).resolves.toEqual({ released: true });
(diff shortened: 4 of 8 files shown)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/db8172599ef9a759f968672b53d9d272c907fd93..5ad84cb265ec5de6a616b3fcc1d5b54c0ea93b52) to address both review findings:

- Bugbot: the «Δήμαρχος» fallback now uses the existing `isMayorRole()` predicate (city-level head with no party/admin-body dimension) instead of the looser `cityId && isHead`, so a party-leader or committee-chair role with a null name can no longer be mislabeled as mayor
- Greptile: the roles-helper import uses the `@/` alias per repo convention

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/5ad84cb265ec5de6a616b3fcc1d5b54c0ea93b52..2e180d87889926c3e74c6b73d7cfad533fc44dc5) to remove duplicated role-selection logic:

- `roleLabelAt` now composes the existing `getSpeakerDisplayInfo` from `@/lib/utils/roles` instead of re-implementing the filter-active-then-sort-by-priority selection inline. Only the label formatting (name → Δήμαρχος derivation → party fallback) stays local.
- The hand-written `RoleForLabel` shape is replaced by the helper's own parameter type.

src/lib/mcp/data.ts
```diff
@@ -43,6 +44,24 @@ const urls = {
...
+ * speech so consumers copy titles instead of guessing them. Display logic is
+ * getSpeakerDisplayInfo (the same the site uses); the fallback states party
+ * membership, a fact, rather than inventing a title.
+ */
+function roleLabelAt(roles: RoleForLabel[] | undefined, at: Date): string | null {
+    if (!roles || roles.length === 0) return null;
+    const { party, role } = getSpeakerDisplayInfo(roles, at);
+    if (role?.name) return role.name;
+    if (role && isMayorRole(role)) return 'Δήμαρχος';
+    if (party?.name) return `παράταξη «${party.name}»`;
```
kouloumos: **12.5% of returned rows carry a role label that omits the speaker's headship.**

I measured 2,035 transcript segments across 36 meetings in 6 cities. 814 rows (40%) return «παράταξη …». Of those, 254 belong to someone who holds an active head role at the meeting date.

There are two mechanisms, and `getRoleText` (`RoleDisplay.tsx:44`) handles both.

**Party leaders — 147 rows.** The highest-priority role is the party role, so `getSpeakerDisplayInfo` returns `role: null`. It also returns `isPartyHead: true`, which this destructure drops.

Μπακογιάννης in `athens/jun10_2026` returns `παράταξη «Αθήνα Ψηλά»`. His badge on that same meeting's transcript page reads `Αθήνα Ψηλά (Επικ.)`.

**Chairs with a null role name — 107 rows.** Here the head role *is* selected, at priority 2. But `role.name` is null and the role is not a mayor role, so the label falls through to party.

Δρούλια chairs `athens/jul20_2026` (6η Δημοτική Κοινότητα) and speaks in 11 of the 30 segments on page 1. She returns «παράταξη «Αθήνα Τώρα»». `getRoleText` would return `Πρόεδρος - 6η Δημοτική Κοινότητα`. The facts sit on the role row. The selects at L377/L449 just do not fetch `administrativeBody`.

A third case is worth knowing. Λουτσέτης holds `Αναπληρωματικό Μέλος` in `Δημοτική Επιτροπή`, and spoke in a `Δημοτική Επιτροπή` meeting. That role loses to his party role, because priority 3 beats 5. This is `getRoleTypePriority`'s own TODO, in the wild. `requireVisibleMeeting` is already being widened, so adding `administrativeBodyId` to that select is one word.

None of this is wrong. The "a fact, never an invented title" rule holds throughout. But the role label omits the headship in 1 row in 8, and the site already shows the fuller label one tab away.

Two smaller points:

- Add a line to the tool descriptions saying the label is resolved **as of the meeting date**. `get_person` returns roles for all time, so a model comparing the two sees an apparent contradiction.
- `name_en` is never read. A role with `name: null, name_en: 'Deputy Mayor'` returns `null`, although the title is stored. `mapRoles` in this file already exposes it.

src/lib/mcp/data.ts
```diff
@@ -294,6 +322,14 @@ export async function mcpGetSubject(subjectId: string, identity: McpIdentity) {
...
             : null,
         votes: subject.votes.map(vote => ({ person: vote.person.name, vote: vote.voteType })),
+        // Precomputed tally so consumers (humans and agents alike) never have
+        // to count the array themselves — miscounted votes are a real failure.
+        voteSummary: subject.votes.length > 0
+            ? subject.votes.reduce<Record<string, number>>((acc, vote) => {
+                acc[vote.voteType] = (acc[vote.voteType] ?? 0) + 1;
+                return acc;
+            }, {})
+            : null,
```
kouloumos: **`calculateVoteResult` already does this**, in `src/lib/utils/votes.ts:16` — 6 call sites and 11 unit tests.

It encodes the rule the schema states: `PRESENT` and `DID_NOT_VOTE` are declarations, not votes, and stay out of `totalVotes`.

The two now disagree. On `cmpshaqkc0441g5e7p6sd4kam` (`athens/may29_2026`) this returns `{"FOR":26,"AGAINST":6,"DID_NOT_VOTE":3}`. An agent that sums the map reports 35 votes cast. `calculateVoteResult` reports `totalVotes: 32`.

The field exists so consumers stop counting the array. It should agree with what the site computes. `passed` and `isUnanimous` come free.

The raw enum also carries a translation trap:

```
ABSTAIN       // ΛΕΥΚΟ — vote with no position (counts as a vote)
DID_NOT_VOTE  // ΑΠΟΧΗ — declined to participate (declaration, not a vote)
```

`ABSTAIN` is the Greek word for *blank ballot*. The Greek word for *abstain* is the other member. Asked «πόσοι απείχαν;», a model reads `ABSTAIN` and answers backwards. Worth one line in the tool description either way.

```suggestion
        voteSummary: subject.votes.length > 0 ? calculateVoteResult(subject.votes) : null,
```

src/lib/mcp/data.ts
```diff
@@ -328,7 +368,16 @@ export async function mcpGetSubjectTranscript(subjectId: string, page: number, i
...
                     select: {
                         speakerTag: {
-                            select: { label: true, person: { select: { id: true, name: true } } },
+                            select: {
+                                label: true,
+                                person: {
+                                    select: {
+                                        id: true,
+                                        name: true,
+                                        roles: { include: { party: true } },
```
kouloumos: **This path repeats the role label on 150 rows; the lighter path memoises it.**

The nested select sits under a `findMany` of 150 utterances (`SUBJECT_TRANSCRIPT_PAGE_SIZE`). The engine serialises the speaker's roles once per row, and `roleLabelAt` re-runs its filter and sort per utterance.

Seventy lines down, `mcpGetTranscript` does the opposite: a deduped `person.findMany`, with the label memoised into a `Map` (L449-452). That path returns 40–100 segments and is consumer-tunable. The asymmetry is inverted.

Measured: one page grew from 37,931 to 46,580 bytes, **+22.8%**. That page carries **3 distinct speakers**, and repeats `Αντιδήμαρχος Οικονομικού Προγραμματισμού` 83 times. Almost all the growth is duplication, not information. `get_transcript` grows 3.3%, because its rows already carry long text.

This matters because the responses are billed and context-budgeted per call, and `instructions.ts` steers agents onto this path ("prefer per-subject transcripts").

Both new selects already fetch `person.id`, then discard it. Emitting `personId` per row, plus a `speakers: { id: { name, role } }` header, would cut the overhead about tenfold. It would also give agents a handle to reach `get_person`, which they cannot do from a transcript today.

If the per-row repetition is deliberate, apply `simplifyRoleName` at minimum. The docx renders `Αντιδήμαρχος` for the same meeting where this emits the full portfolio string.

src/lib/mcp/data.ts
```diff
@@ -43,6 +44,24 @@ const urls = {
         `${currentBaseUrl()}/${cityId}/${meetingId}?t=${Math.floor(seconds)}`,
 };
 
+
+type RoleForLabel = Parameters<typeof getSpeakerDisplayInfo>[0][number];
```
kouloumos: **`RoleWithRelations` already covers this.**

`roleWithRelationsInclude` and `RoleWithRelations` live in `src/lib/db/types/roles.ts:13,21`, with 13 call sites. One of them is `getSubject`, which feeds `mcpGetSubject` on the other path of this same function. So the function reads roles in the house shape on one path, and a narrower hand-rolled shape on the other.

Switch the two inline includes at L377/L449. Then `roleLabelAt` takes `RoleWithRelations[]`, and this alias goes away. That also brings `administrativeBody`, which both the chair label and `getRoleText` need — `RoleWithRelations` is the type `getRoleText` already takes.

src/lib/mcp/gate.ts
```diff
@@ -29,18 +29,19 @@ export async function requireVisibleMeeting(
...
         // all pass through here.
         where: { cityId, id: meetingId, city: { realm: currentRealm() } },
-        select: { released: true },
+        select: { released: true, dateTime: true },
     });
 
     if (!meeting || (!meeting.released && !(await canSeeUnreleased(identity, cityId)))) {
         throw new NotFoundError('Meeting not found');
     }
+
```
kouloumos:
```suggestion
```

`git diff --check` flags this as trailing whitespace.

kouloumos (commented):
I ran this against the preview and prod side by side. Anonymous MCP reads make a real A/B possible, so the numbers below are measured, not inferred from the diff.

Checked and fine:

- Date-awareness works. Μπακογιάννης returns `παράταξη «Αθήνα Ψηλά»` for a 2026 meeting, not his old mayoralty.
- `get_subject` adds zero queries, as described.
- No existing field changed shape or type.
- The extra role fetch costs +40–80ms, after subtracting the host offset I measured on unchanged endpoints.

## The role label already exists in the UI

`getRoleText` (`src/components/persons/RoleDisplay.tsx:44`) already builds this label. It renders on every speaker header in the transcript UI, through `SpeakerSegment.tsx` → `PersonBadge` → `RoleDisplay`.

The two surfaces now disagree. Same speaker, same meeting:

| `athens/jun10_2026` | transcript badge | MCP `role` |
|---|---|---|
| Μπακογιάννης — party leader | `Αθήνα Ψηλά (Επικ.)` | `παράταξη «Αθήνα Ψηλά»` |
| Δρούλια — chairs 6η Δημ. Κοινότητα, role name null | `Πρόεδρος - 6η Δημοτική Κοινότητα` | `παράταξη «Αθήνα Τώρα»` |
| any mayor | `t('mayor')`, localized | `'Δήμαρχος'`, hardcoded |

`getRoleText` handles every case in my first inline comment:

- party head → `` `${party.name} (${t('partyLeader')})` ``
- admin-body head with a null name → `` `${t('president')} - ${administrativeBody.name}` ``
- city head → `t('mayor')`

The `Person` namespace already holds the keys `mayor`, `president`, `partyLeader` and `member`. Routing through them also removes the hardcoded `'Δήμαρχος'`.

Two caveats, so the ask stays fair:

- `getRoleText` is a module-local `const` in a `.tsx`, and it takes `t` from `useTranslations`. So this is an extraction into `roles.ts`, not an import. MCP would need `getTranslations` from `next-intl/server`.
- The badge has a bug this PR does not. `PersonBadge.tsx:37` passes `filterActiveRoles(person.roles)` with no date, so the badge shows roles active *today*. The next line passes `date` to `getPartyFromRoles`. Threading the meeting date is what this PR gets right.

So the fix helps both surfaces: lift `getRoleText` into `roles.ts`, give it the meeting date this PR already threads, and point both at it.

## Tests

The 20 passing tests do not import `data.ts`. Grepping `mcp/data|mcpGetSubject|mcpGetTranscript` across every test file returns nothing.

`roleLabelAt` is pure — no clock, no I/O. One `export` makes it testable. The case I would most want pinned is Bugbot's: an `isHead` role that is not the mayor must not return `Δήμαρχος`. A fixed finding with no regression test can come back silently.

`gate.test.ts` still asserts `{released: true}`, a shape production can no longer return. The return-type annotation is the real guard there, so this is hygiene, not a bug.

---

I would take `calculateVoteResult` and the role label first. Both are small, and both remove a disagreement between MCP and the site.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/2e180d87889926c3e74c6b73d7cfad533fc44dc5..3a9d6b3ba9216c6aa20a2e1aac5641ee7d82e803) to take the review's core items — thanks @kouloumos, the measured A/B made these easy to act on:

- **`getRoleText` extracted** from `RoleDisplay.tsx` into `roles.ts` (taking a `RoleTextTranslator`, satisfied by `useTranslations` and `getTranslations` alike), plus a new `getRoleLabelAt(roles, t, date)` = `getPrimaryRole(filterActiveRoles(roles, date))` + `getRoleText`. MCP and the badges now render the same label from the same code: party leaders get «Αθήνα Ψηλά (Επικεφαλής)», chairs get «Πρόεδρος - 6η Δημοτική Κοινότητα», and the hardcoded 'Δήμαρχος' now comes from the `Person` message catalog (`locale: 'el'` — MCP serves the Greek record).
- **`voteSummary` is now `calculateVoteResult`** — same shape and rules the site computes, with `totalVotes`/`passed`/`isUnanimous`, and PRESENT/DID_NOT_VOTE excluded from totals.
- **`RoleWithRelations`/`roleWithRelationsInclude`** replace the two inline includes and the hand-rolled type — which also fetches the `administrativeBody` the chair label needs.
- **Tests**: 13 new unit tests for `getRoleText`/`getRoleLabelAt`, including the Bugbot regression (an `isHead` admin-body chair must never label as Δήμαρχος) and date filtering. `gate.test.ts` now asserts the real `{released, dateTime}` shape.
- **Tool descriptions** state that role labels resolve as of the meeting date, and spell out the ABSTAIN=ΛΕΥΚΟ vs DID_NOT_VOTE=ΑΠΟΧΗ trap.
- Trailing whitespace in `gate.ts` fixed.

Deferred, deliberately: the per-row label repetition in `get_subject_transcript` (the `speakers` header map is the right call but changes the wire shape consumers are being written against — follow-up), the `PersonBadge` no-date bug, and the `getRoleTypePriority` alternate-member TODO.

src/lib/mcp/data.ts
christosporios: Taken — the label is now `getRoleLabelAt` in `roles.ts`: `getPrimaryRole` over roles active at the meeting date, rendered by the extracted `getRoleText`. Both of your mechanisms are fixed: party heads label as «party (Επικεφαλής)» (priority 3 selects the party role, and `getRoleText` keeps the headship), and null-name chairs label as «Πρόεδρος - body» now that the selects fetch `administrativeBody` via `roleWithRelationsInclude`. The as-of-date line is in the tool descriptions. `name_en` I left site-consistent (the badges don't fall back to it either); the Λουτσέτης priority TODO stays in `getRoleTypePriority`'s court.

src/lib/mcp/data.ts
christosporios: Taken as suggested — `calculateVoteResult(subject.votes)`, null when no votes. The ΛΕΥΚΟ/ΑΠΟΧΗ trap is now spelled out in the get_subject tool description.

src/lib/mcp/data.ts
christosporios: Deferred, deliberately: the `speakers` header + `personId` shape is the right end state, but it changes the wire shape agents are currently being prompted against, so I want to do it as its own PR with the consumer-side updates. The +22.8% measurement makes the case well.

src/lib/mcp/data.ts
christosporios: Taken — both inline includes are `roleWithRelationsInclude` now, `roleLabelAt`/`RoleForLabel` are gone, and the shared helpers take `RoleWithRelations`.

src/lib/mcp/gate.ts
christosporios: Fixed.

kouloumos (approved):
Re-reviewed `3a9d6b3b`. I re-ran the measurements rather than taking the changes on trust.

**Verified:**

- **Headship omission is 0**, on the same 2,035-segment corpus that had 254 rows (12.5%). Live: party leaders read `X (Επικεφαλής)`, chairs `Πρόεδρος - <body>`, Δούκας `Δήμαρχος`.
- **`getRoleText` came across byte-identical** — only `t`'s type narrowed. The badges render what they rendered before, which is what makes this an extraction rather than a rewrite.
- **`voteSummary` agrees with the site**: `{forCount:30, againstCount:8, abstainCount:1, totalVotes:39, passed:true}` on the subject I'd measured at 35-vs-32.
- **`filterActiveRoles`'s optional `at`** leaves every existing caller on today's date, so no badge regression.
- 75 tests pass. Nothing orphaned — `isMayorRole`, `getSpeakerDisplayInfo` and `simplifyRoleName` keep their other call sites. The memoisation in `mcpGetTranscript` survived.

`getRoleText` + `getRoleLabelAt` is the right split: one formats a role, the other picks which role and formats it. `RoleDisplay` needs the first because it renders a badge per role. No notes there.

Approving. Everything below is follow-up — **none of it blocks this PR.**

### 1. `getRoleNameForPerson` is now a fossil

It does selection *and* formatting, and both halves are weaker than the pair that replaced them. No translator, and it never looks at party roles, so it cannot produce `Δήμαρχος` or `(Επικεφαλής)`. Same person, same meeting, today:

```
MCP        getRoleLabelAt       -> "Δήμαρχος"
summarize  getRoleNameForPerson -> speakerRole "Δημοτικό Συμβούλιο", speakerParty "Αθήνα Τώρα"
```

`src/lib/db/utils.ts:62` and `processAgendaInternal.ts:63` still label through it, so the summarize model is still being told the mayor is a council member — the case your first paragraph opens with. Fixed on MCP, still live one pipeline over.

The consolidation is mechanical:

- `getRoleLabelAt` absorbs the three `getRoleNameForPerson` call sites; delete it
- `PersonBadge.tsx:37` passes the date — at which point `RoleDisplay`'s compact path *is* `getRoleLabelAt`

### 2. Transcripts don't know which body is meeting — pre-existing, both surfaces

18.3% of transcript rows (372 of 2,035) omit a named office the speaker holds in the body that is actually meeting: chairs, vice-chairs, secretaries.

The site has the same gap and always has. Αλόγλου is Πρόεδρος of the Δημοτική Επιτροπή in `chania/jul20_2026`, and his badge reads `Αντιδήμαρχος Οικονομικών & Ψηφιακής Διακυβέρνησης` while the speech beneath it is him opening that committee's first sitting. `Πρόεδρος` and `Αντιπρόεδρος` appear nowhere on the page. `RoleDisplay`'s compact path picks by global priority and has no notion of the meeting's body.

Two causes:

- `getRoleTypePriority` scores a named admin-body role **5**, below party membership at 4 — so the vice-chair and the secretary of the sitting rank beneath "member of a party". That's most of the 372.
- City-level roles (mayor 0, deputy mayor 1) outrank a body chairmanship at 2. Βακαλόπουλος chairs the Sparta committee and reads `Δήμαρχος`.

The first is a re-ranking. The second needs a design call — should a committee transcript say `Δήμαρχος`, `Πρόεδρος - Δημοτική Επιτροπή`, or both? Worth its own issue.

Nothing to do here, but it's the payoff from this extraction: both surfaces now share `getRoleText`, so this gets fixed once instead of twice.

**Nit:** `getRoleTranslations` pins `locale: 'el'`. Better than the hardcoded literal, but still a hardcoded choice, and `currentRealm()` is in scope in `gate.ts`. One line whenever `.rs` stops being a demo.

kouloumos:
Correction to my review above: I called `getRoleNameForPerson` "the fossil version… both halves are weaker". That is wrong, and it would misdirect whoever picks the follow-up up.

It is *better* than `getRoleLabelAt` at body offices — Πρόεδρος and Αντιπρόεδρος come through correctly, because it scopes to the administrative body. Those are exactly the rows `getRoleLabelAt` gets wrong. It is worse only at city-level roles, because it applies `administrativeBodyId` as a filter, which removes city roles (they have `administrativeBodyId: null`) before they can be considered. Measured on the summarize payload: 29.8% of speaker-segments drop a city-level office, and the mayor of Athens arrives as an empty string.

So the two fail in opposite directions and the consolidation is not a swap. It is folded into #604, which now covers all four surfaces.

PR #601, opened 2026-08-10
feat(mcp): meeting context on subject records + list_nearby_subjects

christosporios:
Stacked on #600 (which stacks on #599); base retargets as the stack merges.

Two data improvements to the public MCP server, both born from agent testing:

## Meeting context denormalized onto subject records

`get_subject`, `search` results and `list_hot_subjects` now carry `meetingName`, `meetingDate` and `administrativeBody` (the body that met, e.g. «Δημοτική Επιτροπή»; `null` when the record names none). Previously only `get_meeting` had the body — an agent needed a second call just to name it, and in testing labeled the wrong body in the meantime.

Zero extra queries: `requireVisibleMeeting` widens the select of the lookup it already does (returning `name` + `administrativeBody` alongside `released`/`dateTime`), search hits already hydrate `councilMeeting.administrativeBody`, and hot-subject rows already carried `meetingName`/`bodyName` — they were just dropped in the mapping.

## `list_nearby_subjects`

New tool: given `lat`/`lng` (+ `radiusMeters`, default 1000), resolves the municipality by realm-scoped point-in-polygon (`getCityAtPoint`) and returns its recent subjects the way the embed widget does: subjects pinned within the radius first — each with `distanceMeters` — then recent municipality-wide (no-location) subjects filling remaining slots, both groups ranked by the standard recency/discussion blend (`sortByRanking`). Outside coverage returns a friendly note instead of an error.

Implementation: the embed widget's geo ranking in `src/lib/hotSubjects.ts` is extracted into a shared `rankSubjectsNearPoint`, with a new exported `getHotSubjectsNearPoint` (caller-chosen center/radius, cached meetings query); the geohash variant is now a thin wrapper over it — behavior unchanged.

One gotcha fixed along the way: `unstable_cache` revives cached meetings with **string** dates, so the tool re-wraps `dateTime` before formatting — the first (cache-miss) call worked and every subsequent one 500'd until this.

## Testing

- `npx tsc --noEmit` and the MCP + ranking suites (36 tests) pass.
- End-to-end on the dev server: `get_subject` and `list_hot_subjects` carry the new fields (verified against Ζωγράφου/Χαλάνδρι records); `list_nearby_subjects` from Syntagma returns Athens subjects with sane distances (Σταδίου 54 at 199m with the 1km default; the 5km radius surfaces the Ευαγγελισμός metro subject at 1893m), repeat calls exercise the cache-hit path, and an open-sea point returns the not-covered note.
- The `search` mapping change is type-checked against `SearchResultLight` but not exercised live: local Elasticsearch is failing pre-existing (the site's own `/api/search` 500s on this branch's parent too). Worth one search call on the preview deployment.

---

> [!NOTE]
> **Medium Risk**
> New public MCP geo API and stricter location DB error propagation affect embed/MCP behavior on failures; subject response shape changes are additive but client-visible.
> 
> **Overview**
> Adds **`list_nearby_subjects`** to the public MCP server (lat/lng + radius): resolves only **listed** municipalities via **`getListedCityAtPoint`**, ranks recent subjects like the embed widget (in-radius first, then municipality-wide), attaches **`distanceMeters`** through shared **`withDistances`**, and returns **`meetingsScanned` / `oldestMeetingScanned`** so empty results are scoped in time. Uncovered points get a note, including a **lat/lng swap** hint when the transposed point would match.
> 
> **Meeting context** is denormalized onto subject-facing MCP payloads: **`get_subject`**, **`search`**, and **`list_hot_subjects`** now share a **`subjectSummary`** shape with **`meetingName`**, **`meetingDate`**, and **`administrativeBody`**, fed by widening **`requireVisibleMeeting`** (no extra round-trip).
> 
> Geo stack changes: point ranking moves into **`rankSubjectsNearPoint`** / **`getHotSubjectsNearPoint`**; the embed preflight API uses **`withDistances`** and **catches** distance failures. **`filterLocationIdsWithinRadius`** and **`getLocationDistancesFromPoint`** **propagate** DB errors instead of returning empty results; widgets/MCP geohash paths **catch** where degradation is acceptable. **`isoDate`** accepts cache-revived **string** dates (fixes repeat-call 500s).
> 
> The **/mcp** page adds a fourth example prompt (neighbourhood) in all locales. PostHog MCP events gain **`$host`** for dashboard filtering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e130835a0a6e2d088c33a4bdfb91a42d801b9a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds realm-scoped nearby-subject discovery and enriches MCP subject records with meeting context.
- Extracts shared point-based subject ranking and distance mapping.
- Adds the `list_nearby_subjects` MCP tool with municipality coverage detection.
- Adds meeting name, date, and administrative body to subject-facing MCP responses.
- Updates embed fallback behavior, analytics host attribution, tests, instructions, and localized examples.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/mcp/data.ts | Adds the nearby-subject response flow and consolidates subject-list meeting metadata while preserving realm and visibility boundaries. |
| src/lib/mcp/server.ts | Registers and validates the new nearby-subject MCP tool. |
| src/lib/hotSubjects.ts | Extracts point-based ranking, preserves near-before-wide ordering, and centralizes distance attachment. |
| src/lib/db/location.ts | Changes proximity-query failures to propagate while retaining swapped-coordinate correction. |
| src/lib/db/cities.ts | Adds a realm-scoped point lookup restricted to listed municipalities. |
| src/lib/mcp/gate.ts | Widens the visibility-gate projection to return meeting context without weakening access checks. |
| src/app/api/embed/subjects/route.ts | Uses shared distance mapping and deliberately degrades to null distances if measurement fails. |

<sub>Reviews (9): Last reviewed commit: ["feat(mcp): meeting context on subjects; ..."](https://github.com/schemalabz/opencouncil/commit/0198922e98093cedb83859a41ea34e810685a044) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51774223)</sub>

**Context used:**

- Knowledge Base — [MCP assistant integration](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/mcp-assistant.md)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/6027033c3dcabfa67799eba67397b5c979d4bf66..25c7d278dd0f1db9b9c100502cc8becd88a807aa) to rebase onto the updated base branch (#600's PostHog-scrub fix) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/25c7d278dd0f1db9b9c100502cc8becd88a807aa..99c28de754046468620bd845d5c88645cc86b7f7) to rebase onto the updated base (#600's intent-scrub fix) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/99c28de754046468620bd845d5c88645cc86b7f7..86d2bd14b02b484f927d6c39dba322e5489270cf) to rebase onto the updated base (#600's error-path scrub) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/86d2bd14b02b484f927d6c39dba322e5489270cf..e657c21dda09ec09b221d1b20d16d0f7def545d1) to rebase onto the updated base (#600 now stamps `$host` on analytics events) — no changes to this PR's own diff.

christosporios:
Heads-up: #600 merged just before its final force-push, so this PR now carries that one small commit at its base — 98beab58, the `$host` stamping on analytics events (keeps dev/preview MCP traffic out of filtered PostHog dashboards, [context](https://github.com/schemalabz/opencouncil/pull/600#issuecomment-5243211856)). It lands with this PR; the subject-context/nearby diff is unchanged.

src/lib/mcp/data.ts
```diff
@@ -551,6 +562,61 @@ export async function mcpListHotSubjects(args: {
...
+                meetingId: meeting.id,
+                // Re-wrap: on unstable_cache hits the meeting arrives
+                // JSON-revived, with dateTime as a string rather than a Date.
+                meetingDate: isoDate(new Date(meeting.dateTime)),
+                meetingName: meeting.name,
+                administrativeBody: meeting.administrativeBody?.name ?? null,
+                topic: subject.topic?.name ?? null,
+                // Meters from the queried point; null for municipality-wide
+                // subjects, which have no pinned location.
+                distanceMeters: distance != null ? Math.round(distance) : null,
```
kouloumos: `getLocationDistancesFromPoint` already returns `Math.round(r.meters)`, so this round is a no-op.

src/lib/mcp/server.ts
```diff
@@ -101,6 +102,28 @@ export function registerOpenCouncilServer(server: McpServer) {
...
+        'list_nearby_subjects',
+        {
+            title: 'List subjects near a location',
+            description:
+                'Recent council subjects around a geographic point — "what has the council decided ' +
+                'about this neighborhood". Resolves the municipality containing the point, then ' +
+                'returns subjects pinned within the radius first (with their distance in meters), ' +
+                'followed by recent municipality-wide subjects that have no pinned location; each ' +
+                'group is ranked by a recency/discussion blend, like the site\'s own nearby widgets. ' +
+                'Draws from the municipality\'s recent meetings rather than the full archive.',
```
kouloumos: "Recent meetings" reads as a duration, but `HOT_MEETING_WINDOW = 8` makes it a count. For Athens, the meetings behind a 10 km / limit 50 call spanned 20–29 July — a 9-day window, with the oldest 21 days back. For a council that meets monthly the same constant is eight months. An agent can't tell which regime it's in, and there's no location filter on `search` to fall back to, so an empty result has nowhere to go and gets reported as "nothing has been decided here".

Returning the bound alongside the results — `meetingsScanned`, or just the oldest meeting date scanned — lets the agent say "nothing since 20 July" instead of "nothing". That's the difference between a true statement and a false one, and it's cheaper than widening the window.

src/lib/mcp/data.ts
```diff
@@ -551,6 +562,61 @@ export async function mcpListHotSubjects(args: {
...
+ * location) subjects filling remaining slots, each group ordered by the
+ * standard recency/discussion blend.
+ */
+export async function mcpListNearbySubjects(args: {
+    lat: number;
+    lng: number;
+    radiusMeters: number;
+    limit: number;
+}) {
+    const city = await getCityAtPoint(currentRealm(), args.lng, args.lat);
```
kouloumos: This is the map-highlight helper rather than the coverage one. Its JSDoc says so directly — *"intentionally NOT restricted to listed cities (out-of-network δήμοι must resolve too — unlike `getCityIdContainingPoint`)"* — and `getCityIdContainingPoint` (`status = 'listed'`) is what the embed path this tool mirrors uses, via `getCityIdForGeohashCached`.

The effect is that the not-covered note is effectively unreachable. We carry boundaries for far more municipalities than we cover — seeded so the map can highlight any δήμος, including out-of-network ones — so any point on land falls inside *some* polygon and this returns non-null. I probed 18 points around the country: all 18 resolved to a municipality, and the only input that produced the note was one in the middle of the Aegean.

Of 15 non-listed municipalities, six returned real subjects and nine returned a city name with an empty list. Πειραιάς comes back as `{cityId: "piraeus", cityName: "Πειραιάς", subjects: []}`, so an agent reports that we cover Πειραιάς and there's nothing recent — when we don't cover it at all.

Not a disclosure issue: those subjects are already reachable before this PR via `get_subject` and `search` with `cityIds`. It's the claim of coverage that's wrong, not the data.

Neither existing helper is a drop-in — `getCityIdContainingPoint` is listed-only but not realm-scoped, this one is the reverse — so it probably wants a realm-scoped listed-only variant over the `cityCoversPoint` fragment already shared between them. Whatever it becomes can also skip the geometry: this call simplifies and parses a full municipal polygon every time, and only `id` and `name` are read.

src/lib/mcp/data.ts
```diff
@@ -294,6 +297,12 @@ export async function mcpGetSubject(subjectId: string, identity: McpIdentity) {
         description: subject.description,
         cityId: subject.cityId,
         meetingId: subject.councilMeetingId,
+        // Which body met, and when — denormalized so consumers never have to
+        // fetch the meeting just to say «Δημοτική Επιτροπή» instead of
+        // guessing (and getting) the wrong body.
+        meetingName: meeting.name,
```
kouloumos, line 341: Adding the fields here is right, but it's worth stepping back one level, because it isn't only these three fields.

`list_hot_subjects` (`:543`), `list_nearby_subjects` (`:596`) and `search` (`:668`) all emit the *same* subject-summary record — `id`, `title`, `snippet: truncate(description, 300)`, `meetingId`, the three meeting fields, `topic`, `url: urls.subject(…)` — each differing only by a single extra field (`discussionSeconds`, `distanceMeters`, `score`). Three tools hand-build one wire contract from three different upstream shapes, and they've already drifted:

- **nullability** — hot-subjects has all three meeting fields as `string | null`; the other two are non-null
- **date** — `row.meetingDate?.slice(0, 10)` in one, `isoDate(…)` in the others
- **city** — hot-subjects and search carry `cityId`/`cityName` per subject; nearby hoists them to the envelope

That divergence is the argument for fixing the shape rather than tidying it: an agent parsing "a subject" gets three subtly different objects depending on which tool produced it. One `subjectSummary()` builder here, spread by the three list tools with their extra field appended, defines it once. This record (`get_subject`) stays out of it — it's the detail shape, legitimately different.

Where I *wouldn't* go higher: normalizing the three upstreams. `GeneralSubjectRow` is flat with ISO strings and `bodyName`/`topicName` because it was shaped for the landing map, `HotSubject` is nested Prisma, `SearchResultLight` is Elasticsearch-hydrated — they serve different consumers and it's fine that they differ. This layer is the adapter; it just shouldn't be three adapters. (The one thing that does smell upstream is `list_hot_subjects` borrowing a view model built for the map.)

Fair to say this is bigger than the PR that surfaced it — this is just the third instance, which is what turns it into a pattern. Equally happy to see it filed and this landed with the duplication acknowledged.

src/lib/mcp/data.ts
```diff
@@ -551,6 +562,61 @@ export async function mcpListHotSubjects(args: {
...
+        return {
+            cityId: null,
+            note: 'This point is not inside any municipality covered by OpenCouncil.',
+            subjects: [],
+        };
+    }
+
+    const center: [number, number] = [args.lng, args.lat];
+    const top = await getHotSubjectsNearPoint(city.id, center, args.radiusMeters, { limit: args.limit });
+    const distances = await getLocationDistancesFromPoint(
```
kouloumos: Two things about these two calls.

**They fail silently.** `filterLocationIdsWithinRadius` and `getLocationDistancesFromPoint` both catch and return empty. The first collapses the result to municipality-wide subjects only; the second nulls every `distanceMeters` — which the comment just below defines as "municipality-wide subjects, which have no pinned location", so an agent will positively assert that a subject pinned 80 m away isn't tied to an address. They hit the same table with the same predicate, so they tend to fail together, and the combined output is indistinguishable from a genuinely empty neighbourhood. That swallow is right for the widget, where a degraded render beats a crash and the UI claims nothing; it's wrong where the emptiness *is* the claim. Either let them propagate and lean on `run()`, or add variants that signal failure so the tool can say "proximity ranking unavailable".

**And the annotation is a copy.** These six lines are the same as `src/app/api/embed/subjects/route.ts:63-78` — collect non-null `locationId`s off the ranked list, call `getLocationDistancesFromPoint`, look each back up, emit `distanceMeters | null`. Two public surfaces publishing a field of the same name by copy-paste. A `withDistances(top, center)` in `hotSubjects.ts` (where `HotSubject` lives, and which the embed route already imports from) would cover both — and would fix the route's version, which does `(subject.locationId && distances.get(id)) ?? null` and so swallows a distance of exactly 0. Yours is the more correct guard.

src/lib/mcp/data.ts
```diff
@@ -551,6 +562,61 @@ export async function mcpListHotSubjects(args: {
...
+        subjects: top.map(({ subject, meeting }) => {
+            const distance = subject.locationId != null ? distances.get(subject.locationId) : undefined;
+            return {
+                id: subject.id,
+                title: subject.name,
+                snippet: truncate(subject.description, 300),
+                meetingId: meeting.id,
+                // Re-wrap: on unstable_cache hits the meeting arrives
+                // JSON-revived, with dateTime as a string rather than a Date.
+                meetingDate: isoDate(new Date(meeting.dateTime)),
```
kouloumos: The wrap is right, but this is the narrowest place to put it. `isoDate` is typed `(date: Date)` at `:71` and is also called at `:304` and `:676`, so the next `isoDate(somethingCached.dateTime)` fails the same way at runtime. Widening it to `(date: Date | string)` with the `new Date()` inside fixes all of them and lets this line go back to `isoDate(meeting.dateTime)` — keep the explanation, just move it to the definition.

For context, we already carry four independent defences against the same `unstable_cache` Date→string lie, none discoverable from the `Date` type: a re-wrap in the feed route, a `typeof === 'string'` branch inside `formatDate` (under a `date: Date` signature), `RankableSubject.meetingDate: Date | string`, and now this. Widening `isoDate` is the cheap version; typing `createCache` to return a serialized shape is the real fix, and clearly a separate PR.

src/lib/mcp/data.ts
kouloumos: `getLocationDistancesFromPoint` already returns `Math.round(r.meters)`, so this round is a no-op.

src/lib/mcp/server.ts
```diff
@@ -101,6 +102,28 @@ export function registerOpenCouncilServer(server: McpServer) {
         args => run(() => mcpListHotSubjects(args))
     );
 
+    server.registerTool(
+        'list_nearby_subjects',
+        {
+            title: 'List subjects near a location',
+            description:
+                'Recent council subjects around a geographic point — "what has the council decided ' +
```
kouloumos: Two things about the description, which is the prompt the model actually reads.

"what has the council decided" overstates the payload — these are agenda items with `title`, `snippet`, `topic` and meeting context; no decision, no status, no votes. `get_subject`'s own description is careful to list "decision, votes" as things *it* returns. "discussed" would be accurate, with a pointer to `get_subject` for the decision itself.

And `distanceMeters: null` is explained only in a code comment, which the model never sees. Worth a clause naming the field and saying the nulls are municipality-wide filler rather than results near the point — otherwise nothing stops an agent answering "5 things happened near you" when 4 of them are city-wide.

src/lib/mcp/server.ts
kouloumos: "Recent meetings" reads as a duration, but `HOT_MEETING_WINDOW = 8` makes it a count. For Athens, the meetings behind a 10 km / limit 50 call spanned 20–29 July — a 9-day window, with the oldest 21 days back. For a council that meets monthly the same constant is eight months. An agent can't tell which regime it's in, and there's no location filter on `search` to fall back to, so an empty result has nowhere to go and gets reported as "nothing has been decided here".

Returning the bound alongside the results — `meetingsScanned`, or just the oldest meeting date scanned — lets the agent say "nothing since 20 July" instead of "nothing". That's the difference between a true statement and a false one, and it's cheaper than widening the window.

src/lib/mcp/server.ts
```diff
@@ -101,6 +102,28 @@ export function registerOpenCouncilServer(server: McpServer) {
...
+                'about this neighborhood". Resolves the municipality containing the point, then ' +
+                'returns subjects pinned within the radius first (with their distance in meters), ' +
+                'followed by recent municipality-wide subjects that have no pinned location; each ' +
+                'group is ranked by a recency/discussion blend, like the site\'s own nearby widgets. ' +
+                'Draws from the municipality\'s recent meetings rather than the full archive.',
+            inputSchema: z.object({
+                lat: z.number().min(-90).max(90).describe('Latitude (WGS84)'),
+                lng: z.number().min(-180).max(180).describe('Longitude (WGS84)'),
+                radiusMeters: z.number().int().min(50).max(10000).default(1000)
```
kouloumos, line 124: Default radius is 1000 while the widget this description invokes is fixed at `GEO_RADIUS_METERS = 500`. Fine to differ — an agent gets an address rather than a map pin — it's just not stated anywhere.

Separately, the bounds accept transposed arguments: `{lat: 23.7348, lng: 37.9755}` is Athens with lat/lng swapped, and it parses cleanly, matches no polygon, and comes back as "not covered by OpenCouncil". Given `location.ts` already carries a swapped-coordinate repair heuristic for stored points, this seems worth catching — when nothing matches, testing the transposed point and saying so in the note turns a confidently wrong answer into a self-correcting one.

src/lib/hotSubjects.ts
```diff
@@ -48,33 +48,58 @@ export async function getRecentHotSubjects(
...
+    const meetings = await getCouncilMeetingsForCity(cityId, {
+        includeUnreleased: false, limit: HOT_MEETING_WINDOW, administrativeBodyTypes, administrativeBodyIds, timeFilter: 'past',
+    });
+    return rankSubjectsNearPoint(meetings, decodeGeohashToCenter(geohash), GEO_RADIUS_METERS, limit);
+}
+
+/**
+ * Hot subjects near an arbitrary point — the same recent window and
+ * near-first/municipality-wide ordering as {@link getHotSubjectsNearGeohash},
+ * but with a caller-chosen center and radius. Uncached (the coordinate space
```
kouloumos: "Uncached (the coordinate space is unbounded)" and then "the underlying meetings query is the cached public one" are two different subjects in one sentence — a reader scanning for cache behaviour gets the wrong answer from the first clause. Something like "The result is not cached (unbounded coordinate space), but the meetings come from the cached public query — unlike the geohash variant, which must use the uncached one because it runs inside `createCache`" explains both functions at once.

Also `administrativeBodyTypes` / `administrativeBodyIds` at `:95` are accepted and forwarded but dead — `mcpListNearbySubjects` passes only `{ limit }` and the tool exposes no body filter. Fine if it's for symmetry with the two siblings; flagging so it's deliberate.

kouloumos (commented):
Reviewed at `86d2bd14`. I rebased onto current main locally and the resulting tree is byte-identical, so all of this applies unchanged to `e657c21d`.

**On the conflict.** It's an add/add on the analytics module. Main has `a13776b1`, this branch carries `98beab58` — same parent (`3a9d6b3b`), same message, different content (the seven-line `$host` stamping). Since that file doesn't exist at the merge base, git sees both sides adding it, which is why it can't auto-merge. `currentBaseUrl` is already on main, so the delta stands alone:

```bash
git diff a13776b1 98beab58 > /tmp/host-stamp.patch   # capture it before the rebase drops it
git rebase --onto upstream/main 98beab58             # drops the duplicated commit
git apply /tmp/host-stamp.patch && git commit -am "fix(mcp): stamp \$host on analytics events"
```

I dry-ran that: the rebase is clean, the patch applies, and this PR's own diff comes out unchanged. Resolving by hand works too, but leaves a second commit on main with the same message as `a13776b1`.

**Closing your open item.** You noted the `search` mapping was type-checked but never run, and that it was worth one search call on the preview. I ran it: `search` comes back with `administrativeBody` populated (`"1η Δημοτική Κοινότητα"`), so that path works.

**On "behavior unchanged" for the geohash refactor.** I called the embed subjects endpoint with the same geohash on this preview and on #600's and diffed the responses: identical down to the byte — same subjects, same order, same distances. The extraction is a no-op, and the geo cache version doesn't need the bump it asks for when ranking or selection changes.

The rest is one theme; specifics are inline. `list_nearby_subjects` returns the same shape for four different states — genuinely nothing nearby, a municipality we don't cover, a failed proximity query, and something that happened just outside the meeting window. All four come back as an empty or municipality-wide-only list, and an agent will read every one of them aloud as "the council hasn't decided anything near you." The tool has no way to say "I don't know."

Three of those four are worth closing before merge: the point-to-municipality lookup resolves δήμοι we don't publish, so the not-covered note is effectively unreachable and we claim coverage we don't have; the recency window is a count of meetings rather than a span of time, which for a busy council is far narrower than "recent" suggests; and the two proximity queries return empty on failure rather than erroring, so a transient database problem is indistinguishable from a quiet neighbourhood. The fourth — a genuinely empty result — is the only one the current shape describes honestly.

Two things with no line to sit on: the tool is missing from the server instructions clients receive on connect, which name every other discovery entry point including its closest sibling, and from the /mcp page copy in all four locales — where "what has the council decided about my neighborhood" is the most legible thing this PR adds.

**Tests.** Nothing currently covers the hot-subjects ranking module, so the near/wide ordering that ships in the live widget has never been executed by a test. The extraction holds today, and the A/B above confirms it, but nothing would catch it if it stopped holding. Two worth adding, both plain unit tests with the radius filter mocked: `rankSubjectsNearPoint`'s ordering — including the non-obvious part, that a subject whose location falls *outside* the radius lands in neither group and is dropped entirely — and that ranking survives cache-revived string dates, pinning the exact trap behind the 500 you hit.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/e657c21dda09ec09b221d1b20d16d0f7def545d1..5e13083560d3a86b3feaf87768cf0a1eeb2e2be6) taking the whole review — the rebase recipe first, then every finding. Thanks @kouloumos, the measured probes (18 points, the byte-identical A/B, the Πειραιάς case) made these unambiguous to act on.

**Structure**: rebased onto main exactly as prescribed — the duplicated commit is gone and the `$host` delta is its own `fix(mcp)` commit at the base.

**The "can't say I don't know" theme, all four states:**
- *Coverage*: new `getListedCityAtPoint` (realm-scoped, `status='listed'`, selects only `id`/`name` — no geometry parse) replaces the map helper. Πειραιάς-style points now get the honest not-covered note.
- *Silent failures*: `filterLocationIdsWithinRadius` and `getLocationDistancesFromPoint` now **throw**; the MCP path propagates through `run()`, and the embed widget catches at its own boundary (degraded render, as before). The distance annotation is the shared `withDistances()` in `hotSubjects.ts`, used by both surfaces — which also fixes the embed route's `?? null` swallowing a distance of exactly 0.
- *Window honesty*: the response carries `meetingsScanned` + `oldestMeetingScanned`, and the tool description instructs "report an empty list as 'nothing since {oldestMeetingScanned}', never 'nothing ever'".
- *Transposed coordinates*: an unmatched point probes the swapped pair — `{lat: 23.73, lng: 37.98}` now answers "…swapping lat and lng lands in Αθήνα, so the coordinates may be transposed."

**Smaller points, all taken**: shared `subjectSummary()` builder — the three list tools now emit one record (same fields, same nullability, same date format, per-subject `cityId`/`cityName` everywhere) plus their single extra field; `isoDate` widened to `Date | string` with the explanation at the definition; description rewritten ("discussed" with a `get_subject` pointer for decisions/votes, `distanceMeters` null semantics spelled out); the no-op `Math.round` is gone; the dead `administrativeBody*` params dropped from `getHotSubjectsNearPoint`; its JSDoc's cache sentence split as suggested; the tool is now in the server instructions and the /mcp page examples in all four locales («Τι έχει συζητήσει ο δήμος για τη γειτονιά μου;»).

**Tests**: new `src/lib/__tests__/hotSubjects.test.ts` — near-first/wide-fill ordering including the outside-radius-dropped case, ranking over cache-revived string dates (pinning the 500 you referenced), failure propagation for both queries, and the 0-distance guard.

Verified end-to-end on the dev server: Athens center returns `meetingsScanned: 3, oldestMeetingScanned: 2026-01-14` with distances; swapped coords self-correct; open sea gets the note; the embed endpoint is unchanged for the geohash path and still degrades gracefully. `tsc` + all suites green; #603 rebased on top.

src/lib/mcp/data.ts
christosporios: Taken — new `getListedCityAtPoint` in cities.ts: realm-scoped AND `status = 'listed', over the shared `cityCoversPoint` fragment, selecting only `id`/`name` (no geometry simplify/parse), throwing on query failure. Πειραιάς-style points now get the honest not-covered note.

src/lib/mcp/data.ts
christosporios: Taken in this PR — `subjectSummary()` now defines the record once and all three list tools spread it plus their single extra field. Nullability, date format and per-subject `cityId`/`cityName` are identical across the three; the upstreams stay as they are, per your note. get_subject remains the detail shape.

src/lib/mcp/data.ts
christosporios: Taken, both halves — the two queries now throw (the swallow moved to the callers that want it: the embed widget catches at its own boundary and degrades, the geohash compute falls back to municipality-wide-only). The annotation is the shared `withDistances()` in hotSubjects.ts, used by both surfaces, which also fixes the embed route's zero-distance swallow.

src/lib/mcp/data.ts
christosporios: Taken — `isoDate` is now `(date: Date | string)` with the unstable_cache explanation at the definition; the call site went back to plain `isoDate(meeting.dateTime)`. Agreed that typing createCache to return the serialized shape is the real fix, and a separate PR.

src/lib/mcp/data.ts
christosporios: Gone — the mapping now uses `withDistances()`, which passes the already-rounded values through.

src/lib/mcp/server.ts
christosporios: Taken — "discussed", with an explicit pointer to get_subject for the decision and votes, and the description now spells out that `distanceMeters: null` means municipality-wide filler ("count only non-null distances when saying how much happened nearby").

src/lib/mcp/server.ts
christosporios: Taken — the response now carries `meetingsScanned` and `oldestMeetingScanned`, and the description instructs reporting an empty list as "nothing since {oldestMeetingScanned}", never "nothing ever".

src/lib/mcp/server.ts
christosporios: Transposed-argument probe added: when no municipality matches, the swapped pair is checked and the note says so ("swapping lat and lng lands in Αθήνα, so the coordinates may be transposed") — verified live. The 1000m default stays (an address deserves a wider net than a map pin) and is visible in the schema default.

src/lib/hotSubjects.ts
christosporios: Both taken — the JSDoc now separates the two subjects ("The result itself is not cached… but the meetings come from the cached public query — unlike the geohash variant, which must use the uncached one because it runs inside createCache"), and the dead `administrativeBody*` params are gone; the function takes a plain `limit`.

kouloumos (approved):
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/5e130835a0a6e2d088c33a4bdfb91a42d801b9a5..0198922e98093cedb83859a41ea34e810685a044) a fixup to the ranking test — nothing else touched. Re-reviewed the rest at `5e13083` and it all checks out: Πειραιάς and Θεσσαλονίκη now return the honest not-covered note, transposed coords self-correct, `meetingsScanned: 8` / `oldestMeetingScanned: 2026-07-16` come through, and the three list tools emit an identical key set apart from their one extra field. The embed endpoint is byte-identical to base for the geohash path, so the `location.ts` throw change and `withDistances` extraction didn't move the widget.

**The test fix.** `ranks newer meetings first within each group and survives cache-revived string dates` couldn't fail. The fixture is fed newest-first (matching `dateTime desc`), all four subjects carry `_count.contributions: 1`, and `Array.prototype.sort` is stable — so the expected order came out of the input order, not the ranker. Confirmed by setting `adapt`'s `meetingDate` to `null`, making recency completely inert: all five tests still passed.

Split it in two rather than reversing the fixture, because that ordering is load-bearing — `oldestMeetingDate` reads `meetings[length-1]`, which is only the oldest while the input stays newest-first:
- the existing case keeps the fixture and now claims only what it verifies (near-before-wide partition, string dates surviving), with a comment noting it deliberately doesn't discriminate recency
- a new case feeds meetings oldest-first and asserts they come back newest-first

Same mutation now fails exactly that one case and nothing else. 470 tests green.

**One note, not a request:** `getHotSubjectsNearPoint` draws its 8-meeting window across all administrative bodies, so in a δήμος with active δημοτικές κοινότητες those 8 can be entirely community sessions — `meetingsScanned: 8` is then a well-bounded answer over a slice containing no δημοτικό συμβούλιο. That was equally true before this round (the body filters were never actually passed), so nothing regressed; it just interacts with the new window-honesty reporting. Might be worth defaulting to council+committee, or naming the scanned bodies in the envelope.

PR #603, opened 2026-08-10
feat(mcp): advertise highlight tools only on authenticated connections

christosporios:
Top of the stack (#599 → #600 → #601 → this); base retargets as the stack merges.

Anonymous MCP connections previously saw all five highlight tools (`create_highlight`, `generate_highlight_video`, `list_highlights`, `set_highlight_showcase`, `get_highlight`) — every one of which requires an identity end to end, so the only thing an anonymous caller could ever get from them was an auth error. They also cluttered the tool list ChatGPT/Claude users see.

## What changes

- **Anonymous connections** no longer get the highlight suite or the `create_meeting_highlights` prompt advertised; calling one by name returns the standard JSON-RPC "Tool not found". `MCP_INSTRUCTIONS` (static, shared) now tells anonymous users the tools appear once they reconnect with a personal MCP URL, so discoverability is preserved.
- **Authenticated connections** (mcp_ or sk_ token) are unchanged — all 20 tools and 3 prompts.

## How

Serving is stateless — a fresh server instance per request — so registration time *is* per-connection time. The route now:

1. Verifies the bearer token **once** up front (`verifyMcpToken`), instead of leaving it to `withMcpAuth`'s verifier; `withMcpAuth` now receives a closure returning the precomputed result, so there's no duplicate DB lookup or double `lastUsedAt` update.
2. Carries the resolved identity in the per-request `McpRequestContext` (alongside realm/origin), exposed via `currentMcpIdentity()`.
3. `registerOpenCouncilServer` calls the extracted `registerHighlightTools(server)` only when an identity is present. Tool handlers keep reading identity from their `ServerContext` as before.

## Testing

`npx tsc --noEmit` + MCP suite (20 tests) pass. End-to-end against the dev server with a freshly minted token (revoked after):

- anonymous `tools/list`: 15 tools, no highlight tools; `prompts/list`: 2 prompts
- authenticated `tools/list`: all 20 incl. the highlight suite; `prompts/list`: 3
- anonymous `tools/call create_highlight` → `-32602 Tool create_highlight not found`
- authenticated `list_highlights` returns real highlights (URL-token and header-token styles both verified)
- invalid token still 401s (the pre-verified error is re-thrown inside `withMcpAuth`)

---

> [!NOTE]
> **Medium Risk**
> Changes MCP authentication timing and which capabilities are exposed per caller, but behavior is fail-closed for anonymous users and covered by new registration tests.
> 
> **Overview**
> Anonymous MCP clients no longer see the five highlight tools or the `create_meeting_highlights` prompt in `tools/list` and `prompts/list`; authenticated `mcp_` / `sk_` connections still get the full catalog. Anonymous calls to those tool names behave as **not found** instead of surfacing auth errors on tools that were never usable without a token.
> 
> The MCP route **verifies the bearer token once** before handler setup, passes the result into `withMcpAuth` via a closure (avoiding a second DB hit), and stores the resolved identity on per-request `McpRequestContext` via `currentMcpIdentity()`. `registerOpenCouncilServer` only calls `registerHighlightTools` when that identity is present. `identityFromAuthInfo` centralizes unwrapping identity from `AuthInfo` for the route, analytics, and tool context. Static `MCP_INSTRUCTIONS` now tells agents to reconnect with a personal MCP URL if highlight tools are missing.
> 
> `mcp-handler` is pinned to exact **2.1.0**. New registration tests cover anonymous vs user vs service identity and missing request scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 506e37f6e76bbe76a1c122ed7277a79b864a7d54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR makes MCP highlight tools and the highlight-creation prompt conditional on the connection’s resolved identity while preserving anonymous discoverability through static instructions.
- Pre-verifies URL or header bearer tokens and carries the resulting identity in request-local context.
- Extracts highlight registration into an authenticated-only registration path.
- Adds registration tests for anonymous, user-token, and service-key identities.
- Pins `mcp-handler` to version 2.1.0.

The PR appears safe to merge because no eligible blocking failure remains in the provided follow-up-review scope.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/app/api/mcp/[[...token]]/route.ts | Pre-verifies bearer credentials, reuses the authentication result, and seeds request-local identity before invoking the MCP handler. |
| src/lib/mcp/server.ts | Moves highlight tools and their prompt behind a request-local authenticated-identity registration gate. |
| src/lib/mcp/realm-context.ts | Extends MCP request context with the resolved user or service identity. |
| src/lib/mcp/auth.ts | Centralizes extraction of the typed MCP identity from AuthInfo. |
| src/lib/mcp/__tests__/server.test.ts | Tests the intended anonymous and authenticated registration inventories, including service keys and missing request scope. |

<sub>Reviews (8): Last reviewed commit: ["build(deps): pin mcp-handler exactly"](https://github.com/schemalabz/opencouncil/commit/506e37f6e76bbe76a1c122ed7277a79b864a7d54) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51785760)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/b047262d7a2ee3354bd546af0383c47e2d1790ff..b37f33f216587e65d6756efb6f82890bfb22b28b) to rebase onto the updated base (#600's intent-scrub fix, via #601) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/b37f33f216587e65d6756efb6f82890bfb22b28b..e034f511af3b4ec25f2b8c6ea21fae5790f74e83) to rebase onto the updated base (#600's error-path scrub, via #601) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/e034f511af3b4ec25f2b8c6ea21fae5790f74e83..559c2019d224c12826573792dbef3b562dcbeee2) to rebase onto the updated base (#600's `$host` stamping, via #601) — no changes to this PR's own diff.

src/lib/mcp/server.ts
```diff
@@ -301,6 +302,81 @@ export function registerOpenCouncilServer(server: McpServer) {
             )
     );
 
+    // Highlight tools (and their prompt) require authentication end to end,
+    // so anonymous connections don't get them advertised at all —
+    // MCP_INSTRUCTIONS still tells those users how to create a personal URL.
+    // Registration runs per request (a fresh server per POST), which is what
+    // makes per-caller advertisement possible.
+    if (currentMcpIdentity()) {
```
kouloumos: This registers `create_meeting_highlights` above the `--- Prompts ---` block, and the SDK serves `prompts/list` as `Object.entries(this._registeredPrompts)` — insertion order. So authenticated callers go from `weekly_roundup, councillor_profile, create_meeting_highlights` to `create_meeting_highlights, weekly_roundup, councillor_profile`. The set is unchanged, the order isn't, which is a small dent in "authenticated connections are unchanged". Tool order is unaffected either way.

Moving the call below the prompts block fixes it and reads better — right now a call that registers a prompt sits above the header for prompts.

src/app/api/mcp/[[...token]]/route.ts
```diff
@@ -43,34 +42,54 @@ async function handler(
...
         } as RequestInit);
-        return withRealm(() => authedHandler(request));
+    } else if (segments.length !== 0) {
+        return new Response('Not found', { status: 404 });
+    }
+
+    // Verify the token once, up front: the identity steers which tools get
+    // registered (highlight tools are not advertised to anonymous callers),
+    // and withMcpAuth reuses the same result instead of re-verifying.
+    const bearerToken = request.headers.get('authorization')?.match(/^Bearer (.+)$/i)?.[1];
```
Comment (kouloumos) [inline src/app/api/mcp/[[...token]]/route.ts]: `withMcpAuth` still does its own extraction and passes it to the verifier, which the closure below ignores — so there are two parsers and only one is live. They disagree: `Authorization: Bearer  mcp_x` (double space) is 200-anonymous on pr-601 and 401 here, because `(.+)` captures the leading space. I checked both previews.

Failing loudly is arguably better, so not asking you to change the behaviour. The reason to care is the coupling: if a future version short-circuits when *its* extraction yields `undefined`, the closure never runs, `req.auth` is never set, and you'd advertise the full suite while every call fails auth. `^2.1.0` plus active dependabot makes that reachable. Using the library's own rule removes the possibility:

```suggestion
    const [scheme, token] = request.headers.get('authorization')?.split(' ') || [];
    const bearerToken = scheme?.toLowerCase() === 'bearer' ? token : undefined;
```

src/app/api/mcp/[[...token]]/route.ts
```diff
@@ -43,34 +42,54 @@ async function handler(
...
+        baseHandler,
+        async () => {
+            if (authError) throw authError;
+            return authInfo;
+        },
+        { required: false }
+    );
 
-    return new Response('Not found', { status: 404 });
+    const identity = (authInfo?.extra?.identity as McpIdentity | undefined) ?? null;
```
Comment (kouloumos) [inline src/app/api/mcp/[[...token]]/route.ts]: Third hand-written unwrap of `authInfo.extra.identity` with a third `as McpIdentity` — the others are `auth.ts:22` and `analytics.ts:47`, and that one already documents itself as "mirroring identityFromContext". Three makes it a pattern rather than an exception, and they're unchecked casts off an `unknown`.

Worth one primitive in `auth.ts` that all three delegate to:
```ts
export function identityFromAuthInfo(authInfo: AuthInfo | undefined): McpIdentity {
    return (authInfo?.extra?.identity as McpIdentity | undefined) ?? null;
}
```
`identityFromContext` becomes a one-liner over it, `analytics.ts` keeps its `unknown` narrowing but ends in the same call, and this line becomes `identityFromAuthInfo(authInfo)`.

kouloumos (approved):
Verified the split by A/B'ing this preview against pr-601's rather than re-reading the diff: anonymous `tools/list` 19 → 14, `prompts/list` 3 → 2, `tools/call create_highlight` → `-32602`, invalid token 401 on both. I also checked the per-request-factory assumption against the library instead of taking the comment's word for it — both serving legs of `@modelcontextprotocol/server` v2 call `await factory(...)` inside the request chain, so registration genuinely runs inside `mcpRealmStore.run()`. And the five implementations still reject a null identity on their own (`data.ts:798`, `:851`, `:908`, `:939`), so this is advertisement hardening sitting on top of intact enforcement. No blockers.

One incidental finding: the tool counts are environment-dependent. The preview shows 14/19, not 15/20 — the extra one is PostHog's virtual `get_more_tools`, which only exists when `NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN` is set. Doesn't affect anything here, but it's a reason not to write a count-based assertion anywhere.

**There's no automated test for any of this.** `server.ts`, `route.ts` and `realm-context.ts` had zero coverage before and have zero after; the suite passes identically if the `if` is inverted or removed. That matters more than usual because the refactor moved five `registerTool` calls and one `registerPrompt` across a new boundary.

It's cheap — `server.ts`'s SDK import is type-only, so no SDK, no DB, no containers, 0.6s. I wrote it and mutation-tested it:

<details>
<summary><code>src/lib/mcp/__tests__/server.test.ts</code></summary>

```ts
/** @jest-environment node */

// server.ts only closes over the data functions inside tool callbacks — it never
// calls them at registration time — so a blanket stub keeps ../data's Prisma,
// Elasticsearch and next-intl imports out of the test, without a name list that
// goes stale every time a tool is added.
jest.mock('../data', () =>
    new Proxy({ __esModule: true } as Record<string, unknown>, {
        get: (target, prop: string) => (prop in target ? target[prop] : jest.fn()),
    })
);

// auth.ts reaches Prisma (and through it env.mjs, which jest won't transform).
// Same stub gate.test.ts uses; registration never touches the client.
jest.mock('../../db/prisma', () => ({ __esModule: true, default: {} }));

import { Realm } from '@prisma/client';
import type { McpServer } from '@modelcontextprotocol/server';
import { registerOpenCouncilServer } from '../server';
import { mcpRealmStore, requestContext } from '../realm-context';
import type { McpIdentity } from '../auth';

const HIGHLIGHT_TOOLS = [
    'create_highlight', 'generate_highlight_video', 'list_highlights',
    'set_highlight_showcase', 'get_highlight',
];
const USER: McpIdentity = { type: 'user', userId: 'u1' };
const SERVICE: McpIdentity = { type: 'service', keyName: 'bot' };

/** What a connection with this identity would see in tools/list & prompts/list. */
function advertised(identity: McpIdentity, { inRequestScope = true } = {}) {
    const tools: string[] = [];
    const prompts: string[] = [];
    const recorder = {
        registerTool: (name: string) => { tools.push(name); },
        registerPrompt: (name: string) => { prompts.push(name); },
    } as unknown as McpServer;

    const register = () => registerOpenCouncilServer(recorder);
    if (inRequestScope) {
        mcpRealmStore.run(requestContext(Realm.greece, 'opencouncil.gr', identity), register);
    } else {
        register();
    }
    return { tools, prompts };
}

describe('highlight tools are advertised only on authenticated connections', () => {
    it('withholds exactly the highlight suite from anonymous callers', () => {
        const anon = advertised(null);
        const authed = advertised(USER);

        // Asserting the *difference*, not an inventory: adding a public tool
        // needs no test edit, but a tool crossing the boundary in either
        // direction fails here.
        expect(authed.tools.filter((n) => !anon.tools.includes(n))).toEqual(HIGHLIGHT_TOOLS);
        expect(anon.tools).toEqual(authed.tools.filter((n) => !HIGHLIGHT_TOOLS.includes(n)));
        expect(authed.prompts.filter((n) => !anon.prompts.includes(n)))
            .toEqual(['create_meeting_highlights']);
    });

    it('withholds every highlight-named tool, including ones added later', () => {
        expect(advertised(null).tools.filter((n) => n.includes('highlight'))).toEqual([]);
    });

    it('advertises the suite to service keys as well as user tokens', () => {
        // Guards a later `identity?.type === 'user'` narrowing quietly stripping
        // the suite from admin-panel sk_ keys.
        expect(advertised(SERVICE).tools).toEqual(advertised(USER).tools);
        expect(advertised(SERVICE).prompts).toEqual(advertised(USER).prompts);
    });

    it('fails closed when there is no request scope at all', () => {
        expect(advertised(null, { inRequestScope: false }).tools)
            .toEqual(advertised(null).tools);
    });
});
```

</details>

| mutation | result |
|---|---|
| unmodified branch | 4 pass; MCP suite 5 files / 27 tests green |
| gate removed (`registerHighlightTools` called unconditionally) | fails 1, 2 |
| gate narrowed to `?.type === 'user'` | fails 3 |
| new `delete_highlight` added to the **public** block | 1 passes (blind), fails 2 |

The last row is why the name-based check is separate: it's the realistic future mistake, and the difference-based assertion structurally can't see it because the tool shows up in both lists. Assertions are on the *difference* rather than a full inventory so adding a public tool needs no test edit.

I deliberately didn't suggest a route-level integration test. If the per-request-factory assumption ever broke, the outcomes are "anonymous sees the tools and gets `UnauthorizedError`" (i.e. exactly base behaviour) or "authenticated users lose the tools" (loud). Neither is a security hole, so testcontainers to guard a dependency's behaviour isn't proportionate — pinning `mcp-handler` exactly would be the cheaper insurance, given dependabot is active here.

Smaller things, take or leave: `get_subject_transcript`'s description still ends "…are what `create_highlight` needs" (`server.ts:259`), naming a tool anonymous callers can't see — same class as the `MCP_INSTRUCTIONS` fix, just missed. The instructions say "personal MCP URL" but `sk_` service keys unlock the suite too. `requestContext`'s `identity = null` default has exactly one caller that always passes it. And the split left a stray blank at `server.ts:370` and a double blank at `372-373`.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/559c2019d224c12826573792dbef3b562dcbeee2..eca72988bfe5061d0c31bea7be93c8bc02b16a15) to rebase onto the updated base (#601's review-response changes) — no changes to this PR's own diff.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/eca729880ff554da3e689a8faa03b575fe9e842e..1638e000998f1b06825a3ad1652fee79ee16a067) taking the whole review — and thank you for writing the test, @kouloumos; it's in verbatim as `src/lib/mcp/__tests__/server.test.ts`, all four green (and the mutation table is a model for how to justify a test's shape).

- **Prompt order**: the gated registration moved below the prompts block — authenticated callers get `weekly_roundup, councillor_profile, create_meeting_highlights` again (verified live), and the highlight-prompt registration no longer sits above the prompts header.
- **Bearer extraction**: replaced with your split-based rule, so the two parsers can't disagree — `Bearer` + double space is now 200-anonymous here too, matching the library's own extraction (verified live).
- **`identityFromAuthInfo`**: added to `auth.ts` as the one place the `as McpIdentity` cast lives; `identityFromContext`, `analytics.ts` and the route all delegate to it.
- **`mcp-handler` pinned exactly** (own commit) — agreed that's the proportionate insurance for the two withMcpAuth couplings, rather than an integration test guarding a dependency's behavior.
- **The nits**: `get_subject_transcript`'s description now says "highlight creation … (available on authenticated connections)" instead of naming the hidden tool; the instructions mention service keys unlock the suite too; `requestContext`'s `identity` default is gone (the one caller always passes it); stray blanks cleaned by the move.

Good catch on the environment-dependent tool counts (`get_more_tools` riding on the PostHog token) — noted, and the difference-based assertions dodge it by construction.

src/lib/mcp/server.ts
christosporios: Taken — the gated call now sits below the prompts block (with a comment noting it must register last so the base orders are identical for every caller, merely extended). Verified live: authenticated prompts/list is back to `weekly_roundup, councillor_profile, create_meeting_highlights`.

src/app/api/mcp/[[...token]]/route.ts
Comment (christosporios) [reply src/app/api/mcp/[[...token]]/route.ts]: Taken as suggested — split-based extraction, so both parsers follow the same rule. Verified: `Bearer` + double space is now 200-anonymous here, matching pr-601. The exact `mcp-handler` pin (separate commit) covers the future-version coupling you describe.

src/app/api/mcp/[[...token]]/route.ts
Comment (christosporios) [reply src/app/api/mcp/[[...token]]/route.ts]: Taken — `identityFromAuthInfo` in auth.ts is now the single home of that cast; `identityFromContext`, `analytics.ts` and the route all delegate to it.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/1638e000998f1b06825a3ad1652fee79ee16a067..506e37f6a71a29c46b57a29e10a86ec7ae4a17b8) to rebase onto main now that #601 is merged. The conflict was this branch still carrying the pre-merge copy of #601's commit — main's version had the test-file improvements added at merge time (the sharpened string-dates test description plus the new recency-ordering test), which this branch now inherits instead. No changes to this PR's own diff; `tsc` and all suites green, PR shows mergeable again.

PR #606, opened 2026-08-10
fix(mcp): resolve /mcp page locale from params, not the request default

christosporios:
opencouncil.rs/mcp (and .fr/mcp) rendered Greek content under a correct `<html lang>`.

The page is under the `[locale]` tree and fully translated (`sr/mcp.json` is complete) — but it called `getTranslations` without a locale and took no `params` at all. Layouts and pages render **in parallel** in the App Router, so the page's `getTranslations` raced the `[locale]` layout's `setRequestLocale` and fell back to the app default (`el`). Every other `(other)`-group page dodges this by awaiting `params` first, which happens to sequence it after the layout.

Fix: both `generateMetadata` and the page now read `locale` from `params` and pass it to `getTranslations` explicitly — race-proof rather than ordering-dependent.

## Testing

Verified against the dev server with spoofed Host headers — title, `<h1>` and `<html lang>` now agree on all three realms:

- `opencouncil.rs` → «Питајте свог AI асистента о својој општини» (was Greek)
- `opencouncil.gr` → «Ρωτήστε τον AI βοηθό σας για τον δήμο σας»
- `opencouncil.fr` → « Interrogez votre assistant IA sur votre commune »

`npx tsc --noEmit` passes.

---

> [!NOTE]
> **Low Risk**
> Localized copy/metadata only on one page; no auth, data, or API behavior changes.
> 
> **Overview**
> Fixes wrong-language copy on `/mcp` (e.g. Serbian host showing Greek) when `<html lang>` was already correct.
> 
> **`generateMetadata` and the page** now accept `params`, await `locale`, and call `getTranslations({ locale, namespace: ... })` instead of relying on implicit request locale. That avoids a race with the `[locale]` layout’s `setRequestLocale` under parallel App Router rendering, which previously fell back to the default locale (`el`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a61e4491c62479bc00da4c8859e2aab5db87c8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

Fixes the MCP page’s translation race by resolving the locale directly from route parameters.
- Passes the explicit route locale to page translation lookup.
- Applies the same locale handling to generated metadata.
- Keeps authentication, token loading, and canonical alternate behavior unchanged.

The PR appears safe to merge, with page content and metadata now deterministically using the locale selected by the route.

The parent locale layout validates supported locale values, the explicit next-intl lookup form is already established in the repository, and the change does not alter authentication or token access behavior.

| Filename | Overview |
|----------|----------|
| src/app/[locale]/(other)/mcp/page.tsx | Explicitly supplies the validated route locale to both translation lookups, consistently fixing localized metadata and page content without introducing an actionable defect. |

<sub>Reviews (1): Last reviewed commit: ["fix(mcp): resolve /mcp page locale from ..."](https://github.com/schemalabz/opencouncil/commit/6a61e4491c62479bc00da4c8859e2aab5db87c8c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51884800)</sub>

### DIFF
--- a/src/app/[locale]/(other)/mcp/page.tsx
+++ b/src/app/[locale]/(other)/mcp/page.tsx
@@ -10,8 +10,15 @@ import { McpTokenManager } from "@/components/mcp/McpTokenManager";
 import { buildCanonicalAlternates } from "@/lib/utils/hreflang";
 import { ArrowRight, ArrowUpRight } from "lucide-react";
 
-export async function generateMetadata(): Promise<Metadata> {
-    const t = await getTranslations("mcp.metadata");
+// Both entry points read the locale from params and pass it explicitly:
+// without it, getTranslations races the [locale] layout's setRequestLocale
+// (layouts and pages render in parallel) and falls back to the default
+// locale — opencouncil.rs/mcp rendered Greek.
+export async function generateMetadata(props: {
+    params: Promise<{ locale: string }>;
+}): Promise<Metadata> {
+    const { locale } = await props.params;
+    const t = await getTranslations({ locale, namespace: "mcp.metadata" });
     return {
         title: t("title"),
         description: t("description"),
@@ -36,8 +43,9 @@ function SectionHeading({ children }: { children: React.ReactNode }) {
     );
 }
 
-export default async function McpPage() {
-    const t = await getTranslations("mcp");
+export default async function McpPage(props: { params: Promise<{ locale: string }> }) {
+    const { locale } = await props.params;
+    const t = await getTranslations({ locale, namespace: "mcp" });
     const user = await getCurrentUser();
     const tokens = user ? await listUserMcpTokens(user.id) : [];
     const serverUrl = mcpBaseUrl();


kouloumos:
I verified the commit `6a61e449` after the merge. Your three realm results reproduce. The routes `/lat/mcp` and `/en/mcp` give identical output before the change and after it.

I then looked for the cause behind the cause. The follow-up is #615. This comment records the mechanism. It also corrects two statements in the code comment that this pull request adds.

## How the failure works

next-intl keeps one locale value for each request. Only one line writes this value. That line is `setRequestLocale(locale)` at `src/app/[locale]/layout.tsx:46`.

The proxy skips the next-intl middleware for the hosts `opencouncil.rs` and `opencouncil.fr`. See `src/proxy.ts:167-183`. On these two hosts the layout is the only source of the locale.

next-intl also caches each configuration. The cache key is the locale argument. A call to `getTranslations("mcp")` sends no locale. The cache key for that call is `undefined`. A call to `getTranslations({ locale, namespace: "mcp" })` uses the key `sr`.

A request to `opencouncil.rs/mcp` runs in this order:

1. The page component calls `getTranslations("mcp")`. The call sends no locale.
2. next-intl reads the request locale. The value is empty. next-intl selects the default locale `el`.
3. next-intl caches this Greek configuration under the key `undefined`.
4. The layout calls `setRequestLocale("sr")`. The request locale is now correct.
5. The layout calls `getMessages()` at line 48. This call also sends no locale.
6. next-intl finds the cache entry for the key `undefined`. It returns the Greek configuration. It does not read the request locale again.

Step 6 causes the wide effect. The locale is correct after step 4. No later call reads it.

The effect also reaches the client components. `getMessages()` at `layout.tsx:48` supplies `NextIntlClientProvider`. I counted the rendered strings in the response for `opencouncil.rs`. Before the fix: `Αρχική` 9, `Почетна` 0. After the fix: `Αρχική` 0, `Почетна` 9. The header and the footer were Greek. The attribute `<html lang>` was correct throughout.

## Two corrections to the code comment

**The comment names `generateMetadata` as a cause.** I tested the four combinations:

| | `<title>` | body | client components |
|---|---|---|---|
| both implicit | Greek | Greek | Greek |
| metadata explicit, page implicit | Serbian | Greek | Greek |
| metadata implicit, page explicit | Serbian | Serbian | Serbian |
| both explicit (merged) | Serbian | Serbian | Serbian |

Row 3 shows the important result. The implicit call in `generateMetadata` returns the correct locale on its own. The page component is the only cause. Please treat this order as an observation. I measured it on one route. The Next.js documentation does not state it.

**The comment states the rule "await `params` first".** That rule is not correct. `ProfilePage` at `profile/page.tsx:15` takes no props. `UnsubscribePage` at `unsubscribe/page.tsx:18` takes only `searchParams`. Both pages are safe. Each page awaits some other value before the translation call.

## What #615 does instead

#615 changes the proxy. The rewrite resolves the target path through next-intl first. It then copies the request headers that next-intl sets. The locale header is therefore present on `opencouncil.rs` and `opencouncil.fr`.

The layout is no longer the only source of the locale. An implicit `getTranslations` resolves correctly on any page. I verified this with the fix from this pull request removed.

#615 also localizes the `/unsubscribe` title. That page declared a static Greek string. Production served that title on every realm and every locale.

## I withdraw the earlier recommendation

The first version of this comment asked for an explicit locale in six more files. I withdraw that request.

I checked all six paths on production. Each one renders the correct language today. Those six edits fix instances. #615 removes the cause. The explicit locale in this pull request also becomes unnecessary. It stays correct, so it needs no revert.

---

**Edit:** I have revised this comment twice. The first version asked for the six-file change. The second version stated the execution order of `generateMetadata` as a fact. This version records that order as an observation, and it replaces the six-file request with #615.

PR #609, opened 2026-08-11
feat(mcp): tool annotations and categories

christosporios:
When adding the OpenCouncil MCP to Claude, the permissions screen showed one tool under "Read-only" and nineteen under "Other" — because the read-only grouping comes from the MCP spec's `ToolAnnotations.readOnlyHint`, which we never set (the one correctly-grouped tool was PostHog's virtual `get_more_tools`, which ships its own annotation).

## What changes

- All sixteen read tools declare `{ readOnlyHint: true, openWorldHint: false }`.
- The three writes (`create_highlight`, `generate_highlight_video`, `set_highlight_showcase`) declare `readOnlyHint: false, destructiveHint: false` — they create, render, or toggle, never delete — plus `idempotentHint: true` on the latter two, where re-calling with the same arguments is a designed no-op. `openWorldHint: false` everywhere: these tools only touch OpenCouncil's own data.
- Every tool carries `_meta: { category }` (`discovery` / `directory` / `meetings` / `highlights`). The PostHog MCP SDK reads exactly this key into `$mcp_tool_category`, so tool-call analytics gain a grouping dimension for free.

Result: Claude's permission screen shows the sixteen read tools under "Read-only — always allow", and only the three highlight actions require per-use approval.

## Not included (checked, not possible today)

Server-level branding (`title`/`websiteUrl`/`icons` on `serverInfo`) — `createMcpHandler`'s option type only forwards `name`/`version`. Per-tool `icons` would also need a square icon asset (`public/logo.png` is 1606×1354). Both possible later; neither affects the permissions grouping.

## Testing

`npx tsc --noEmit` + MCP suite pass. Verified live on the dev server through the full stack (including the PostHog schema-injection wrapper): anonymous `tools/list` shows all 15 advertised tools with `readOnlyHint: true`; the authenticated list adds the three writes with their exact annotations; `_meta.category` rides along on every tool.

---

> [!NOTE]
> **Low Risk**
> Declarative metadata only; tool execution, auth, and data access are unchanged.
> 
> **Overview**
> Adds MCP **tool annotations** and **category metadata** so clients can correctly group permissions and PostHog can dimension tool usage.
> 
> Read tools now declare `readOnlyHint: true` (plus `openWorldHint: false`), so clients like Claude can put them under always-allow read-only. The three highlight writes declare `readOnlyHint: false` and `destructiveHint: false`, with `idempotentHint: true` on `generate_highlight_video` and `set_highlight_showcase`.
> 
> Every tool also gets `_meta.category` (`discovery` / `directory` / `meetings` / `highlights`) for `$mcp_tool_category` analytics. A regression test asserts every registered tool has a boolean `readOnlyHint` and a known category.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cbb2e2ed8b4aa5ea0708eaf0f9f20d60451d4b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds MCP safety annotations and analytics categories to registered tools without changing their handlers.
- Marks public and authenticated read tools as read-only.
- Describes the three highlight actions as non-destructive and classifies two as idempotent.
- Assigns stable discovery, directory, meetings, and highlights categories.
- Adds an invariant test requiring every advertised tool to expose a read-only hint and recognized category.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/mcp/server.ts | Adds permission-related MCP annotations and analytics category metadata to every registered tool. |
| src/lib/mcp/__tests__/server.test.ts | Extends the registration recorder and verifies that all authenticated tools declare a read-only hint and recognized category. |
| src/lib/mcp/data.ts | Documents the create-only highlight behavior on which the non-destructive annotations rely. |

<sub>Reviews (2): Last reviewed commit: ["feat(mcp): tool annotations and categori..."](https://github.com/schemalabz/opencouncil/commit/0cbb2e2ed8b4aa5ea0708eaf0f9f20d60451d4b4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52184213)</sub>

**Context used:**

- Knowledge Base — [MCP assistant integration](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/mcp-assistant.md)

### DIFF
--- a/src/lib/mcp/server.ts
+++ b/src/lib/mcp/server.ts
@@ -49,6 +49,15 @@ async function run(fn: () => Promise<unknown>): Promise<CallToolResult> {
     }
 }
 
+/**
+ * Tool grouping, stamped into each tool's `_meta`. Not decorative: the
+ * PostHog MCP analytics SDK reads exactly `_meta.category` into
+ * $mcp_tool_category, so these strings become analytics dimensions — keep
+ * them stable, and keep this union the only place they are defined.
+ */
+type ToolCategory = 'discovery' | 'directory' | 'meetings' | 'highlights';
+const category = (category: ToolCategory) => ({ category });
+
 const paginationShape = {
     page: z.number().int().min(1).default(1).describe('Page number, starting at 1'),
 };
@@ -58,6 +67,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'search',
         {
             title: 'Search subjects',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('discovery'),
             description:
                 'Full-text and semantic search over council meeting subjects (agenda items). ' +
                 'Filter by city, person, party, topic or date range. Omit the query to list ' +
@@ -86,6 +97,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'list_hot_subjects',
         {
             title: 'List hot subjects',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('discovery'),
             description:
                 'The most-discussed subjects across all municipalities over a recent period, ranked ' +
                 'by debate time — start here for "what is happening in the councils", a weekly ' +
@@ -107,6 +120,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'list_nearby_subjects',
         {
             title: 'List subjects near a location',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('discovery'),
             description:
                 'Recent council subjects around a geographic point — "what has the council discussed ' +
                 'about this neighborhood" (for the decision and votes, follow up with get_subject). ' +
@@ -134,6 +149,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'fetch',
         {
             title: 'Fetch a record',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('discovery'),
             description:
                 'Fetch the full content of a single record by id. Accepts a subject id (default), or ' +
                 'prefixed ids: "city:{cityId}", "person:{personId}", "party:{partyId}", "meeting:{cityId}/{meetingId}".',
@@ -148,6 +165,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'list_cities',
         {
             title: 'List municipalities',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('directory'),
             description: 'List the municipalities available on OpenCouncil, with ids and counts.',
             inputSchema: z.object({}),
         },
@@ -158,6 +177,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'get_city',
         {
             title: 'Get municipality',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: category('directory'),
             description: 'Get a municipality profile, including its political parties.',
             inputSchema: z.object({ cityId: z.string().min(1) }),
         },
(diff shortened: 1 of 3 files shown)


src/lib/mcp/server.ts
```diff
@@ -58,6 +58,8 @@ export function registerOpenCouncilServer(server: McpServer) {
         'search',
         {
             title: 'Search subjects',
+            annotations: { readOnlyHint: true, openWorldHint: false },
+            _meta: { category: 'discovery' },
```
kouloumos: `_meta` is `Record<string, unknown>` in the SDK, and `@posthog/mcp` takes any non-empty string, so `'meeting'` or `'Directory'` compiles, ships, and quietly becomes its own `$mcp_tool_category` value. The four strings have no definition site — they exist only at the 19 use sites.

```ts
type ToolCategory = 'discovery' | 'directory' | 'meetings' | 'highlights';
const category = (category: ToolCategory) => ({ category });
```

then `_meta: category('discovery')`. Also gives the union a natural home for a note that this feeds PostHog, which nothing currently says.

kouloumos (approved):
The annotations match what the handlers do, including the three write tools. `destructiveHint: false` on the video render holds because the MCP surface has no way to change a highlight's content: `mcpCreateHighlight` calls `upsertHighlightCore` without an `id` (`src/lib/mcp/data.ts:882`), so the upsert's `where: { id: id || 'new' }` never matches and it always takes the create branch, and there's no edit tool in the suite. The update branch, which does `highlightedUtterances: { deleteMany: {} }`, is unreachable from MCP. A re-render can therefore only swap the format of content that cannot change.

Nothing in the code records that, though. Add a `highlightId` to `create_highlight` so agents can revise a selection and both write annotations become wrong at once: `create_highlight` starts deleting utterances while advertising additive-only, and `generate_highlight_video`'s `destructiveHint: false` stops holding too, since a re-render could then replace content rather than just reformat it. Worth a comment at that call recording why no id is passed.

None of this is tested. The recorder in `src/lib/mcp/__tests__/server.test.ts:35` takes only the tool name and discards the config, so a tool added later without annotations lands back in the "Other" bucket exactly as before, with no type error, no runtime error and no failing test. Nothing else in the repo reads these fields, so that file is the only place that can notice. Capturing the second argument is enough, and the assertion wants to be an invariant rather than an inventory so that adding a tool needs no test edit:

```ts
it('gives every tool a readOnlyHint and a known category', () => {
    const { tools, meta } = advertised(USER);
    expect(tools.filter((name) =>
        typeof meta[name].annotations?.readOnlyHint !== 'boolean'
        || !CATEGORIES.includes(meta[name]._meta?.category ?? '')
    )).toEqual([]);
});
```

On this branch it passes, and it fails for a tool registered with no annotations, a category typo (`'meetings'` to `'meeting'`), a dropped `_meta`, and a missing `destructiveHint` on a write tool. Filtering into an array rather than asserting in a loop makes Jest name the offending tool.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/729935f1e718b77b96cd7c79623ac03be19c20ae..0cbb2e2ed8b4aa5ea0708eaf0f9f20d60451d4b4) taking all three review points:

- **The fragile invariant is now written down**: a comment at `mcpCreateHighlight`'s `upsertHighlightCore` call records that no `id` is passed on purpose — MCP can only ever create, the delete-carrying update branch is unreachable, and both write-tool `destructiveHint: false` annotations lean on exactly that. Anyone adding a `highlightId` parameter will read why it falsifies both.
- **Your invariant test is in**: the recorder now captures the config, and `tool metadata › gives every tool a readOnlyHint and a known category` asserts the invariant with the filter-into-array shape so Jest names offenders. Verified it fails on a tool with no annotations and on a category typo; passes on the branch (5/5 in the file).
- **Categories have a definition site**: `ToolCategory` union + `category()` helper in server.ts, with the doc note that these strings feed `$mcp_tool_category` — a typo is now a compile error instead of a stray analytics dimension.

src/lib/mcp/server.ts
christosporios: Taken — `ToolCategory` union + `category()` helper, with the PostHog note at the definition site. A typo like `'meeting' is now a compile error.

PR #610, opened 2026-08-11
fix(i18n): stop Greek text and .gr links leaking into other realms

christosporios:
Greek text and `.gr` links were leaking into the non-Greek realms. Reported from opencouncil.rs with screenshots; this fixes everything in that report, plus the highlight email found while tracing it.

Two root causes, both predating the multi-realm work:

- **Strings typed directly into components** instead of the message catalogs, so they render in Greek whatever the page language is.
- **Links built from `NEXTAUTH_URL`**, which is one build-time host, so they always point at opencouncil.gr — sometimes with a hardcoded `/el` on top.

## What's in here

**`i18n(ui)`** — the presentation view (all of it), the transcript's "jump to current time" button, the DOCX/audio export buttons, and the in-person operator panel. That last one was hardcoded *English*, so Greek admins saw English too. `formatDateTime` also defaults to Greek and was never passed a locale, which is why the presentation slide showed a Greek date under a Serbian meeting title.

**`i18n(metadata)`** — `generateMetadata` built city page titles by hand in Greek, so a Serbian city's browser tab read `Δημοτικοί Σύμβουλοι | Niš | OpenCouncil`. Covers the three tabs a visitor lands on. This is also what search engines index for `.rs` and `.fr`.

**`feat(explain)`** — `/explain` is a long article about how *Greek* municipalities work, priced in euros and citing Διαύγεια. There is no Serbian equivalent to translate to, so the page now 404s outside `el` and every entry point hides itself: footer link, landing info panel, and the map badge (which sits on the Athens office and whose popup exists only to link there). `hasExplainPage` is the single decision point, so a visible link can't outlive the page it points at.

**`fix(auth)`** — a reader who signed in on opencouncil.rs got a Greek magic link and landed **signed out** on opencouncil.gr/profile. next-auth rewrites every request URL to `NEXTAUTH_URL`'s origin (`reqWithEnvURL`), so the Greek domain is the only origin the auth code ever sees. That broke two things: the post-verification target is baked into the link as an absolute `.gr` URL (we repointed the link but not that parameter, so the token verified on `.rs` and then redirected to `.gr`, where the session cookie didn't exist), and Auth.js' default `redirect` callback discarded any target that isn't that origin. Both fixed; origins are compared parsed, since `opencouncil.gr` is a string prefix of `opencouncil.gr.evil.com`.

**`fix(email)`** — the highlight-ready email linked to `${NEXTAUTH_URL}/el/${cityId}/...`, which 404s for a non-Greek city and costs a redirect hop even on `.gr`.

## Why email copy isn't in `messages/`

Both localized emails are rendered outside any next-intl request scope — one from Auth.js' provider callbacks, one from a task callback — and neither can load the catalogs (`@/i18n/request` reads them off disk with `fs`, and `auth.config.ts` is pulled into the middleware bundle). So their copy lives in typed per-locale tables next to each template, sharing one lookup helper. `EmailLocale` is derived from `REALMS`, so adding a realm in a new language won't compile until that language's email copy exists.

## Verification

Checked against the running dev server in `el` / `sr` / `sr-Latn`: page titles, the presentation view, the admin panel, and the rendered emails in all three languages. `sr-Latn` is transliterated from `sr` at load time, so the new catalogs need no Latin entries.

Each commit passes tsc and the full test suite on its own. New tests cover the magic-link rewrite (including the lookalike-origin case), the email locale resolution, and the realm base URL in production vs preview vs local.

## Out of scope

The notification email/SMS, the unsubscribe link, the transcript email and `src/lib/auth/invite.ts` (the admin-invite sign-in email) all have the same `NEXTAUTH_URL` + hardcoded-Greek problem, but their bodies are fully Greek too — each is a job about the size of this one. `invite.ts` in particular needs the same treatment as the highlight email rather than a subject-line swap: `sendInviteEmail` has no request, so the realm has to come from the invited user's city. Agreed to handle separately.

Also unrelated but worth flagging: viewing a Greece city under a `?realm=serbia` override 500s. It reproduces on `main`, and it's the preview-only escape hatch, not production.

---

> [!NOTE]
> **Medium Risk**
> Changes touch authentication redirects and magic-link/callback URL rewriting across production domains; mistakes could break sign-in or open redirect issues, though the allowlist and parsed-origin checks mitigate that.
> 
> **Overview**
> Stops Greek UI strings, Greek SEO metadata, and **opencouncil.gr** URLs from appearing on Serbian, French, and other non-Greek realms.
> 
> **UI and SEO** — Hardcoded Greek (and some English) in the presentation view, transcript “jump to current time”, meeting export toasts, and in-person operator admin panel move into **next-intl** catalogs (`presentation`, `transcript`, `admin`). City **people/parties** tab and city home **generateMetadata** use new `metadata.*` message files so titles and descriptions follow the page locale.
> 
> **/explain** — Greek-only article now **404s** outside `el` (metadata + page body). Footer, landing info panel, and map “explain” badge use **`hasExplainPage`** so links and popups do not point at a missing page.
> 
> **Auth** — Magic-link emails use **`localeForRequest`** (host, realm override, Serbian script cookie) and localized **AuthEmail** copy. **`signInUrlForRequest`** also rewrites embedded **`callbackUrl`** onto the sign-in host (parsed origin compare, not string prefix). **`redirect`** callback in **`auth.ts`** allows redirects to any **`isKnownRealmHost`** domain so sign-in on `.rs`/`.fr` does not dump users on `.gr` signed out.
> 
> **Transactional email** — Highlight-ready email uses realm locale, **`emailBaseUrlForRealm`**, and localized templates; auth/highlight copy lives in template-local tables (outside `messages/`) because Auth.js and task callbacks are outside next-intl.
> 
> **Misc** — User dropdown uses locale-aware **`useRouter`** from `@/i18n/routing` for sign-in navigation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f075e18b8e4ed0c2574d889a9eff21629aad6253. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR localizes previously hardcoded UI and metadata content and makes links, authentication redirects, transactional emails, and the Greek-only explainer realm-aware.
- Adds localized presentation, transcript, export, operator, metadata, and email copy.
- Rewrites magic-link and callback URLs for the originating realm.
- Restricts `/explain` and its entry points to the Greek realm.
- Builds highlight-email links from the associated city’s realm.

The PR appears safe to merge because no eligible blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/auth.ts | Adds parsed, HTTPS-only redirect support for canonical realm apex domains while retaining same-origin and relative redirects. |
| src/lib/auth/signInUrl.ts | Rewrites magic-link and embedded callback origins to the validated request realm host. |
| src/auth.config.ts | Selects localized authentication-email copy and the request-specific sign-in URL. |
| src/lib/email/highlightComplete.ts | Generates localized highlight-ready emails using the meeting city’s realm and canonical base URL. |
| src/lib/explain/availability.ts | Centralizes the Greek-realm availability decision for the explainer page and its entry points. |
| src/components/layout/Footer.tsx | Makes the explainer footer link conditional on the server-resolved realm. |
| src/components/landing/v2/LandingV2.tsx | Propagates realm-based explainer availability to landing-page links, panels, and map markers. |
| src/app/[locale]/(other)/explain/page.tsx | Returns a real 404 for the Greek-specific explainer outside the Greek realm. |
| src/app/[locale]/(city)/[cityId]/(other)/(tabs)/people/page.tsx | Replaces hardcoded Greek people-page metadata with locale-specific catalog messages. |

<sub>Reviews (3): Last reviewed commit: ["fix(email): point the highlight-ready em..."](https://github.com/schemalabz/opencouncil/commit/f92dc853d3f90b3e089914721cb5791ef7c62dcb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52231792)</sub>


src/app/[locale]/(city)/[cityId]/(other)/(tabs)/people/page.tsx
```diff
@@ -27,29 +29,18 @@ export async function generateMetadata(props: { params: Promise<{ cityId: string
...
-            'κόμματα',
-            'παρατάξεις',
-            cityName,
-            'OpenCouncil',
-            'λίστα συμβούλων',
-            'πολιτικές παρατάξεις'
-        ],
-        authors: [{ name: `Δήμος ${cityName}` }],
+        keywords: [...(t.raw('keywords') as string[]), cityName, 'OpenCouncil'],
+        authors: [{ name: t('author', { cityName }) }],
```
Comment (kouloumos) [inline src/app/[locale]/(city)/[cityId]/(other)/(tabs)/people/page.tsx:41]: `metadata.people.author` isn't in any of the four catalogs — only `metadata.city.author` is. next-intl logs the miss and returns the dotted path rather than throwing, so it renders:

```
opencouncil.gr/athens/people    <meta name="author" content="Δήμος Αθήνα">
pr-610 preview /athens/people   <meta name="author" content="metadata.people.author">
```

Greek production included, not just the new realms. Each locale's existing `metadata.city.author` is the same `Δήμος {cityName}` string this used pre-PR, so it copies straight across.

`translations.test.ts` can't catch this — it compares the four catalogs to each other, and the key is missing from all four equally.

src/lib/explain/availability.ts
```diff
@@ -0,0 +1,14 @@
...
+ * it is generated from translated content, so on any other locale it renders as
+ * a wall of Greek text — which is what it did on opencouncil.rs.
+ *
+ * Single source of truth for "does this locale have an explain page": the page
+ * itself 404s when it returns false, and every entry point into it (footer link,
+ * landing info panel, landing map badge) hides itself. Keep them agreeing —
+ * a visible link to a 404 is worse than either alone.
+ */
+export function hasExplainPage(locale: string): boolean {
+    return locale === 'el';
```
kouloumos: Cyprus's default locale is `el`, so this returns true on opencouncil.cy — the article, the footer link, and the map badge (which sits on the Athens office) all still render there. On the preview, `?realm=cyprus` serves the page at 200 with Διαύγεια and the euro pricing intact; fr and rs correctly 404.

The page is about Greek local government rather than about the Greek language, so I think the predicate wants the realm. That would also make the sitemap gate fall out naturally — `sitemap.ts` has a `Realm` and no locale, which is why gating it on a locale reads awkwardly.

Defensible either way as "Greek readers can read it" — just flagging that the page's own docstring argues the other side.

src/auth.ts
```diff
@@ -21,6 +22,25 @@ declare module "next-auth" {
...
+         * cross-origin and drops the user on opencouncil.gr. Allow any of our own
+         * domains through, so a sign-in that started on .rs/.fr finishes there —
+         * `isKnownRealmHost` is a strict allowlist, not a suffix check on
+         * attacker-supplied input.
+         */
+        redirect({ url, baseUrl }) {
+            if (url.startsWith('/')) return `${baseUrl}${url}`;
+            try {
+                const { origin, host } = new URL(url);
+                if (origin === baseUrl || isKnownRealmHost(host)) return url;
```
kouloumos: `isKnownRealmHost` matches any subdomain, which is wider than this needs. Driving the signout action against both builds:

```
callbackUrl                                 main        this branch
https://opencouncil.rs/profile              rejected    allowed    ← the point
https://data.opencouncil.gr/x.png           rejected    allowed
https://evil.example.com/phish              rejected    rejected
https://opencouncil.gr.evil.example.com/x   rejected    rejected
```

The lookalike defence holds. But `data.opencouncil.gr` is the Spaces bucket, and `/api/upload/presigned-url` takes `contentType` from the request body unvalidated behind `withServiceOrUserAuth` — so a city admin could park HTML there and have our own auth endpoint redirect to it. I didn't upload anything to prove that last hop, so treat the chain as read rather than demonstrated.

`origin === baseUrl || isRealmApexHost(host)` keeps previews working via same-origin and drops the subdomains. Worth pinning the scheme too — `http://opencouncil.gr/x` passes today.

src/lib/auth/signInUrl.ts
```diff
@@ -43,7 +32,33 @@ export function signInUrlForRequest(url: string, request: Request): string {
     const proto =
         firstHeaderValue(request.headers.get('x-forwarded-proto')) ??
         target.protocol.replace(/:$/, '');
```
kouloumos: This takes `x-forwarded-proto` at face value. Replicating the function's logic against a `.rs` host:

```
x-forwarded-proto: https   →  https://opencouncil.rs/…?token=…
x-forwarded-proto: http    →  http://opencouncil.rs/…?token=…
x-forwarded-proto: ftp     →  ftp://opencouncil.rs/…?token=…
(absent)                   →  falls back to the link's own scheme ✓
```

The docstring above makes a careful case that Host is attacker-controllable and allowlists it, then trusts the sibling header that decides whether the sign-in token travels in cleartext. Hard to reach behind Cloudflare, but the asymmetry with your own threat model is the thing. Taking the forwarded value only when it reads `https` leaves the local-http-dev fallback intact.

src/components/meetings/admin/MeetingOperator.tsx
```diff
@@ -102,18 +104,16 @@ function MeetingOperatorInner({
...
             });
         } catch (error) {
             toast({
-                title: "Error",
+                title: t("errorTitle"),
                 description:
-                    error instanceof Error
-                        ? error.message
-                        : "Failed to update operator",
+                    error instanceof Error ? error.message : t("updateFailed"),
```
kouloumos: `t("updateFailed")` can't be reached. The only throw in this handler is `new Error("Failed to update operator")` a few lines up, and a failed `fetch` or `res.json()` is an `Error` too — so the branch that always runs is `error.message`, and the four new `updateFailed` entries are unused.

Net effect on the realm this commit is for: a Serbian admin gets a Greek toast title over an English body, on the one path where the panel actually has something to say.

`description: t("updateFailed")` for the `!res.ok` case, keeping `error.message` for genuinely unexpected throws.

src/components/meetings/MeetingExportButtons.tsx
```diff
@@ -64,15 +66,15 @@ export function MeetingExportButtons({
...
       toast({
-        title: "Εξαγωγή επιτυχής",
-        description: "Το αρχείο ήχου κατέβηκε με επιτυχία.",
+        title: t('successTitle'),
+        description: t('audioSuccess'),
       });
     } catch (error) {
       console.error('Error exporting audio:', error);
-      const errorMessage = error instanceof Error ? error.message : 'Άγνωστο σφάλμα';
+      const errorMessage = error instanceof Error ? error.message : t('unknownError');
```
kouloumos: Same shape as the operator panel — every throw site here is an `Error` carrying an English message, the common one being `'No audio URL available for this meeting'` from `lib/export/meetings.tsx`. So `t('unknownError')` never renders, and the interpolation produces `Извоз звучног записа није успео: No audio URL available for this meeting`.

src/components/presentation/PresentationView.tsx
```diff
@@ -123,10 +125,10 @@ export default function PresentationView({
     const currentSubject = index > 0 ? agendaSubjects[index - 1] : null;
     const positionLabel = index === 0
         ? "INTRO"
```
kouloumos: `"INTRO"` is the one string left behind in this file — it renders untranslated in both `el` and `en` on the preview. The nav zones' `aria-label` further down (`"Previous slide"` / `"Next slide"`) is the same, and that one is screen-reader-visible on every locale.

While you're in the deck: `SlideHeader` renders `{city.name}` and `alt={city.name}` raw, so `/en/present/chania/…` shows `Χανιά` under an otherwise fully English slide — its siblings moved to `getLocalizedName` in this commit.

src/lib/email/components/BaseTemplate.tsx
```diff
@@ -10,15 +10,35 @@ import {
...
 
 export const BaseTemplate = ({
     children,
-    previewText = 'OpenCouncil - Ψηφιακή Δημοκρατία'
+    locale = DEFAULT_LOCALE,
+    // Destructured after `locale` so the default can read it: a template that
+    // passes a locale but no preview text still gets its own language.
+    previewText = `OpenCouncil - ${emailCopy(TAGLINE, locale, DEFAULT_LOCALE)}`,
 }: BaseTemplateProps): React.ReactElement => (
     <Html>
```
kouloumos: The new `locale` doesn't reach `<Html>` — `@react-email/html` defaults to `lang="en"`, so every localized email ships English `lang` around Greek, Serbian and French copy. `<Html lang={locale}>`.

src/lib/auth/requestLocale.ts
```diff
@@ -0,0 +1,29 @@
...
+    const realm = effectiveRealm(
+        hostFromRequest(request),
+        readCookie(request, REALM_OVERRIDE_COOKIE),
+    );
+
+    if (realm === 'serbia' && readCookie(request, SERBIAN_SCRIPT_COOKIE) === 'latn') {
+        return 'sr-Latn';
+    }
+
+    return REALMS[realm].defaultLocale;
```
kouloumos, line 34: This can only return `el`, `fr`, `sr` or `sr-Latn` — never `en`, since `en` isn't any realm's default. But `en` is reachable on every realm and never redirected away, so someone signing in from `opencouncil.gr/en` still gets the Greek email, and the `en` block in `AuthEmail` (and in `TAGLINE`) is unreachable.

The magic link's `callbackUrl` carries the locale-prefixed path they came from and `signInUrlForRequest` already parses it, if you want the signal. Otherwise dropping the `en` entries and saying email locale is realm-derived by design would at least make the tables honest.

kouloumos (commented):
Two things I'd fix before this merges, both small.

The first is a missing catalog key on the councillors tab — inline below. next-intl doesn't throw on a missing key, so it ships silently, and it's every locale rather than just the new realms.

The second isn't in the diff, which is probably why nobody's caught it: `src/app/sitemap.ts` emits `/explain` unconditionally and is realm-scoped, so opencouncil.fr and opencouncil.rs now advertise a URL that 404s. I checked all four realms against the preview — fr and rs serve 404 with the entry still in their sitemap, gr and cy serve 200. `hasExplainPage`'s docstring claims every entry point hides itself; the sitemap is the one entry point no human ever looks at. Gating it needs the realm's default locale rather than a locale, which ties into the note I left on the gate itself.

Two structural things. `src/lib/auth/invite.ts` — the other sign-in email — has exactly the bug you're fixing here, `NEXTAUTH_URL` plus a hardcoded Greek subject, and it isn't in the out-of-scope list; small enough to take now or worth naming there so it doesn't get lost.

And the reasoning for keeping email copy out of `messages/` doesn't hold as written. Next 16 runs the proxy on Node — "Proxy always runs on Node.js runtime", in its own build code — so `fs` isn't the constraint. The real one is that `next-intl/server` imports `next/headers` at module top level, which does justify it for the auth email. It doesn't for the highlight email: that renders from a plain Node route handler, and passing a locale to `getTranslations` short-circuits `getRequestLocale()` entirely, which `lib/mcp/data.ts` already relies on. Not worth reworking now, but the comments will be load-bearing for whoever adds the third email.

The auth work itself is right, and the parsed-origin comparison earns its keep — I probed the lookalike origin against both builds and it's correctly rejected in favour of the base URL. Two hardening notes inline.

On what's measured vs. read: everything above was checked against the preview with production as the baseline, signed in as superadmin. Serbian and French were reached with `?realm=` on a `.gr` preview host, since staging has no non-Greek city — so realm-domain behaviour specifically (sign-out cross-domaining, the operator panel itself) is code-reading, not runtime.

I have the two fixes plus mutation-tested coverage for the gap that let the missing key through, if that's useful — `translations.test.ts` compares the catalogs to each other, and a key absent from all four equally passes it.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/f075e18b8e4ed0c2574d889a9eff21629aad6253..65a62990b5925313aa15539a6106c5364f968c93) to address the review — all of it, squashed into the commits that introduced each problem.

**The missing key.** You're right, and it was mine — I moved `authors` to `t('author')` under `metadata.people` but only defined `author` under `metadata.city`. Copied each locale's existing string across; `/athens/people` is back to `content="Δήμος Ζωγράφου"` on the Greek realm and `Град …` on `?realm=serbia`. Your point about `translations.test.ts` being structurally unable to catch this stands — a key absent from all four passes it. **Yes please** to the mutation-tested coverage; that gap will keep producing this class of bug and it's worth its own PR rather than being buried here.

**`hasExplainPage` now takes a `Realm`.** You convinced me: the page is about Greece, not about the Greek language, and Cyprus having `el` as its default made a locale check quietly wrong. And it does make the sitemap fall out — `sitemap.ts` had a realm and no locale, which is exactly why gating it read awkwardly.

```
                greece   cyprus   france   serbia
page            200      404      404      404
sitemap entry   1        0        0        0
footer link     1        0        0        0
```

That meant threading the realm to the footer and the landing rather than reading it in the browser, so the link is absent from the first painted HTML rather than disappearing after hydration. The footer takes it from its three server layouts; the landing already resolved a realm for its data queries and now passes it through the existing `layoutProps` bundle to `InfoPanel`.

**Both hardening notes taken.**

`redirect` is now `origin === baseUrl || (https && isRealmApexHost(host))`. You're right that `isKnownRealmHost` was wider than the job needs, and `data.opencouncil.gr` is a good reason not to hand our own auth endpoint a subdomain wildcard. Previews keep working through the same-origin branch.

`x-forwarded-proto` may now only *upgrade* the scheme — an explicit `https` is taken, anything else keeps the link's own scheme, so the local-http-dev fallback survives. The asymmetry you pointed at was real: allowlisting the Host and then trusting its sibling header to decide whether the token travels in cleartext.

**The unreachable error strings.** Both correct, and worse than dead keys — on the one path where each panel has something to say, a Serbian admin got a Greek title over an English body. The operator panel now throws `t("updateFailed")` so the translated copy is what `error.message` carries. The audio export needed a bit more: all four of its failures were English `Error`s from `lib/export/meetings.tsx`, so they now carry an `AudioExportError` with a `reason`, and the component translates the expected ones while leaving genuine surprises as-is.

**Presentation leftovers** — `INTRO`, both nav `aria-label`s, and `SlideHeader`'s raw `city.name`/`alt` all fixed. The header one was a straightforward miss: its siblings moved to `getLocalizedName` in the same commit.

**`<Html lang>`** — fixed, and thank you. Every localized email was shipping inside `lang="en"`.

**On the email-copy reasoning: you're right and my comment was wrong.** I checked rather than took it on trust — `"Proxy always runs on Node.js runtime"` is in Next's own `get-page-static-info.js`, so `fs` was never the constraint. The real one is that `server/react-server/RequestLocale.js` does a top-level `import { headers } from "next/headers"` and `getConfig` pulls it in statically, so the module graph is the problem rather than the call. And `getConfig`'s `get requestLocale()` does short-circuit when handed an explicit locale, so the highlight email genuinely could live in `messages/`. Both docstrings now say so — the highlight one explicitly notes it's a consistency choice rather than a technical necessity, since that's the comment the third email's author will read.

**`localeForRequest` can only return a realm default** — documented rather than fixed. `/en` readers getting a Greek email is real but it's not what this PR is for, and the `callbackUrl` signal deserves its own change.

**`invite.ts`** — added to the out-of-scope list in the description with the other Greek-bodied emails. It needs the same treatment as the highlight email (no request in `sendInviteEmail`, so the realm has to come from the invited user's city), which is more than a subject-line swap.

Also folded in a native speaker's correction to the Serbian magic-link disclaimer: `Уколико сте грешком примили ову поруку, можете је слободно занемарити.` / `Ukoliko ste greškom primili ovu poruku, možete je slobodno zanemariti.`

On your last point — noted that realm-domain behaviour is code-read rather than measured, since staging has no non-Greek city. Same limitation here: everything above was checked locally against `?realm=` overrides, so the sign-out cross-domaining and the operator panel are still reasoned rather than exercised on a real `.rs` host.

Each of the five commits still passes tsc and the full suite on its own.

christosporios:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/65a62990b5925313aa15539a6106c5364f968c93..f92dc853d3f90b3e089914721cb5791ef7c62dcb) — Bugbot is right, and it caught an inconsistency in my own fix rather than just a leftover.

`t("updateFailed")` only covered the `!res.ok` branch. A rejected `fetch` or a bad `res.json()` still landed in `error.message`, so the Serbian admin got "Failed to fetch" — the same English-body-under-a-Greek-title problem I'd just claimed to fix, one branch over.

The framing I used for the export buttons was wrong here. Every failure in that handler means exactly one thing to the reader — the change didn't save — and the underlying message is either browser boilerplate or my own status text. Neither is worth showing over a sentence in their language, so the catch is now unconditionally `t("updateFailed")` with the detail going to `console.error`, and the `!res.ok` throw carries the status code for that log instead of translated copy (which was a bit sneaky of me).

Applied the same reasoning back to the audio export, where I'd left `error.message` for the non-`AudioExportError` case: those are bugs rather than things to tell an admin about, so they now show `t('unknownError')` and the detail stays in the console line that was already there. That also makes `unknownError` reachable, which was the original complaint about it being a dead key.

Net: no raw `error.message` reaches a toast in either panel.

Restacked #611 and #613 on top. tsc clean and the full suite green on all three, and each of #610's five commits still passes on its own.

kouloumos:
Re-reviewed the merged state and verified it in production. Both blockers are fixed, and the three surfaces now agree on every realm:

| realm | `/explain` | sitemap entry | footer link |
|---|---|---|---|
| greece | 200 | 1 | present |
| cyprus | 404 | 0 | absent |
| france | 404 | 0 | absent |
| serbia | 404 | 0 | absent |

`/athens/people` serves `Δήμος Αθήνα` again, and all four landing pages render clean.

Two of your fixes are better than what I suggested. On the operator panel I proposed restoring `t("updateFailed")` on the `!res.ok` branch and keeping `error.message` elsewhere — which left the same defect one branch over, as Bugbot found. "Every failure here means one thing: it didn't save" is the right question and I didn't ask it, having already established the defect class in the comment next door. And typing `hasExplainPage` on `Realm` makes the agreement compiler-enforced; I argued for the realm on semantics and reached for a test as the enforcement mechanism, which was the more expensive answer.

The catalog guard is #614 — the `t()`-to-catalog axis, plus coverage for the computed ``t(`audioFailure.${...}`)`` key your fix introduced, which a static scan would otherwise skip.

One extra, only because it turns out to be free: the hand-written `sr-Latn` block in `AuthEmail`, and the `TAGLINE` entry, are exactly what `transliterateCatalog` produces from the `sr` block — I checked all five strings and the tagline against the merged copy, corrected disclaimer included, and they match byte-for-byte. That's the same function `src/i18n/request.ts` already uses to derive the entire `sr-Latn` UI catalog, so the email tables could join it and the native speaker's correction would only have needed applying once. Plain-string tables only — the highlight email's closures aren't derivable, and it has no `sr-Latn` entry.

PR #611, opened 2026-08-11
feat(realm): show each realm its own contact number

christosporios:
> Part of stack #612, on top of #610 — review that one first. The base retargets to `main` on its own once #610 merges (`gh stack sync`).

Serbia has a domestic toll-free line, **0800 301167**. Every realm was showing the Athens office number, which a Serbian visitor cannot call for free and — in the international form we printed it — may not reach at all. Every other realm keeps the +30 number.

## Where the number lives

It joins `domain`, `defaultLocale` and `country` in `REALMS`. Realms already own this class of per-country fact, and the type requires a number, so a new realm can't ship without deciding what it dials.

`telHref` derives the dialable form by stripping display spacing and nothing else. Whether a number works internationally is a property of the number itself: the Serbian line is domestic-only, so its leading `0` stays rather than being rewritten to `+381`.

| realm | shown | dialled |
|---|---|---|
| serbia | `0800 301167` | `tel:0800301167` |
| greece / cyprus / france | `+30 211 198 0212` | `tel:+302111980212` |

## How each surface gets the realm

The footer takes it as a **prop from its server layouts** rather than reading `window.location` after mount. The layouts rendering the footer are all server components, so this costs nothing. `CountrySwitcher` now takes the same prop and drops its own host read — the footer resolves the realm once for both.

That switcher change is a **bugfix, not a tidy-up**. It resolved the realm in a lazy initialiser and marked the output `suppressHydrationWarning`, and React does not reconcile suppressed text — so the server's value is final. The footer on opencouncil.rs reads "Ελλάδα" today and keeps reading it. The same reasoning is why `useRealm` resolves in an effect rather than suppressing: an effect corrects the `tel:` href too.

The error and not-found boundaries genuinely have no server parent to take it from, so they keep reading the browser, through a shared `useRealm()` hook (lazy initializer, so the correct value is there on the first client render rather than a paint later).

The landing's contact rows were a module-level constant, so they become `footerGroups(realm)`. `/explain` already had the realm to hand.

## One cosmetic change

The Greek number is now written grouped — `+30 211 198 0212` — everywhere. That is how `/explain` already showed it and how the Serbian number reads; the footer previously ran it together as `+30 2111980212`. One canonical display form beat keeping two.

## Verified

Footer and `/about` on the Greek realm render `+30 211 198 0212` → `tel:+302111980212`; under `?realm=serbia` they render `0800 301167` → `tel:0800301167`. Tests cover both realms' numbers, the href derivation for each, and that every realm has a number and a dialable href — so the checks fail rather than defaulting to Greece when a realm is added.

## `NEXT_PUBLIC_CONTACT_PHONE` is gone

Second commit. That variable held **the same** Athens number the realm config now does, so the contact number had two sources of truth — and `/about`, which is translated and served on every domain, was still showing the Greek number to Serbian readers through it.

Adding a Serbian value to the env files would not have fixed that: `NEXT_PUBLIC_*` is baked in at build time and one deployment serves all four realms, so it can only ever hold one number. Same reason CLAUDE.md gives for not building server URLs from `NEXT_PUBLIC_*`.

All seven readers now take the number from the realm. Most already had one to hand — the brochure resolves the realm forty lines above the line that used the env var, and `/about`'s `Hero` already used it for the domain name. The Greek-language admin invite has no realm in its flow and names the Greek office line explicitly.

Removed from the env schema, the nix dev shell, the preview build workflow and the docs. `contact-badge.tsx` read it and was imported by nothing, so it's deleted. `NEXT_PUBLIC_CONTACT_EMAIL` and `_ADDRESS` stay — one inbox and one address serve every realm, so a single build-time value is still the right shape for them.

Nothing to do on deploy: the variable simply stops being read. It can be dropped from the App Platform config and the `NEXT_PUBLIC_CONTACT_PHONE` GitHub secret whenever convenient.

## Left alone

Three places still hardcode the Greek number, all Greek-only documents rather than site chrome: the offer-letter DOCX letterhead, the transcript email, and the consultation-comment email. Their bodies are entirely Greek, so they belong with the other Greek-bodied emails we agreed to handle separately.

---

> [!NOTE]
> **Medium Risk**
> Wide but mostly display-only changes; medium risk from wrong `tel:` on Serbian/error pages during the initial `useRealm` render and from any missed call site still assuming a single global phone.
> 
> **Overview**
> **Contact numbers are per-realm** in `REALMS` (Serbia gets domestic `0800 301167`; Greece/Cyprus/France share the Athens line). **`getRealmContactPhone`** and **`telHref`** drive display and `tel:` links everywhere site chrome showed a hardcoded or env-baked Greek number.
> 
> **Surfaces wired to the request realm:** footer, about CTAs, landing nav drawers, consultations (`SourcesList`, print footer), brochure (shared inbox default `hello@opencouncil.gr`), and **`CountrySwitcher`** now take **`realm` from the server** so opencouncil.rs does not flash the Greek number on first paint.
> 
> **Client-only boundaries** (`error`, `not-found`) use new **`useRealm()`**, which starts as `greece` and updates from the host in an effect—so non-Greek domains may briefly show the Greek line until hydration.
> 
> **Removed `NEXT_PUBLIC_CONTACT_PHONE`** from `env.mjs`, Nix prod build, preview workflow, and docs; deleted unused **`contact-badge.tsx`**. Tests assert each realm’s number and `telHref` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c91007b5a83adcd294db89a97d2e5041a7caada8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR moves contact phone configuration into the per-realm registry and removes the former build-time phone environment variable.
- Assigns Serbia its domestic toll-free number while retaining the Athens office number for other realms.
- Passes server-resolved realms through footer, landing, about, consultation, and brochure surfaces.
- Adds shared display-to-`tel:` formatting and realm-number tests.
- Uses browser realm resolution for client-only error and not-found boundaries.

The PR does not yet appear safe to merge because Serbian error and not-found responses initially expose a visible and actionable Greek office number.

The browser-only boundaries still initialize their realm as Greece during server rendering and the first client paint, so their phone text and dial target remain incorrect on Serbian hosts until a post-mount effect runs.

**Files Needing Attention:** src/hooks/useRealm.ts, src/app/error.tsx, src/app/not-found.tsx, src/app/[locale]/not-found.tsx

| Filename | Overview |
|----------|----------|
| src/lib/realm.ts | Adds required per-realm contact numbers and a shared helper for generating normalized `tel:` links. |
| src/hooks/useRealm.ts | Browser realm resolution remains deferred until after mount, leaving Serbian error and not-found pages with the Greek contact number in their initial HTML and first paint. |
| src/app/error.tsx | Makes the error-page contact realm-aware, but its rendered phone remains incorrect on Serbian hosts until the realm effect runs. |
| src/app/not-found.tsx | Makes the root not-found contact realm-aware while retaining the outstanding initial-render problem on Serbian hosts. |
| src/app/[locale]/not-found.tsx | Makes the localized not-found contact realm-aware while retaining the outstanding initial-render problem on Serbian hosts. |
| src/components/layout/Footer.tsx | Uses the server-provided realm for both the contact number and country switcher, avoiding browser-dependent first-paint resolution. |
| src/components/landing/v2/navLinks.ts | Replaces the static contact group with realm-derived phone text and dialable links. |
| src/env.mjs | Removes the obsolete build-time contact-phone variable after callers migrate to realm configuration. |

<sub>Reviews (7): Last reviewed commit: ["refactor(realm): drop NEXT\_PUBLIC\_CONTAC..."](https://github.com/schemalabz/opencouncil/commit/49259c8efca11686546ebb3d9288269ae04516f2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52236483)</sub>

**Context used:**


christosporios:
Restacked onto #610's rewritten history. One deliberate change while resolving: the Footer's `realm` prop and the layouts that pass it have **moved down into #610**, because its `/explain` gate now keys on the realm and needs it. This PR keeps only what's actually about the contact number — `CountrySwitcher` taking the same prop and dropping its own host read, the `useRealm()` hook for the error/not-found boundaries, and the number itself. Commit message updated to match. tsc clean, 1385 tests.

src/components/about/Hero.tsx
```diff
@@ -8,7 +8,7 @@ import ShineTitle from './ShineTitle'
 import { env } from '@/env.mjs'
```
kouloumos: This import is dead now. It's the only `env` reference left in the file, and both uses moved to the realm. `CTAFooter` had the same import removed.

src/lib/__tests__/realm.test.ts
```diff
@@ -192,3 +194,34 @@ describe('getRealmGeocoding', () => {
...
+        expect(getRealmContactPhone('france')).toBe('+30 211 198 0212');
+        expect(getRealmContactPhone('cyprus')).toBe('+30 211 198 0212');
+    });
+
+    it('covers every realm, so a new one cannot ship without a number', () => {
+        for (const realm of ALL_REALMS) {
+            expect(getRealmContactPhone(realm)).toMatch(/\d/);
+        }
+    });
+});
```
kouloumos, line 213: The `satisfies` clause is already doing the work this describes — a realm without `contactPhone` won't compile, and that's a better guard than a test. What these two assertions add on top is thinner than the Verified section implies. I added a fifth realm sharing `OFFICE_PHONE` and the suite stayed green at 33/33: `ALL_REALMS` is `Object.keys(REALMS)` at runtime and ts-jest runs `isolatedModules`, so nothing typechecks during a test run and `/\d/` is satisfied. The explicit list above is the only thing pinning actual values.

Folding both into one object comparison makes the tests carry the property rather than the type carrying it alone — a fifth realm then has to be added to the expectation, which is where someone confirms the sharing was deliberate. Green unmutated; against that fifth realm it fails with `+ "portugal": "+30 211 198 0212"`.

```suggestion
describe('getRealmContactPhone', () => {
    it('pairs every realm with its number, so a new one cannot quietly inherit the Athens line', () => {
        expect(Object.fromEntries(ALL_REALMS.map((r) => [r, getRealmContactPhone(r)]))).toEqual({
            greece: '+30 211 198 0212',
            france: '+30 211 198 0212',
            cyprus: '+30 211 198 0212',
            serbia: '0800 301167',
        });
    });
});
```

The `telHref` loop at 222-226 has a smaller version of the same issue — the `+` is optional in `/^tel:\+?\d+$/`, so it passes even if the `+` gets stripped, and the two explicit cases above cover everything it does. The trunk-zero one at 219 is the strongest assertion here; it's the only thing that would catch someone adding an E.164 normalisation later.

src/lib/realm.ts
```diff
@@ -15,16 +15,37 @@ import type { Country } from '@/lib/apiTypes';
  * client components (the footer country-switcher) and the edge/middleware bundle
  * (`proxy.ts`). The request-scoped resolver lives in `realm.server.ts`.
  */
+/**
+ * The Athens office line — the number every realm shows unless it has one of
+ * its own. Written the way it should be displayed; `telHref` derives the
+ * dialable form.
+ */
+const OFFICE_PHONE = '+30 211 198 0212';
```
kouloumos: This lands between the `REALMS` doc comment and `REALMS` itself, so TypeScript now attaches that comment to `OFFICE_PHONE`. Hovering `REALMS` gives you nothing, and hovering `OFFICE_PHONE` gives you the multi-tenancy explanation, including the rule about keeping `next/headers` out of this module. Moving the constant above the doc block, or below `REALMS`, puts it back.

src/hooks/useRealm.ts
```diff
@@ -0,0 +1,24 @@
...
+import { realmForBrowser } from '@/lib/realm.client';
+
+/**
+ * The realm a client component should render for, read from the browser's host.
+ *
+ * Only for components with no server parent to take the realm from — error and
+ * not-found boundaries. Anything rendered inside the app tree should receive it
+ * as a prop from a server component (`getRealm()`), which costs no hydration
+ * mismatch; this necessarily renders the default realm on the server and the
+ * real one on the client, so mark the output `suppressHydrationWarning`.
```
kouloumos: This line tells the next person to use `suppressHydrationWarning`, and that's the construction that's currently broken in production. React doesn't reconcile suppressed text, so the server's value stays on screen permanently — which is why opencouncil.rs still says "Ελλάδα" in the footer, from this same pattern on `CountrySwitcher`.

It doesn't bite here because these boundaries never server-render at all: with JavaScript off, a 404 paints a completely empty body, on production as well as the preview. I verified that for both not-found pages. I couldn't trigger `error.tsx`, so that one I'm assuming matches.

That's a real property, but an undocumented one, and a Next upgrade or a rendering-mode change would revoke it without anyone noticing. Either note that the suppression depends on these boundaries being client-only, or read the realm in an effect instead — then the server and the first client render agree, no suppression is needed, and the second paint updates both the text and the `href`.

src/components/landing/v2/LandingAside.tsx
```diff
@@ -44,6 +45,7 @@ export function LandingAside({
     cities: LandingListCity[];
 }) {
     const t = useTranslations('landingV2');
+    const realm = useRealm();
```
kouloumos: The realm is already two hops up. The page resolves it, `LandingV2` takes it as a prop, and `layoutProps` already carries `explainAvailable`, which is derived from it. Adding `realm` next to that in `LayoutProps` and destructuring it in both layouts gets it here and to `MobileHeader` without introducing anything new.

No rush — I opened both menus on both realms and they show the right number. That works because Radix doesn't mount dropdown or sheet content while it's closed: nothing is in the DOM until you open it, by which point the browser read has resolved. But that's the only thing making it correct. Adding `defaultOpen` or `forceMount`, or moving the contact rows into the rail, would break it, and there's no `suppressHydrationWarning` here, so it would fail silently.

Unrelated and not yours: `ScriptSwitcher` still reads the host in a post-mount effect, and it renders immediately after `<CountrySwitcher realm={realm} />`, where the prop is already in scope.

src/components/consultations/SourcesList.tsx
```diff
@@ -17,6 +19,7 @@ export default function SourcesList({
     consultationId,
     cityId
 }: SourcesListProps) {
+    const realm = useRealm();
```
kouloumos: This file has no `"use client"` of its own. It works because `ConsultationDocument` has one and pulls it into the client bundle. But `components/consultations/index.ts` re-exports it, and that barrel is imported by the server page — so if anyone later writes `import { SourcesList }` from the barrel in a server component, it throws at render. Worth adding the directive regardless of the realm question.

On the realm: the consultation layout already awaits `getRealm()` for the footer, so the prop is available a level up. Low stakes either way — an Athens consultation 404s under the Serbian realm, so this can only ever resolve Greek.

src/components/landing/v2/navLinks.ts
```diff
@@ -46,15 +48,29 @@ export const FOOTER_GROUPS: FooterGroup[] = [
...
             { label: 'hello@opencouncil.gr', href: 'mailto:hello@opencouncil.gr', icon: 'mail' },
         ],
-    },
-];
+    };
+}
+
+/** Every group in display order, contact last — what the popover/drawer render. */
+export function footerGroups(realm: Realm): FooterGroup[] {
+    return [...FOOTER_GROUPS, contactGroup(realm)];
```
kouloumos: `FOOTER_GROUPS` is still exported but no longer holds all the groups. Nothing outside this file reads it, so nothing breaks now. The risk is the next person wiring up a footer-like surface: they reach for the obviously-named exported constant, get a menu with no contact row, and nothing warns them. Renaming it to `STATIC_GROUPS` and dropping the export, or folding all three symbols into one `footerGroups(realm)`, closes that off.

src/app/[locale]/(bare)/brochure/page.tsx
```diff
@@ -54,7 +55,7 @@ export default async function BrochurePage() {
                         coveredBodyTypes: bodyTypesByCity[city.id],
                     }))}
                     contactEmail={env.NEXT_PUBLIC_CONTACT_EMAIL ?? "christos@opencouncil.gr"}
-                    contactPhone={env.NEXT_PUBLIC_CONTACT_PHONE ?? "+30 6980586851"}
+                    contactPhone={getRealmContactPhone(realm)}
```
Comment (kouloumos) [inline src/app/[locale]/(bare)/brochure/page.tsx:62]: These two fallbacks used to be a pair — one person's email and the same person's mobile, printed next to each other in the PDF. The email keeps its fallback and the phone loses one, so a build without the env var now prints that email beside the office landline. Deliberate?

kouloumos (approved):
Your Verified section covers the footer and `/about`. I extended it to `/brochure`, `/explain`, the 404 pages, the desktop popover and the mobile drawer, on both realms — all correct. I also checked the thing the criterion actually turns on, which that section doesn't claim: the number is in the HTML the server sends, not just after hydration. `/sr/about` with JavaScript off already contains `tel:0800301167` three times. `error.tsx` is the one surface I couldn't reach, since I couldn't force a real error.

On `CountrySwitcher` — it isn't a flash it was tolerating. `suppressHydrationWarning` stops React reconciling that text at all, so the server's value stays put. opencouncil.rs reads "Ελλάδα" today and still does six seconds after hydration. That makes it a bugfix rather than a tidy-up.

The one thing I'd push back on is where the new hook gets used. You describe it as being for the error and not-found boundaries, which genuinely have no server parent. But three other call sites use it too — the two landing menus and `SourcesList` — and those do have one. All three work today and I checked each in the browser, but in every case that's down to something nobody wrote down: the Radix menu happens to be closed while the server renders, consultations happen to be Greek-only. The JSDoc also tells the next person to wrap the output in `suppressHydrationWarning`, which is the same construction that leaves "Ελλάδα" on opencouncil.rs — so the diff removes that pattern from `CountrySwitcher` and then recommends it.

The rest is small: a dead import, two test assertions that lean on the type system more than the Verified section suggests, and a JSDoc block that's ended up attached to the wrong declaration.

One housekeeping thing. The greptile block in the description still says "not yet safe to merge" and lists the boundary files. There's nothing behind it — that thread is resolved, cursor's summary is pinned to a commit the restack orphaned, and no review threads are open. Worth clearing out so it doesn't read as an outstanding objection.

christosporios:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/21af07f9e04bf18965f5cb3a655a0d7aeed9b945..c91007b5a83adcd294db89a97d2e5041a7caada8) — all eight, squashed into the two commits.

**The `suppressHydrationWarning` finding is the one that mattered, and you're right that it makes this a bugfix.** I checked `main` and the pattern is exactly as you describe: `CountrySwitcher` resolves in a lazy initialiser and suppresses, so React never reconciles that text and the server's "Ελλάδα" is final. Not a flash. I've rewritten that part of the description.

I took your second option for the hook rather than documenting the dependency. `useRealm` now reads in an effect: server and first client render both start at the home realm, and the second paint corrects the text *and* the `href` — which suppression never would, and which is what the two bots were complaining about. The `suppressHydrationWarning` on the three phone spans is gone, so nothing in the diff recommends the pattern it removes. It no longer matters whether those boundaries server-render.

Verified in the HTML the server sends, which is the thing that actually turns on this:

```
/corrections               greece                      serbia
country label              Ελλάδα                      Србија
footer tel:                tel:+302111980212           tel:0800301167
/this-does-not-exist       no tel: in server HTML at all (renders after mount)
```

**All three misplaced hook call sites now take the prop.** `realm` rides in `LayoutProps` next to `explainAvailable`, so `LandingAside` and `MobileHeader` get it without anything new. `SourcesList` takes it through `ConsultationViewer` → `ConsultationDocument`, and the page already had `getRealm()` a few lines up. `useRealm` is now used by exactly the two boundaries its docstring claims.

On the `"use client"` directive there: removing the hook removes the hazard rather than papering over it. `SourcesList` has no hooks and no handlers left, so importing it from the barrel in a server component is now fine. Adding the directive would force it into the client bundle for no reason. Happy to add it if you'd still rather be explicit.

**The tests were as weak as you showed.** I applied your object comparison and dropped the `telHref` loop. Confirmed both directions: green unmutated at 31/31, and against a fifth realm sharing `OFFICE_PHONE` it fails with exactly `+ "portugal": "+30 211 198 0212"`. The trunk-zero assertion stays.

**`FOOTER_GROUPS`** is now `STATIC_GROUPS` and unexported, so `footerGroups(realm)` is the only way in.

**The doc comment** is back on `REALMS` — moved `OFFICE_PHONE` above the block.

**Dead `env` import** in `Hero.tsx` gone.

**On the brochure pair — not deliberate, good catch.** I changed the email fallback to `hello@opencouncil.gr` so both defaults are the company's, matching the office line. A personal address printed beside a switchboard number was the wrong half to keep.

**Description cleaned** — the stale Greptile block is out, and the `CountrySwitcher` paragraph now says bugfix.

Two things I couldn't reach, same as you: `error.tsx` (couldn't force a real error) and the landing menus on a realm domain — their contact rows aren't in the server HTML at all, because Radix doesn't mount closed menu content, so the prop change there is defensive rather than observable. That's precisely the fragility you flagged.

`ScriptSwitcher` reading the host in an effect is noted and left alone — not this PR.

PR #613, opened 2026-08-11
feat(city): collapse status and officialSupport into pending|demo|supported

christosporios:
> Was part of stack #612; #610 and #611 have merged, so this now sits directly on `main`. **Contains a destructive, irreversible migration** — read the backfill section before merging.

We had no way to express "show this city, but it is not a customer" — which is exactly what opencouncil.rs needs, so the Serbian landing page, MCP and search return something before we have a Serbian customer.

## The problem

Two fields disagreed about what they meant:

- `status` (`pending|unlisted|listed`) decided whether a city was **published**.
- `officialSupport` (Boolean) decided both whether it got the **support badge** *and* whether it existed **on the landing map at all** — `getMapCitiesCached` filtered `officialSupport = true`, and `getPetitionedMapCitiesCached` filtered the same field `false`.

So `listed` + no support — which is exactly what the Belgrade seed fixture is — was a contradictory state: a public page, a directory card, a sitemap entry, search results and MCP visibility, but **no map pins, no outline, no count bubble**, and a *petition* popup when you clicked it, as if we'd never heard of the place.

## The model

| | `pending` | `demo` | `supported` |
|---|---|---|---|
| Landing map (pins, outline, bubble) | – | ✅ | ✅ |
| Directory, search, MCP, sitemap | – | ✅ | ✅ |
| Official-support badge | – | – | ✅ |
| Petition route open (CTA, selector, submissions) | ✅ | ✅ | – |
| Petition *map layer* and leaderboard | ✅ | – | – |
| Customer stats, coverage, ops lists | – | – | ✅ |

The one row that isn't a straight ladder is the petition map. A `demo` city accepts petitions and records them, but it is already drawn on the landing map as covered — putting it in the petition layer too would render the same municipality twice with opposite meanings. Its petition counts surface in the admin expansion table instead. That's commented at both queries.

Every call site now goes through `isPublic` / `isCustomer` / `isOutOfNetwork` / `isPetitionable` in [`src/lib/cityStatus.ts`](src/lib/cityStatus.ts) rather than comparing the field itself — comparing it directly is how the old field came to mean two things. The four raw-SQL sites use `Prisma.Sql` fragments next to the existing `cityCoversPoint` helper, since the compiler can't see inside a template literal.

**`grep -rn "officialSupport\|'unlisted'\|'listed'" src/ prisma/*.ts scripts/` is the review gate.** What survives is comments, the seed's legacy branch, and the deliberately-derived wire field. That grep caught four sites tsc could not: the `/api/cities/at` consumers use inline structural types on a `fetch().then()`, so they'd have silently read `undefined` and shown **petition popups over every municipality**.

## The backfill — destructive, no reverse

```
officialSupport = true  ->  supported
everything else         ->  pending
```

Verified against a scratch PostGIS database by replaying the real upgrade path — seed the pre-migration schema, insert all six field combinations, then `prisma migrate deploy`:

| before | after | |
|---|---|---|
| `listed` + support | `supported` | |
| `listed`, no support | `pending` | **goes dark** |
| `unlisted`, no support | `pending` | |
| `unlisted` + support | `supported` | **becomes published** |
| `pending` | `pending` | |

Two rows to check on production before merging:

- **`listed` without support** stop being published. Accepted, but they keep resolving at `/{cityId}` (`getAllCityIds` is status-agnostic) while leaving the sitemap, search, MCP and the directory — so they stay in Google's index. The SEO follow-up is a separate decision.
- **`unlisted` + `officialSupport = true`** becomes `supported`, i.e. **publicly visible**. There are none in dev; if production has any, they get published by this migration.

```sql
SELECT realm, status, "officialSupport", count(*) FROM "City" GROUP BY 1,2,3 ORDER BY 1,2;
```

Dev is 13 `unlisted`/no-support + 10 `listed`/support, so it migrates to 10 `supported` + 13 `pending` with nothing currently-visible going dark. @kouloumos ran the same query on staging: zero rows in either flagged class.

## Behaviour changes that follow from the model, not the mechanics

- **`demo` cities stay petitionable.** The petition CTA (`CityHeader`) and the petition selector both key off `isPetitionable` now, so a demo city routes to `/{id}/petition`. Showing what OpenCouncil looks like somewhere is what makes asking for it worthwhile.
- **`unlisted` folds into `pending`**, so the City Creator's populate route leaves an imported city unpublished rather than half-published.
- **`scripts/import_offers.ts`** created stubs as `unlisted` and promoted referenced cities to it. With no unpublished-but-not-pending state, promoting would *publish* every city with an offer — so stubs stay `pending` and the promote step is gone.

## Seed — the thing most likely to break previews

`prisma/seed_data.json` is gitignored, lives in `schemalabz/opencouncil-seed-data`, still carries the old columns, and **every PR preview database is seeded from it**. The old code spread `...city` straight into `createMany`, so one legacy row would throw `Unknown argument 'officialSupport'` and bring the preview up empty. `cityStatusFromDump()` now reads both shapes and applies the same backfill rule.

Verified by running the real seed against a migrated scratch DB with the current legacy dump: 10 Greek cities → `supported`, Belgrade → `demo`. Belgrade is now the demo fixture, so the whole feature is exercised in dev and previews before Niš exists.

## The public wire

`officialSupport` **stays** on `/api/cities/all` and on MCP, derived as `status === 'supported'`, so those consumers don't break — pinned by a test. It is **not** re-added to `/api/cities` (POST response) or `/api/cities/{cityId}`, which are admin-shaped routes returning the raw Prisma row; they now carry `status` and nothing else. `/api/cities/at` likewise returns `status` instead. The unavoidable break everywhere is the `status` enum's *values* changing from `pending|unlisted|listed` to `pending|demo|supported`; `cityStatusSchema` regenerates the OpenAPI spec automatically.

## Also

Renames the GeoJSON `officialSupport` feature property to `borderOnlyHover`. It was presentation state that never came from city data — `useMapFeatures` is its only producer and writes `false` four times — and it would have made the review grep useless. The branch it fed was consequently dead in every one of the seven `<Map>` call sites, so it's deleted rather than left as a false suggestion that some caller sets it.

Nothing here creates Niš; that's an admin step after deploy. Nor does it guard against a realm holding both a `demo` and a `supported` city — that guard, and adding `status` to the public payloads, are the follow-up.

## Verification

`npx tsc --noEmit` clean, lint clean, **1400 unit tests across 104 suites**. Two integration tests fail (`notifications.admin-queries` temporal, `task-handlers` subject IDs) — both reproduce identically on the base branch. Migration and seed re-verified against the scratch PostGIS DB after the review round.

---

> [!NOTE]
> **High Risk**
> Irreversible schema migration with documented lossy backfill can unpublish or newly publish cities; wide changes to public visibility, search, and API enum values affect SEO and integrations.
> 
> **Overview**
> **Replaces the split between `City.status` (`pending|unlisted|listed`) and `City.officialSupport` with one `CityStatus` enum: `pending`, `demo`, and `supported`.** The `officialSupport` column is dropped; semantics live in `src/lib/cityStatus.ts` (`isPublic`, `isCustomer`, `isOutOfNetwork`, `isPetitionable` plus Prisma `WHERE` fragments). **Demo** cities are published everywhere a customer city is (map, directory, search, MCP, sitemap) but without the official-support badge; **supported** is the customer state (badge, ops lists, marketing counts).
> 
> A **destructive, irreversible migration** backfills: `officialSupport = true` → `supported`, everything else → `pending` (so former `listed` without support goes dark; nothing auto-promotes to `demo`). Admin UI drops the official-support checkbox and uses **Demo** / **Supported** status options; `getCities` collapses `includeUnlisted`/`includePending` into **`includeNonPublic`**.
> 
> Call sites across landing, sitemap, petitions, reviews, seed/import scripts, and map behavior are updated (e.g. map coverage uses public status, petition layer stays **pending-only**, populate import leaves cities **`pending`**). **`/api/cities/all`** and MCP still expose **`officialSupport`** derived as `status === 'supported'` (tested). Seed gains **`cityStatusFromDump()`** for legacy gitignored dumps; Belgrade fixture is **`demo`**.
> 
> **Risk:** Enum value changes break consumers that still expect `unlisted`/`listed`; production should be checked for edge backfill cases (`listed` without support, `unlisted` with support) before merge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fc8391650812e85537fae05ce37131c9976dd2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR consolidates city publication and customer-support state into the `pending`, `demo`, and `supported` statuses.
- Adds and migrates the unified city-status model.
- Centralizes visibility, customer, petition, map, search, and administrative filtering through shared status predicates.
- Preserves the legacy `officialSupport` field on compatibility-facing API and MCP payloads by deriving it from status.
- Updates seed/import paths, administration interfaces, landing-map behavior, translations, and tests for the new model.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| prisma/migrations/20260811190506_city_status_demo_supported/migration.sql | Destructively replaces the legacy status/support combination with the unified enum using the documented backfill. |
| src/lib/cityStatus.ts | Centralizes public, customer, out-of-network, and petitionable status semantics. |
| src/lib/db/cities.ts | Updates city visibility and petition-map queries; the previously reported admin filtering and aggregate behavior are resolved or explicitly intentional. |
| src/app/api/cities/all/route.ts | Restores wire compatibility by deriving `officialSupport` from each city's unified status. |
| prisma/seed.ts | Normalizes both legacy and current seed-data shapes into the new city-status enum. |
| scripts/render_brochure_pdf.tsx | Uses the unified customer predicate while retaining alias-based imports. |

<a href="https://app.greptile.com/ide/claude-code?prompt=Greploop%20schemalabz%2Fopencouncil%20PR%20%23613%3A%20work%20through%20Greptile's%20open%20review%20comments%2C%20then%20keep%20reviewing%20and%20fixing%20until%20it%20comes%20back%20clean%20at%205%2F5%20with%20zero%20unresolved%20comments.%0AStart%20by%20reading%20the%20comments%20off%20the%20PR%20itself.%20On%20GitHub%2C%20use%20paginated%20%60gh%20api%20graphql%60%20to%20query%20%60pullRequest.reviewThreads%60%20with%20each%20thread's%20%60isResolved%60%20value%20and%20inline%20%60comments%60.%20%60gh%20pr%20view%20--comments%60%20only%20includes%20conversation%20comments%2C%20so%20do%20not%20use%20it%20as%20the%20findings%20source.%20They%20are%20not%20listed%20in%20this%20prompt%20on%20purpose%3A%20the%20PR%20is%20current%2C%20a%20pasted%20copy%20would%20not%20be.%20Skip%20anything%20already%20resolved%2C%20and%20if%20you%20judge%20a%20comment%20wrong%2C%20say%20so%20rather%20than%20changing%20code%20to%20satisfy%20it.%0A%0APrefer%20the%20Greptile%20CLI%2C%20which%20reviews%20the%20working%20tree%20with%20no%20push%20and%20no%20CI%20run.%20Fall%20back%20to%20PUSH%20LOOP%20only%20where%20a%20step%20below%20says%20to.%0A1.%20Run%20%60command%20-v%20greptile%60.%20Missing%3A%20go%20to%20PUSH%20LOOP%2C%20and%20pitch%20the%20CLI%20at%20the%20end.%0A2.%20Run%20%60greptile%20whoami%60.%20It%20exits%200%20even%20when%20signed%20out%2C%20so%20read%20the%20output%3A%20if%20it%20says%20%22Not%20signed%20in%22%2C%20ask%20the%20user%20to%20run%20%60greptile%20login%60%20%28it%20opens%20a%20browser%2C%20do%20not%20run%20it%20yourself%29%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%0A3.%20Look%20for%20%60greploop%2FSKILL.md%60%20under%20%60.agents%2Fskills%2F%60%20or%20%60.claude%2Fskills%2F%60%2C%20both%20in%20the%20repo%20root%20and%20in%20the%20user's%20home%20directory.%20Found%3A%20follow%20that%20skill%2C%20never%20pushing%20between%20rounds%2C%20and%20you%20are%20done.%0A4.%20Missing%3A%20run%20%60greptile%20skills%20list%60.%20If%20it%20fails%20as%20an%20unknown%20command%2C%20this%20CLI%20predates%20the%20skill%2C%20so%20tell%20the%20user%20to%20run%20%60greptile%20update%60%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%20Otherwise%20ask%20the%20user%20before%20running%20%60greptile%20skills%20install%20greploop%60%2C%20then%20follow%20the%20skill%20it%20writes.%0A%0APUSH%20LOOP%20%28only%20when%20a%20step%20above%20sends%20you%20here%29%3A%0A%0AGREPLOOP%20ENABLED%3A%20After%20fixing%20all%20issues%2C%20commit%20and%20push%20your%20changes.%20Greptile%20will%20automatically%20re-review%20the%20PR%20via%20webhook.%20Wait%20for%20the%20review%20to%20complete%2C%20then%20read%20the%20new%20review%20comments%20and%20confidence%20score.%20If%20the%20confidence%20score%20is%20below%205%2F5%20or%20there%20are%20unresolved%20comments%2C%20fix%20those%20issues%20and%20push%20again.%20Repeat%20until%20the%20confidence%20score%20is%205%2F5%20with%20zero%20unresolved%20comments.%20Maximum%205%20iterations.%20Do%20NOT%20open%20a%20new%20PR%20%E2%80%94%20keep%20pushing%20to%20the%20same%20branch.%0A%0AIF%20YOU%20ENDED%20UP%20ON%20PUSH%20LOOP%20BECAUSE%20THE%20CLI%20WAS%20MISSING%3A%20once%20the%20loop%20is%20done%2C%20print%20the%20block%20below%20and%20nothing%20else%20about%20the%20CLI.%20Reproduce%20it%20character%20for%20character%20inside%20a%20fenced%20code%20block%2C%20so%20the%20borders%20line%20up.%20Do%20not%20reword%20it%2C%20translate%20it%2C%20summarize%20it%2C%20add%20to%20it%2C%20or%20drop%20the%20box.%20Print%20it%20once%2C%20and%20do%20not%20bring%20the%20CLI%20up%20again%20later%20in%20the%20session.%0A%0A%E2%95%AD%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AE%0A%E2%94%82%20%20You%20can%20run%20greploops%20faster%20locally%20with%20our%20CLI.%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%E2%94%82%0A%E2%94%82%20%20Install%20it%20at%20https%3A%2F%2Fwww.greptile.com%2Fcli%2C%20or%20I%20can%20install%20it%20for%20you.%20%20%E2%94%82%0A%E2%95%B0%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AF%0A%0AIf%20they%20take%20you%20up%20on%20it%2C%20install%20with%20%60npm%20install%20-g%20greptile%60%20%28or%20%60brew%20install%20greptileai%2Ftap%2Fgreptile%60%29%2C%20then%20%60greptile%20skills%20install%20greploop%60.%20Leave%20%60greptile%20login%60%20to%20them%2C%20it%20opens%20a%20browser.&repo=schemalabz%2Fopencouncil&pr=613&platform=github">Fix All in Greploop</a>

<sub>Reviews (11): Last reviewed commit: ["feat(city): collapse status and official..."](https://github.com/schemalabz/opencouncil/commit/3fc8391650812e85537fae05ce37131c9976dd2c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52262551)</sub>

**Context used:**

- Knowledge Base — [API documentation (code-first OpenAPI spec)](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/api-documentation.md)


christosporios:
Restacked onto #611. Only import-adjacency conflicts, resolved by keeping both sides. tsc clean, 1384 tests, and the `grep officialSupport` review gate still comes back with nothing but comments and the derived wire field.

christosporios:
Restacked, plus the three real findings from the bots.

**`getCities({ includeUnlisted: true })` was silently dropping every ex-`unlisted` city** from admin meetings, admin people, the Elasticsearch status page and the offer form. With `unlisted` folded into `pending`, both flags mean the same thing, so the filter now applies only when neither is set. Three assertions in `landing.test.ts` had encoded the broken behaviour and now assert the fixed one.

**`/api/cities/all` was dropping `officialSupport`** despite the description promising it stays. I'd derived it for MCP and missed this endpoint. Derived at the response boundary now, so `CityMinimalWithCounts` stays a faithful `Pick`.

**Import style** in `render_brochure_pdf.tsx` fixed to `@/`.

**The petition-map finding I'm not taking.** Both bots say demo cities should appear in the petition results because they're petitionable. They shouldn't: a demo city is already drawn on that map as covered, with pins and an outline, so adding it to the petition layer renders the same municipality twice with opposite meanings. Its petitions are still recorded and still show in the admin expansion table, which splits on `isCustomer`. Added a comment on the query so this doesn't get re-raised.

tsc clean, lint clean, 1382 tests, and the `grep officialSupport` gate still returns only comments.

christosporios:
Good catch — and it's the sibling of the `getCities` bug from the last round. I fixed one auth path and left the other behind.

`GET /api/cities?includeUnlisted=true` dispatches to `getAllCitiesAsServiceKey()` for bearer auth and to `getCities({ includeUnlisted: true })` for a session, and the route comment says they are the same superadmin-equivalent view. After the last fix they were not: the session path returned every status, the service-key path returned only public ones. Two credentials for the same access level, seeing different data.

The old filter was `status IN (listed, unlisted)` — "everything except pending". With `unlisted` folded into `pending`, that same intent now means no status filter at all, so the function drops its `where` entirely. Docstring rewritten to say the two paths must not drift, and why the old filter stopped expressing what it meant.

tsc clean, lint clean, 1382 tests.

prisma/migrations/20260811190506_city_status_demo_supported/migration.sql
```diff
@@ -0,0 +1,44 @@
...
+ALTER TABLE "City" ADD COLUMN "status_new" "CityStatus_new";
+
+-- Migrate existing data.
+-- officialSupport = true -> 'supported' (a customer city: published and badged).
+UPDATE "City" SET "status_new" = 'supported' WHERE "officialSupport" = true;
+
+-- Everything else -> 'pending'. DELIBERATE AND LOSSY: 'unlisted' folds into
+-- 'pending', and today's (status = 'listed', officialSupport = false) cities stop
+-- being published. Nothing is promoted to 'demo' here — that is set by hand
+-- afterwards, one city per realm. There is no reverse migration.
```
kouloumos, line 20: "Set by hand afterwards" makes this a runbook dependency worth writing down before merge: if production's beograd row is `listed` without support, opencouncil.rs is completely empty from the moment this runs until the hand-promotion happens. Suggest putting the exact `UPDATE "City" SET status = 'demo' WHERE id = 'beograd'` in the deploy plan next to the census query.

Related deploy note: migrations run at build time while old pods serve, and the old code raw-selects `officialSupport` in `getMapCitiesCached`/`getCityAtPoint` — so the landing map and `/api/cities/at` will 500 for the window between migrate and cutover. Same class as the previous enum swap, but this one hits the landing page.

src/lib/db/cities.ts
```diff
@@ -476,30 +498,27 @@ export async function getCities({ includeUnlisted = false, includePending = fals
     // Build where clause based on user permissions
     let whereClause: any = {};
 
+    // Visibility is binary now that 'unlisted' is gone: a city is public (demo or
```
kouloumos: After the collapse these two flags produce identical results whenever either is set — the only cell that still differs is `includePending` alone, which no caller uses and which skips the auth check entirely. The docstring warning against drift is the weakest of the available remedies; a single `includeNonPublic` flag that both widens and gates auth makes the drift bug this PR already fixed once unwritable, and it's a mechanical rename of 11 call sites (the comment says ~23, the real count is 11, 6 of which pass a flag).

Related consequence worth a deliberate yes: the `includeUnlisted: true` call sites (admin/meetings, admin/people, ES status, offer form) now return every mass-imported pending stub — ~330 rows on production-shaped data instead of a few dozen. Probably fine, but nobody's seen those dropdowns at that size.

src/lib/db/cities.ts
```diff
@@ -535,18 +554,9 @@ export async function getCitiesWithCouncilMeetings({ includeUnlisted = false, in
...
-        if (!includeUnlisted && !includePending) {
-            statusFilter = 'listed';
-        } else if (!includePending) {
-            statusFilter = { in: ['listed', 'unlisted'] };
-        } else if (!includeUnlisted) {
-            statusFilter = { in: ['listed', 'pending'] };
-        }
-        // If both are true, show all (no filter)
-
+        // Same binary ladder as getCities above.
```
kouloumos: This comment isn't true — `getCities` drops the filter when either flag is set, this one only checks `includePending`, so `{includeUnlisted: true}` authorizes and then still hides the ex-unlisted cities: the exact bug the last round fixed in its sibling.

Before porting the fix here, worth asking if the function should exist: it has zero callers on this branch and on main, and as a `"use server"` export, `{includePending: true}` hands any browser every city with all `councilMeetings` — unreleased included — with no auth check. I'd delete it and `CityWithCouncilMeeting` rather than migrate dead code carrying a live exposure.

src/components/cities/OfficialSupportBadge.tsx
```diff
@@ -33,6 +35,8 @@ export function OfficialSupportBadge({
         lg: 'w-3.5 h-3.5'
     };
 
+    const officialSupport = isCustomer(status);
```
kouloumos, line 38: Confirming intent: the matrix row "badge: –" in practice renders the *negative* badge — verified on the preview by flipping Vrilissia to demo: its page shows "Χωρίς επίσημη υποστήριξη του δήμου" with the petition CTA beside it (Belgrade renders the same in Serbian). That's arguably what keeps a demo honest, but it's a different thing from no badge; if "no badge" was literal, this needs a `null` branch for demo.

src/components/onboarding/selectors/MunicipalitySelector.tsx
```diff
@@ -27,13 +28,13 @@ export function MunicipalitySelector({
         {
```
kouloumos, line 28: Wording cluster to sweep in the follow-up: this `isPublic` group renders under "ΥΠΟΣΤΗΡΙΖΟΜΕΝΟΙ ΔΗΜΟΙ", so a demo city sits under a heading that claims exactly what demo exists not to claim — while its own page badge says the opposite. Same family: `/{id}/petition`'s headline tells a demo city's visitors it "is not yet in the OpenCouncil network" next to a page full of its meetings, and `city-mini-card.tsx:37` still hardcodes an English "unlisted" label for non-public cities.

src/components/landing/v2/hooks/useMapFeatures.ts
```diff
@@ -59,14 +59,14 @@ export function useMapFeatures({
                 });
             }
         }
-        // Blue-gray overlay over the filtered municipality. officialSupport:false keeps the
-        // map's built-in hover from clearing the fill on supported cities.
+        // Blue-gray overlay over the filtered municipality. borderOnlyHover:false keeps
```
kouloumos: The deferral rationale doesn't survive its own grep: no `<Map>` consumer ever wrote this property truthy on main either — all four producers hardcoded `false`, and `undefined` is equally falsy at the only read site. So the `if (borderOnlyHover)` branch was already unreachable for all six consumers before this PR, and deleting it (plus these four writes) is pure dead-code removal, not a shared hover-behaviour change. The kept branch also still carries the old "supported city" comments in map.tsx, which re-import the vocabulary the rename purges. Fine as a follow-up, but it's safe now.

prisma/seed.ts
```diff
@@ -415,21 +415,47 @@ async function seedTopics(topics: any[]) {
...
+ *   current: { status: 'pending' | 'demo' | 'supported' }
+ *
+ * Legacy rows follow the production backfill exactly, so a preview database ends
+ * up with the same cities visible as production. `officialSupport` is the reliable
+ * discriminator: the column is gone, so a regenerated dump cannot carry it.
+ *
+ * Delete the legacy branch once the dump has been regenerated.
+ */
+function cityStatusFromDump(city: { status?: unknown; officialSupport?: unknown }): CityStatus {
+  if (typeof city.officialSupport === 'boolean' || LEGACY_CITY_STATUSES.has(city.status as string)) {
```
kouloumos: The discriminator works, but only because both branches agree at the `'pending'` overlap — a current-shape `{status:'pending'}` row takes the legacy branch. The positive form is strictly simpler and drops the Set, the `typeof`, and the unchecked cast:

```ts
if (status === 'demo' || status === 'supported') return status;
return officialSupport === true ? 'supported' : 'pending';
```

— and "delete the legacy branch" becomes a one-liner. If you want it pinned, extracting this to an import-safe module is what unblocks testing it (seed.ts runs `main()` at import).

src/components/landing/v2/hooks/useMapPopups.tsx
```diff
@@ -245,10 +247,10 @@ export function useMapPopups({
             fetch(`/api/cities/at?lng=${lng}&lat=${lat}`)
                 .then((r) => (r.ok ? r.json() : null))
                 .catch(() => null)
-                .then((city: { id: string; name: string; officialSupport: boolean; geometry: GeoJSON.Geometry } | null) => {
+                .then((city: { id: string; name: string; status: CityStatus; geometry: GeoJSON.Geometry } | null) => {
```
kouloumos: This inline shape (and its sibling in useMapMarkers) is the third hand-copy of the `/api/cities/at` response this PR had to edit — `CityAtPoint` is exported from `@/lib/db/cities` and type-only imports are client-safe (the pattern cityStatus.ts itself documents). `Pick<CityAtPoint, ...>` at both sites makes the next migration a one-file change.

prisma/schema.prisma
```diff
@@ -80,10 +79,23 @@ enum Realm {
...
+/// value changes that sort.
 enum CityStatus {
-  pending   
-  unlisted
-  listed
+  /// Not published anywhere. Petitionable. The default for imported δήμοι.
+  pending
+  /// Published everywhere a customer city is — landing map, directory, search,
+  /// MCP, sitemap — but not a customer: no official-support badge, and still
+  /// petitionable. Only allowed in a realm that has no `supported` city.
```
kouloumos: Until the follow-up guard lands, suggest phrasing this as policy ("one per realm, set by hand") — as written it reads as a system invariant, and nothing enforces it.

src/lib/cityStatus.ts
```diff
@@ -0,0 +1,52 @@
+import type { CityStatus, Prisma } from '@prisma/client';
+
+/**
+ * Predicates over `City.status`, so the meaning of each state lives in one place
+ * instead of being re-derived as a raw comparison at ~40 call sites — which is
+ * how the field this replaced (`officialSupport`) came to mean two different
+ * things depending on who was reading it.
+ *
+ * Type-only Prisma import, so this is safe in client components too.
+ */
```
kouloumos, line 14: These predicates are the load-bearing center of the PR and ship untested — and they're pure, so there's no obstacle. Below is a drop-in `src/lib/__tests__/cityStatus.test.ts`: full truth table, an enum-growth tripwire, the public/out-of-network partition that getCities' "visibility is binary" comment relies on, and where-fragment↔predicate consistency. Each test is mutation-verified — I introduced the six bugs it claims to catch (isPublic loses demo, isCustomer accepts demo, isOutOfNetwork includes demo, isPetitionable excludes demo, PUBLIC_CITY_WHERE loses demo, CUSTOMER_CITY_WHERE→'demo') one at a time and watched it fail on every one. The consistency tests are what catch the last two — the exact drift this header worries about.

One more line worth adding here: a pointer to the raw-SQL twins in `cities.ts` (they point here, but not back).

<details><summary>src/lib/__tests__/cityStatus.test.ts</summary>

```ts
import { CityStatus } from '@prisma/client';
import {
    isPublic,
    isCustomer,
    isOutOfNetwork,
    isPetitionable,
    PUBLIC_CITY_WHERE,
    CUSTOMER_CITY_WHERE,
} from '../cityStatus';

const ALL_STATUSES = Object.values(CityStatus);

describe('cityStatus predicates', () => {
    // The full truth table, one row per status. Every visibility decision in the
    // app reduces to one of these cells; a wrong cell is a wrong page.
    it.each([
        // status        isPublic  isCustomer  isOutOfNetwork  isPetitionable
        ['pending' as const, false, false, true, true],
        ['demo' as const, true, false, false, true],
        ['supported' as const, true, true, false, false],
    ])('%s: public=%s customer=%s outOfNetwork=%s petitionable=%s',
        (status, pub, customer, oon, petitionable) => {
            expect(isPublic(status)).toBe(pub);
            expect(isCustomer(status)).toBe(customer);
            expect(isOutOfNetwork(status)).toBe(oon);
            expect(isPetitionable(status)).toBe(petitionable);
        });

    it('covers every CityStatus value (extend the table when the enum grows)', () => {
        expect(ALL_STATUSES.sort()).toEqual(['pending', 'demo', 'supported'].sort());
    });

    // getCities' "visibility is binary" comment relies on this: every status is
    // either public or out-of-network, never both, never neither. A fourth enum
    // value added without updating the predicates breaks this immediately.
    it('isPublic and isOutOfNetwork partition the enum', () => {
        for (const status of ALL_STATUSES) {
            expect(isPublic(status)).toBe(!isOutOfNetwork(status));
        }
    });

    it('petitionable is exactly the non-customers', () => {
        for (const status of ALL_STATUSES) {
            expect(isPetitionable(status)).toBe(!isCustomer(status));
        }
    });
});

describe('where-clause fragments agree with the predicates', () => {
    // These fragments are spread into Prisma queries at ~10 call sites; if they
    // drift from the predicates, the DB filters one set and the UI another.
    it('PUBLIC_CITY_WHERE selects exactly the statuses isPublic accepts', () => {
        expect([...PUBLIC_CITY_WHERE.status.in].sort())
            .toEqual(ALL_STATUSES.filter(isPublic).sort());
    });

    it('CUSTOMER_CITY_WHERE selects exactly the statuses isCustomer accepts', () => {
        expect(ALL_STATUSES.filter((s) => s === CUSTOMER_CITY_WHERE.status))
            .toEqual(ALL_STATUSES.filter(isCustomer));
    });
});
```

</details>

src/app/api/cities/all/route.ts
```diff
@@ -1,11 +1,17 @@
...
     try {
         const cities = await getAllCitiesMinimal();
-        
-        return NextResponse.json(cities);
+
+        // `officialSupport` is derived here rather than in CityMinimalWithCounts,
+        // which stays a faithful Pick of the model. Its meaning survived the move
+        // to CityStatus unchanged, so third parties keep the field they have.
+        return NextResponse.json(
+            cities.map((city) => ({ ...city, officialSupport: isCustomer(city.status) })),
```
kouloumos, line 13: The compat promise this line implements is unpinned by any test — here's a drop-in route test (mutation-verified: dropping the derivation and counting demo as supported both fail it; the pattern follows `profile/__tests__/route.test.ts`).

<details><summary>src/app/api/cities/all/__tests__/route.test.ts</summary>

```ts
/** @jest-environment node */
jest.mock('@/lib/db/cities', () => ({
    getAllCitiesMinimal: jest.fn(),
}));

import { GET } from '../route';
import { getAllCitiesMinimal } from '@/lib/db/cities';

const mockGetAllCitiesMinimal = getAllCitiesMinimal as jest.MockedFunction<typeof getAllCitiesMinimal>;

describe('GET /api/cities/all', () => {
    beforeEach(() => {
        jest.clearAllMocks();
        jest.spyOn(console, 'error').mockImplementation(() => {});
    });

    // The City model no longer has officialSupport, but third-party consumers of
    // this endpoint keep the field: it must be derived per city as "is a customer".
    it('derives officialSupport on the wire: true only for supported cities', async () => {
        mockGetAllCitiesMinimal.mockResolvedValue([
            { id: 'a', status: 'supported' },
            { id: 'b', status: 'demo' },
            { id: 'c', status: 'pending' },
        ] as unknown as Awaited<ReturnType<typeof getAllCitiesMinimal>>);

        const res = await GET();
        const body = await res.json();

        expect(body.map((c: { id: string; officialSupport: boolean }) => [c.id, c.officialSupport])).toEqual([
            ['a', true],
            ['b', false],
            ['c', false],
        ]);
        // status still present alongside the derived field
        expect((body[0] as { status: string }).status).toBe('supported');
    });
});
```

</details>

src/lib/__tests__/landing.test.ts
```diff
@@ -98,12 +97,10 @@ describe('getCities', () => {
 
         const result = await getCities({ includeUnlisted: true });
 
+        // includeUnlisted now widens to every status: with 'unlisted' folded into
+        // 'pending', a superadmin asking for non-public cities must get them.
         expect(prisma.city.findMany).toHaveBeenCalledWith(
```
kouloumos, line 101: These rewritten assertions are genuinely stronger than what they replaced, but the regular-user ones still use `objectContaining`, which ignores sibling keys: I mutated `getCities` to AND a top-level `status` filter alongside the OR — hiding admins' pending cities — and all 11 tests pass; this superadmin `where: {}` assertion is the only guard for filter-conjunction bugs. Capturing the call and asserting the full where object with `toEqual` closes it.

Also unpinned: the `includePending: true`-alone cell (no auth check, no filter — the comment in getCities relies on it, no test exercises it).

src/lib/db/cities.ts
```diff
@@ -659,8 +669,7 @@ export async function getSupportedCitiesWithLogos(): Promise<Array<{ id: string;
     try {
         const cities = await prisma.city.findMany({
             where: {
-                officialSupport: true,
-                status: 'listed',
+                status: 'supported',
```
kouloumos: Straggler sweep: the review-gate grep hunts the old vocabulary, so it can't see raw comparisons in the new one. Five sites escaped the centralization: this one and `render_brochure_pdf.tsx:53` spell `status: 'supported'` inline in files that import `CUSTOMER_CITY_WHERE`; `CityHeader.tsx:192,218` and `landing.tsx:34` keep raw `'pending'` comparisons that are `isOutOfNetwork` in disguise. All semantically correct today — which is exactly the drift condition the predicates exist to prevent. An `OUT_OF_NETWORK_CITY_WHERE` counterpart would close the set (the SQL twin exists; the Prisma twin doesn't).

```suggestion
                ...CUSTOMER_CITY_WHERE,
```

src/app/api/cities/[cityId]/route.ts
```diff
@@ -89,12 +89,9 @@ export async function PUT(request: Request, props: { params: Promise<{ cityId: s
         }
 
         // Only include admin-only fields if user is superadmin. Realm (tenant)
-        // and language are tenant-level like status/officialSupport, so a
-        // city-scoped admin must not be able to move a city to another realm.
+        // and language are tenant-level like status, so a city-scoped admin must
```
Comment (kouloumos) [inline src/app/api/cities/[cityId]/route.ts:92]: Wire-contract check: `/api/cities` and `/api/cities/{cityId}` lose `officialSupport` entirely — only `/api/cities/all` and MCP got the derivation shim. The OpenAPI schemas followed consistently, so this may be intended; it's just not named in the description the way the `/all` compat is. Worth one sentence there, and a confirmation that nothing external reads these two.

kouloumos (commented):
Ran this through a deep review — code agents across the diff plus runtime verification on the preview: drove Vrilissia through supported→demo→pending→supported via the real PUT flow, checked both realms' landing/directory/city pages, petition flows, sitemap realm-scoping, and the wire compat on `/api/cities/all` and MCP. The model holds up everywhere I could reach it: demo publishes without claiming support, pending goes dark but keeps resolving, revalidation fires on every flip.

One thing to settle before merge: **the cache-versioning story contradicts itself.** The rationale comment in `cities.ts` says a warm data cache would keep serving pre-migration rows, and eight keys got `-v2` bumps for it — but `getCityCached` (`['city', id, 'basic']`, `src/lib/cache/queries.ts:48`) and `getAllCitiesMinimalCached` (`['cities','all',realm]`, `:161`) were left unversioned, both TTL-less, and both store rows that embed the old enum literals. A stale `status:'listed'` row makes every new predicate misread it: Athens's own page would show the "no official support" badge plus a petition button until someone edits Athens. The counterweight: `cache-handler.mjs` namespaces the prod cache per `SOURCE_COMMIT`, so a normal deploy starts cold and *none* of the nine keys need bumps. Either the namespace is trusted — then I'd drop the eight `-v2` bumps and the rationale comment — or it isn't for some flow you care about (hand-applying the migration ahead of a deploy?), and then these two keys plus `geohashCity` need the same treatment. Half-mitigating an asserted threat is the one option I'd rule out.

Also ran your migration census on staging: zero rows in both classes you flagged (`listed` without support, `unlisted` with support) — production still needs the same check.

Small stuff with no diff line to hang a comment on: `scripts/import_offers.ts:13` still documents the deleted promote step; `src/lib/openapi/routes/cities.ts:86-97` descriptions still say "listed/unlisted"; `src/components/admin/elasticsearch/Status.tsx:143` labels its toggle "Show Unlisted Cities"; `src/components/cities/CityForm.tsx:26` keeps a now-unused `Checkbox` import; `prisma/import_administrations.ts:136` has a stray-indented closing brace from the deleted line; and `city-ranking-table.tsx:8-10` / `landingData.ts:6,16` each gained a second `@prisma/client` import line next to an existing one.

christosporios:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/da52e762a3844d76ea6eb37985038f880b816fe9..982166eb3e4c02591a7d679e2ff63f98c2a8afea) with the review round, rebased onto `main` now that #610 and #611 have merged. Migration and seed re-verified against the scratch PostGIS DB afterwards, since both files changed. tsc clean, lint clean, 1400 tests / 104 suites.

### Changed

- **`getCities`' two flags collapsed into one `includeNonPublic`.** You were right that the docstring was the weakest available remedy — the two flags produced identical results in every cell a caller used, and the only distinct cell (`includePending` alone) skipped the auth check. One flag now both widens and gates. Six call sites updated; the HTTP `?includeUnlisted=true` query parameter is kept as-is, since that one is a published contract rather than an internal argument.
- **`getCitiesWithCouncilMeetings` deleted** rather than fixed. Zero callers here and on `main`, and as a `"use server"` export it handed any browser every city with its unreleased meetings, with no auth check on the path that mattered. Its type went with it.
- **The five stragglers centralized.** `cities.ts:640`, `render_brochure_pdf.tsx:53`, `CityHeader.tsx:192,218`, `landing.tsx:34` — all now go through `CUSTOMER_CITY_WHERE` / the new `OUT_OF_NETWORK_CITY_WHERE`. Added a pointer in the `cityStatus.ts` header to the raw-SQL twins in `cities.ts`, since those are the ones no grep in the new vocabulary will find either.
- **`borderOnlyHover` branch deleted.** Your grep beats my rationale: no consumer ever wrote it truthy on `main` either, so the branch was unreachable before this PR too. Four writes and the branch are gone.
- **Seed discriminator in the positive form**, exactly as written. The `Set`, the `typeof` and the cast are gone.
- **`Pick<CityAtPoint, …>` at both `/api/cities/at` consumers**, replacing the hand-copied shapes.
- **Schema comment rephrased as policy** ("one per realm, set by hand") rather than as an invariant.
- **All three test files added**, effectively verbatim — plus an `OUT_OF_NETWORK_CITY_WHERE` consistency case in the first, since that fragment is new in this round. `landing.test.ts` moved to a full `toEqual` on the where object; I re-ran your mutation (ANDing a top-level status filter alongside the OR) and it now fails three tests where it previously passed all eleven. Same for the `/all` route test: dropping the derivation fails both cases.
- **The `-v2` cache-key suffixes are gone.** `cache-handler.mjs` namespaces every key as `oc:${envKey}:${buildKey}:` where `buildKey` is `SOURCE_COMMIT || BUILD_ID || buildId` — so a deploy already gets a fresh keyspace and the suffixes bought nothing. The description's claim about them went too.

### Confirming, not changing

- **The negative badge on `demo`.** Intended — "no badge" in the matrix meant "not the *support* badge", and a demo city saying "χωρίς επίσημη υποστήριξη" next to its petition CTA is the honest reading. Matrix row relabelled in the description so it stops implying a literal `null`.
- **The wording cluster** ("ΥΠΟΣΤΗΡΙΖΟΜΕΝΟΙ ΔΗΜΟΙ", the petition headline, `city-mini-card`'s English "unlisted") — agreed on all three, deferring to the follow-up with the guard. They're copy, they don't gate the migration, and they read better decided together.
- **`/api/cities` and `/api/cities/{cityId}` losing `officialSupport`** is intended: those return the raw admin-shaped row, and only `/all` and MCP have external consumers worth a shim. Now named explicitly in the description, in its own "public wire" section.
- **The petition-map exclusion** (Greptile's remaining flag) is deliberate and commented at both queries: a demo city is already drawn on that map as covered, so adding it to the petition layer would render the same municipality twice with opposite meanings. The description's matrix now splits "petition route open" from "petition map layer" so the table stops reading as a contradiction.

### Deliberately not addressed

Christos' call on both: the opencouncil.rs-goes-empty window doesn't matter because nothing points at opencouncil.rs yet, and the migrate-to-cutover 500 window on the landing map is acceptable — same class as the previous enum swap, and we take it.

src/lib/db/citiesAdmin.ts
```diff
@@ -16,22 +16,28 @@ const CITY_COUNT_SELECT = {
...
 
 /**
- * Returns listed + unlisted cities (superadmin-equivalent view) without consulting
- * the session. The caller is responsible for authenticating the request out-of-band
- * (e.g. via a service API key at the API-route layer).
+ * Every city regardless of status — the superadmin-equivalent view, without
+ * consulting the session. The caller authenticates the request out-of-band (a
+ * service API key at the API-route layer).
+ *
+ * Must return exactly what `getCities({ includeUnlisted: true })` returns for a
```
kouloumos: Stale after the rename: this says `getCities({ includeUnlisted: true })`, and the argument is now `includeNonPublic`. Worth keeping exact here specifically — this is the docstring whose whole job is to state that the bearer path and the session path must return the same set.

prisma/seed.ts
```diff
@@ -415,21 +415,43 @@ async function seedTopics(topics: any[]) {
...
 }
 
+/**
+ * A dumped city's visibility, as the current CityStatus.
+ *
+ * The dump lives in a separate, gitignored repo (schemalabz/opencouncil-seed-data)
+ * and lags this schema — and PR preview databases are seeded from it — so this has
+ * to read both shapes:
+ *   legacy:  { status: 'pending' | 'unlisted' | 'listed', officialSupport: boolean }
+ *   current: { status: 'pending' | 'demo' | 'supported' }
```
kouloumos, line 425: The paragraph below still says `officialSupport` is "the reliable discriminator", but the function now checks `status ∈ {demo, supported}` first and only falls through to `officialSupport` — that sentence describes the previous strategy. Both go away when the dump is regenerated, so it's only worth a line now if you're touching the file anyway.

src/lib/db/cities.ts
```diff
@@ -463,43 +478,44 @@ export async function getAllCityIds(realm?: Realm): Promise<string[]> {
...
 
     // Validate permissions
-    if (includeUnlisted && !currentUser) {
-        throw new UnauthorizedError("Not authorized to view unlisted cities");
+    if (includeNonPublic && !currentUser) {
+        throw new UnauthorizedError("Not authorized to view non-public cities");
     }
 
     // Build where clause based on user permissions
     let whereClause: any = {};
```
kouloumos: `any` survived the rewrite. Pre-existing, but the function is effectively new now and `Prisma.CityWhereInput` types it without any other change.

kouloumos (approved):
Everything from the last round is in, and the two items that needed a decision rather than a fix both got one.

The mutation that motivated the test tightening now fails three tests where it previously passed all eleven. The deleted helper has no remaining references anywhere, the hover branch and its four writes are gone, and on the preview an anonymous request for non-public cities gets a 401 where the cell you removed would have answered it. tsc clean, 1400 tests across 104 suites, migration unchanged and still the newest one.

Collapsing to a single auth-gating flag is the right call — it closes the cell that had no auth check, which was the part that actually worried me. Dropping the cache suffixes rather than adding more is also the right read: the per-build namespace already covers it, and half-versioning was the only outcome worth avoiding. Greptile's remaining flag on the petition aggregates I'd leave as it stands — the double-draw reasoning holds.

Does this cover demoing a candidate δήμος in Greece? The scenario I have in mind: we process a demo meeting for a prospect and want to show them their own city on the map during a presentation. As far as I can tell `demo` is now the only way to do that — the map query is public-only with no auth branch, so there's no logged-in preview — and flipping a Greek city to `demo` and back works today, immediately, since no guard exists yet. Worth noting it wasn't possible at all before: map pins required `officialSupport = true`, which also badges the city as officially supported.

Three nits inline, none blocking.

christosporios:
[Force-pushed](https://github.com/schemalabz/opencouncil/compare/982166eb3e4c02591a7d679e2ff63f98c2a8afea..72390a80bc8f5364c5aae222d20ca625d67d5ea0) — all three, they're right.

- **`citiesAdmin.ts` docstring** now says `getCities({ includeNonPublic: true })`. Agreed on why this one matters more than the usual stale reference: it's the docstring asserting that the bearer path and the session path return the same set, so a name that no longer resolves is exactly the kind of drift that let the two diverge in the first place.
- **The seed comment's "reliable discriminator" sentence** rewritten to describe what the function does now — only the new values match positively, and everything else, including a legacy `'listed'` or `'unlisted'`, falls through to `officialSupport`. Same conclusion (a regenerated dump can't carry the column, so the fallthrough is safe), stated as the reason the *fallthrough* is safe rather than as the discriminator.
- **`whereClause` is `Prisma.CityWhereInput`.** Worth noting this one wasn't free: the `any` was hiding a real looseness. `administers.filter(a => a.cityId).map(a => a.cityId)` is `(string | null)[]`, so typing the variable fails to compile. Correct at runtime — the filter guarantees it — but the compiler couldn't see it. Swapped to a map-then-narrow so the type is earned rather than asserted:

  ```ts
  const administerableCityIds = currentUser?.administers
      .map(a => a.cityId)
      .filter((id): id is string => id !== null) ?? [];
  ```

tsc clean, lint clean, 1400 tests / 104 suites.

kouloumos:
LGTM!

PR #614, opened 2026-08-12
test(i18n): guard the catalogs against the code that reads them

kouloumos:
Follow-up to #610, which shipped `metadata.people.author` — a key `people/page.tsx` asks for and no catalog defines. next-intl does not throw on a miss. It logs and returns the dotted path, so production served `<meta name="author" content="metadata.people.author">` in every locale until a reviewer caught it.

`translations.test.ts` could not have caught it. It compares the four catalogs to each other, and the key was missing from all four equally. Nothing checked the catalogs against the *code*. That is the axis this adds.

### What it does

A static scan over `src/**/*.tsx?` resolves each `t()` call to the namespace its translator is bound to, then checks the key against `messages/en`. The other catalogs are pinned to `en` by `translations.test.ts`, and `sr-Latn` is derived from `sr`.

A static scan rather than a render: the keys are spread over ~200 components, and most need a request scope, a database or a map canvas to render. The suite runs in ~40 ms with no mocks.

| | |
|---|---|
| keys checked against the catalog | 2100 |
| computed groups found | 29, across 46 call sites |
| computed groups with a member check | 5 |
| calls the scan reads but cannot attribute | 12 |

### How a translator is resolved

Four binding shapes are in use. Each one needs its own rule.

| shape | example | rule |
|---|---|---|
| one binding per component scope | `mcp/page.tsx`, `SpeakerSegment.tsx` | the nearest **preceding** binding of the same name |
| one binding, called above itself | `og/route.tsx` binds `t` once to `og`, then passes it into seven image components declared above the binding | a name bound exactly once applies wherever it is called |
| several bindings in one array pattern | `const [t, tAccount] = await Promise.all([getTranslations("Profile"), getTranslations("account")])` | pair the names with the array elements by position |
| no binding — a shared helper takes the translator as its `t` parameter | `subjects.ts`, `meetingPages.ts`, `roles.ts`, `administrativeBodies.ts`, `time.ts`, `AdminSection.tsx` | an `@translationNamespace <ns...>` JSDoc tag, which the scan reads as a binding |

Nearest-preceding alone is not enough, and a whole-file map is worse. `List.tsx` takes `t` as a prop while it binds the hook to `tCommon`. A per-file map reports seven healthy keys as missing and loses the real one.

Five of the six helpers already stated their namespace, in prose:

> Pass a translator scoped to the `Subject` namespace (e.g. `useTranslations('Subject')`).

The contract was there. Only the machine could not read it. The tag sits next to that prose. It is checked against the catalog; nothing checks that the callers honour it, which is the same trust the prose asked for.

### Computed keys

A key such as ``t(`audioFailure.${reason}`)`` has a leaf that only exists at runtime. The scan checks the prefix, which catches the group being renamed or dropped. It says nothing about the members, and a member is what renders.

`COMPUTED_GROUPS` therefore lists all 29 groups. Each one either names the type that supplies its members, or records why no type states the set. A new ``t(`group.${x}`)`` fails the test until someone makes that call. A second assertion drops a registration once its call site is gone.

Five groups have a member check. `AdministrativeBodyType` is the reason the Prisma schema is a member source: it is an enum that feeds three catalog groups in three namespaces across six call sites, and a schema migration is the realistic way it drifts.

The other 24 carry a reason. Most take their leaf from a const array beside the call, which no type describes. One is deliberate: `runSummarize.unavailable.` has three union members and two catalog entries, because the branch only runs when the value is not `available`.

### The gap this found

`formatDateRange` reads `from`, `until` and `present`. Person.tsx and Party.tsx each passed their own translator, so the same three words had to exist under both namespaces. `Person` held two of them. `Party` held none.

This is latent, not live. The helper returns an empty string when both dates are null, and the roles on the pages checked carry no dates, so no live page shows a leak today. A role with a start date, which the admin role editor sets, makes the copy render as literal text.

`d7b51def` fixes the cause rather than the symptom: copy that belongs to a shared helper belongs to a shared namespace, so the three keys move to `Common` and both call sites pass `tCommon`. `Person.from` and `Person.until` are reachable only through this helper, so `Common` replaces them. `RolesList.present` has no reader at all and goes with them. Three keys in one namespace, in place of nine spread over three.

RolesList.tsx keeps a second, divergent copy of this formatter — a different date format, and it returns `null`. Merging it changes what the admin role editor renders, so that belongs in its own change.

### Mutations

Every assertion was checked against the bug it claims to catch.

| mutation | result |
|---|---|
| `metadata.people.author` removed from all four catalogs (= what shipped) | fails, names the key and `people/page.tsx:42` |
| a typo in `og/route.tsx` — single binding, called above itself | fails |
| a typo in `profile/page.tsx` — `Promise.all` binding | fails |
| a typo in `subjects.ts` — annotated helper | fails |
| `Common.present` removed again | fails |
| a 5th `AudioExportFailure` member with no copy | fails |
| a 4th `AdministrativeBodyType` value in `prisma/schema.prisma` | fails, all three groups |
| a new ``t(`group.${x}`)`` nobody registered | fails, and also when its prefix already exists |
| a registration whose call site is gone | fails |
| a second `t` binding in `og/route.tsx` | fails the cap on unattributable calls |
| clean tree | 10 assertions pass |

The cap on unattributable calls is absolute (25, against 12 today) rather than a share of the total. A proportional cap grows with the codebase, so it can only catch a collapse, never the loss of one file.

### Not covered

- 12 calls on a translator whose namespace is assembled at runtime — `useTranslations(`about.demos.map.byRealm.${scenario}`)`. Nothing static resolves it.
- 24 computed groups whose leaf comes from a const array beside the call. Each carries a written reason instead of a check.
- `t.has()`, which is the deliberate "might not exist" escape hatch.
- A key present in `en` and missing elsewhere. That stays `translations.test.ts`'s job.

### Notes

Source changes: the copy move in `d7b51def` (four catalogs, plus a `tCommon` binding in Person.tsx and Party.tsx). The rest is the test plus six JSDoc tags.

A tag may name several namespaces, and then the key must exist under every one. No tag needs that today. One that did would be worth a second look — it reports that two callers scope the same helper differently, so one string is being kept in two places. `formatDateRange` read that way until its copy moved.

<h3>Greptile Summary</h3>

The PR adds a static catalog-reference test and moves shared date-range labels into the `Common` namespace.

- Scans TypeScript translation calls and checks attributed keys against the English catalog.
- Registers computed translation groups and validates selected groups against unions or Prisma enums.
- Adds namespace annotations for shared translation helpers.
- Updates person and party pages to use the shared date-range translations.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/__tests__/translation-key-references.test.ts | Adds static checks for literal and computed translation-key references, translator bindings, and catalog membership. |
| src/lib/formatters/time.ts | Documents `formatDateRange` as using the shared `Common` translation namespace. |
| src/components/persons/Person.tsx | Uses a `Common` translator when formatting role date ranges. |
| src/components/parties/Party.tsx | Uses the existing `Common` translator when formatting former-member date ranges. |
| messages/en.json | Moves date-range labels into the shared `Common` namespace. |
| messages/el.json | Moves date-range labels into the shared `Common` namespace. |
| messages/fr.json | Moves date-range labels into the shared `Common` namespace. |
| messages/sr.json | Moves date-range labels into the shared `Common` namespace. |

<sub>Reviews (5): Last reviewed commit: ["test(i18n): cap unattributable calls at ..."](https://github.com/schemalabz/opencouncil/commit/2c776663e0354b3268656d4057941893ade830df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52488278)</sub>

**Context used:**

- Knowledge Base — [Application presentation system](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/application-presentation-system.md)


kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/fa43278c4ddb5e5d6014342ce3a8c49a60ee4cf8..29126d16e4fcf68eccd657c8bd740eee7472146d) to rebase onto `main` and close the coverage gaps a second review found. The scan read 2029 keys and skipped 35; it now reads 2100 and skips 12.

- Three binding shapes it could not resolve are handled: a name bound once but called above its binding (`f4072307` — 21 of the recovered calls are in `og/route.tsx`, where a leaked path renders into a PNG), an array pattern over `Promise.all` (`422bd1b3`), and a shared helper that takes the translator as a parameter (`0e40a69b`).
- `COMPUTED_GROUPS` (`42e0a7c4`) replaces the one-entry union table. All 29 groups are listed — five with a member check, 24 with a reason. A new computed group fails the test until it is registered, and a registration fails once its call site is gone.
- `107a807b` adds four catalog keys the new coverage exposed: `Party.from`, `Party.until`, `Party.present` and `Person.present`. Latent, not live — `formatDateRange` returns an empty string when both dates are null, and the pages I checked show no leak today.
- The cap on unattributable calls is absolute rather than a share of the total (`29126d16`), so losing one file to a second binding fails instead of passing quietly.

Assertions go from 4 to 10. The mutation table in the body covers each one.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/29126d16e4fcf68eccd657c8bd740eee7472146d..2c776663e0354b3268656d4057941893ade830df) to fix the cause of the date-range gap instead of the symptom.

The previous revision added `Party.from`, `Party.until`, `Party.present` and `Person.present`, which left the same three words in three namespaces. `d7b51def` moves them to `Common` and passes `tCommon` at both call sites. `Person.from` and `Person.until` are reachable only through `formatDateRange`, so `Common` replaces them rather than joining them, and the unused `RolesList.present` goes too — three keys in one namespace in place of nine over three.

The annotation on the helper collapses from `@translationNamespace Person Party` to `Common`. That is the property worth keeping: a tag naming two namespaces reports that one string is being kept in two places.

RolesList.tsx still holds a divergent copy of the formatter — a different date format, returns `null`, no `present` branch. Merging it changes what the admin role editor renders, so I left it out.

Same eight commits, each green on its own. Assertions unchanged at 10; `Common.present` removed reproduces the gap.

Issue #625, opened 2026-08-14
Improve search quality: relevance floor, phrase search, and the /search experience


kouloumos:
### Concept

The team says that search "is not good", but we never wrote down what is wrong. Because of this we cannot hand anyone a starting point. This issue records reproducible examples, and it defines the work that follows from them.

Public search fuses two retrievers with RRF. A keyword retriever runs over the Greek-analysed fields. A semantic retriever runs over the `name.semantic` and `description.semantic` embeddings. Semantic retrieval returns its nearest neighbours always. It has no concept of "no match". Every query therefore fills every page. When the number of real matches is small, the page fills with unrelated subjects.

### User story

As a citizen, I want search to show me only the subjects that relate to my query, and to tell me when it finds nothing, so that I can trust the results on the page.

### Evidence

All four examples reproduce against the production index today.

**1. A nonsense query returns 115 results.** Query: `διαστημόπλοιο κοάλα` ("spaceship koala"). Page one holds `Πρόστιμο KOLA DRITAN`, `Επανεξέταση απόφασης ΚΟΛΩΝΑΚΙ ΕΣΤΙΑΣΗ` and `Διάσωση συγκροτήματος Columbia`. Run each retriever alone: the keyword retriever returns **0** hits, the semantic retriever returns **115**.

**2. A real query goes to noise inside page one.** Query: `υδρονομείς`. The keyword retriever finds **7** relevant subjects. The interface reports **104**. Positions 5 to 10 already show `Γνωμοδότηση για υδροηλεκτρικό έργο`, `Υποστελέχωση δασικών υπηρεσιών` and `Προβλήματα αντιπλημμυρικών έργων Πειραιά`.

**3. A qualifying word disappears.** Query: `κολυμβητήριο Μαρουσίου`. The top ten hold pools in Γουδί, Νέα Χώρα, Άργος and Χανιά. Μαρούσι is not one of the municipalities that we cover, so the correct answer is that we hold nothing for that qualifier. Instead the page fills with pools from four other municipalities. The `multi_match` clause uses `operator: 'or'`, so `Μαρουσίου` stays optional.

**4. A councillor surname returns unrelated subjects.** Query: `Μαλτέζος`. The first hit is `Μίσθωση οικοπέδων για χώρους στάθμευσης Μύλοι`.

### Two measurements that constrain the work

A score threshold cannot supply a relevance floor today. Two measurements show why.

[...]

PR #628, opened 2026-08-14
feat(elastic): index administrative body and discussion duration metrics

pkakelas:
## Why
Search ranks a subject by text relevance alone. A subject that the council debated for
an hour ranks the same as a passing mention. Elasticsearch holds no signal about the
size of a discussion, and no signal about the kind of meeting.
This PR indexes the signals. A follow-up PR uses them to rank the results.

This is the first out of the many PRs solving [625](https://github.com/schemalabz/opencouncil/issues/625)

## What
Three new fields on the `subjects` index:
| Field | Source |
|-------|--------|
| `contributor_count` | Number of `SpeakerContribution` rows on the subject |
| `discussion_speaking_seconds` | Time the council spent speaking about the subject |
| `administrative_body_id` / `administrative_body_type` | The body of the meeting: council, committee, or community |

`SubjectSearchView` precomputes the two metrics as scalar columns, because a nested
array costs a nested query to aggregate at search time. `discussion_speaking_seconds`
repeats `getDiscussionSecondsForSubjects()` in `src/lib/db/subject.ts`, so the index
agrees with the number the subject page shows: tagged `SUBJECT_DISCUSSION` utterances
first, procedural segments excluded, and the legacy `SubjectSpeakerSegment` table only
as a fallback.
The new `MeetingAdministrativeBodyView` flattens the `Subject` → `CouncilMeeting` →
`AdministrativeBody` join, which PGSync cannot express as one relationship.
No application code reads the new fields yet. The search API exposes no filter on them.

**Caution: We need re-indexing on each env after merging this PR**

<h3>Greptile Summary</h3>

The PR adds administrative-body metadata, city realm, contributor counts, and discussion speaking time to subject search documents.
- Adds Elasticsearch mappings and PGSync relationships for the new scalar fields.
- Adds PostgreSQL helper views and validation queries for metrics, administrative bodies, and realms.
- Moves utterance discussion-status persistence into the subject transaction so subject writes and indexed metrics are synchronized.
- Documents schema migration, backfill, verification, and refresh behavior.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| elasticsearch/schema.json | Adds keyword and numeric mappings plus PGSync relationships for realm, administrative-body metadata, and subject metrics. |
| elasticsearch/views.sql | Adds helper views that compute subject metrics and flatten administrative-body and city-realm data. |
| elasticsearch/validate-views.sql | Extends view validation to cover metric integrity, administrative-body enum values, and city realms. |
| src/lib/db/utils.ts | Validates, deduplicates, batches, and transactionally persists summarize-task utterance discussion statuses. |
| src/lib/tasks/summarize.ts | Passes discussion statuses into the transactional subject-save operation and removes the later per-utterance persistence pass. |
| elasticsearch/README.md | Documents the new indexed fields and operational schema-deployment, backfill, and refresh requirements. |

<sub>Reviews (7): Last reviewed commit: ["docs(elastic): document the deploy path ..."](https://github.com/schemalabz/opencouncil/commit/e21aae0d7272850ec3d57635b9cfb6c8c7b464c4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53347278)</sub>

### DIFF
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -164,6 +164,26 @@ PGSync requires helper views to denormalize complex relationships and handle Pos
 4. **SpeakerContributionSearchView** - Denormalizes speaker contributions
 5. **SubjectMetricsView** - Precomputes the per-subject discussion metrics
 6. **MeetingAdministrativeBodyView** - Flattens the Subject → CouncilMeeting → AdministrativeBody join and casts the type enum to text. Exposes only the id and the type, because search results hydrate the full body from PostgreSQL
+7. **CitySearchView** - Casts the `Realm` enum to text. Every other `City` column reaches the index as a direct column
+
+### Realm, Not Country
+
+The index stores `city_realm` (`greece`, `france`, `cyprus`, `serbia`), not an ISO country code.
+
+PostgreSQL holds no country column. `REALMS` in `src/lib/realm.ts` maps a realm to its country, and
+`getRealmCountry()` reads that map. A `CASE` expression in SQL would copy the map into a second place
+that no test covers, so a new realm would sync an empty or wrong country. A caller that needs the
+country resolves it from `city_realm` in application code.
+7. **CitySearchView** - Casts the `Realm` enum to text. Every other `City` column reaches the index as a direct column
+
+### Realm, Not Country
+
+The index stores `city_realm` (`greece`, `france`, `cyprus`, `serbia`), not an ISO country code.
+
+PostgreSQL holds no country column. `REALMS` in `src/lib/realm.ts` maps a realm to its country, and
+`getRealmCountry()` reads that map. A `CASE` expression in SQL would copy the map into a second place
+that no test covers, so a new realm would sync an empty or wrong country. A caller that needs the
+country resolves it from `city_realm` in application code.
 
 ### Discussion Metrics
 
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -161,6 +161,40 @@ PGSync requires helper views to denormalize complex relationships and handle Pos
 1. **LocationSearchView** - Converts PostGIS geometry to GeoJSON format
 2. **IntroducedByPartyView** - Resolves party affiliation through the `Role` table
 3. **SubjectSpeakerSegmentSearchView** - Denormalizes speaker segments with concatenated utterances
+4. **SpeakerContributionSearchView** - Denormalizes speaker contributions
+5. **SubjectMetricsView** - Precomputes the per-subject discussion metrics
+6. **MeetingAdministrativeBodyView** - Flattens the Subject → CouncilMeeting → AdministrativeBody join and casts the type enum to text. Exposes only the id and the type, because search results hydrate the full body from PostgreSQL
+
+### Discussion Metrics
+
+`SubjectMetricsView` precomputes two scalar fields, because a nested array costs a nested query to
+aggregate at search time. They exist for score rescoring. The search API exposes no filter on them:
+
+| Field | Meaning |
+|-------|---------|
+| `contributor_count` | Number of `SpeakerContribution` rows on the subject |
+| `discussion_speaking_seconds` | Time the council spent speaking about the subject |
+
+`contributor_count` counts rows, the same measure as `getContributionCount()` in `src/lib/utils.ts`.
+The count comes from `SpeakerContribution`, the model that replaces `SubjectSpeakerSegment`, so a
+subject that predates the contribution pipeline reports zero contributors.
+
+`discussion_speaking_seconds` repeats `getDiscussionSecondsForSubjects()` in `src/lib/db/subject.ts`, so
+the index agrees with the number the subject page shows:
+
+- **Primary source**: utterances tagged `SUBJECT_DISCUSSION`. The summarize task writes these tags.
+- **Excluded**: procedural segments, which are not part of a discussion.
+- **Fallback**: `SubjectSpeakerSegment`, and only for a subject with no tagged utterance at all. A
+  subject whose tagged utterances are all procedural reports 0 instead of falling back.
+
+Both sources sum the parts rather than measure the span from the first mention to the last, so a
+subject that the council revisits reports the time spent on it. The name avoids the word "duration"
+because `calculateMeetingDurationMs()` uses it for a wall-clock span.
+
+> **Never name the root table in `base_tables`.** `SubjectMetricsView` is keyed on the subject id, so
+> it is tempting to declare `base_tables: ["Subject"]`. That is what `SubjectSearchView` did, and it
+> made PGSync route `Subject` DELETE events as child re-syncs, so deleted subjects stayed in the
+> index. Declare only the tables the view actually reads.
 4. **SpeakerContributionSearchView** - Denormalizes speaker contributions with party resolution
 
 ### Create the Views
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -11,6 +11,7 @@ This document describes how OpenCouncil uses Elasticsearch to provide powerful s
 4. [Configure PostgreSQL Views](#configure-postgresql-views)
 5. [Configure the Ingest Pipeline](#configure-the-ingest-pipeline)
 6. [Set up PGSync](#set-up-pgsync)
+   - [Deploying a Schema Change](#deploying-a-schema-change)
 7. [Sync Data](#sync-data)
 8. [Search Examples](#search-examples)
 9. [Best Practices & FAQ](#best-practices--faq)
@@ -300,7 +301,9 @@ Expected output: `"description": "Ενημέρωση από τον Δήμαρχ�
 
 ### Schema Configuration
 
-> **Important**: The sync configuration is defined in `elasticsearch/schema.json`. Changing this schema effectively changes the structure of documents in Elasticsearch and requires re-indexing the entire index.
+> **Important**: The sync configuration is defined in `elasticsearch/schema.json`. A change to this file
+> changes the structure of the documents in Elasticsearch. Some changes need a new index; most do not.
+> See [Deploying a Schema Change](#deploying-a-schema-change).
 
 The `elasticsearch/schema.json` file defines:
 - **Index mapping**: Elasticsearch field types, analyzers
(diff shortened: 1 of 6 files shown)


elasticsearch/README.md
```diff
@@ -716,6 +859,20 @@ A: PGSync's live sync receives WAL events with the base table's original column
 
 This ensures bootstrap (reads from view) and live sync (reads from WAL) both work correctly.
 
+**Q: When do the discussion metrics refresh?**  
+A: PGSync rebuilds the whole document when it processes a change, so the metrics recompute on the next
+re-sync of the parent `Subject`. A change to a child table does not always trigger that re-sync (see
+[Testing Live Sync](#testing-live-sync-important)). The aggregates read `SpeakerContribution` and
+`Utterance`, which the view does not declare as base tables, so they refresh through the sibling nodes that
```
kouloumos: This is not what happens, and the sibling nodes are already declared the way this describes: `speaker_contributions` lists `SpeakerContribution`, `speaker_segments` lists `Utterance`, and the re-sync still does not fire. Adding those tables to `SubjectSearchView` would not help either — its primary key is a subject id, while a contribution's WAL event carries the contribution's own id, so it could never match.

With the daemon running, tagging an utterance with no accompanying subject write leaves the document behind: Postgres reported 60 seconds, the document stayed at 0, and the counters read `Db: [4] => Elasticsearch: [1]` — the single write being an unrelated transaction that did touch a subject. So `contributor_count` is fine, because contributions are written in the same transaction as their subject, and `discussion_speaking_seconds` is the one that goes stale.

The cause is a commit boundary rather than anything in this PR. `saveSubjectsForMeeting` wraps subjects, contributions and highlights in one transaction, which is exactly why the contribution count is correct. The utterance tagging is the one piece that runs after that transaction commits, so nothing writes `Subject` again and the document keeps its pre-tagging numbers. The summarize task is the only writer that sets those tags, so moving the tagging inside the same transaction fixes it structurally — no parent touch, no trigger. Grouping the updates by `(subjectId, discussionStatus)` makes that cheap: on staging it collapses ~2,456 statements per meeting into ~46.

One judgement call comes with it. Tagging failures are currently swallowed per utterance, so a partly-tagged meeting still keeps its saved subjects; inside a transaction a tag failure would roll the subjects back too. Worth deciding deliberately.

Whichever way that lands, this answer needs rewording to match.

elasticsearch/schema.json
```diff
@@ -379,6 +397,58 @@
...
+            "foreign_key": { "parent": ["id"], "child": ["id"] }
+          }
+        },
+        {
+          "table": "MeetingAdministrativeBodyView",
+          "schema": "public",
+          "base_tables": ["CouncilMeeting", "AdministrativeBody"],
+          "primary_key": ["id", "cityId"],
+          "columns": ["administrative_body_id"],
+          "label": "administrative_body_id",
```
kouloumos: **This node can drop the view entirely.**

Look at what the view does for this column:

```sql
SELECT cm.id, cm."cityId",
       ab.id AS administrative_body_id,      -- <- this one
       ab.type::text AS administrative_body_type
FROM "CouncilMeeting" cm
LEFT JOIN "AdministrativeBody" ab ON ab.id = cm."administrativeBodyId";
```

The join condition is `ab.id = cm."administrativeBodyId"`, so `ab.id` is by definition equal to `cm."administrativeBodyId"` — a column that already sits on `CouncilMeeting`. The join cannot change the value, only reproduce it. (It could in principle yield NULL for a dangling reference, but that is a real foreign key, so it cannot dangle.)

Meanwhile `CouncilMeeting` is already a node in this schema — `meeting_released` reads `columns: ["released"]` straight off the table. So this can be the same shape, with no view involved:

```json
{
  "table": "CouncilMeeting",
  "schema": "public",
  "columns": ["administrativeBodyId"],
  "label": "administrative_body_id",
  "relationship": {
    "variant": "scalar", "type": "one_to_one",
    "foreign_key": { "parent": ["councilMeetingId", "cityId"], "child": ["id", "cityId"] }
  }
}
```

The payoff is not only one less node. A view-backed node only re-syncs when PGSync can match a WAL event to the view's primary key; a plain `CouncilMeeting` node is matched by PGSync's normal table routing, so moving a meeting to a different body propagates on its own.

I ran it. With this change, the bootstrap produces exactly the same value, and reassigning a meeting to another body updated the document within seconds — `administrative_body_id` and `administrative_body_type` both followed. `administrative_body_type` still needs the view, because the type lives on `AdministrativeBody` and is an enum that has to be cast to text.

elasticsearch/schema.json
```diff
@@ -379,6 +397,58 @@
...
+          "relationship": {
+            "variant": "scalar",
+            "type": "one_to_one",
+            "foreign_key": { "parent": ["id"], "child": ["id"] }
+          }
+        },
+        {
+          "table": "MeetingAdministrativeBodyView",
+          "schema": "public",
+          "base_tables": ["CouncilMeeting", "AdministrativeBody"],
```
kouloumos: **`AdministrativeBody` does not belong in `base_tables` here, and listing it is misleading rather than harmless.**

What `base_tables` means to PGSync: "if a row in one of these tables changes, this node may be affected". When a WAL event arrives, PGSync takes the changed row's key and looks it up against the node's declared `primary_key`.

This node declares `primary_key: ["id", "cityId"]`. Those are **`CouncilMeeting`'s** id and cityId — the view keeps them under the original names for exactly the WAL reason documented in the FAQ.

So the two base tables behave very differently:

- **`CouncilMeeting`** — the event carries the meeting's `id` and `cityId`, which is what the key holds. Correct match.
- **`AdministrativeBody`** — the event carries the **body's** `id`, and the body's `cityId`. PGSync compares that body id against a column holding *meeting* ids. It is not that the lookup is disabled; it runs and finds nothing.

That is why a body edit never shows up, and it is why I would not describe it as a documented limitation. It is a key that happens to always miss. Nothing guarantees it keeps missing: `CouncilMeeting.id` is not always a cuid — the seed data alone has `jan14_2026`, `dec17_2025` — so a body id colliding with a meeting id in the same city is a value collision, not a type impossibility. On that day PGSync would re-sync a different meeting's subjects. I checked the current data and there are zero collisions, so this is latent, not live.

Suggested change: `"base_tables": ["CouncilMeeting"]`.

This composes with the other comment. If `administrative_body_id` reads straight off `CouncilMeeting`, the view is left serving only `administrative_body_type`, and `CouncilMeeting` is the only table whose events can route to it.

I ran it with `base_tables` narrowed. Reassigning a meeting to a different body still propagates within seconds. Editing the body a meeting points at produces no sync — and now it produces no event at all, rather than an accepted event that quietly resolves to nothing: the counters stayed at `Db: [5] => Elasticsearch: [4]` across the edit. Worth saying plainly in the FAQ that a body rename or type change reaches the index only when the meeting or the subject is next written.

elasticsearch/validate-views.sql
```diff
@@ -125,6 +185,72 @@ FROM "LocationSearchView";
...
+-- ============================================================================
+\echo '7. Validating MeetingAdministrativeBodyView...'
+SELECT
+  COUNT(*) AS total_meetings,
+  COUNT(administrative_body_id) AS with_body,
+  COUNT(*) - COUNT(administrative_body_id) AS without_body,
+  CASE
+    WHEN COUNT(CASE WHEN administrative_body_id IS NOT NULL AND administrative_body_type IS NULL THEN 1 END) > 0
+      THEN 'FAIL: Body present but type missing'
+    WHEN COUNT(CASE WHEN administrative_body_type NOT IN ('council', 'committee', 'community') THEN 1 END) > 0
```
kouloumos: Check 8 reads the realms from `pg_enum` on purpose, so a new realm passes on its own. The same argument applies here — a fourth `AdministrativeBodyType` would fail this check for no reason.

elasticsearch/views.sql
```diff
@@ -235,17 +330,38 @@ LEFT JOIN "IntroducedByPartyView" ibp ON ibp.person_id = s."personId" AND ibp.ci
...
 \echo ''
 
--- Check 6: SubjectSearchView - verify description stripping works
+-- Check 6: SubjectSearchView - verify description stripping and metrics
 \echo '6. Checking SubjectSearchView data...'
 SELECT 
   COUNT(*) AS total_subjects,
-  COUNT(description) AS subjects_with_description
+  COUNT(description) AS subjects_with_description,
+  COUNT(*) FILTER (WHERE contributor_count > 0) AS subjects_with_contributors,
```
kouloumos, line 369: Not yours, but it blocks these checks from ever being seen: the verification block further down selects `segment_id` and `contribution_id`, neither of which exists on those views. Both are on `main`. Under `ON_ERROR_STOP` the script stops there, and the harness runs it with output discarded — so checks 6, 8 and 9 that this PR adds had never actually run. They pass once you get past it. I will send a separate fix unless you would rather it came here.

kouloumos (commented):
The SQL is correct. I built a fixture that isolates each branch of the metric: tagged utterances spread over a substantive segment, a procedural one, and one with no summary row; a subject with only legacy rows; a subject whose tagged utterances are all procedural and which also carries a legacy row; and a subject with no timing source at all. Every value matched, including the case that has to report zero rather than fall back to the legacy table. Running `getDiscussionSecondsForSubjects` over the same database returns the same numbers for all of them. On the sync side every new node resolves, the bootstrap succeeds, and the documents carry the right values — including a null administrative body and two different realms.

Your re-index question has a clean answer: an additive field does not need one. Elasticsearch accepts new fields on a live mapping. What it will not do is let PGSync deliver them, because PGSync sends the mapping block only when it creates the index — bootstrapping against an existing index with a deliberately conflicting type declared succeeds, and the mapping does not move. Skip the mapping call and dynamic mapping picks the types instead, so a string becomes `text` rather than `keyword`, permanently. Filling the documents that already exist costs no inference at all when the write carries only the new fields, against tens of minutes of serialized embedding for a full re-sync. I pushed that as a commit on this branch, replacing the line that said every schema change needs a full re-index: it now carries a rehearsal step, the order of operations, and the sub-field case where the mapping call succeeds and the data silently never arrives.

One behavioural thing is worth settling together before this ships, and it is not something you introduced. A change to an utterance's discussion tag never reaches the index. Both halves check out against a running daemon: a contribution written in the same transaction as its subject syncs fine, so `contributor_count` is right on every real path; an utterance tag written on its own does not, so `discussion_speaking_seconds` keeps whatever value it had. That bites here because the summarize task saves the subjects first and tags the utterances afterwards with no further subject write, so a freshly summarized subject would index zero seconds and hold it. The transcript editor writes the same columns and has the same gap. There are several ways to close it and they differ a lot in scope, so I would rather agree an approach than pick one in a review comment.

kouloumos:
**What changed, and why**

Four commits on `elasticsearch/`, and one on the summarize path.

**The view verification blocks selected columns that do not exist.** `SubjectSpeakerSegmentSearchView` exposes `id`; `segment_id` is a name PGSync creates at index time via `transform.rename`. Same for `contribution_id`. So both scripts errored partway and psql abandoned everything after — which is why checks 6, 8 and 9 that this PR adds had never been observed running. The harness invokes `views.sql` with output discarded, so the error was invisible. Both scripts now run to completion and every check reports PASS. This bug is on `main`, not yours; it is here because it was hiding your own checks.

**`administrative_body_id` no longer needs the view.** The join condition is `ab.id = cm."administrativeBodyId"`, so `ab.id` is that column verbatim — the join can only reproduce it. It now reads straight off `CouncilMeeting`, the way `meeting_released` does. That also buys correct routing: moving a meeting to another body is a `CouncilMeeting` write and now propagates on its own, which I confirmed against a running daemon.

**`MeetingAdministrativeBodyView` is scoped to `CouncilMeeting` events.** With the id gone the view serves only the type, and `AdministrativeBody` cannot route to it anyway: the node's key holds the *meeting's* `(id, cityId)`, so a body's WAL event was being compared against meeting ids and always missed. Now it produces no event rather than an accepted one that resolves to nothing. A body's type change reaches the index when the meeting or subject is next written, and the FAQ says so; a rename needs no sync at all, since the index stores no names.

**Body types now come from `pg_enum`**, matching the reasoning already applied to realms in check 8.

---

**The tagging change.**

The discussion tags are now applied inside `saveSubjectsForMeeting`'s transaction instead of after it. Some history, because the current shape was deliberate.

Originally the tags were written as `prisma.$transaction(updateOperations)` — an array of singular `update()` calls, batched purely for performance. `update()` throws `P2025` when the row is gone, and in an array transaction one throw aborts every statement. Christos hit that on staging (`da5d8dc5`): a summarize request snapshots utterance ids, the model takes minutes, and anything that deletes an utterance in that window — an editor removing one, deleting an empty segment, splitting a segment — leaves the response naming a row that no longer exists. One stale id discarded the tags for the whole meeting. His fix replaced the batch with per-row updates and try/catch, plus dedup and an existence pre-check.

That resilience is preserved here, by a different route. Grouping into `updateMany` by `(subjectId, discussionStatus)` means a missing id matches nothing instead of raising — I reproduced all three paths against a real database, and the old one wrote zero tags where the new one writes every valid tag. It also collapses ~2,456 statements per meeting into ~46, so the work is smaller than it is today, not larger. The existence pre-check becomes unnecessary rather than merely moved; it could never close the race anyway, since a row can be deleted between the check and the update.

One new guard: `discussionStatus` arrives from the task server and nothing on this path validated it. An unknown value is a `PrismaClientValidationError`, which inside a transaction would discard the subjects too — so unknown statuses are now dropped with a warning before the transaction runs. The editor path already validates the same field; the task path did not.

Why move it at all: PGSync re-indexes a subject only when the `Subject` row changes. Written after the transaction commits, the tags never reach the index, so every newly summarized subject would report `discussion_speaking_seconds: 0` forever. With the tags inside the transaction, a live test took a subject from 0 to 50 seconds in the index.

One thing worth agreement: a tagging failure now rolls the subjects back. That property did not exist before — tags and subjects were never in one transaction — so it is something we are adding, not something being undone. It is recoverable, since `TaskStatus.responseBody` keeps the response and the reprocess button re-runs the handler.

---

**After this merges**

These fields are additive, so no re-index. The procedure is in `elasticsearch/README.md` under "Deploying a Schema Change"; the specifics for this change:

1. **Views** — run `views.sql` against the production database.
2. **Rehearse the mapping** on a throwaway copy of the live mapping. For this PR it returns `acknowledged`, which is the additive path — I have already run it.
3. **Apply the mapping** to `subjects`, before any document carries the new fields. Skipping this is the one irreversible mistake here: dynamic mapping would type `city_realm` and `administrative_body_id` as `text` instead of `keyword`, and that cannot be changed without a real re-index. `location_id` is already in that state.
4. **Deploy `schema.json` and restart the PGSync daemon** — before the backfill, not after. A live-sync write from the old schema is a full-document `index` operation and would drop the new fields from any document it touches.
5. **Backfill** the existing documents with a bulk `update` carrying only the five new fields. That costs no inference at all, measured on the model's own counter, against tens of minutes of serialized embedding for a full re-sync. Roughly a minute for the 11,430 documents.
6. **Verify** — the mapping matches `schema.json` with no undeclared fields, a `terms` aggregation on `administrative_body_type` works (it fails on a `text` field, so it proves the keyword mapping landed), and a semantic query returns its usual results.

Only the production index needs this. Staging and preview leave `ELASTICSEARCH_INDEX` unset and read it, and the local harness recreates its index on every run.

@christosporios — is there a failure you saw in that window that was *not* `Record to update not found`?

kouloumos:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/822e349ef9640167f702135785c8f0629f47a2d0..9f8d8f5e9838e82aa348cd4371187a2f9e1d19b9>) to rebase onto main now that #346 has merged, and to fix a conflict between the two that was more than textual.

**#346 removed `SubjectSearchView`, and this branch was building on it.**

That view is gone for a reason. Its commit message says it plainly: the view declared `base_tables: ["Subject"]` — the root table — which made PGSync route `Subject` DELETE events as child re-syncs instead of root deletions, so deleted subjects were never removed from the index. This branch added two more nodes with that same shape, so merging it as it was would have restored the bug.

I confirmed it with the deletion recipe rather than reasoning about it. Same probe, same database, only the schema differing:

| schema | after `DELETE` | counters |
|---|---|---|
| this branch, before | still `HTTP 200` past 60s | `Db: [2] => Elasticsearch: [1]` — event accepted, no ES operation |
| #346 | `HTTP 404` at t+15s | `Db: [2] => Elasticsearch: [2]` |
| this branch, now | `HTTP 404` at t+15s | `Db: [2] => Elasticsearch: [2]` |

**What changed here.** The metrics moved to a new `SubjectMetricsView` — `id`, `contributor_count`, `discussion_speaking_seconds`, nothing else. The aggregate SQL is unchanged. What is different is the declaration: `base_tables` now names only the tables the aggregates actually read, never `Subject`. That matches every other view-backed node in the schema — `LocationSearchView` lists `["Location"]`, `SpeakerContributionSearchView` lists its own source tables. `SubjectSearchView` was the only one that named the root, and that was the defect. The README now carries a warning next to the view so the next subject-keyed view does not repeat it.

Your two commits stayed yours, with only that forced change. The rest of the series is unchanged, except `fix(elastic): select a column that exists in the segment sample query`, which narrowed: #346 already fixed the `contribution_id` half, so it now covers `segment_id` only and the message says so.

**Verified on the rebased branch**, all through `pgsync-test.sh`: `views.sql` and `validate-views.sql` both run to completion under `ON_ERROR_STOP` with every check PASS; the bootstrap reproduces all four metric branches exactly (90 / 45 / 0 / 0, including the legacy fallback and the all-procedural subject that must report 0 rather than fall back); a subject write with its tags in one transaction moves a document from 0 to 50s; and deletion propagates. `tsc` is clean and the 151 unit tests pass.

The `schema.json` diff is 68 added lines and no deletions. An earlier push of this rebase had reformatted the whole file — that was mine, and it is fixed; if you looked at the diff in the last hour and saw ~700 changed lines, that is why.

Nothing else about the change has moved. The post-merge deploy steps in my earlier comment still apply as written.

elasticsearch/README.md
kouloumos: Rewritten in d929bdf0. The answer now says the metrics depend on their data committing together with a `Subject` write, and that the summarize task applies the discussion tags inside that transaction.

elasticsearch/schema.json
kouloumos: Done in a45faad4 — it reads `CouncilMeeting.administrativeBodyId` directly, no view. Confirmed against a running daemon that moving a meeting to another body now propagates on its own.

elasticsearch/schema.json
kouloumos: Narrowed to `["CouncilMeeting"]` in a45faad4. A body edit now raises no event at all, rather than one that is accepted and resolves to nothing.

elasticsearch/validate-views.sql
kouloumos: Now reads `pg_enum` (974adaa6), matching the realm check.

elasticsearch/views.sql
kouloumos: `segment_id` fixed in 82801b51; the `contribution_id` half was fixed by #346. Both scripts now run to completion under `ON_ERROR_STOP`.

kouloumos:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/9f8d8f5e9838e82aa348cd4371187a2f9e1d19b9..e21aae0d7272850ec3d57635b9cfb6c8c7b464c4>) to tidy the history and fix one stale reference. The tree is unchanged apart from a single word.

- The FAQ still said "after you change the aggregates in `SubjectSearchView`". That view no longer exists; it now says `SubjectMetricsView`. The other mentions of the old name are deliberate — they explain why it was removed.
- My two documentation commits became one, and it now sits after the summarize fix, because it describes behaviour that only holds once that fix is in.

Seven commits: your two, then the `administrative_body_id` refactor, the segment-query fix, the `pg_enum` check, the summarize transaction fix, and the documentation.

Re-verified after the rewrite: no node names the root table, both SQL scripts run to completion under `ON_ERROR_STOP` with no failures, `tsc` is clean, and the 151 unit tests pass.

One thing you should know rather than discover: your two commits no longer contain exactly what you wrote. #346 removed `SubjectSearchView`, which they were built on, so during conflict resolution I rewrote those hunks to use `SubjectMetricsView`. The aggregate SQL is untouched — only the view name and its `base_tables` declaration changed. They still carry your authorship, so please check they read the way you would have written them.

PR #632, opened 2026-08-15
Notis PR 2: databases, views, rollout and shared-cookie auth

christosporios:
PR 2 of the Notis plan (`notis-db`, PRD §9). Notis gets its memory and its integration seams. Everything ships dark: no user-visible behavior changes until an admin flips a user in the release panel.

## What this adds

- **Notis's own Postgres.** A Prisma schema under `services/notis/prisma` with its own migration history: subscriptions, wakes, messages, journal, scheduled wakes, queue, and the processed-event dedup log. `userId` references the main database by id only. Each `NotisWake` row stores its event, full outcome, delivery and trace verbatim, so the admin viewer replays DB conversations into the same `WakeRecord` shape the playground uses.
- **The only pre-rollout main-DB migration.** `User.notisEnabledAt`, five read-only views (`notis_users`, `notis_fanout_targets`, `notis_meeting_events`, `notis_sessions`, `notis_admin_sessions`), and the `notis_reader` role. City rows carry `realm`, `language` and `timezone`. The role can `SELECT` the five views and nothing else, so the service physically cannot write the main database. Every statement is idempotent, so the integration suite executes the migration file on top of a `prisma db push` schema.
- **Matching-engine exclusion.** Users with `notisEnabledAt` set lose the message-medium delivery at creation, at the send boundary (pre-existing pending deliveries become `skipped`, a new delivery status), and in the signup welcome. Email stays untouched. Nobody is served by both WhatsApp paths.
- **Release panel** at `/admin/notis` in the main app: coverage (enabled X of Y eligible), a searchable per-user toggle, and «enable next batch» — N random eligible users, clamped to the remainder. The flip writes the timestamp and nothing else.
- **Shared-cookie auth** replaces the interim HMAC gate. See the design note below.
- **Playground real users.** `/api/admin/real-users` serves `notis_fanout_targets` grouped per user; the wizard gains «διάλεξε πραγματικό χρήστη» and prefills name, cities, topics and locations. The profile now tracks `seedProfileFromPreferences` — the same mechanical rule PR 6's migration enrollment will use — until the admin edits it by hand.
- **Real admin queries.** The mock modules (`_lib/metrics`, `_lib/conversations`, `_lib/wakes`) become database queries. Without `NOTIS_DATABASE_URL` the service keeps the honest-zeros playground-only mode.
[...]

PR #634, opened 2026-08-16
feat(notis): inbound — Bird webhook, live queue, agent replies (PR 3)

christosporios:
PR 3 of the [Notis plan](https://github.com/schemalabz/opencouncil/pull/598) (PRD §9), stacked on #632. The queue tables from PR 2 get their first code. A rollout-enabled user who texts the WhatsApp number now gets a real agent conversation. Their message opens the 24h window, so every reply is free-form — no Meta template dependency.

## What this adds

**Notis webhook** — `POST /api/webhooks/bird` on the notis service, with its own Bird subscription and signing key (HMAC ported from the main app). Routing, in order: outbound events reconcile delivery status on notis messages (forward-only, `read` is a real status here); inbound events dedupe by `birdMessageId`; the gate serves only users with `notisEnabledAt` set; first contact enrolls them (origin `inbound`, profile seeded from preferences, no transition template — they opened the conversation); a bare ΣΤΟΠ unsubscribes deterministically with no wake; everything else becomes a live-lane queue row, and `after()` kicks the drainer so Bird gets its 200 before the 30–60s wake runs.

**Live queue** — `queue-core.ts` is the state machine (DB-injected, so the main repo's integration tests drive it against real Postgres). The claim is one atomic `UPDATE … FOR UPDATE SKIP LOCKED` with stale-claim recovery and strict per-subscription serialization. Attempts count at claim time; `attempts > 3` fails terminally with an alert.

**Drainer** — assemble state (profile + last 30 journal entries + live cities from the fanout view, snapshot fallback) → `runWake` → ONE transaction persists the wake, journal entry, profile/unsubscribe deltas, scheduled-wake rows, pending outbound messages, and the queue item's `done` — **before Bird is called**. Once the model has run, a retry never re-runs it. Sends carry the message row id as `Idempotency-Key`; a sweeper (60s) re-claims stale work and re-sends messages stuck between commit and send, without double delivery.

**Deterministic unsubscribe** — strict bare keyword only (`stop`/`στοπ`, punctuation-tolerant). Longer messages («θέλω να απεγγραφώ», «γιατί έγινε διακοπή νερού;») go to the agent, which carries `unsubscribe_user`. The main app's broad-regex + Haiku-verify flow is deliberately not ported: the agent IS the verifier for anything beyond a bare ΣΤΟΠ.

**Main app** — the webhook now skips notis-served phones entirely (no reply, no unsubscribe handling, no `Message` row). Without this, an enabled user's reply draws two answers, and every notis outbound event creates a stray main-DB row. Legacy outbound reconciliation stays.

**Shared rails** — the template-vs-freeform decision moved out of the playground client into `agent/delivery.ts`; the simulator and the real send path now call the same function. The conversation viewer renders wake-less ΣΤΟΠ exchanges (synthesized from their journal entries) so the thread matches the ΣΤΟΠ badge.

## Design decisions (agreed before implementation)

1. **Inbound gate = `notisEnabledAt` only.** During rollout both webhook subscriptions receive every event. A STOP must land where it can act: an old-path STOP flips a main-DB flag that notis's read-only role physically cannot write. So notis stays SILENT for old-path users and unknown numbers — the main app answers them, as today. The PRD's "any matching user gets the agent" arrives when the old path retires.
2. **ΣΤΟΠ pre-step is strict.** Deterministic action only when the message IS the keyword. No second model dependency in the pre-step, no false positives on «διακοπή νερού».

## Schema (notis DB only; no main-DB migration)

`SubscriptionOrigin` gains `inbound`; `NotisSubscription` gains `birdConversationId` (the conversation replies are sent into) and a phone index; `NotisMessage.birdMessageId` becomes unique (dedupe + status-update key).

## Testing

- Unit (notis, 122 tests): signature vectors, ported extractor cases, stop matcher, delivery rails, drainer ordering invariants with fakes (wake committed before Bird, idempotency key = message id, failure → pending retry), webhook routing decisions.
- Integration (main repo, testcontainers): queue claim semantics on real Postgres — SKIP LOCKED under parallel claims, per-subscription serialization, stale reclaim, the attempts ladder, migration idempotency. Plus the main-app gate (`isNotisServedPhone`) in both stored phone formats.
- Golden: new `user_message` fixture (`athens-jul15-petralona-parking-reply`), recorded live, replayed deterministically.
- Live dev E2E via `scripts/send-test-webhook.ts`: enrollment + real wake ($0.17, 42s, grounded 3-message reply over prod MCP), bare ΣΤΟΠ → unsubscribed with no wake, unknown number → silence, old-path user → silence. Verified in the admin conversations viewer (thread, ΣΤΟΠ exchange, trace inspector).

## Deploy checklist (on merge)

1. Run the notis migration (`prisma migrate deploy` for `services/notis`) per environment.
2. Set on the notis app: `BIRD_API_KEY`, `BIRD_WORKSPACE_ID`, `BIRD_WHATSAPP_CHANNEL_ID` (same values as the main app), and a FRESH `BIRD_WEBHOOK_SECRET`.
3. Create the second Bird webhook subscription: URL `https://notis.opencouncil.gr/api/webhooks/bird`, the fresh signing key, service `Conversations`, events `conversation.created` + `conversation.updated`. **Production only** — staging gets no Bird subscriptions (decision 2026-08-16: the staging webhooks were removed from Bird, because Bird fans every event out to all subscriptions and staging would answer real users). Test inbound on staging with `scripts/send-test-webhook.ts` and the playground.
4. Verify one live event's signature in the notis logs.
5. On the first real reply, confirm `Idempotency-Key` is honored on the conversations endpoint (replay returns `Idempotency-Replay`).
6. PRD gate: at least two weeks of inbound-only operation precede any proactive send (PR 4).

## Known limitations (deliberate, noted for later PRs)

- No re-subscribe path after ΣΤΟΠ (the profile toggle lands in PR 5; `unsubscribed` is terminal until then). An unsubscribed reader's normal message still gets an answer — replying in-window is not a proactive send.
- The playground kill switch is localStorage-only and does not cover reactive replies (it governs proactive sends, PR 4).
- Subscription status has two states (`active`/`unsubscribed`); `paused_no_phone` was dropped in #632. **Notis owns the state** — ΣΤΟΠ and `unsubscribe_user` write it directly, and there is no second copy anywhere. The main app's profile checkbox becomes a client of it through an authenticated notis API (PR 5); the PR 4 poller refreshes phones and unsubscribes when a phone is gone, and never re-activates anyone. Re-activation is only the checkbox toggle.
- Template sends from the drainer are unreachable in PR 3 (`user_message` is always in-window); the path fails loudly with an alert if ever hit.

---

> [!NOTE]
> **High Risk**
> Touches live WhatsApp ingress/egress, dual-webhook rollout gating, and transactional queue delivery—user-facing messaging failures or duplicate replies would be high impact; Greptile also flags transient Bird 408/429 as non-retryable in `bird.ts`.
> 
> **Overview**
> **Inbound WhatsApp for Notis (PR 3)** wires a dedicated `POST /api/webhooks/bird` on the notis service with its own Bird subscription and signing key. Inbound traffic is deduped on `birdMessageId`, gated to users with `notisEnabledAt`, auto-enrolled on first message (`inbound` origin, no intro template), handled with a strict bare **ΣΤΟΠ** path without waking the agent, and otherwise enqueued on the **live** lane with `after()` draining so Bird gets a fast 200. Outbound webhook events reconcile delivery status (including `read`) and store `failureReason` for ops.
> 
> **Persistence and sending** extend the notis schema (`inbound` origin, `birdConversationId`, unique `birdMessageId`, one-running-per-sub queue index, failure reason). The drainer commits wake/journal/messages before calling Bird, uses message row ids as `Idempotency-Key`, and a 60s sweeper recovers stale queue work and pending sends. Shared **`decideDelivery`** centralizes template vs free-form rules for playground and production.
> 
> **Main app** skips inbound handling entirely for notis-served phones so one message never gets two replies; legacy outbound reconciliation is unchanged.
> 
> **Admin and docs** replace the overview with range-based metrics (charts, period deltas, delivery/failure breakdown), paginated searchable conversations and filtered wake feeds with avatars and real delivery ticks in the chat viewer (including synthesized ΣΤΟΠ journal rows). Bird rollout setup is documented; `send-test-webhook.ts` exercises the path locally. Agent prompt and golden fixtures cover natural-language unsubscribe and follow-up replies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a56e47a9a3452ab7de92d063bfaf38791325912. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds the Notis inbound Bird webhook, transactional live queue, agent replies, retryable outbound delivery, and rollout gating in the main application.
- Adds inbound enrollment, deduplication, strict STOP handling, and queue creation.
- Persists wake results before Bird delivery and retries pending sends with stable idempotency keys.
- Adds the Notis schema migrations, administration views, tests, and deployment configuration.

The PR is not yet safe to merge because the resend claim can expire before its Bird request finishes, allowing concurrent sends and conflicting delivery state.

The claim timestamp starts before the Bird timeout, but both use the same 30-second interval. A concurrent sweep can reclaim the message during that timing gap, and the original sender can overwrite the new owner's result because result updates are not fenced.

**Files Needing Attention:** services/notis/src/lib/queue.ts and services/notis/src/lib/bird.ts

| Filename | Overview |
|----------|----------|
| services/notis/src/lib/queue.ts | Implements transactional wake processing and outbound retries, but the send-claim expiration can overlap the request it is intended to protect. |
| services/notis/src/lib/bird.ts | Adds bounded Bird sends and correctly classifies network errors, 5xx, 408, and 429 as retryable. |
| services/notis/src/app/api/webhooks/bird/handlers.ts | Adds inbound routing, rollout gating, deduplication, enrollment, and deterministic STOP handling. |
| services/notis/src/lib/queue-core.ts | Adds atomic queue claims, stale recovery, claim fencing, and per-subscription serialization. |
| services/notis/prisma/schema.prisma | Extends subscriptions and messages for inbound conversations, delivery reconciliation, and send ownership. |

<a href="https://app.greptile.com/ide/claude-code?prompt=Greploop%20schemalabz%2Fopencouncil%20PR%20%23634%3A%20work%20through%20Greptile's%20open%20review%20comments%2C%20then%20keep%20reviewing%20and%20fixing%20until%20it%20comes%20back%20clean%20at%205%2F5%20with%20zero%20unresolved%20comments.%0AStart%20by%20reading%20the%20comments%20off%20the%20PR%20itself.%20On%20GitHub%2C%20use%20paginated%20%60gh%20api%20graphql%60%20to%20query%20%60pullRequest.reviewThreads%60%20with%20each%20thread's%20%60isResolved%60%20value%20and%20inline%20%60comments%60.%20%60gh%20pr%20view%20--comments%60%20only%20includes%20conversation%20comments%2C%20so%20do%20not%20use%20it%20as%20the%20findings%20source.%20They%20are%20not%20listed%20in%20this%20prompt%20on%20purpose%3A%20the%20PR%20is%20current%2C%20a%20pasted%20copy%20would%20not%20be.%20Skip%20anything%20already%20resolved%2C%20and%20if%20you%20judge%20a%20comment%20wrong%2C%20say%20so%20rather%20than%20changing%20code%20to%20satisfy%20it.%0A%0APrefer%20the%20Greptile%20CLI%2C%20which%20reviews%20the%20working%20tree%20with%20no%20push%20and%20no%20CI%20run.%20Fall%20back%20to%20PUSH%20LOOP%20only%20where%20a%20step%20below%20says%20to.%0A1.%20Run%20%60command%20-v%20greptile%60.%20Missing%3A%20go%20to%20PUSH%20LOOP%2C%20and%20pitch%20the%20CLI%20at%20the%20end.%0A2.%20Run%20%60greptile%20whoami%60.%20It%20exits%200%20even%20when%20signed%20out%2C%20so%20read%20the%20output%3A%20if%20it%20says%20%22Not%20signed%20in%22%2C%20ask%20the%20user%20to%20run%20%60greptile%20login%60%20%28it%20opens%20a%20browser%2C%20do%20not%20run%20it%20yourself%29%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%0A3.%20Look%20for%20%60greploop%2FSKILL.md%60%20under%20%60.agents%2Fskills%2F%60%20or%20%60.claude%2Fskills%2F%60%2C%20both%20in%20the%20repo%20root%20and%20in%20the%20user's%20home%20directory.%20Found%3A%20follow%20that%20skill%2C%20never%20pushing%20between%20rounds%2C%20and%20you%20are%20done.%0A4.%20Missing%3A%20run%20%60greptile%20skills%20list%60.%20If%20it%20fails%20as%20an%20unknown%20command%2C%20this%20CLI%20predates%20the%20skill%2C%20so%20tell%20the%20user%20to%20run%20%60greptile%20update%60%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%20Otherwise%20ask%20the%20user%20before%20running%20%60greptile%20skills%20install%20greploop%60%2C%20then%20follow%20the%20skill%20it%20writes.%0A%0APUSH%20LOOP%20%28only%20when%20a%20step%20above%20sends%20you%20here%29%3A%0A%0AGREPLOOP%20ENABLED%3A%20After%20fixing%20all%20issues%2C%20commit%20and%20push%20your%20changes.%20Greptile%20will%20automatically%20re-review%20the%20PR%20via%20webhook.%20Wait%20for%20the%20review%20to%20complete%2C%20then%20read%20the%20new%20review%20comments%20and%20confidence%20score.%20If%20the%20confidence%20score%20is%20below%205%2F5%20or%20there%20are%20unresolved%20comments%2C%20fix%20those%20issues%20and%20push%20again.%20Repeat%20until%20the%20confidence%20score%20is%205%2F5%20with%20zero%20unresolved%20comments.%20Maximum%205%20iterations.%20Do%20NOT%20open%20a%20new%20PR%20%E2%80%94%20keep%20pushing%20to%20the%20same%20branch.%0A%0AIF%20YOU%20ENDED%20UP%20ON%20PUSH%20LOOP%20BECAUSE%20THE%20CLI%20WAS%20MISSING%3A%20once%20the%20loop%20is%20done%2C%20print%20the%20block%20below%20and%20nothing%20else%20about%20the%20CLI.%20Reproduce%20it%20character%20for%20character%20inside%20a%20fenced%20code%20block%2C%20so%20the%20borders%20line%20up.%20Do%20not%20reword%20it%2C%20translate%20it%2C%20summarize%20it%2C%20add%20to%20it%2C%20or%20drop%20the%20box.%20Print%20it%20once%2C%20and%20do%20not%20bring%20the%20CLI%20up%20again%20later%20in%20the%20session.%0A%0A%E2%95%AD%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AE%0A%E2%94%82%20%20You%20can%20run%20greploops%20faster%20locally%20with%20our%20CLI.%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%E2%94%82%0A%E2%94%82%20%20Install%20it%20at%20https%3A%2F%2Fwww.greptile.com%2Fcli%2C%20or%20I%20can%20install%20it%20for%20you.%20%20%E2%94%82%0A%E2%95%B0%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AF%0A%0AIf%20they%20take%20you%20up%20on%20it%2C%20install%20with%20%60npm%20install%20-g%20greptile%60%20%28or%20%60brew%20install%20greptileai%2Ftap%2Fgreptile%60%29%2C%20then%20%60greptile%20skills%20install%20greploop%60.%20Leave%20%60greptile%20login%60%20to%20them%2C%20it%20opens%20a%20browser.&repo=schemalabz%2Fopencouncil&pr=634&platform=github">Fix all with Greploop</a> <a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Aservices%2Fnotis%2Fsrc%2Flib%2Fqueue.ts%3A389%0A**Send%20claim%20expires%20too%20early**%0A%0AIf%20another%20resend%20sweep%20runs%20while%20a%20Bird%20request%20approaches%20its%2030-second%20timeout%2C%20the%20sweep%20can%20reclaim%20the%20message%20because%20%60sendingAt%60%20starts%20before%20the%20request%20timeout%20but%20uses%20the%20same%20expiration%20interval.%20Both%20requests%20can%20then%20overlap%2C%20and%20their%20unfenced%20result%20updates%20can%20leave%20a%20delivered%20reply%20pending%20or%20failed%2C%20causing%20another%20resend%20or%20an%20incorrect%20terminal%20status.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=634&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
services/notis/src/lib/queue.ts:389
**Send claim expires too early**

If another resend sweep runs while a Bird request approaches its 30-second timeout, the sweep can reclaim the message because `sendingAt` starts before the request timeout but uses the same expiration interval. Both requests can then overlap, and their unfenced result updates can leave a delivered reply pending or failed, causing another resend or an incorrect terminal status.

---

`````

</details>

<sub>Reviews (18): Last reviewed commit: ["Merge feat/notis-db into feat/notis-inbo..."](https://github.com/schemalabz/opencouncil/commit/9bdca5a2561f8dcc5ab2a4e1522727979d32675a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53780592)</sub>

> Greptile also left **1 inline comment** on this PR.

### CODE REVIEW

**Comment by christosporios** [comment]
Pushed three commits on top of the initial stack:

- `e3571d6b` — the conversation viewer now renders wake-less ΣΤΟΠ exchanges (synthesized from their journal entries). Found during browser verification: the thread dropped the deterministic pre-step's messages and contradicted the ΣΤΟΠ badge.
- `5d855ffb` — outbound messages store a `failureReason` (new idempotent migration). The webhook and the send path both already had the text; now it persists.
- `c477dd25` — the overview page becomes a period dashboard: 7/14/30-day and 3-month windows (default 7), every metric with its change versus the preceding period, wake/delivery/cost breakdowns, and the 5 latest inbound messages with links into their conversations.

**Comment by christosporios** [comment]
Follow-up to the overview: the four KPI cards now carry their per-day series as a faint background area chart (recharts, Athens-local buckets, zero-filled). Hovering a card swaps the roles — the number recedes, the chart comes forward, and a cursor tooltip reads out individual day values. Commit `924b2461`.

**Comment by christosporios** [comment]
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/924b24612b7db1c7a60e26d51e22987abe6d8517..a1181c29604f37504aa2f4f8b64c7c87380ce5b4) to squash a wording fixup into the overview commit: the Wakes panel now says «απόφαση σιωπής / απόφαση μηνύματος» instead of «σιωπή / απάντησε». No other change to the diff.

**Comment by christosporios** [comment]
Three more commits from panel review:

- `6de71e3b` — settings page reflects PR 3 (Bird / WhatsApp env section with live badges, updated upcoming-PRs list); the sidebar drops the 'Συνδεδεμένη βάση' label.
- `d103dd4a` — the overview range picker gains 1h and 24h; the chart series generalizes to minute/hour/day buckets (verification caught hour keys not truncating to :00, which silently zero-filled every sub-day chart — now tested). Inbound feed polished (avatars, mini-bubbles); panel copy says χρήστες and απόφαση σιωπής / απόφαση μηνύματος.
- `5b40317a` — the conversation thread shows real per-message delivery status: tick ladder on Notis bubbles and a red 'Δεν παραδόθηκε' line with the stored failure reason. Previously a failed send was indistinguishable from a delivered one.

**Comment by christosporios** [comment]
User-spotted timezone bug in the overview charts, fixed in `b4f5db13`: the stored timestamps are naive UTC, and a single `AT TIME ZONE 'Europe/Athens'` interprets instead of converts — every bucket landed six hours early (2× the offset). Now the double conversion (`AT TIME ZONE 'UTC' AT TIME ZONE 'Europe/Athens'`). Hour tooltips also name their full floored interval («13:00–14:00») so the bucket boundary is unambiguous.

**Comment by christosporios** [comment]
Panel review continued in `aea1843c`: the conversations list now answers who/what/cost at a glance (last message inline, wakes count, per-conversation cost, failed deliveries in red, search); the wakes feed puts the rationale front and center with decision filter chips and amber health marks (repair nudges, truncation, missing finish_wake). Every emoji left the panel UI.

**Comment by christosporios** [comment]
Added the event-type filter to the wakes feed (`d16ac986`): a second chip group that composes with the decision filter — links preserve both dimensions, counts are cross-filtered, and only event types that actually exist get a chip (agenda/summary/scheduled appear as PR 4 produces them).

**Comment by christosporios** [comment]
Small polish in `92b01ce2`: table rows in Συνομιλίες and Wakes are vertically centered — single-line cells no longer hug the top edge next to multi-line ones.

**Comment by christosporios** [comment]
Two more from panel review:

- `f279b6f1` — real pagination on Συνομιλίες (50/page) and Wakes (100/page): prev/next preserving search and filters, true totals, page clamping. Both lists were silently capped before; they now handle thousands.
- `3bc43026` — facehash avatars (deterministic local SVG, seeded by subscription id, brand palette) replace the initial circles in the conversations list and overview feed.

**Comment by christosporios** [comment]
Facehash avatars extended to the playground in `ebe6adeb`: the casting-sheet portrait (a fictional citizen's face morphs as you type the name; a real user keeps their stable face) and the real-user picker rows. All surfaces now seed by the main-database userId — the first pass mixed seeds and gave the same person two faces.

**Comment by christosporios** [comment]
Ran a max-effort self-review (10 finder angles, adversarial verification, gap sweep) and applied all 15 findings in four commits:

- `386cfdb4` — queue concurrency: claim ownership fencing (a post-commit failure could resurrect a done item and double-send; a reclaimed slow wake could double-run), a partial unique index enforcing one running wake per subscription at the database (the NOT EXISTS guard is snapshot-raceable), STALE_CLAIM_MS 5→15 min with an explicit 3-min Anthropic per-call timeout, and single-event items only until PR 4's coalescing gets per-event progress.
- `b7a563dd` — webhook: the rollout gate now applies to existing subscriptions (rollback no longer double-replies), enrollment upserts by userId (a changed phone permanently dropped every message incl. ΣΤΟΠ via P2002→200), the P2002 duplicate-ack is scoped to birdMessageId (a ΣΤΟΠ racing a wake commit was silently discarded), BIRD_SMS_CHANNEL_ID added so SMS events stop classifying as WhatsApp, and a prod instance without a database 503s + alerts instead of acking events into the void. Inbound also bumps updatedAt so the conversations list truly sorts by activity.
- `da930d4f` — panel: DST day-skip fixed in the chart bucketing (reproduced + tested), ?range=constructor no longer 500s (prototype-chain `in`), last-message query becomes SQL DISTINCT ON (Prisma's distinct fetched entire chat histories), delivery-status pairing gets a deterministic tiebreaker, and the wakes feed stops fetching full traces (~hundreds of KB × 100/page).
- `44ad4c93` — the narrowed unsubscribe coverage gets its safety net: an explicit prompt rule plus a recorded golden fixture where «Θέλω να απεγγραφώ» must produce unsubscribe_user.

All suites green (notis 137, main unit + integration incl. two new queue-fence/index tests), migrations applied to the dev DB, fixes verified live in the browser.

File: services/notis/src/lib/queue.ts [hunk 1]
```diff
@@ -0,0 +1,342 @@
...
+    if (result.success) {
+      await db.notisMessage.update({
+        where: { id },
+        data: { status: "sent", birdMessageId: result.messageId },
+      });
+    } else {
+      await db.notisMessage.update({
+        where: { id },
+        data: { status: "failed", failureReason: (result.error ?? "unknown error").slice(0, 300) },
+      });
```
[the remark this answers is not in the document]
christosporios: Fixed in `07f2f181`. Bird send results now carry a `retryable` flag: network errors and 5xx leave the row `pending` with the reason recorded, and the sweeper retries it under the same idempotency key. The sweeper gives up with an alert after one hour of retries. 4xx responses, Bird-side rejections and missing configuration stay terminal.

services/notis/src/app/api/webhooks/bird/handlers.ts
```diff
@@ -0,0 +1,286 @@
...
+  sent: 1,
+  delivered: 2,
+  read: 3,
+  failed: 3,
+};
+
+export function isForwardProgression(current: MessageStatus, next: MessageStatus): boolean {
+  if (current === "read" || current === "failed") return false;
+  return STATUS_RANK[next] > STATUS_RANK[current];
+}
```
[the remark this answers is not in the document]
christosporios: Fixed in `07f2f181`. `isForwardProgression` now rejects `delivered → failed` — a failure report after the handset confirmed delivery is a stale replay. `delivered → read` stays allowed (the reason notis cannot copy the main app's blanket terminal-delivered rule), and `sent/pending → failed` still lands. Covered by a new test.

christosporios:
Addressed both review comments in `07f2f181`: transient Bird failures now stay `pending` and retry under the same idempotency key (terminal after 1h with an alert), and a stale failure replay can no longer regress a `delivered` message.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/07f2f181048387688d8f39f46062241c2b408af7..e4323e8e7f642e3b16baef167b6ba4ba428f5ebd) to add a note to docs/bird-setup.md: webhook subscriptions are production-only. Staging gets no Bird subscriptions, because Bird sends every event to every subscription and staging would answer real users. The two staging subscriptions were removed from Bird today. Staging inbound tests use `scripts/send-test-webhook.ts`. The deploy checklist above is updated to match.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/e4323e8e7f642e3b16baef167b6ba4ba428f5ebd..2fb66daaa9f06800fb41fe4d32975d7a47fab013) for two reasons:

- Rebase onto the new #632 tip, which drops `paused_no_phone` from `SubscriptionStatus` (see the design note there). The PR 3 diff is unchanged by the rebase.
- New commit `2fb66daa`: the integration helper now waits on the postgres "ready to accept connections" log line instead of port probes. The postgis image is amd64-only; on Apple-silicon Docker Desktop the emulated container reports an empty `/proc/net/tcp`, so every internal port probe timed out and each suite failed at startup. The log wait works on every architecture and cuts the local suite from ~8 min to ~2 min.

Suites: notis 139 ✓, root tsc ✓, integration 80/82 (the 2 failures are the documented pre-existing ones that also fail on `main`).

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/2fb66daaa9f06800fb41fe4d32975d7a47fab013..0a56e47a9a3452ab7de92d063bfaf38791325912) to rebase onto the new #632 tip (wake `delivery`/`usage` json → columns). The drainer now writes `deliveryMode`/`deliveryTemplate` and the five token counters directly; the conversation viewer reads them. No standalone changes — the PR 3 diff only adapts to the schema. Suites: notis 139 ✓, tsc ✓, notis integration 15/15 (migration replayed twice), dev-DB backfill verified.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/0a56e47a9a3452ab7de92d063bfaf38791325912..3dbe42234e9c2662aefa712704b4959c81df1f7b) to fix the CI failures (Nix Checks and preview build). Both failed for one reason: `tests/integration/notis-queue.test.ts` imports `queue-core.ts` by relative path, which imports the gitignored notis Prisma client — and the root typecheck (including the one inside `next build`) followed that chain without generating the client. The fix, folded into the queue-suite commit:

- `tsconfig.json` now excludes `tests/` — `next build` typechecks the app, not the test suite, so app builds (preview and DO) never reach notis code.
- `tsconfig.jest.json` keeps `tests/` — the Nix `types` check still covers the suite and the queue core.
- The flake's check derivations now run `npm run prisma:generate -w notis`, same as the notis CI job already does.

Verified locally: the jest-scope typecheck passes, and the build-scope typecheck passes with `services/notis/generated` deleted (the exact CI condition). CLAUDE.md's quick-check command is updated to the jest-scope form.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/3dbe42234e9c2662aefa712704b4959c81df1f7b..4fddf77d0a411c3c59c6c6303849bde1b1a5eee7) to rebase onto the new #632 tip (subscription-state ownership recorded in the schema comment, plus latest `main`). The PR 3 diff is unchanged. The known-limitations note in the body is updated to the same design: notis owns the state, the main app is a client.

services/notis/src/lib/stop.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function isBareStop(body: string | null | undefined): boolean {
+  if (!body) return false;
+  const normalized = body
+    .normalize("NFD")
+    .replace(/[\u0300-\u036f]/g, "")
+    .toLowerCase()
+    // Strip surrounding punctuation, symbols and whitespace so «ΣΤΟΠ!» and
+    // « stop. » still count as bare.
+    .replace(/^[^a-zα-ω]+|[^a-zα-ω]+$/gu, "");
```
kouloumos: `;` is the question mark on a Greek keyboard, so «ΣΤΟΠ;» is "STOP?" — and the strip makes it identical to the bare keyword. Measured against this code:

```
UNSUBSCRIBES  "ΣΤΟΠ;"  "ΣΤΟΠ?"  "stop?"  "stop 2"  "-> stop <-"
-> agent      "stop stop"  "stopping"  "θέλω να απεγγραφώ"
```

Stripping `!` is emphasis-tolerance; stripping `;` inverts the meaning in the language this ships in. The trailing-digit cases are implausible as real input — they're just evidence the class is too wide.

What makes this the one I'd fix first is that there's no way back: `status: "active"` is written only in the create branch of `enrollFromInbound`, the update branch doesn't restore it, and no agent tool re-subscribes. Excluding `;` and `?` sends these to the agent, which §3 already directs to let people go immediately. Worth weighting because §2.1 makes `unsubscribed` terminal until an explicit re-subscribe, and that path is PR 5.

services/notis/src/lib/queue.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  const { outcome, trace } = await runWake(state, event, deps);
+
+  const delivery =
+    outcome.messages.length > 0
+      ? decideDelivery(event.type, lastInbound?.createdAt.toISOString(), new Date(event.at))
+      : undefined;
+
+  const outboundIds = await db.$transaction(async (tx) => {
+    const { _max } = await tx.notisJournalEntry.aggregate({
```
kouloumos: This and `handleBareStop` both allocate `seq` as `MAX(seq)+1` with no lock, and ΣΤΟΠ runs outside the queue's serialization entirely. Overlap them and one dies on `NotisJournalEntry_subscriptionId_seq_key` — reproduced on real Postgres. Because the two transactions take `NotisSubscription` and `NotisJournalEntry` in opposite order they can also deadlock outright (`40P01`).

§4.2's at-least-once contract survives this: when the wake loses, its rollback undoes `completeItem`, `failItem`'s fence still matches, the item returns to `pending` and the sweeper re-runs it — nothing was sent, so there's no duplicate delivery. What it costs is a second paid model call for one event, and when ΣΤΟΠ is the loser instead, a 500 that delays the unsubscribe by a Bird retry. It's also the case the body's stronger claim ("once the model has run, a retry never re-runs it") doesn't cover.

Both paths already `UPDATE` the subscription row, so moving that update above the aggregate lets the existing row lock serialize them — verified, the second transaction blocks then gets the next seq cleanly. It holds the lock slightly longer, which matters given the next comment.

services/notis/src/lib/queue.ts
```diff
@@ -0,0 +1,376 @@
...
+  };
+
+  const { outcome, trace } = await runWake(state, event, deps);
+
+  const delivery =
+    outcome.messages.length > 0
+      ? decideDelivery(event.type, lastInbound?.createdAt.toISOString(), new Date(event.at))
+      : undefined;
+
+  const outboundIds = await db.$transaction(async (tx) => {
```
kouloumos: No `transactionOptions` on the client, so this runs under Prisma's 5,000 ms default. Blowing it gives the same outcome as the seq collision: `P2028`, full rollback after the model ran, `failItem`, re-run. The write count isn't the risk (~2,700 statements would fit) — lock contention on the subscription row is, and every ordinary inbound webhook takes that lock too.

Also worth knowing that `P2028` doesn't fire at 5s; Prisma only notices on the next query, so a 30s lock wait hangs the full 30s and then dies.

services/notis/src/lib/queue.ts
```diff
@@ -0,0 +1,376 @@
...
+    await db.notisMessage.updateMany({
+      where: { id: { in: messageIds } },
+      data: { status: "failed", failureReason: "no birdConversationId on subscription" },
+    });
+    await alert(`subscription ${sub.id} has no birdConversationId — cannot deliver replies`);
+    return;
+  }
+
+  for (const id of messageIds) {
+    const message = await db.notisMessage.findUnique({ where: { id } });
```
kouloumos: Check-then-act with no lock, and nothing stops two runs overlapping: `setInterval(sweep, 60_000)` never awaits the prior run, `RESEND_STALE_AFTER_MS` is 120s, and this `fetch` sets no timeout — unlike `anthropic.ts`, which this PR does bound to three minutes. A row that qualifies at sweep T still qualifies at T+60s while the first `sendText` is unreturned, because the row is only updated after it resolves. Measured two concurrent calls for one row carrying the same key.

The duplicate is a duplicate API call, not a proven duplicate message — but that rests on Bird deduplicating two *concurrent* requests, which is the weakest case for any idempotency implementation and is the item your own deploy checklist still lists as unverified. A conditional claim (`pending → sending`, act on `count === 1`) is the same pattern `claimNext` already uses; an `AbortSignal.timeout` shorter than the sweep interval is the cheap version.

services/notis/src/lib/queue-core.ts
```diff
@@ -0,0 +1,165 @@
...
+ *  already-completed item is left untouched. */
+export async function failItem(
+  db: Db,
+  id: string,
+  attempts: number,
+  error: string,
+): Promise<void> {
+  await db.notisWakeQueue.updateMany({
+    where: { id, status: "running", attempts },
+    data: { status: "pending", claimedAt: null, lastError: error.slice(0, 2000) },
```
kouloumos: `runAfter` isn't advanced here, and the drain loop calls `claimNext` again on the next iteration — so all three attempts land against the same outage inside one `drainQueue` call. A 30-second Anthropic blip therefore drops the message permanently; the attempt budget buys nothing in the case it exists for. `failItem` already receives `attempts`, so a backoff is computable right here.

services/notis/src/lib/queue-core.ts
```diff
@@ -0,0 +1,165 @@
...
+            c.status = 'pending'::"QueueItemStatus"
+            OR (c.status = 'running'::"QueueItemStatus"
+                AND c."claimedAt" < now() - make_interval(secs => ${staleSeconds}))
+          )
+          AND NOT EXISTS (
+            SELECT 1 FROM "NotisWakeQueue" r
+            WHERE r."subscriptionId" = c."subscriptionId"
+              AND r.status = 'running'::"QueueItemStatus"
+              AND r.id <> c.id
+              AND r."claimedAt" >= now() - make_interval(secs => ${staleSeconds})
```
kouloumos: This excludes only *fresh* running rows, so a subscription with a stale running row still offers its pending rows as candidates. If a pending row sorts earlier, the UPDATE violates `NotisWakeQueue_one_running_per_sub`, `isUniqueViolation` swallows it to `null`, and `drainQueue` reads that as "queue empty" and breaks — starving every *other* subscription too, permanently.

Not reachable today: `enqueueLiveWake` is the only `runAfter` writer and always uses `new Date()`, so the stale row always sorts first. It needs clock skew across instances — or PR 4's poller, which enqueues at past times by design. Dropping the `claimedAt >=` clause fixes it; I checked that the reclaim then works and the drain moves on to the next subscription.

services/notis/src/lib/bird.ts
```diff
@@ -0,0 +1,99 @@
...
+        }),
+      });
+
+      if (!response.ok) {
+        const errorText = await response.text();
+        console.error(`Bird send failed (${response.status}):`, errorText);
+        return {
+          success: false,
+          status: response.status,
+          retryable: response.status >= 500,
```
kouloumos: Confirming Greptile's open thread is still live at this head — 408 and 429 classify as terminal, so a rate-limited reply is never retried by the sweeper.

src/app/api/webhooks/bird/route.ts
```diff
@@ -308,6 +309,15 @@ export async function POST(request: Request) {
         return NextResponse.json({ ok: true });
     }
 
+    // Notis-served users belong to the notis webhook subscription: no reply,
+    // no unsubscribe handling, no Message row (notis's database is the record
+    // of those conversations). Legacy outbound reconciliation stays above —
+    // a Message row that exists here is this app's send regardless of flag.
+    if (await isNotisServedPhone(fields.phone)) {
```
kouloumos: This returns before `handleUnsubscribeIfApplicable`, so a ΣΤΟΠ from a notis-served user no longer clears `NotificationPreference.notifyByPhone`. Per §2.1 that's correct as an endpoint — notis holds the only copy and PR 6 drops the flag.

The gap is the window before PR 5. `deliver.ts` skips these users while `notisEnabledAt` is set, so the opt-out is honoured; `setNotisEnabled(user, false)` is where it bites, because the old path then sees `notifyByPhone: true` and resumes, and the release panel shows no sign the person ever sent ΣΤΟΠ. §9 says narrowing never unsubscribes anyone — this is the mirror case, where narrowing re-subscribes someone who did.

services/notis/src/lib/queue.ts
kouloumos: Check-then-act with no lock, and nothing stops two runs overlapping: `setInterval(sweep, 60_000)` never awaits the prior run, `RESEND_STALE_AFTER_MS` is 120s, and this `fetch` sets no timeout — unlike `anthropic.ts`, which this PR does bound to three minutes. A row that qualifies at sweep T still qualifies at T+60s while the first `sendText` is unreturned, because the row is only updated after it resolves. Measured two concurrent calls for one row carrying the same key.

The duplicate is a duplicate API call, not a proven duplicate message — but that rests on Bird deduplicating two *concurrent* requests, which is the weakest case for any idempotency implementation and is the item your own deploy checklist still lists as unverified. A conditional claim (`pending → sending`, act on `count === 1`) is the same pattern `claimNext` already uses; an `AbortSignal.timeout` shorter than the sweep interval is the cheap version — `e146f742` just did exactly that for the alert webhook, which leaves the Bird send as the only unbounded outbound call in the service.

services/notis/src/lib/queue-core.ts
kouloumos: `runAfter` isn't advanced here, and the drain loop calls `claimNext` again on the next iteration — so all three attempts land against the same outage inside one `drainQueue` call. A 30-second Anthropic blip therefore drops the message permanently; the attempt budget buys nothing in the case it exists for. `failItem` already receives `attempts`, so a backoff is computable right here.

services/notis/src/lib/queue-core.ts
kouloumos: This excludes only *fresh* running rows, so a subscription with a stale running row still offers its pending rows as candidates. If a pending row sorts earlier, the UPDATE violates `NotisWakeQueue_one_running_per_sub`, `isUniqueViolation` swallows it to `null`, and `drainQueue` reads that as "queue empty" and breaks — starving every *other* subscription too, permanently.

Not reachable today: `enqueueLiveWake` is the only `runAfter` writer and always uses `new Date()`, so the stale row always sorts first. It needs clock skew across instances — or PR 4's poller, which enqueues at past times by design. Dropping the `claimedAt >=` clause fixes it; I checked that the reclaim then works and the drain moves on to the next subscription.

services/notis/src/app/api/webhooks/bird/handlers.ts
```diff
@@ -0,0 +1,290 @@
...
+    if (hasMainDb()) {
+      const flag = await mainDb().notisUserRow.findUnique({
+        where: { id: sub.userId },
+        select: { notisEnabledAt: true },
+      });
+      if (!flag?.notisEnabledAt) {
+        return { action: "ignored", reason: "user rolled back to the old path" };
+      }
+    }
+    if (fields.conversationId && sub.birdConversationId !== fields.conversationId) {
```
kouloumos: This refreshes `birdConversationId` but never `phone`, and `NotisSubscription.phone` is only ever written by `enrollFromInbound`. So after someone changes their number in their profile, a message from the *old* number: the main app's gate looks up `User.phone`, misses, and falls through to `sendUnsupportedReply`; notis still matches the subscription here and answers properly. Two replies, one contradicting the other, repeating on every message.

From the new number it self-heals via the upsert, and a ΣΤΟΠ from the old number is safe because the main app's phone-mismatch guard suppresses its reply. Selecting `phone` alongside `notisEnabledAt` and treating a mismatch like a rollback would close it.

services/notis/src/lib/__tests__/fake-db.ts
```diff
@@ -0,0 +1,182 @@
...
+          return existing;
+        }
+        const row: Row = { id: id("sub"), status: "active", unsubscribedAt: null, ...create };
+        store.subscriptions.set(row.id as string, row);
+        calls.push("subscription-upserted:create");
+        return row;
+      },
+    },
+    notisJournalEntry: {
+      findMany: async ({ take }: { take?: number } = {}) => {
```
kouloumos: `findMany` takes only `take`, `aggregate` takes no arguments, and `notisMessage.findFirst` ignores its `where` too — but the three real calls in `runOneWake` all scope by `subscriptionId`. I removed that scoping from the journal window, the seq aggregate and the last-inbound lookup at once, and all 139 tests still passed. That mutation is one reader's history feeding another reader's wake, and a seq allocated from a stranger's sequence. A small `matchesWhere` (scalar equality plus `lt`/`in`) closes the class.

services/notis/src/lib/bird.ts
kouloumos: Confirming Greptile's open thread is still live at this head — 408 and 429 classify as terminal, so a rate-limited reply is never retried by the sweeper.

services/notis/src/lib/queue.ts
kouloumos, line 141: This and `handleBareStop` both allocate `seq` as `MAX(seq)+1` with no lock, and ΣΤΟΠ runs outside the queue's serialization entirely. Overlap them and one dies on `NotisJournalEntry_subscriptionId_seq_key` — reproduced on real Postgres. Because the two transactions take `NotisSubscription` and `NotisJournalEntry` in opposite order they can also deadlock outright (`40P01`).

§4.2's at-least-once contract survives this: when the wake loses, its rollback undoes `completeItem`, `failItem`'s fence still matches, the item returns to `pending` and the sweeper re-runs it — nothing was sent, so there's no duplicate delivery. What it costs is a second paid model call for one event, and when ΣΤΟΠ is the loser instead, a 500 that delays the unsubscribe by a Bird retry. It's also the case the body's stronger claim ("once the model has run, a retry never re-runs it") doesn't cover.

Both paths already `UPDATE` the subscription row, so moving that update above the aggregate lets the existing row lock serialize them — verified, the second transaction blocks then gets the next seq cleanly. It holds the lock slightly longer, which matters given the next comment.

services/notis/src/lib/queue.ts
kouloumos: No `transactionOptions` on the client, so this runs under Prisma's 5,000 ms default. Blowing it gives the same outcome as the seq collision: `P2028`, full rollback after the model ran, `failItem`, re-run. The write count isn't the risk (~2,700 statements would fit) — lock contention on the subscription row is, and every ordinary inbound webhook takes that lock too.

Also worth knowing that `P2028` doesn't fire at 5s; Prisma only notices on the next query, so a 30s lock wait hangs the full 30s and then dies.

services/notis/src/lib/queue.ts
kouloumos, line 257: Check-then-act with no lock, and nothing stops two runs overlapping: `setInterval(sweep, 60_000)` never awaits the prior run, `RESEND_STALE_AFTER_MS` is 120s, and this `fetch` sets no timeout — unlike `anthropic.ts`, which this PR does bound to three minutes. A row that qualifies at sweep T still qualifies at T+60s while the first `sendText` is unreturned, because the row is only updated after it resolves. Measured two concurrent calls for one row carrying the same key.

The duplicate is a duplicate API call, not a proven duplicate message — but that rests on Bird deduplicating two *concurrent* requests, which is the weakest case for any idempotency implementation and is the item your own deploy checklist still lists as unverified. A conditional claim (`pending → sending`, act on `count === 1`) is the same pattern `claimNext` already uses; an `AbortSignal.timeout` shorter than the sweep interval is the cheap version — `e146f742` just did exactly that for the alert webhook, which leaves the Bird send as the only unbounded outbound call in the service.

services/notis/src/lib/queue-core.ts
kouloumos: `runAfter` isn't advanced here, and the drain loop calls `claimNext` again on the next iteration — so all three attempts land against the same outage inside one `drainQueue` call. A 30-second Anthropic blip therefore drops the message permanently; the attempt budget buys nothing in the case it exists for. `failItem` already receives `attempts`, so a backoff is computable right here.

services/notis/src/lib/queue-core.ts
kouloumos: This excludes only *fresh* running rows, so a subscription with a stale running row still offers its pending rows as candidates. If a pending row sorts earlier, the UPDATE violates `NotisWakeQueue_one_running_per_sub`, `isUniqueViolation` swallows it to `null`, and `drainQueue` reads that as "queue empty" and breaks — starving every *other* subscription too, permanently.

Not reachable today: `enqueueLiveWake` is the only `runAfter` writer and always uses `new Date()`, so the stale row always sorts first. It needs clock skew across instances — or PR 4's poller, which enqueues at past times by design. Dropping the `claimedAt >=` clause fixes it; I checked that the reclaim then works and the drain moves on to the next subscription.

services/notis/src/app/api/webhooks/bird/handlers.ts
kouloumos, line 283: This refreshes `birdConversationId` but never `phone`, and `NotisSubscription.phone` is only ever written by `enrollFromInbound`. So after someone changes their number in their profile, a message from the *old* number: the main app's gate looks up `User.phone`, misses, and falls through to `sendUnsupportedReply`; notis still matches the subscription here and answers properly. Two replies, one contradicting the other, repeating on every message.

From the new number it self-heals via the upsert, and a ΣΤΟΠ from the old number is safe because the main app's phone-mismatch guard suppresses its reply. Selecting `phone` alongside `notisEnabledAt` and treating a mismatch like a rollback would close it.

services/notis/src/lib/__tests__/fake-db.ts
kouloumos: `findMany` takes only `take`, `aggregate` takes no arguments, and `notisMessage.findFirst` ignores its `where` too — but the three real calls in `runOneWake` all scope by `subscriptionId`. I removed that scoping from the journal window, the seq aggregate and the last-inbound lookup at once, and all 139 tests still passed. That mutation is one reader's history feeding another reader's wake, and a seq allocated from a stranger's sequence. A small `matchesWhere` (scalar equality plus `lt`/`in`) closes the class.

services/notis/src/lib/bird.ts
kouloumos: Confirming Greptile's open thread is still live at this head — 408 and 429 classify as terminal, so a rate-limited reply is never retried by the sweeper.

src/app/api/webhooks/bird/route.ts
kouloumos, line 329: This returns before `handleUnsubscribeIfApplicable`, so a ΣΤΟΠ from a notis-served user no longer clears `NotificationPreference.notifyByPhone` — and notis cannot write it, by design. Per §2.1 that is the intended end state: notis holds the only copy.

Where it bites is the rollout toggle. `deliver.ts` skips these users while `notisEnabledAt` is set, so the opt-out is honoured; clearing it via `setNotisEnabled(user, false)` hands them back to the old path, which sees `notifyByPhone: true` and resumes messaging someone who explicitly asked it to stop. The release panel shows no sign that person ever sent ΣΤΟΠ, so there is nothing to warn whoever flips it. §9 says narrowing never unsubscribes anyone — this is the mirror case.

services/notis/src/app/admin/(panel)/_lib/conversations.ts
```diff
@@ -126,10 +238,55 @@ export async function getConversation(id: string): Promise<ConversationDetail |
...
+        messages: entry.messages,
+        scheduledWakes: [],
+        journalAppend: entry,
+        ...(entry.unsubscribed ? { unsubscribe: { reason: entry.rationale } } : {}),
+      },
+      ...(entry.messages.length > 0 ? { delivery: { mode: "freeform" as const } } : {}),
+      ...(deliveries.length > 0 ? { deliveries } : {}),
+    });
+  }
+  records.sort((a, b) => a.event.at.localeCompare(b.event.at));
```
kouloumos: Sorting on `event.at` means a wake's replies are placed by when it was *triggered*, not when they were sent — and `WhatsAppChat.tsx:545,557` stamp the bubbles from the same field, so they display that time too.

Observed on a live thread. A question arrived 18:56:44 and started a 36s wake; the reader sent ΣΤΟΠ at 18:56:49; the confirmation went out at 18:56:49; the four agent replies were written at 18:57:20. The replies carry the label 18:56:44, so they sort *before* the ΣΤΟΠ, and the thread reads question → answer → ΣΤΟΠ. What actually reached the reader was question → ΣΤΟΠ → «δεν θα σου ξαναγράψω εδώ» → four more messages 31 seconds later.

Everything is on screen, so nothing is hidden — but the anomaly stops looking like one, which matters because this viewer is where you would go to answer "did we message someone after they unsubscribed?".

Today the error is one wake duration. Once wakes fire on world events, `event.at` becomes the event's own time, so a reply written now could sort under a meeting from last week. Sorting and stamping on each message's own `createdAt` fixes both; `MessageDelivery` (`_lib/records.ts:53`) would need to carry it, since line 217 currently drops it.

services/notis/src/app/admin/(panel)/_lib/metrics.ts
```diff
@@ -1,93 +1,377 @@
...
+  }
+
+  const db = notisDb();
+
+  const [current, previous, series, recent, subscriptions, unsubscribed] = await Promise.all([
+    periodStats(db, currentFrom, now),
+    periodStats(db, previousFrom, currentFrom),
+    bucketedSeries(db, currentFrom, now, bucket),
+    db.notisMessage.findMany({
+      where: { direction: "inbound" },
```
kouloumos: No date filter here, so «Τι λένε οι χρήστες» ignores the selected range. Backdated a message three days and it still appeared under `?range=1h`, directly beneath an ΕΛΗΦΘΗΣΑΝ counter reading 0 for the same window. Every other number on the page obeys the range.

services/notis/src/app/admin/(panel)/_lib/metrics.ts
```diff
@@ -1,93 +1,377 @@
...
+/** Chart resolution per range: sub-day windows need sub-day buckets. */
+export type BucketUnit = "minute" | "hour" | "day";
+
+export const RANGES = {
+  "1h": { ms: HOUR_MS, label: "1 ώρα", short: "1ω", bucket: "minute" as BucketUnit },
+  "24h": { ms: DAY_MS, label: "24 ώρες", short: "24ω", bucket: "hour" as BucketUnit },
+  "7d": { ms: 7 * DAY_MS, label: "7 ημέρες", short: "7ημ", bucket: "day" as BucketUnit },
+  "14d": { ms: 14 * DAY_MS, label: "14 ημέρες", short: "14ημ", bucket: "day" as BucketUnit },
+  "30d": { ms: 30 * DAY_MS, label: "30 ημέρες", short: "30ημ", bucket: "day" as BucketUnit },
+  "90d": { ms: 90 * DAY_MS, label: "3 μήνες", short: "3μ", bucket: "day" as BucketUnit },
```
kouloumos: The header concatenates a fixed «τελευταίες» onto these labels, which renders «τελευταίες 3 μήνες» (μήνες is masculine) and «τελευταίες 1 ώρα» (singular). Only 24 ώρες and the day ranges read correctly. Moving the whole phrase into `RANGES` would fix both.

kouloumos (commented):
Read this against the PRD. The transaction ordering, the claim fence, and keeping the notis role read-only against the main database all hold up under concurrent load, and §4.2's at-least-once contract is satisfied. Most of what I found sits at the seams — between the two services, between the deterministic ΣΤΟΠ path and the queue, and between committing and sending — which is the expected shape for the PR that joins them.

The one I'd want settled rather than patched is where a withdrawal of consent lives. A ΣΤΟΠ from a notis-served reader is recorded in notis alone, and the main app's copy keeps its old value — which is the intended end state, but it means the rollout toggle can put someone back in reach of the old path after they asked to be left alone. Detail inline.

Separately, and only so it isn't lost: losing a phone number becomes an unsubscribe rather than a pause. I gather that is a wrinkle you already carry today, so I am not treating it as this PR's problem.

On review state: of the four bot threads reading as resolved, three were resolved by the bot itself after a force-push with no reply from you, and the code they point at is unchanged at this head. Greptile's 408/429 classification is also still live.

I ran the inbound path against a local stack — notis on its own database, the main schema with the views, no Bird credentials so every send fails at the boundary. The gate, dedupe, signature rejection, enrollment and both ΣΤΟΠ paths behave as documented, and a real wake produced a properly grounded four-message answer for $0.08. Three of the comments below came out of that.

christosporios:
Pushed two commits, both follow-through from the review of #632:

- `e146f742` — every operational alarm carries a 5s timeout. The janitor raises them beside a database lock and the poller inside a tick, so a wedged webhook must not become a wedged caller. (The janitor's own alarm also moved out of its transaction in #632.)
- `9ee631b8` — the `cities` snapshot is gone at its source, so the three places here that read or wrote it follow: inbound enrollment stops storing a copy, wake assembly reads the view (an outage now fails the item into a retry instead of telling the agent this reader follows nothing), and the conversation panel takes names from the view.

Rebased onto the updated #632.

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/9ee631b8d936915f47ec83605d8b7237304163bb..f4571f546f5e00c454007628e066f4c5f13e78e0>) to rebase onto `main` after #642 merged. Same diff; the only resolution was a duplicate `.gitignore` rule for `services/notis/generated/` that #642 added first — main's wording is kept.

christosporios:
Thank you — the local run and the mutation test earned their keep; two of these were only visible that way.

**Decisions taken**, both yours:

**Where consent lives: documented, deferred to the rollout PRs.** You're right that this is the case §9's promise misses, and we're not adding a second copy of the opt-out to close it. The reason is that the copies would then be able to disagree in the unsafe direction too, and the answer we want — one place that can say "is this reader subscribed" — is exactly what PR 5 builds when the profile checkbox becomes a client of the notis API. So it is written down instead: at the early return in `route.ts` and in the notis README's known gaps, stating that clearing `notisEnabledAt` for an individual needs a look at their notis subscription first. Until PR 5, that is a procedure rather than a guarantee, and calling it that seemed better than pretending otherwise.

**The sweeper claims at the database.** New `sendingAt` column, claimed in a fenced `updateMany`, and the run acts only when `count === 1`. The send is bounded too (`AbortSignal.timeout`, same 30s as the claim's staleness window), so a hung call cannot hold a claim it will never release. "No duplicate message" no longer rests on Bird deduplicating two simultaneous requests.

**Everything else in your list is fixed**, and each fix is re-established in #639 where that code is rewritten:

- **Seq collision and deadlock.** Both writers already `UPDATE` the subscription, so the persist now takes that lock FIRST and the second transaction blocks until the seq is free — your suggestion exactly. The transaction also carries an explicit 30s budget, and I noted your point that P2028 does not fire at the deadline.
- **`failItem` backoff** — 1, 4, 9 minutes. The integration test that walked attempts in a tight loop now asserts the delay and fast-forwards past it, which reads better than the old loop did.
- **The claim-starvation guard** is fixed here rather than in #639, where the poller makes it reachable. Dropped the `claimedAt >=` clause as you suggested.
- **408 and 429 are retryable**, which also closes Greptile's and Cursor's threads on `bird.ts:75`. Two of the other bot threads are closed by the same commit: the give-up sweep is fenced on `pending` so it cannot rewrite a successful send, and `delivered` no longer ranks below `failed` — `failed` sits with `sent`, so only a still-`sent` row can turn out to have failed. A successful retry also clears the failureReason of the earlier attempt.
- **The phone seam.** The gate now selects `phone` alongside `notisEnabledAt` and treats a mismatch like a rollback: notis stays silent, the main app answers, and the new number self-heals through the upsert. Writing a test for the "refresh the stored phone" half showed that branch was unreachable for a genuinely different number — the lookup is by phone, so the old subscription is never found by the new one — so it now only canonicalizes a stored value that differs by the leading `+`, and says so.
- **The fake database has a real where engine.** Your mutation is the test I added: two readers, the stranger with the longer history, asserting the new entry continues *our* sequence and that the stranger's rationale never reaches the model. I checked it fails under exactly the mutation you described — the engine alone did not catch it, because no test had a second reader.
- **Thread ordering.** `MessageDelivery` carries the row's `createdAt`, and both the sort and the bubble stamps use it. Your live example is in the comment, because it explains the stakes better than the rule does.
- **Metrics**: the messages panel obeys the range, and the range labels carry their whole phrase so «τους τελευταίους 3 μήνες» and «την τελευταία ώρα» agree.

On the phone-loss-is-an-unsubscribe wrinkle: agreed it predates this PR, and it stays on the list for the rollout work.

Verification: 153 unit tests here and 212 in #639, 33 notis integration tests against real Postgres, both typecheck scopes clean. The stack is rebased onto the current main.

christosporios:
Merged `feat/notis-db` forward (`42a238ae`, plain merge, no conflicts) to pick up current main plus the paired-preview wiring — see [the #632 comment](https://github.com/schemalabz/opencouncil/pull/632#issuecomment-5385384474) for what that brought. Verified on this branch: notis 21 suites / 153 tests, root units, integration 11 suites / 84 tests, tsc clean.

christosporios:
Merged `feat/notis-db` forward (`59c859d0`) to pick up the two review fixes — the RSC-bypass panel guards (#1) and migrate-on-start (#2); see the [#632 comment](https://github.com/schemalabz/opencouncil/pull/632#issuecomment-5385784756). Conflicts were in the panel files PR 3 also touched; each resolution keeps PR 3's version and grafts the per-page guard. The `server-only` imports needed a jest mapping (`server-only`/`client-only` → empty module), added to this app's jest config. Verified: notis 22 suites / 155 tests, root units, integration, tsc clean.

christosporios:
Merged the build fix forward (`76e1802a`). See the [#632 note](https://github.com/schemalabz/opencouncil/pull/632) — `server-only` removed from the two mixed `_lib` modules so the production build passes. Verified locally: build + tsc + tests green.

christosporios:
Merged `feat/notis-db` forward (`6032e040`) — the shared notification-status rule (#13) and the preview sign-in allowlist (#14); see [the #632 comment](https://github.com/schemalabz/opencouncil/pull/632). No conflicts; verified green on this branch.

PR #636, opened 2026-08-17
feat(reviews): auto-run summarize when human review completes

christosporios:
## Summary

Summarize follows human review in the pipeline, but a reviewer had to start it by hand from the meeting admin panel. A reviewed transcript could sit without a summary for days.

`markHumanReviewComplete` now starts summarize when the reviewer opts in. Both dialogs that complete a review show the option, together with an explanation of what happens to the notifications that summarize creates.

## What changes

**`autoTriggerTask` (`src/lib/tasks/autoTrigger.ts`)**

`handleTranscribeResult` already chained `fixTranscript` inside a catch-all guard: the transcribe import succeeded, so a throw would report that successful step as failed. Human review needs the same guard. The guard, the log line and the Discord failure alert move into one helper that both chains use.

`startTask` now throws `TaskAlreadyExistsError` when the idempotency guard blocks a pipeline task. `autoTriggerTask` reports that state as a skip, not as a failure. A caller needs no pre-check of its own, so two checks cannot disagree.

**The completion dialogs**

`useReviewCompletion` and `ReviewCompletionOptions` hold the option state and the fetch that feeds it. `CheckboxCard` replaces the markup that the option checkboxes repeated, and it lives in `packages/ui` beside the other shared primitives.

## Notification safety

Summarize creates the notifications for a meeting. For a `NOTIFICATIONS_AUTO` administrative body, `handleSummarizeResult` releases them at once, with no second confirmation. The dialog therefore states the outcome for each notification behavior.

The option defaults to off when summarize would notify subscribers about a meeting that is not released yet. That send has no undo, and the notification links to a page that the public cannot open.

## Review notes

This branch went through a review pass. The following defects were found and fixed here:

- A retry could not finish a review that stopped after the review record. The record commits before the follow-up actions, so a timeout between the two left a complete review with no summary and no transcript email. The follow-up actions now run on every call. Each one keeps its own idempotency guard.
- Both dialogs could confirm with options they never loaded, which completed the review with every follow-up silently off.
- A race with a manual summarize produced a red Discord "failed" alert for a benign duplicate.
- `useReviewCompletion` reported `isLoading: false` on the first frame, and it kept the previous meeting's answers on reopen.
- A failed load in `SendTranscriptButton` read as "no contact emails configured", which points an operator at the wrong problem.
- The cache revalidation ran before the follow-up tasks existed, so the task list of the meeting could stay stale until the backend reported a result.

Findings that this PR reports but does not change:

- Notifications ignore `CouncilMeeting.released` in `createNotificationsForMeeting` and `releaseNotifications`. This PR lowers the risk with the default above, but the gap is older than this branch and deserves its own fix.
- The completion action loads the whole transcript twice on the request path (once for the summarize body, once for the DOCX). The two follow-ups now run in parallel, but moving them off the response path is a separate change.
- A re-transcribed meeting cannot be re-summarized from this dialog, because the idempotency guard has no `force` escape hatch here. The meeting admin panel still offers one.
- The superadmin reviews panel is untranslated English, so the translated option cards now sit inside an English dialog.

## Testing

- `npx tsc --noEmit` — clean.
- `npm test` — 105 suites, 1416 tests pass.
- `npx eslint src packages` — clean.
- `npm run build` — compiles and typechecks. Page-data collection needs Elasticsearch credentials that this environment does not have.
- Each commit typechecks and passes the suite on its own.

`src/lib/tasks/__tests__/markHumanReviewComplete.test.ts` covers the opt-in, the skip, the failed trigger, the retry, and `getReviewCompletionState` including its authorization check.

https://claude.ai/code/session_012RPsGPoMLgBiP99ncR5fQH

---
_Generated by [Claude Code](https://claude.ai/code/session_012RPsGPoMLgBiP99ncR5fQH)_

<h3>Greptile Summary</h3>

The PR lets reviewers start summarization when they complete a human review.
- Adds review-completion options for summarization and transcript delivery.
- Reports follow-up outcomes without reverting a completed review.
- Centralizes automatic pipeline triggering and duplicate-task handling.
- Reloads meeting state after follow-up tasks start.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/tasks/humanReview.ts | Completes the review once, runs selected follow-ups on each request, and returns each follow-up outcome. |
| src/lib/tasks/autoTrigger.ts | Centralizes guarded pipeline triggering and treats an existing task as a benign skip. |
| src/lib/tasks/tasks.ts | Introduces a typed error for pipeline tasks blocked by the existing idempotency check. |
| src/components/reviews/useReviewCompletion.ts | Loads completion state, resets stale choices, and derives safe option defaults. |
| src/components/reviews/CompleteReviewDialog.tsx | Presents the new options and prevents confirmation before those options load. |
| src/components/admin/reviews/MarkReviewCompleteButton.tsx | Uses the shared completion flow and preserves visible follow-up failures. |

<sub>Reviews (4): Last reviewed commit: ["feat(reviews): auto-run summarize when h..."](https://github.com/schemalabz/opencouncil/commit/faf48d78114bfd81ff8b71001924d84812e6ebdf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53976768)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/7e9fdd8f0ba705e66296bdc3b4c3b7d832fa7ca8..f8616bad5f0ea9b0d76ba1e4b4d4f0b1b7e9c4a1) to separate the two loads of the completion dialog:

- The review statistics and the completion options now load and fail on their own. A failure of the statistics reported itself through a shared error, which hid the whole body of the dialog — the option checkboxes included. The reviewer then saw an error and no way to complete the review.
- The confirm button no longer blocks on a statistics failure. The statistics only report the work that the reviewer did. The button still blocks while the options are absent, because the reviewer must see the notification warning before summarize starts.
- Each failure names the part that failed, in all four locales.

The preview showed this state on a large council meeting: `getMeetingReviewStats` failed, and the dialog rendered only the generic production error. That function loads every utterance of a meeting together with its edit rows. It is older than this branch, and this branch does not change it. The change above stops it from blocking the review completion, but the underlying load deserves its own fix.

---

christosporios:
Correction: the compare link above carries a wrong head SHA. The correct one is [7e9fdd8..f8616ba](https://github.com/schemalabz/opencouncil/compare/7e9fdd8f0ba705e66296bdc3b4c3b7d832fa7ca8..f8616bad2839ffd52463fab80585a6dc22ede6d0).

---

src/lib/tasks/humanReview.ts
```diff
@@ -1,76 +1,163 @@
...
     });
-    
+
+    if (!meeting) {
+        throw new Error(`Council meeting ${cityId}/${meetingId} not found`);
+    }
+
+    const created = existingReview ?? await createHumanReviewRecord(cityId, meetingId, meeting, manualReviewTime);
+
+    await runReviewFollowUps(cityId, meetingId, meeting, created.id, { sendTranscript, runSummarize });
```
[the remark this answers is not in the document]
christosporios: The window is real, but it sits inside `sendTranscriptToMunicipality`, and this branch does not open it.

That function checks `transcriptSent`, renders the DOCX, calls Resend, and only then writes the `transcriptSent` record (`src/lib/tasks/sendTranscript.ts:35-128`). A process that dies between the send and the write leaves no record, so a later attempt sends again. The admin "Send Transcript to Municipality" button reaches that same window on `main` today, because it calls the function directly with no other guard.

This branch adds a second path to it. The follow-ups now run on every call, which is the fix for a worse defect: the review record commits before the follow-ups, so a timeout between the two used to leave a complete review with no summary and no transcript email, and the retry returned success without running anything.

I am not closing the window here. The correct fix is a claim: write the record first as pending, send, then mark it succeeded or failed. `checkTaskIdempotency` already treats a non-terminal record as "already running", so a retry would skip. That trades a rare double send for a rare permanent block, because a process that dies after the claim leaves a pending record that blocks every later send. It needs a stale-claim recovery to be correct, it changes a shared task helper that this branch does not otherwise touch, and it deserves its own change with its own tests.

---

src/lib/tasks/tasks.ts
```diff
@@ -86,11 +86,7 @@ export const startTask = async (taskType: MeetingTaskType, requestBody: any, cou
...
         const idempotency = await checkTaskIdempotency(taskType, cityId, councilMeetingId, options);
         if (!idempotency.proceed) {
-            throw new Error(
-                idempotency.blockedReason === 'already_succeeded'
-                    ? `A ${taskType} task has already succeeded for this council meeting`
-                    : `A ${taskType} task is already running for this council meeting`
-            );
+            throw new TaskAlreadyExistsError(taskType, idempotency.blockedReason ?? 'already_running');
         }
     }
```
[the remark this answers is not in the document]
christosporios: The race is real. `startTask` reads with `checkTaskIdempotency` and then writes with `prisma.taskStatus.create`, and `TaskStatus` carries no unique constraint — only `@@index([councilMeetingId, cityId])` (`prisma/schema.prisma:277`). Two requests can pass the read before either writes.

This branch does not introduce it. The diff of that block changes one thing: `startTask` throws `TaskAlreadyExistsError` instead of a generic `Error`. The read-then-write structure matches `main`, where a double click on Summarize in the meeting admin panel reaches the same window.

This branch does widen the exposure, in two ways worth stating:

- Summarize now starts from a routine reviewer action, not only from a deliberate admin click.
- `requestSummarize` builds the request body before it calls `startTask`, and that body load reads the whole transcript. Two overlapping requests therefore spend seconds in the same slow load and arrive at the read-write window together, so the window is easier to hit than its width suggests.

I am not fixing it here. The sound fix is a guarantee in the database: a partial unique index on `(cityId, councilMeetingId, type)` for rows that are not in a terminal state. Prisma expresses a partial index only as raw SQL in a migration, the constraint then applies to every pipeline task type, and a stale non-terminal row would block every later run for that meeting. That last point makes it need the same stale-claim recovery as the thread above.

Both P1 comments on this PR point at one missing piece: the task system has no atomic claim. That belongs in a single change that adds the claim and the recovery together, with tests and a migration tested against a real database.

---

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/f8616bad2839ffd52463fab80585a6dc22ede6d0..23e3b8879bee715b879ed232eb9429389c9dcad5) to fix the alignment of the option cards and to name the task in the label:

- `DialogContent` centers its content by default, so the option cards inherited centered text: the checkbox sat left of a centered label. `CheckboxCard` now sets `text-left` and fills the width of its container, which fixes it in every dialog that uses the card.
- The body of the completion dialog aligns left. The header keeps the centered style of the dialog.
- The label of the option now reads `Run summarize task`, and it keeps `summarize task` in all four locales. The label names the task that the option starts, which matches the term that the meeting admin panel uses.

---

src/lib/tasks/humanReview.ts
```diff
@@ -111,28 +198,58 @@ export async function markHumanReviewComplete(
...
+ * Send the transcript to the municipality and record the outcome.
+ * sendTranscriptToMunicipality reports a failure in its result and never throws,
+ * so an unreported failure would leave the reviewer with a success message.
+ */
+async function sendReviewedTranscript(cityId: string, meetingId: string) {
+    const result = await sendTranscriptToMunicipality(cityId, meetingId);
+
+    if (result.skipped) {
+        console.log(`[humanReview] Transcript sending skipped (no contact emails) for ${cityId}/${meetingId}`);
+        return;
```
kouloumos: The JSDoc above this function names this exact case, then the branch returns without telling anyone. Resend rejects the send or the DOCX render throws → `{success:false, error}` → this logs and returns → `markHumanReviewComplete` resolves → the dialog fires the success toast. The reviewer is told the transcript reached the municipality. Same for the summarize leg: `autoTriggerTask` swallows everything that isn't a duplicate, so a down task server also gets a green toast.

Base did this for the transcript too, so it isn't new — but this is the change that puts a second follow-up on the same path, and combined with the completed-review gate (body) a stranded meeting has no signal at all outside Discord.

Returning the outcomes would cover it — `{ review, followUps: { summarize: 'started'|'skipped'|'failed', transcript: 'sent'|'skipped'|'failed' } }` — with both dialogs raising a warning toast on a failure. `autoTriggerTask` can return a small union without weakening its no-throw contract.

src/components/reviews/CompleteReviewDialog.tsx
```diff
@@ -227,7 +222,10 @@ export function CompleteReviewDialog({
           </Button>
           <Button
             onClick={handleConfirm}
-            disabled={isLoading || isSubmitting}
+            // Never confirm options that the dialog did not load: the reviewer would
+            // start summarize — and the notifications it releases — without seeing
+            // the warning. A failure of the statistics alone still allows the review
+            disabled={isSubmitting || completion.isLoading || !completion.state}
```
kouloumos: Two things on this line.

Gating confirm on the loaded options is right, but `!completion.state` never clears after a failed load, so the button is dead for the life of the dialog, with no retry and nothing saying that reopening re-runs the fetch. Base was `disabled={isLoading || isSubmitting}` and `isLoading` was already false after a failure — the reviewer could still complete the review, with the transcript simply not sent. I blocked the action on the preview to check: "Failed to fetch", Mark Complete disabled and unresponsive. A transient failure of a newly-added call now blocks work that used to go through.

Second, `isLoadingStats` isn't in this condition. If the statistics are slower than the options — likely on a long meeting with a lot of edit history — confirm is clickable while the stats spinner is still turning, so the review can complete before the manual-time checkbox has rendered and the override is silently unavailable. Base's combined `Promise.all` made that impossible.

A `reload()` off the hook covers the first; adding `isLoadingStats` covers the second.

src/components/admin/reviews/MarkReviewCompleteButton.tsx
```diff
@@ -121,24 +101,30 @@ export function MarkReviewCompleteButton({
...
             >
               Cancel
             </Button>
-            <Button 
+            <Button
               onClick={handleMarkComplete}
-              disabled={isLoading}
+              // Never confirm without the loaded options: the reviewer would
+              // complete the review with every follow-up silently off
+              disabled={isBusy || completion.isLoading || !completion.state}
```
kouloumos, line 138: Same dead-button-after-a-failed-load as the completion dialog — whatever retry affordance you add there should land here too.

kouloumos (commented):
The split into two independent loads is the right call, and the notification card is doing real work — I drove all four behaviour renderings and both unavailable states on the preview and they say what the code does. The states staging can't produce (AUTO + unreleased, already-summarized) I reached by rewriting the response payload.

What's worth a second look is what the split did to the failure model. Two of its consequences aren't in the description: an options-load failure now leaves confirm permanently dead where it used to still complete the review, and confirm is no longer gated on the statistics load, so a fast reviewer can confirm before the manual-time override has rendered. Both are inline.

On the retry — it does what it says inside the page session, since the dialog stays open on error and clicking again re-runs the follow-ups. After a reload there's no way back in: `determineReviewStatus` returns `'completed'` as soon as a succeeded `humanReview` task exists, and both entry points hide the dialog on that (`MarkReviewCompleteButton.tsx:65-73`, `EditingModeBar.tsx:199`). So a review that committed with a failed follow-up is stranded with no affordance — and since the action resolves successfully when a follow-up fails, the reviewer was told it worked. Those two together are the thing I'd want closed before this ships; the second half is inline on `humanReview.ts`.

Both Greptile P1s look correctly deferred to me — no argument there.

src/lib/tasks/tasks.ts
```diff
@@ -86,11 +86,7 @@ export const startTask = async (taskType: MeetingTaskType, requestBody: any, cou
     if (TASK_CONFIG[taskType].requiredForPipeline) {
         const idempotency = await checkTaskIdempotency(taskType, cityId, councilMeetingId, options);
         if (!idempotency.proceed) {
-            throw new Error(
-                idempotency.blockedReason === 'already_succeeded'
-                    ? `A ${taskType} task has already succeeded for this council meeting`
-                    : `A ${taskType} task is already running for this council meeting`
-            );
+            throw new TaskAlreadyExistsError(taskType, idempotency.blockedReason ?? 'already_running');
```
kouloumos: Nothing asserts this throws a `TaskAlreadyExistsError` rather than an `Error`. I replaced this line with a plain `Error` carrying the identical message strings and ran the three task suites: 49/49 still pass. `checkTaskIdempotency.test.ts` asserts on message text, and the new humanReview test synthesizes the error from a mock, so neither can see the producer. If this regresses, `autoTrigger`'s `instanceof` stops matching and the manual-summarize race goes back to a red "failed" alert — the thing the change set out to fix.

Also `?? 'already_running'` quietly defeats the exhaustiveness the mapper in `humanReview.ts` is built for: a future third reason compiles fine and gets relabelled "already running" in both the message and the alert. It's unreachable today, which is the argument for a throw rather than a default.

src/components/reviews/useReviewCompletion.ts
```diff
@@ -0,0 +1,104 @@
...
+            })
+            .catch((err) => {
+                if (cancelled) return;
+                console.error('Failed to load review completion state:', err);
+                setError(err instanceof Error ? err.message : 'Failed to load review completion state');
+            })
+            .finally(() => {
+                // No cancelled guard: a close mid-flight would otherwise leave the
+                // dialog loading for the rest of the life of the component
+                setIsLoading(false);
```
kouloumos, line 106: Dropping the guard here reopens the misreport this PR fixes, in a narrower window. Close the dialog mid-flight and reopen: cleanup sets `cancelled` for request 1, effect 2 sets `isLoading` true and starts request 2, then request 1's `finally` clears `isLoading` while request 2 is still going. The render is `isLoading:false, state:null, error:null`. In `SendTranscriptButton` that falls through to "No contact emails configured for this administrative body." — I hit this on the preview against Σαμοθράκη, which has three. Here and in `MarkReviewCompleteButton` it's an empty dialog body with a disabled confirm and no spinner or error to explain it.

The load measures 93–121 ms, so the window is about that wide and it self-corrects when request 2 lands; I reproduced it by throttling the action. Base had no cancellation guard at all, so this isn't a regression — it's the last bit of the fix.

The stated reason for dropping the guard doesn't seem to bite: the effect early-returns on close without touching `isLoading`, so it's only read while the dialog is open. `if (cancelled) return;` should be safe.

src/components/reviews/ReviewCompletionOptions.tsx
```diff
@@ -0,0 +1,116 @@
...
+                        <>
+                            <p>{t('runSummarize.help')}</p>
+                            {!canSummarize && <p>{t(`runSummarize.unavailable.${summarizeAvailability}`)}</p>}
+                            {canSummarize && runSummarize && (
+                                <p className={cn('flex items-start gap-1.5', notification.className)}>
+                                    <NotificationIcon className="h-4 w-4 shrink-0 mt-0.5" />
+                                    <span>{t(`runSummarize.notifications.${notification.key}`)}</span>
+                                </p>
+                            )}
+                            {canSummarize && notifiesBeforeRelease && (
```
kouloumos: This gate doesn't include `runSummarize`, unlike the notification line above it. Since `notifiesBeforeRelease` implies AUTO, and the hook defaults the box off in exactly that case, the default render for an unreleased AUTO meeting is an unchecked checkbox carrying an amber bell warning that subscribers get notified about a page they can't open — a warning about something the reviewer hasn't asked for. Tick the box and two amber bell lines stack, reading as two separate sends.

Both meetings on staging are `NOTIFICATIONS_DISABLED`, so I rewrote the payload to reach this; screenshots if useful.

`canSummarize && runSummarize && notifiesBeforeRelease`, rendering the before-release string instead of `notifications.auto` rather than after it.

src/lib/tasks/humanReview.ts
```diff
@@ -1,76 +1,163 @@
...
+    ]);
+
+    const blockedReason = summarizeIdempotency.blockedReason;
 
     return {
         contactEmails: meeting?.administrativeBody?.contactEmails || [],
         administrativeBodyName: meeting?.administrativeBody?.name || null,
+        notificationBehavior: meeting?.administrativeBody?.notificationBehavior || null,
+        summarizeAvailability: blockedReason ? SUMMARIZE_AVAILABILITY[blockedReason] : 'available',
+        released: meeting?.released ?? false,
```
kouloumos: Every field here optional-chains off `meeting`, so a meeting that doesn't exist and a meeting with no administrative body produce the same object. A superadmin passes `checkUserAuthorization` before the meeting-existence check, so with a stale `meetingId` this renders a normal dialog with summarize enabled and defaulted on — then confirm throws "not found" from this same file. Worth an `if (!meeting) throw` here so the two functions agree.

src/lib/tasks/humanReview.ts
```diff
@@ -1,76 +1,163 @@
...
+    // and no later event corrects it until the backend reports a result.
+    revalidateMeetingsForCity(cityId);
+
+    return created;
+}
+
+/** Refresh the meeting lists of a city. A stale list never justifies a failed review. */
+function revalidateMeetingsForCity(cityId: string) {
+    try {
+        revalidateTag(`city:${cityId}:meetings`, 'max');
```
kouloumos: `revalidateMeeting` in `src/lib/cache` is this same try/catch with the same rationale in its JSDoc, and `processAgenda.ts` and `summarize.ts` already use it. It also busts `city:${cityId}:meeting:${meetingId}` — no practical difference today, since every meeting-scoped entry also carries `city:X:meetings`, but it's the already-shared version.

src/lib/tasks/__tests__/markHumanReviewComplete.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    const state = await getReviewCompletionState(CITY_ID, MEETING_ID);
+
+    expect(state.summarizeAvailability).toBe(availability);
+  });
+
+  it('reports no body as no recipients, no notifications and unreleased', async () => {
+    mockMeetingFindUnique.mockResolvedValue(null);
+
+    await expect(getReviewCompletionState(CITY_ID, MEETING_ID)).resolves.toEqual({
```
kouloumos, line 307: This case is named for "no body" but the fixture is `findUnique -> null`, which is "no meeting". The two differ on `released`, and the genuine no-body case — a real meeting row with `administrativeBody: null` — isn't covered.

While in here: the `sendTranscript` gate has no test at all. `runSummarize` gets opt-in, opt-out and no-caller; `sendTranscript` only gets opt-in paths. Flipping its default to `true`, or making the call unconditional, leaves all 16 green — and it's the leg that can't be recalled.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/23e3b8879bee715b879ed232eb9429389c9dcad5..faf48d78114bfd81ff8b71001924d84812e6ebdf) to act on the review. Every point held up when I checked it against the code.

- `markHumanReviewComplete` returns `{ review, followUps }`. `autoTriggerTask` returns `'started' | 'skipped' | 'failed'`, and it still never throws. Both dialogs report a failed follow-up: the completion dialog raises a destructive toast that names the action, and the admin dialog holds the dialog open with the same message.
- `useReviewCompletion` exposes `reload()`. Both dialogs show a retry control beside the load error, so a transient failure no longer kills the confirm button for the life of the dialog.
- The confirm button of the completion dialog waits for the statistics too, so the reviewer cannot confirm before the manual-time override renders. A statistics failure leaves both flags false, so it still allows the review.
- The before-release warning replaces the plain `NOTIFICATIONS_AUTO` line and only renders when the box is ticked. The two amber lines can no longer stack, and an unticked box carries no warning.
- The `cancelled` guard is back on the `finally`. Your reasoning is better than mine: the effect early-returns while the dialog is closed, so nothing reads a flag that a cancelled request left set.
- `startTask` no longer defaults `blockedReason`. A blocked task with no reason now throws its own error.
- `getReviewCompletionState` throws for a meeting that does not exist, so it agrees with `markHumanReviewComplete`.
- `revalidateMeeting` from `src/lib/cache` replaces the local copy.

Tests: 155 in the task suites, 1425 in the repo. The new `startTask` test asserts the type, not the message — I checked it by replacing the throw with a plain `Error` carrying the identical strings, which turns it red, then restored it. The `getReviewCompletionState` case now uses a real meeting row with `administrativeBody: null`, and a separate case covers the missing meeting. `sendTranscript` gets opt-in, opt-out and no-caller cover, plus cases that separate `skipped` from `failed` on both legs.

One item I did not change: the completed-review gate. `determineReviewStatus` returns `'completed'` from the `humanReview` task alone, so both entry points still hide the dialog after a reload. What that gate combined with was the silent success, and that half is gone — the reviewer now sees which follow-up failed. The recovery paths also exist for an admin: `SendTranscriptButton` renders on `transcriptSent`, not on review status, and summarize has the button in the meeting admin panel. Making the completion dialog itself reachable after a completed review changes what `'completed'` means for the reviews dashboard, which reads to me like its own change rather than part of this one. Happy to be told otherwise.

---

src/lib/tasks/tasks.ts
christosporios: Both done, and I reproduced your experiment to check that the new test is load-bearing rather than just present.

`checkTaskIdempotency.test.ts` now asserts the type and the reason, not the message. I replaced the throw with a plain `Error` carrying the identical strings, exactly as you did: the two new cases go red. Restoring the typed throw turns them green again.

`?? 'already_running'` is gone. A blocked task with no reason now throws its own error, so a third reason reaches neither the message nor the Discord alert wearing the wrong label.

---

src/components/reviews/useReviewCompletion.ts
christosporios: Restored. Your reasoning is better than the comment I had written there: the effect early-returns while the dialog is closed, so nothing reads a flag that a cancelled request left set, and the stuck-loading case I was guarding against cannot happen.

The comment on the `finally` now records the race you measured — a cancelled request clearing the flag while its replacement is in flight — rather than the reason I dropped the guard for.

---

PR #637, opened 2026-08-17
feat(search): Improve search by cutting off irrelevant results and introducing ordering

pkakelas:
## Summary

Second PR solving https://github.com/schemalabz/opencouncil/issues/625.
Depends on https://github.com/schemalabz/opencouncil/pull/628.

Search ranked results by text relevance only, so a strong text match from a rarely-discussed community meeting could outrank a well-attended council debate. This PR adds a post-relevance ranking layer, tightens lexical matching, and indexes the new fields ranking needs.

## What changed
- Elasticsearch views/schema: new MeetingAdministrativeBodyView and CitySearchView; index administrative_body_type, city_realm, contributor_count, discussion_speaking_seconds on subjects.
- Retrieval: fuzzy matching + phrase boosts on the lexical retriever; require 75% of query terms on 3+ term queries; add a raw score cutoff (semanticMinScore) so unrelated queries can return zero semantic hits.
- Ranking: wrap lexical, semantic, and the filter-only browse path in a script_score function that boosts council > committee > community (reusing ADMIN_BODY_TIER), longer discussions, and recent meetings. Replaces the filter-only path's plain date sort — MCP search tool docs updated to match.

Deployment note
Existing documents need reindexing before the ranking function has data to score against.

<h3>Greptile Summary</h3>

The PR replaces rank fusion with a scored lexical and semantic search query. It also adds post-relevance ranking and a search evaluation harness.
- Adds document-level lexical coverage, fuzzy matching, phrase boosts, and a semantic score cutoff.
- Ranks similar results by administrative-body tier, discussion duration, and meeting recency.
- Updates search filters, request configuration, tests, evaluation scripts, and Elasticsearch documentation.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/search/query.ts | Builds the new lexical, semantic, and metadata-ranking query; the previous two-term behavior is now explicitly documented as intentional. |
| src/lib/search/index.ts | Integrates the revised query and preserves search-result hydration behavior. |
| src/lib/search/filters.ts | Updates derived-filter handling for the revised search flow. |
| src/lib/search/__tests__/query.test.ts | Tests coverage gates, tiered scoring, semantic fallback, ranking, and deterministic browsing behavior. |
| scripts/search-eval.ts | Adds a read-only evaluation harness for recall, semantic cutoff, and field-tier ordering. |
| src/app/api/search/route.ts | Configures the API search path to use the revised semantic fallback. |

<sub>Reviews (6): Last reviewed commit: ["fix(search): raise the semantic cutoff t..."](https://github.com/schemalabz/opencouncil/commit/5318e3dd81a5299aab5af3b0f94ddc566306e6b2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53985610)</sub>

### DIFF
--- a/src/lib/ranking/subjects.ts
+++ b/src/lib/ranking/subjects.ts
@@ -60,7 +60,7 @@ export const DEFAULT_SUBJECT_RANKING_WEIGHTS: SubjectRankingWeights = {
 };
 
 /** Δημοτικό Συμβούλιο > Δημοτική Επιτροπή > Δημοτική Κοινότητα. */
-const ADMIN_BODY_TIER: Record<AdministrativeBodyType, number> = {
+export const ADMIN_BODY_TIER: Record<AdministrativeBodyType, number> = {
     council: 1,
     committee: 0.5,
     community: 0,
--- a/src/lib/search/query.ts
+++ b/src/lib/search/query.ts
@@ -282,9 +282,24 @@ const NAME_FUZZY_PREFIX_LENGTH = 2;
  * hit actually outranks a title match. Run it after changing a base, a k, or a
  * multiplier weight — the arithmetic alone will not warn you.
  *
- * Last run (Aug 2026, 9.1k released docs): zero inversions, 15 of 16 measurable
- * queries structurally safe. The one exception is "Δούκας" at 1.47 against a
- * 1.484 ceiling — currently held by metadata, not by the constants.
+ * Last run (Aug 2026, 9.1k released docs): zero inversions, 19 of 27 measurable
+ * queries structurally safe against a 1.484 ceiling.
+ *
+ * The eight inside the multiplier's reach are almost all MULTI-TERM, and the
+ * split is sharp: one-word queries measure 1.50-2.57, multi-term queries
+ * 1.11-1.86. The partial share above is why. A multi-term query hands its name
+ * tier's partial share to any document that carries ONE of its terms in the
+ * title, and the greek stemmer creates that case constantly, because it does
+ * not always map an inflection to the stem the query produces: σχολικές stems
+ * to σχολικ while Σχολικών stems to σχολ, so `Κατάργηση Σχολικών Επιτροπών`
+ * takes the partial share of "σχολικές επιτροπές" and never the strict clause.
+ * "χώροι πρασίνου" is the tightest measured pair at 1.11.
+ *
+ * So the corpus correlation holds, but it holds with far less room on a
+ * multi-term query than the earlier one-word-only query set suggested. Nothing
+ * is inverted today, and only metadata separates those eight pairs.
+ * ("Δούκας" at 1.47 predates the harness fixes and re-measures the same, so it
+ * was never an artifact of them.)
  *
  * Within one tier the multiplier is meant to decide, and k per tier keeps the
  * log tiebreak's spread near +-5%, well under the multiplier's reach. The
--- a/src/lib/search/query.ts
+++ b/src/lib/search/query.ts
@@ -95,9 +95,39 @@ const NAME_FUZZINESS = 'AUTO:4,10';
 const NAME_FUZZY_PREFIX_LENGTH = 2;
 
 /**
+ * A person-name query answers with the subjects the person is responsible for
+ * (FIELD_TIER.introducer). The subjects they only spoke in answer the same
+ * query weakly, so they belong in the tail: the speakerName tier is the
+ * lowest in the query, under the transcript's.
+ *
+ * The boost decides only where those subjects rank, never whether they are
+ * found — a should-clause either matches or it does not. Measured on the
+ * production index (Aug 2026, 9.1k released docs), the recall it adds is the
+ * point of the clause, and it is the same at every boost:
+ *   "Χάρης Δούκας"  38 -> 110 hits    "Μαλτέζος"  120 -> 216
+ *   "του Δούκα"     94 -> 154         topic queries: no change at all
+ * Junk queries stay empty, because no name matches their terms.
+ *
+ * The speakerName tier in FIELD_TIER is calibrated on how far the clause may
+ * reorder the first page (measured at the pre-tier boost 0.1, re-checked after
+ * the tier flattening): for the worst case in the index — an Athens mayor who
+ * speaks in 91 subjects and introduced 34 — it moves 2 subjects into the top
+ * 10; every other measured query moves 0-1. Two slots on the single heaviest
+ * speaker is the intended dose: the page still leads with what the person
+ * introduced, and the rest of what they spoke in follows behind it.
+ *
+ * The earlier decision here was to leave the field unsearched, on the reasoning
+ * that a mayor speaks in nearly every subject. The measurements above do not
+ * support that: a speaker reaches 2-3x the subjects they introduce, not the
+ * whole index, because a subject only carries contributions once the summarize
+ * task has run on it. The personIds filter is still the right tool for
+ * "everything this person spoke about" — it is exact, and it does not rank.
+ */
+
+/**
  * Field-tier score flattening: every lexical clause is rescored to
  * base + k * log1p(bm25), so WHICH field matched (the tier: name >
- * description/introducer > transcript) sets the score level,
+ * description/introducer > transcript > speaker name) sets the score level,
  * raw BM25 shrinks to a small within-tier tiebreak, and the post-relevance
  * multiplier (admin body, discussion length, recency; up to ~1.42x) decides
  * among same-tier matches.
@@ -134,6 +164,7 @@ const FIELD_TIER = {
     // exact term), the only name-tier signal for typo queries.
     fuzzyName: { base: 4, k: 2 },
     transcript: { base: 6, k: 3 },
+    speakerName: { base: 0.3, k: 0.2 },
 } as const;
 
 // Rescores a clause to its tier band: base + k * log1p(bm25). boost_mode
@@ -426,12 +457,10 @@ function buildLexicalShouldClauses(
         flattenToTier(termClause('description', queryText), FIELD_TIER.descriptionTerm),
         // Person-name queries ("Χάρης Δούκας", "Μαλτέζος") are a recurring
         // pattern in the logged user searches. This field covers the subjects
-        // the person introduced — a deliberate authorship signal, unlike
-        // speaker_contributions.speaker_person_name, which is intentionally not
-        // searched: a mayor speaks in nearly every subject, so matching it
-        // would flood a name query with everything they ever commented on (the
-        // personIds filter serves that need). The greek analyzer also stems
-        // name declensions, so "του Δούκα" matches "Δούκας".
+        // the person introduced — the strongest authorship signal. The subjects
+        // they only spoke in match through the much weaker nested speaker-name
+        // clause below. The greek analyzer also stems name declensions, so
+        // "του Δούκα" matches "Δούκας".
         flattenToTier(termClause('introduced_by_person_name', queryText), FIELD_TIER.introducer),
         ...(extractedFilters.locationName
             ? [flattenToTier(termClause('location_text', queryText), FIELD_TIER.locationText)]
@@ -473,7 +502,26 @@ function buildLexicalShouldClauses(
                     _source: ['speaker_contributions.contribution_id']
                 }
             }
-        }, FIELD_TIER.transcript)
+        }, FIELD_TIER.transcript),
+        // Subjects the person spoke in, at the bottom of the field tier (see
+        // FIELD_TIER.speakerName and the calibration note above it). A separate
+        // nested clause rather than a second field on the transcript clause
+        // above: inner_hits marks which contributions matched the query text,
+        // and a speaker-name match would add contributions whose text does not
+        // contain the query at all.
+        flattenToTier({
+            nested: {
+                path: 'speaker_contributions',
+                query: {
+                    match: {
+                        'speaker_contributions.speaker_person_name': {
+                            query: queryText,
+                            minimum_should_match: LEXICAL_MINIMUM_SHOULD_MATCH
+                        }
+                    }
+                }
+            }
+        }, FIELD_TIER.speakerName)
     ];
 }
 
--- a/src/lib/search/query.ts
+++ b/src/lib/search/query.ts
@@ -1,5 +1,5 @@
 import { estypes } from '@elastic/elasticsearch';
-import { SearchRequest, ExtractedFilters } from './types';
+import { SearchRequest, ExtractedFilters, Location } from './types';
 import { env } from '@/env.mjs';
 import { ADMIN_BODY_TIER } from '@/lib/ranking/subjects';
 
@@ -8,6 +8,12 @@ const DEFAULT_RANK_CONSTANT = 60;
 
 const SEMANTIC_NAME_BOOST = 2.0;
 const SEMANTIC_DESCRIPTION_BOOST = 1.5;
+
+// Additive score for subjects pinned within an AI-extracted location's radius
+// (see buildLocationBoostClauses). Small next to the name/description boosts:
+// proximity breaks ties among text matches, it does not outrank a better text
+// match.
+const LOCATION_BOOST = 2;
 
 /**
  * Raw-score cutoff for the semantic retriever, measured against the production
@@ -43,6 +49,16 @@ const DEFAULT_SEMANTIC_MIN_SCORE = 3.23;
  * before this applies, so "συνταγή για μουσακά" is a 2-term query.
  */
 const LEXICAL_MINIMUM_SHOULD_MATCH = '2<75%';
+
+/**
+ * Apostrophe-like characters normalized to the ASCII apostrophe before any
+ * clause sees the query: the single quotes mobile keyboards auto-substitute
+ * (U+2018/U+2019), the modifier apostrophe (U+02BC), the Greek tonos (U+0384)
+ * and acute accent (U+00B4) that official minutes use as apostrophes in names
+ * like ΔΙ΄ΕΥΧΩΝ, and the Greek koronis/psili (U+1FBD/U+1FBF). Without this,
+ * the same query succeeds or fails depending on which keyboard typed it.
+ */
+const APOSTROPHE_VARIANTS = /[‘’ʼ΄´᾽᾿]/g;
 
 /**
  * Typo tolerance is restricted to the name field, exact-only elsewhere.
@@ -252,39 +268,29 @@ export function buildFilters(request: SearchRequest): estypes.QueryDslQueryConta
         });
     }
 
-    // Add location filter if specified
-    if (request.locations && request.locations.length > 0) {
-        if (request.locations.length === 1) {
-            // Single location case
-            filters.push({
-                geo_distance: {
-                    distance: `${request.locations[0].radius}km`,
-                    'location_geojson': {
-                        lat: request.locations[0].point.lat,
-                        lon: request.locations[0].point.lon
-                    }
-                }
-            });
-        } else {
-            // Multiple locations case
-            filters.push({
-                bool: {
-                    should: request.locations.map(loc => ({
-                        geo_distance: {
-                            distance: `${loc.radius}km`,
-                            'location_geojson': {
-                                lat: loc.point.lat,
-                                lon: loc.point.lon
-                            }
-                        }
-                    })),
-                    minimum_should_match: 1
-                }
-            });
-        }
-    }
-
     return filters;
+}
+
+// Location proximity clauses. Only the AI filter-extraction path produces
+// `locations` (no UI, API or MCP caller passes them). They must NOT become a
+// hard filter on a text search: only ~45% of subjects carry a location pin,
+// and a geo_distance filter drops every pin-less document. A query like
+// "παλαιστίνη" — extracted as a location and geocoded somewhere — would then
+// return zero results even though subjects carry it in the title. As `should`
+// clauses on the scoring query, nearby pinned subjects rank higher and
+// everything else still matches on text alone.
+function buildLocationBoostClauses(locations: Location[] | undefined): estypes.QueryDslQueryContainer[] {
+    if (!locations || locations.length === 0) return [];
+    return locations.map(loc => ({
+        geo_distance: {
+            distance: `${loc.radius}km`,
+            'location_geojson': {
+                lat: loc.point.lat,
+                lon: loc.point.lon
+            },
+            boost: LOCATION_BOOST
+        }
+    }));
 }
 
 // Transcripts are long enough that a bare OR match lets an off-topic query match
@@ -311,6 +317,23 @@ function buildLexicalShouldClauses(
     extractedFilters: ExtractedFilters
 ): estypes.QueryDslQueryContainer[] {
     return [
+        // The standard tokenizer keeps an intra-word ASCII apostrophe inside
+        // its token (δι'ευχών → δι'ευχ) but splits on the Greek tonos that
+        // official minutes use in names (ΔΙ΄ΕΥΧΩΝ → δι, ευχ), so each indexed
+        // spelling is reachable by only one shape of the query. The main
+        // clauses below match the intact-token form; this space-split variant
+        // matches the split-token form. Same fields and boosts as the main
+        // multi_match — whichever shape the index holds, the score tier is
+        // the same.
+        ...(queryText.includes("'") ? [{
+            multi_match: {
+                query: queryText.replace(/'/g, ' '),
+                fields: ['name^4', 'description^3'],
+                type: 'best_fields' as const,
+                operator: 'or' as const,
+                minimum_should_match: LEXICAL_MINIMUM_SHOULD_MATCH
+            }
+        }] : []),
         {
             multi_match: {
                 query: queryText,
@@ -447,12 +470,20 @@ export function buildSearchQuery(
     // retrievers (they require a query) and rank the filtered set by administrative
     // body, discussion length, and recency alone. Used e.g. for "everything a person
     // spoke about" or "all subjects in a date range".
-    const queryText = mergedRequest.query?.trim();
+    const queryText = mergedRequest.query?.trim().replace(APOSTROPHE_VARIANTS, "'");
     const filters = buildFilters(mergedRequest);
+    const locationBoosts = buildLocationBoostClauses(mergedRequest.locations);
     // Computed once so every ranking function built for this request — both RRF
     // retrievers below, or the filter-only branch — scores recency identically.
     const nowMillis = Date.now();
     if (!queryText) {
+        // With no text to score, a location can only act as a filter. Unreachable
+        // today (locations only come from AI extraction, which requires query
+        // text), but kept so an explicit filter-only location request stays a
+        // location browse rather than being ignored.
+        const browseFilters = locationBoosts.length > 0
+            ? [...filters, { bool: { should: locationBoosts, minimum_should_match: 1 } }]
+            : filters;
         return {
             index: env.ELASTICSEARCH_INDEX,
             size: request.config?.size || 10,
@@ -460,9 +491,24 @@ export function buildSearchQuery(
             track_total_hits: true,
             // A filter-only bool query scores every hit 0, so 'replace' ranks on the
             // function alone instead of nudging a (nonexistent) relevance score.
-            query: applyRanking({ bool: { filter: filters } }, 'replace', nowMillis)
+            query: applyRanking({ bool: { filter: browseFilters } }, 'replace', nowMillis)
         };
     }
+
+    const lexicalCore: estypes.QueryDslQueryContainer = {
+        bool: {
+            should: buildLexicalShouldClauses(queryText, extractedFilters),
+            minimum_should_match: 1,
+            filter: filters
+        }
+    };
+    // Proximity lifts pinned subjects near the extracted location, inside `must`
+    // so a location-only match without any text match cannot surface. Lexical arm
+    // only: the semantic arm's min_score calibration must keep seeing raw
+    // semantic scores, and RRF carries the lexical arm's ordering across.
+    const lexicalQuery: estypes.QueryDslQueryContainer = locationBoosts.length > 0
+        ? { bool: { must: [lexicalCore], should: locationBoosts } }
+        : lexicalCore;
 
     return {
         index: env.ELASTICSEARCH_INDEX,
@@ -474,13 +520,7 @@ export function buildSearchQuery(
                 retrievers: [
                     {
                         standard: {
-                            query: applyRanking({
-                                bool: {
-                                    should: buildLexicalShouldClauses(queryText, extractedFilters),
-                                    minimum_should_match: 1,
-                                    filter: filters
-                                }
-                            }, 'multiply', nowMillis)
+                            query: applyRanking(lexicalQuery, 'multiply', nowMillis)
                         }
                     },
                     ...(request.config?.enableSemanticSearch ? [
--- a/src/lib/search/query.ts
+++ b/src/lib/search/query.ts
@@ -85,6 +85,68 @@ const SEMANTIC_MAPPED_SCALE = 320;
 const LEXICAL_MINIMUM_SHOULD_MATCH = '2<75%';
 
 /**
+ * Fields whose content is a subset of a mixed query by construction, and the
+ * gate that makes LEXICAL_MINIMUM_SHOULD_MATCH mean what it was meant to mean.
+ *
+ * The term requirement above is evaluated per field, and a per-field threshold
+ * silently assumes every field could hold the whole query. That holds for
+ * name/description/transcript. It fails for the entity fields
+ * (introduced_by_person_name, location_text, speaker_person_name), whose whole
+ * content is a name or a place. There the threshold is never a real test — it
+ * is either trivially met or impossible:
+ *   - A councillor's two-word name satisfies 2-of-3 of "Ιωάννης Μαλτέζος
+ *     υδρονομείς" on its own. Measured on the production index, that clause
+ *     matched all 93 subjects the person introduced, while the name and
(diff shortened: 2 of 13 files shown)


src/lib/search/query.ts
```diff
@@ -1,6 +1,173 @@
...
+// duration — a subject's own discussion length, not the whole session's.
+const DISCUSSION_LENGTH_BOOST_WEIGHT = 0.03;
+
+// Exponential decay: at RECENCY_DECAY_SCALE_DAYS old, a meeting keeps ~37% (1/e) of
+// the recency boost. The multiplier floors at 1.0 (no boost), never goes below it —
+// an old meeting loses the recency edge, it isn't penalized for its age.
+const RECENCY_BOOST_WEIGHT = 0.1;
+const RECENCY_DECAY_SCALE_DAYS = 365;
+
+const RANKING_SCRIPT = `
```
kouloumos: This string is the largest piece of new logic here, and nothing executes it. TypeScript cannot see
inside it, and no test reads `script.source` — `rankingScriptParams` asserts on `script.params` only.

I mutation-tested the suite. It catches the two hard ones: moving `min_score` onto the outer
`function_score`, and re-adding `sort: meeting_date desc` beside the new `function_score`. Both fail
as they should. These five do not:

| mutation | suite |
|---|---|
| `doc['administrative_body_type']` -> `doc['administrative_body']` | 24/24 pass |
| drop the trailing `;` (stands in for any Painless syntax error) | 24/24 pass |
| `params.councilWeight` -> `params.council`, in the script source only | 24/24 pass |
| swap the `discussionWeight` and `recencyWeight` values | 24/24 pass |
| recency `1 + w*exp(...)` -> `1 - w*exp(...)` | 24/24 pass |

The first three are not ranking bugs. `doc[]` on an unmapped field throws per document, so the shard
fails and every search returns an error — the filter-only browse path included, since it goes through
`applyRanking` too. I hit exactly this on the live index during the review: `script_exception`, all
shards failed.

Two tests close the first and third, need no change to `query.ts`, and read both values off the
built query. They need the same `jest.mock('@/env.mjs', ...)` the suite already uses, otherwise the
file will not load:

```ts
const script = builtQuery.query.function_score.functions[0].script_score.script;

// kills the params rename
const referenced = [...script.source.matchAll(/params\.(\w+)/g)].map(m => m[1]);
expect([...new Set(referenced)].sort()).toEqual(Object.keys(script.params).sort());

// kills the field-name typo
const mapping = JSON.parse(readFileSync('elasticsearch/schema.json', 'utf8'))[0].mapping;
for (const f of [...script.source.matchAll(/doc\['([^']+)'\]/g)].map(m => m[1])) {
    expect(mapping[f]).toBeDefined();
}
```

I ran both: green on this branch, and each one fails on its own mutation while the other stays green.
Neither catches the syntax error or the two value swaps. Those need the script to actually run, which
means an Elasticsearch instance — worth a separate conversation, not this PR.

src/lib/search/query.ts
```diff
@@ -1,6 +1,173 @@
...
+ * (ES's combination form treats clause counts <= the leading integer as
+ * all-required); a query of 3+ terms requires 75% of them, so a single stray
+ * matching term no longer surfaces a low-relevance hit. Requiring both terms
+ * of a 2-term query is deliberate: measured on the production index, OR-ing
+ * them floods the results with one-word matches (a query like "πάρκα Κυψέλης"
+ * would surface every park in every city), while requiring both keeps the
+ * results on-subject. Stopwords do not count: the greek analyzer drops them
+ * before this applies, so "συνταγή για μουσακά" is a 2-term query.
+ */
+const LEXICAL_MINIMUM_SHOULD_MATCH = '2<75%';
```
kouloumos: Not re-opening `2<75%` — the comment above settles it. This is about the query type it sits on.

On `best_fields`, Elasticsearch rewrites the `multi_match` into a `dis_max` of per-field `match`
queries and applies `minimum_should_match` to each field separately. Their docs call it out: "the
`operator` and `minimum_should_match` parameters are applied to each field individually, which is
probably not what you want." So "75% of the query's terms" means 75% within `name` alone, or within
`description` alone. Terms split across the two no longer combine.

Measured on the live index, isolating documents that hold one term in each field:

```
docs with "Καταστροφή" ONLY in name and "Εγκρίθηκε" ONLY in description: 2
  best_fields, operator or (base):   2
  best_fields '2<75%' (this branch): 0
  cross_fields '2<75%':              2
```

The size of it is small — I checked before assuming it mattered. Switching to `cross_fields` recovers
1 to 3 documents per query (`ανάπλαση πλατείας` 44 -> 45, `άδεια μουσικής` 42 -> 45) and holds every
junk query at 0. So this is a cheap correctness fix, not a recall lever.

The reason I would still take it is #625 item 4. Under `best_fields` with a per-field
`minimum_should_match`, adding `city_name` or `introduced_by_person_name` to the field list cannot
help a split query, because a term matching in a new field can never satisfy the requirement in a
different one. `cross_fields` leaves that door open.

src/lib/search/query.ts
```diff
@@ -120,6 +287,141 @@ export function buildFilters(request: SearchRequest): estypes.QueryDslQueryConta
...
+    extractedFilters: ExtractedFilters
+): estypes.QueryDslQueryContainer[] {
+    return [
+        {
+            multi_match: {
+                query: queryText,
+                fields: [
+                    'name^4',
+                    'description^3',
+                    ...(extractedFilters.locationName ? ['location_text^3'] : []),
```
kouloumos: This boost is now inert for the queries that switch it on.

The field only enters the field list when the extractor found a `locationName`, which needs a query
with more than a place name in it. It then inherits `LEXICAL_MINIMUM_SHOULD_MATCH`, so it needs 75%
of the query's terms inside the address text. An address never contains `κολυμβητήριο`.

```
location_text "κολυμβητήριο Μαρουσίου":  operator or = 25   '2<75%' = 0
location_text "πλατεία Κοραή Πειραιά":   operator or = 238  '2<75%' = 4
```

So it fires only when the whole query is the place name — the one case where the extractor is least
likely to produce a `locationName` in the first place.

`cross_fields` (see the other comment) fixes this as a side effect, since the location term would
then count toward the requirement from whichever field holds it. Otherwise this clause wants its own
`minimum_should_match`, or none.

src/lib/search/query.ts
```diff
@@ -1,6 +1,173 @@
...
+/**
+ * Raw-score cutoff for the semantic retriever, measured against the production
+ * index (scripts/search-eval.ts). multilingual-e5 trains with a low InfoNCE
+ * temperature, so its cosine similarities bunch up near the top of the range
+ * and unrelated text still scores highly. Measured bands (Aug 2026, 9.1k
+ * released docs):
+ *   - on-topic queries, top hits:        3.25 - 3.31
+ *   - paraphrase queries (no shared
+ *     stems with the docs), top hits:    3.23 - 3.275
+ *   - off-topic queries, best hit:       <= 3.218
```
kouloumos: This number and the one in `query.test.ts` disagree — the test comment says `<= 3.226`. That matters
more than a usual comment drift, because this is the one measurement in the PR that nobody can
re-derive from the code, and the gap decides whether the cutoff has 0.012 of headroom or 0.004.

My measurement puts the worst off-topic query at 3.220 (`Taylor Swift συναυλία`), so the test
comment is the accurate one. The worst paraphrase I found is `ζώα χωρίς ιδιοκτήτη` at 3.231, one
hit clearing the cutoff — so the real separating band is about 0.011 wide.

Worth stating the structural half here too, since it is invisible from the constant. The service
hard-codes cosine, so each clause is bounded: `name.semantic` at most 2.0, `description.semantic` at
most 1.5. 3.23 is 92.3% of the 3.5 ceiling, so a document matching on `name.semantic` alone can
never clear it, whatever its similarity. The cutoff is "both sub-fields contribute", not "the match
is good enough". That is harmless today — I sampled 2,000 subjects and none has an empty description
— but it is not what the constant reads like.

src/lib/ranking/subjects.ts
```diff
@@ -60,7 +60,7 @@ export const DEFAULT_SUBJECT_RANKING_WEIGHTS: SubjectRankingWeights = {
 };
 
 /** Δημοτικό Συμβούλιο > Δημοτική Επιτροπή > Δημοτική Κοινότητα. */
-const ADMIN_BODY_TIER: Record<AdministrativeBodyType, number> = {
+export const ADMIN_BODY_TIER: Record<AdministrativeBodyType, number> = {
```
kouloumos: Exporting this shares the magnitudes, not only the order, and the two consumers treat them
differently.

`rankSubjects` z-scores this tier. A z-score is invariant to affine rescaling, so rewriting the table
as `{council: 10, committee: 5, community: 0}` changes nothing here, and the tests in this file still
pass — they assert `council > committee > community` and nothing about size. The same edit takes
search's admin multiplier from 1.15 to 2.5, which stops being a nudge and starts dominating text
relevance.

Nothing at this definition says search reads the values, and no test in either file can fail on it.

Declaring search's own weights against `Record<AdministrativeBodyType, number>` and keeping only the
ordering shared removes the coupling. A test asserting the two orderings agree keeps the property
you actually want, and fails loudly if someone reorders the tiers.

src/lib/search/query.ts
kouloumos, line 524: This string is the largest piece of new logic here, and nothing executes it. TypeScript cannot see inside it, and no test reads `script.source` — `rankingScriptParams` asserts on `script.params` only.

I mutation-tested the suite. It catches the two hard ones: moving `min_score` onto the outer `function_score`, and re-adding `sort: meeting_date desc` beside the new `function_score`. Both fail as they should. These five do not:

| mutation | suite |
|---|---|
| `doc['administrative_body_type']` → `doc['administrative_body']` | 24/24 pass |
| drop the trailing `;` (stands in for any Painless syntax error) | 24/24 pass |
| `params.councilWeight` → `params.council`, in the script source only | 24/24 pass |
| swap the `discussionWeight` and `recencyWeight` values | 24/24 pass |
| recency `1 + w*exp(...)` → `1 - w*exp(...)` | 24/24 pass |

The first three are not ranking bugs. `doc[]` on an unmapped field throws per document, so the shard fails and every search returns an error — the filter-only browse path included, since it goes through `applyRanking` too. I hit exactly this on the live index during the review: `script_exception`, all shards failed.

Two tests close the first and third. They need no change to `query.ts`, they read both values off the built query, and they need the same `jest.mock('@/env.mjs', ...)` the suite already uses or the file will not load:

```ts
const script = builtQuery.query.function_score.functions[0].script_score.script;

// kills the params rename
const referenced = [...script.source.matchAll(/params\.(\w+)/g)].map(m => m[1]);
expect([...new Set(referenced)].sort()).toEqual(Object.keys(script.params).sort());

// kills the field-name typo
const mapping = JSON.parse(readFileSync('elasticsearch/schema.json', 'utf8'))[0].mapping;
for (const f of [...script.source.matchAll(/doc\['([^']+)'\]/g)].map(m => m[1])) {
    expect(mapping[f]).toBeDefined();
}
```

I ran both: green on this branch, and each one fails on its own mutation while the other stays green. Neither catches the syntax error or the two value swaps — those need the script to actually run against Elasticsearch, which is a separate conversation, not this PR.

src/lib/search/query.ts
kouloumos, line 109: Not re-opening `2<75%` — the comment above settles it. This is about the query type it sits on.

On `best_fields`, Elasticsearch rewrites the `multi_match` into a `dis_max` of per-field `match` queries and applies `minimum_should_match` to each field separately. Their docs call it out: "the `operator` and `minimum_should_match` parameters are applied to each field individually, which is probably not what you want." So "75% of the query's terms" means 75% within `name` alone, or within `description` alone. A subject whose title holds two of the words and whose description holds the third now matches neither way.

Measured on the live index, isolating documents that hold one word in each field:

```
docs with "Καταστροφή" ONLY in name and "Εγκρίθηκε" ONLY in description: 2
  best_fields, operator or (base):   2
  best_fields '2<75%' (this branch): 0
  cross_fields '2<75%':              2
```

I checked the size of it before assuming it mattered, and it is small. Switching to `cross_fields` brings back 1 to 3 documents per query (`ανάπλαση πλατείας` 44 → 45, `άδεια μουσικής` 42 → 45) and still holds every junk query at 0. So this is a cheap correctness fix, not a big win in results returned.

The reason I would still take it is the entity-aware work later on. Under `best_fields` with a per-field `minimum_should_match`, adding `city_name` or `introduced_by_person_name` to the field list cannot help a query whose words are spread across fields, because a word matching in a new field can never satisfy the requirement in a different one. `cross_fields` leaves that door open.

src/lib/search/query.ts
kouloumos: This boost is now switched off for the queries that turn it on.

The field only enters the field list when the extractor found a `locationName`, which needs a query with more than a place name in it. It then inherits `LEXICAL_MINIMUM_SHOULD_MATCH`, so it needs 75% of the query's words inside the address text. An address never contains `κολυμβητήριο`.

```
location_text "κολυμβητήριο Μαρουσίου":  operator or = 25   '2<75%' = 0
location_text "πλατεία Κοραή Πειραιά":   operator or = 238  '2<75%' = 4
```

So it fires only when the whole query is the place name — the one case where the extractor is least likely to produce a `locationName` to begin with.

`cross_fields` (see the other comment) fixes this as a side effect, since the place name would then count toward the requirement from whichever field holds it. Otherwise this clause wants its own `minimum_should_match`, or none.

src/lib/search/query.ts
kouloumos: This number and the one in `query.test.ts` disagree — the test comment says `<= 3.226`. That matters more than usual comment drift, because this is the one measurement in the PR that nobody can re-derive from the code, and the gap decides whether the cutoff has 0.012 of headroom or 0.004.

My measurement puts the worst off-topic query at 3.220 (`Taylor Swift συναυλία`), so the test comment is the accurate one. The worst paraphrase I found is `ζώα χωρίς ιδιοκτήτη` at 3.231, with one hit clearing the cutoff. So the real gap between "nonsense" and "a real query phrased differently" is about 0.011 wide.

Worth stating the structural half here too, since it is invisible from the constant. The service hard-codes cosine, so each clause is bounded: `name.semantic` at most 2.0, `description.semantic` at most 1.5. 3.23 is 92.3% of the 3.5 ceiling, so a document matching on `name.semantic` alone can never clear it, however good that match is. The cutoff means "both sub-fields contribute", not "the match is good enough". That is harmless today — I sampled 2,000 subjects and none has an empty description — but it is not what the constant reads like.

src/lib/ranking/subjects.ts
kouloumos, line 63: Exporting this shares the magnitudes, not only the order, and the two consumers treat them differently.

`rankSubjects` z-scores this tier. A z-score is invariant to affine rescaling, so rewriting the table as `{council: 10, committee: 5, community: 0}` changes nothing here, and the tests in this file still pass — they assert `council > committee > community` and nothing about size. The same edit takes search's admin multiplier from 1.15 to 2.5, which stops being a nudge and starts dominating text relevance.

Nothing at this definition says search reads the values, and no test in either file can fail on it.

Declaring search's own weights against `Record<AdministrativeBodyType, number>` and keeping only the ordering shared removes the coupling. A test asserting that the two orderings agree keeps the property you actually want, and fails loudly if someone reorders the tiers.

kouloumos (approved):
The three examples this PR targets all resolve on the live index. I ran the base query builder and this branch's builder against production, same index, same moment:

| query | base | this branch |
|---|---|---|
| `διαστημόπλοιο κοάλα` | 118 | **0** |
| `υδρονομείς` | 117 | **8** |
| `κολυμβητήριο Μαρουσίου` | 141 | **2** |
| `Μαλτέζος` | 138 | 40, first hit still unrelated |

`Μαλτέζος` is the entity-aware case, so it stays open by design. The typo path works too: `ανακίκλωση` returned unrelated fee-waiver subjects before, and returns actual recycling subjects now.

One thing to settle before merge. The ranking script carries no test, and a mistake in it does not degrade the ranking — it takes search down. I hit this during the review: a mutated field name made the cluster return `script_exception` and fail all shards, so every query errored, the filter-only browse path included. I left the detail and two tests inline.

The other thing worth weighing is the cost of the tighter matching, because it does not fall evenly across cities. I ran the 19 queries from your harness that expect results, once per city. With no city filter, every one of them still returns something. Filtered to Zografou, 4 of the 19 now return nothing at all. Smaller municipalities hold fewer subjects, so requiring more of the query's words, plus the semantic cutoff, leaves less to match against. This does not block the PR. It belongs in the decision about where the cutoff sits, and it is the kind of thing a citizen in a small municipality notices first.

Two smaller things. The branch still carries a `fixup!` commit. And `elasticsearch/README.md:913` says the app queries use "no phrase matching" as part of explaining what raw REF markup would cost — this PR adds two `match_phrase` clauses, so that clause is now inaccurate. Nothing follows from it: `strip-refs` already removes the markup before indexing, and I confirmed zero indexed documents contain a REF token in `name` or `description`. Worth a one-line edit for accuracy, no more.

src/lib/search/query.ts
```diff
@@ -1,6 +1,297 @@
...
+ * among same-tier matches.
+ *
+ * Raw BM25 must not rank within a tier: measured on the production index
+ * ("δάνειο"), two title matches differed by 26% in summed BM25 purely through
+ * title length normalization (3 vs 4 stemmed tokens) and description term
+ * repetition — noise, not relevance — which buried a recent, heavily-debated
+ * subject under an old, briefly-discussed one. Under the flattening the same
+ * pair differs by ~2% before the multiplier, so recency and discussion decide.
+ *
+ * Tier levels are chosen so the multiplier can reorder within a tier but not
```
christosporios: You say that tiers are non-overlapping here

src/lib/search/query.ts
```diff
@@ -85,39 +376,253 @@ export function buildFilters(request: SearchRequest): estypes.QueryDslQueryConta
...
+}
+
+// Lexical should-clauses: BM25 match on title/description/transcripts.
+function buildLexicalShouldClauses(
+    queryText: string,
+    extractedFilters: ExtractedFilters
+): estypes.QueryDslQueryContainer[] {
+    // Name and description sit in different tiers, so each field gets its own
+    // clause (a shared best_fields multi_match could not carry two bases).
+    // Matching several fields sums their tiers — more evidence, higher score —
```
christosporios, line 877: you say that tiers _are_ overlapping here, which contradicts another comment I commented on in the same file.

src/lib/search/query.ts
```diff
@@ -85,39 +376,253 @@ export function buildFilters(request: SearchRequest): estypes.QueryDslQueryConta
...
+// and a geo_distance filter drops every pin-less document. A query like
+// "παλαιστίνη" — extracted as a location and geocoded somewhere — would then
+// return zero results even though subjects carry it in the title. As `should`
+// clauses on the scoring query, nearby pinned subjects rank higher and
+// everything else still matches on text alone.
+function buildLocationBoostClauses(locations: Location[] | undefined): estypes.QueryDslQueryContainer[] {
+    if (!locations || locations.length === 0) return [];
+    return locations.map(loc => ({
+        geo_distance: {
+            distance: `${loc.radius}km`,
```
christosporios: i think this is meters -- so you're querying for 1000-times the real radius here

christosporios (changes requested):
Thank you for this -- it's desperately need. I've never touched elastic, so it's taking me a while to review this and understand how it works.

However I do think there are important flaws that need to be fixed before I have another look.

**1. Scores from different fields add up, but the constants assume they don't.**
The nine lexical clauses sit in one `bool.should`, so a document collects every tier it matches, not the highest one. A document matching introducer + description + phrase + transcript + speaker name reaches base 57.3. A real title match reaches 64. The importance multiplier then decides between them, which is the one thing line 142 says it cannot do (is that a leftover comment from a previous approach?) See my two comments on lines 142 and 455.

**2. The location radius is 1000x too large.**
See inline, query.ts on line 394.

**3. The 75% word rule applies per field, so it blocks the wrong things.**
A councillor's full name (e.g. "Μαρία Παπαδοπούλου") is two words, so it satisfies a three-word query on its own. Every subject that person introduced then qualifies and the topic word stops affecting the order. The same rule makes the `location_text` clause unreachable, since an address never contains the topic word. @kouloumos raised the general form of this.

**4. The semantic cutoff needs both sub-fields, not a good match.**
3.23 sits at 92% of the 3.5 ceiling, and a title-only match caps at 2.0. So a perfect paraphrase of a subject's title can never clear the cutoff. It reads like a quality threshold but behaves as an agreement threshold.

**5. Filter-only search lost date ordering.**
Admin body spans 1.15 and recency only 1.10, so a five-year-old council subject outranks yesterday's community one. Nothing can request newest-first any more, and the MCP tool description now points at ranking that works this way.

Two guards also need a look: the test protecting the location fix in bbc29ee3 reads `bool.filter` one level too high and asserts against an empty array, and `search-eval.ts` counts failures but always exits 0.

christosporios (commented):
I ran a multi-angle review over the branch at 56eaff29. Every item below is verified against the code. I leave out the reindex order, the nested-boundary recall cut, and the semantic candidate pool — we covered those separately.

## 1. A malformed `locationName` crashes the search

`spellingsOf` calls `String.prototype.replace` on `extractedFilters.locationName` (`src/lib/search/query.ts:838`). `aiChat<ExtractedFilters>` is an unchecked cast over model output, so nothing validates the type. A non-string value throws.

Reproduced on this branch:

```
buildSearchQuery({ query: 'πάρκα' }, { ...NO_EXTRACTED_FILTERS, locationName: ['Άργος'] })
TypeError: text.replace is not a function
```

On `main` the value only met a truthiness test, so the same output was inert.

The throw happens inside `buildSearchQuery`, which `src/lib/search/index.ts:134` calls **outside** the two try/catch blocks at lines 105-111 and 119-123. Those blocks exist to keep a bad extraction non-fatal. The request therefore returns a 500 and `sendErrorAdminAlert` posts to Discord.

A `typeof locationName === 'string'` guard restores the previous behaviour.

## 2. The location boosts sum on the scored path

The browse path collapses the clauses (`src/lib/search/query.ts:1037`):

```ts
? [...filters, { bool: { should: locationFilters, minimum_should_match: 1 } }]
```

The scored path spreads them (`src/lib/search/query.ts:1108`):

```ts
? { bool: { must: [textCore], should: locationBoosts } }
```

`main` collapsed them too, and in filter context, so they never scored.

`processFilters` calls `getCities()` with no realm argument and geocodes the name in every municipality (`src/lib/search/filters.ts:151-165`). Adjacent Attica cities bias Google Places toward the same landmark, so about ten near-identical entries are normal. A subject within 2 km of K of them collects K × `LOCATION_BOOST`. At K=10 that is +20 points. That equals the whole `namePhrase` tier and exceeds `descriptionTerm` (15).

The comment at `src/lib/search/query.ts:6-9` states the opposite rule: proximity breaks ties, it does not outrank a better text match. Collapse the scored clauses the same way the browse path does.

## 3. The alternate spellings reach only three clause families

`spellingsOf`/`anySpelling` are wired into the `name`, `description` and `location_text` term clauses. The two `match_phrase` clauses, the fuzzy name clause, both transcript clauses, both speaker-name clauses and the introducer clause all use the raw `queryText`. `introduced_by_person_name` takes `[queryText]` at `src/lib/search/query.ts:826` while its neighbours take `spellingsOf`.

Two results follow.

A score gap. For `τιμολόγια Δ.Ε.Υ.Α.Χ.`, the subjects indexed plain as `Τιμολόγια ΔΕΥΑΧ` reach the 40-point `nameTerm` tier only. An identical subject that the index holds dotted also collects `namePhrase` (20) and `descriptionPhrase` (8). Index punctuation decides a 28-point gap.

A dropped hit. `buildCoverageGate` emits its nested transcript alternative once per spelling (`src/lib/search/query.ts:729`). A subject that names the utility only in its debate passes the gate on the glued spelling. The transcript scoring clause at line 886 then queries the dotted token, the fuzzy clause cannot bridge `δ.ε.υ.α.χ` to `δευαχ` at `prefix_length` 2, and `minimum_should_match: 1` drops the document. The gate pays for a nested query per spelling that cannot produce a hit.

## 4. `nonTitleStack` leaves out the partial name share

`src/lib/search/__tests__/query.test.ts:307` models a non-title document as collecting no name credit. The strict/partial split changed that. The `name` tier splits into strict 26 and partial 14, and the partial clause carries `minimum_should_match: 1`, so one query term fires it.

The test pins `nonTitleStack` at 57.3 and a ratio of 64 / 57.3 = 1.117. A document that matches one title term plus the same stack collects 14 + 57.3 = 71.3, against 64 for a pure title match. The real ratio is 0.90. The stack already outranks the title match, and no metadata multiplier is necessary.

`FIELD_TIER` at `src/lib/search/query.ts:252-260` cites this arithmetic to conclude that name dominance is a property of the corpus. The pin behind that conclusion measures a worst case the code no longer has.

The comment at line 304 ("Every clause a title match fires") also contradicts `src/lib/search/query.ts:249-251`, which says a real title match fires six clauses and reaches about 150-210.

## 5. Search reads the `ADMIN_BODY_TIER` magnitudes

`ADMIN_BODY_WEIGHT` multiplies the raw tier values (`src/lib/search/query.ts:397-404`). The comment above says only the tier order is shared. The other consumer z-scores the values (`src/lib/ranking/subjects.ts:136`), and a z-score is invariant to an affine rescale.

Rewrite the table as `{council: 1, committee: 0, community: -1}`. `rankSubjects` is unchanged and its tests still pass, because they assert only `council > committee > community`. In search, `ADMIN_BODY_WEIGHT.community` and `DEFAULT_ADMIN_BODY_WEIGHT` become 0.85. That is a penalty, and it contradicts the floor the comment promises at lines 402-404. `rankingMultiplierRatio` also assumes every floor is 1.0, so it understates the true ratio and `--tier-margin` reports unsafe pairs as safe.

Declare the search weights against `Record<AdministrativeBodyType, number>` and share only the ordering. A test that asserts the two orderings agree keeps the property you want.

## 6. The eval harness mis-measures itself in three places

**The tier-margin check counts a partial name match as a title match.** `scripts/search-eval.ts:369` uses `startsWith('nameTerm')`, which also matches `nameTerm#2` — the partial clause `stampTierNames` stamps at line 322. A subject that matches one title term of a multi-term query scores base 14 and is still labelled `title: true`. It becomes `worstTitle` against a legitimate 51-point non-title stack, so the check prints an inversion and sets `process.exitCode = 1` for correct behaviour. The same mislabel hides the real inversion in item 4, because such a document can never reach the non-title side. Every current entry in `TIER_MARGIN_QUERIES` is one word, which masks this — `2<75%` collapses to `msm 1` there and both halves fire together.

**The bare run still contains the semantic arm.** `runTierQuery` builds with `enableSemanticSearch: true` and strips only `applyRanking` (`scripts/search-eval.ts:345-353`). Semantic-only hits carry no `matched_queries`, so they join `bareSplit.others`. `bareRatio` then measures `SEMANTIC_MAPPED_BASE`/`SCALE` rather than the field tiers, which the doc block says it measures. The `Δούκας` figure of 1.47 quoted at `src/lib/search/query.ts:276-279` may be this artifact.

**`--mode semantic` prints inflated scores.** `runQuery` replaces the dis_max branches at line 202 but sends `{...q}`, and `q.query` is still the `applyRanking` function_score with `boost_mode: multiply`. A document at the cutoff maps to 26 and prints between 26.0 and about 38.6. That flag exists to calibrate `SEMANTIC_MAPPED_BASE` and `SEMANTIC_MAPPED_SCALE`, so the numbers behind 26 and 320 are metadata-inflated. Which documents survive is correct, because `min_score` gates before the outer multiply.

## 7. `elasticsearch/README.md` still documents the deleted design

The README describes hybrid search as an `rrf` retriever with `rank_window_size` and `rank_constant`. The strings survive at lines 61-62, 710, 791-792, 808-809 and 819. This PR removes both knobs from `SearchConfig` and from `src/app/api/search/route.ts`. The documented `_score` scale is also wrong: RRF gives about 0.016-0.033, the flattened tiers give 25-210.

## A note on the tests

Items 4, 5 and 6 share one cause. The suite asserts the shape of the emitted query, not the calibration. I mutation-tested it, and all 57 tests stay green for each of these edits:

- `ADMIN_BODY_BOOST_WEIGHT` 0.15 → 2.0, `DISCUSSION_LENGTH_BOOST_WEIGHT` 0.03 → 1.0, `RECENCY_BOOST_WEIGHT` 0.1 → 5.0 together. `MAX_RANKING_MULTIPLIER_RATIO` goes from 1.484 to about 122.
- `RECENCY_DECAY_SCALE_DAYS` 365 → 1.
- `scaleTier` stops scaling `k`. Every within-tier log spread doubles.
- `LOCATION_BOOST` 2 → 200, or `FIELD_TIER.locationText` back to the description tier.
- `from: 0` hardcoded on the scored return, or `track_total_hits: false`.

The single assertion that couples the multiplier to the tiers (`query.test.ts:317`) is also monotone in the wrong direction. A wider multiplier raises the right-hand side and makes the assertion easier to pass.

`RANKING_SCRIPT` has no assertion on its source, so a wrong `doc[...]` field name or a Painless syntax error also ships green.

christosporios:
Separate from the review above, and **not a change request on this branch** — `src/lib/search/index.ts` is not in this diff. I raise it here because it sits next to the filter work, and it is easy to miss.

The AI-extracted filters override the caller's explicit filters. `src/lib/search/index.ts:126-131`:

```ts
// Merge with explicit filters
const mergedRequest: SearchRequest = {
    ...request,
    cityIds:   processedFilters.cityIds   || cityIds,
    dateRange: processedFilters.dateRange || request.dateRange,
    locations: processedFilters.locations || request.locations
};
```

The extracted value sits on the left of every `||`. It therefore wins whenever it is truthy. The explicit filter is the fallback, not the authority. The comment states the opposite relationship.

What this reaches:

- **City pages.** `src/components/search/SearchPage.tsx:131` sends `cityIds: cityId ? [cityId] : undefined`. On `/αθηνα`, the query `πάρκα Χαλανδρίου` lets the extractor replace Athens with Chalandri. The page still shows Athens. Nothing tells the user that the scope changed.
- **The MCP `search` tool.** Its schema describes `cityIds` as "Restrict to these city IDs (see list_cities)". An assistant that sets it can lose the restriction.
- **Date ranges.** An explicit `dateRange` loses to the range the model infers from the query text.
- **Empty arrays.** `[]` is truthy in JavaScript. If the extractor returns `cityIds: []`, that value wins the `||`, and `buildFilters` adds no city filter for a zero-length array. The search then widens to every city.

Two notes for scope.

This PR improves one third of it. On `main` an extracted location replaced the caller's location AND became a hard `geo_distance` filter. This branch makes it a proximity boost, so that path now degrades instead of excluding. `cityIds` and `dateRange` keep the old behaviour.

`buildSearchQuery` applies the same precedence a second time (`src/lib/search/query.ts:1018-1022`) to a request that `index.ts` already merged. The result is the same value, so nothing breaks. The rule lives in two files, which is worth collapsing whenever someone changes it.

A caller that states a filter has more information than an extractor that guesses one. The direction I would take: let the explicit filter win, and let extraction fill only the fields the caller left empty. Happy to open this as its own issue if you prefer to track it separately.

christosporios:
One more that sits next to this work, and again **not a change request on this branch**.

The filter extractor still asks the model for `isLatest`, and nothing reads the answer. A grep over `src/` finds only the prompt rule, the type field, and `isLatest: null` fixtures.

It was implemented once. `bf78ea5f feat(search): automatic filter derivation from query` added the field together with its handler:

```ts
// If it's a "latest" query, we need to get the most recent meeting date
if (extractedFilters.isLatest && extractedFilters.cityIds?.length === 1) {
    const latestMeeting = await prisma.councilMeeting.findFirst({
        where: { cityId: extractedFilters.cityIds[0] },
        orderBy: { dateTime: 'desc' },
    …
```

`491ea716 refactor(search): modular structure for search logic` removed the handler. The prompt rule survived (rule 6, "Για ερωτήσεις 'τελευταία', ορίστε isLatest σε true"), the `ExtractedFilters` field survived (`src/lib/search/types.ts:82`), and the per-search model cost survived.

Two effects today.

Every text search asks Claude to compute `isLatest` and then discards the value.

A query with recency intent degrades to a plain relevance search over all time. The word `τελευταία` stays in the query text and scores as an ordinary content term against titles and transcripts. The extractor exists to lift such a word out of the text and into a filter.

The plumbing to restore it is intact and this branch does not touch it. `extractedFilters.dateRange` already flows through `buildSearchQuery` (`src/lib/search/query.ts:1021`) into the `meeting_date` range filter in `buildFilters`. A restored handler only needs to set `dateRange` from the meeting lookup.

Two ways to close it: restore the handler, or drop `isLatest` from the prompt and the type. Either is fine. The current state pays for the field and gets nothing. Happy to open it as its own issue.

christosporios (approved):
Verified the fixes against the branch at 5318e3dd. All seven items are closed.

I re-ran the six mutations that the suite let through before. Every one now fails, including the `doc['administrative_body_type']` field-name typo — so `RANKING_SCRIPT` finally has coverage. 96 tests pass, `tsc` is clean.

Three of the fixes improve on what I asked for:

- The zod schema validates the extraction at the boundary, and per-field `.catch(null)` keeps one malformed field from discarding the whole result. `aiChat<unknown>` removes the unchecked cast that caused the problem.
- `ADMIN_BODY_BOOST_SHARE` gives search its own scale, and the test asserts that the two orderings agree. That keeps the property we want and drops the coupling.
- The tier-margin check now carries multi-term queries, and the measurements show they run tighter than the one-word ones (1.11-1.86 against 1.50-2.57). That is the case the check exists for.

The location boost as a single `constant_score` is neater than the collapse I suggested.

Thanks also for `5318e3dd`. Raising the cutoff to 0.934 on a re-measurement is the right call, and the note records why the old value stopped separating.

Approving and merging! I'll open a follow up PR to fix these two adjacent issues, not introduced by this PR:

- AI-extracted filters override the caller's explicit filters
- `isLatest` is useless, extracted on every search and read by nothing

PR #639, opened 2026-08-17
feat: Notis proactive — poller, batch lane, template sends, rails, kill switch (PR 4)

christosporios:
PR 4 of the [Notis plan](https://github.com/schemalabz/opencouncil/pull/598) (PRD §9), stacked on #634. After PR 3, notis only ever replies. This PR makes it speak first — and lands **paused**: `proactivePaused` reads as true when the setting row is absent, so a fresh deployment starts dark. Unpausing after the inbound-only gate IS the launch. The launch posture has two instruments: gradual release as the throttle (blast radius = who you flagged), the kill switch as the brake.

## What this adds

**The five-minute poller** (`lib/poller.ts`) — every OpenCouncil↔notis link is a pull, so an hour of downtime just means catching up. Four phases, in order:
- **Enrollment.** A rollout-enabled user with phone delivery on and no subscription gets one: profile seeded from preferences, origin `transition`, and the `demos_transition` intro. While paused the phase is skipped entirely, so an enrollee always receives the intro. A subscription is never resurrected.
- **Reconciliation.** Refresh phone and name (write only on change). A removed phone unsubscribes, with a `system` journal entry. The poller never re-activates anyone — re-activation stays with the user (PR 5 checkbox). A missing user row stays the janitor's domain.
- **Scheduled fires.** The agent's `schedule_wakeup` notes come due: fenced on `firedAt`, clamped to active hours, enqueued on the batch lane. Notes of unsubscribed readers are consumed without a wake.
- **Meeting events.** Only `released` meetings (unreleased taskIds are not recorded, so a later release fires naturally). Dedup strictly by `taskId`. The audience comes straight from the live `notis_fanout_targets` view, so a dropped city or a removed rollout flag takes effect the same tick. One editorial pass per event, cached on `NotisProcessedEvent` and embedded in every fan-out; the processed row and all enqueues commit in one transaction. At most 4 editorial passes per tick. `{"seedOnly": true}` marks the backlog consumed for a quiet start.

**Batch lane with coalescing** — `enqueueBatchWake` appends to the subscription's pending batch row (earliest `runAfter` wins), backed by a partial unique index, so a reader in three cities gets ONE wake holding all three briefs. `claimNext` serves both lanes, live first. **Multi-event wakes** go through the agent core: `runWake` takes `events[]`, the prompt renders each event, and the single-event output is byte-identical, so all recorded fixtures replay unchanged. The wake row keeps the primary event in `event` and the full array in a new `events` column.

**Template send path** — `sendTemplate` into the existing conversation, or a cold `createConversationWithTemplate` (Bird requires an initialMessage; a 409 returns the existing conversation id, which the subscription adopts). Templates are addressed by the five new `BIRD_WHATSAPP_TEMPLATE_DEMOS_*` project-id env vars. The retry sweeper covers template sends too.

**Hard rails at the send boundary** (PRD §6), driven by three wake classes:
- **Reactive** (the reader spoke): bypasses every rail — a 03:00 question gets a 03:00 answer.
- **Reply-continuation** (a promised follow-up to their question): cap-exempt, but still held by pause and quiet hours.
- **Unprompted** (meeting news, proactive follow-ups): all rails apply.

The rails: weekly cap of 3 unprompted messages per rolling week, counting only rows that reached or will reach the reader (`pending`/`sent`/`delivered`/`read`); quiet hours 23:00–09:00 Athens with a 45-minute release jitter, applied at enqueue time and never dropping a message; the kill switch (`NotisSetting.proactivePaused`, absent ⇒ paused), which defers items BEFORE the model runs, so a paused system spends no model dollars; and an unsubscribed-mid-flight guard. A rail-stopped message lands as status `suppressed` with the reason in `failureReason`.

**SMS fallback** — a failed WhatsApp delivery of a proactive template fires the same text once as a notify-only SMS. The sms row is inserted first with a unique `fallbackForId`, so a replayed failure webhook can never fire a second SMS. Skipped while paused.

**Scheduled-wake origin** — closes the documented template gap: a follow-up promised while answering the reader rides `demos_followup`; a self-scheduled follow-up to proactive news rides `demos_update_news` and never pretends to answer a question.

**Panel: control** — the kill switch moves from localStorage to the database as THE proactive switch: closed, nothing proactive sends and nobody enrolls; replies and the playground continue. Confirm dialogs in both directions.

**Panel: observability** — a new **Σύστημα** page and a rails strip on the overview, built for the "1000 users" question *what is notis about to do, and why*:
- Heartbeat strip: countdown to the next poller tick, the current active/quiet phase, the switch state, and queue depth. Every countdown renders a bar that drains across its TRUE window (phase start → phase end, tick → tick), not from page load.
- The wake queue by lane (live/batch) with per-item events, attempts, and failure reasons, worst first.
- Readers at the weekly cap.
- The scheduled ledger: every undelivered `schedule_wakeup` note — who, when, why, follow-up vs proactive, with quiet-hour holds labeled.
- The editorial ledger: each digested meeting with its brief — headline, admin body · date · meeting name, per-subject scores as bars (Τ/Π/Ε/Ν/Χ dimensions with hover detail), cost, and a deep link to the meeting.
- Conversation view: upcoming-wake chips («θα επανέλθει» + countdown), suppressed sends with reason, «— εστάλη SMS» markers, coalesced captions.
- `scripts/seed-dashboard-demo.sql` populates every dashboard state with `fake-` prefixed rows for panel work; its cleanup script removes them.

## Schema (notis DB only; no main-DB migration)

`NotisSetting` (settings); `NotisProcessedEvent` += `brief`, `briefCostUsd`, then `meetingName`, `meetingDate`, `adminBodyName` (second migration); `NotisScheduledWake` += `origin`; `MessageStatus` += `suppressed`; `NotisMessage` += `channel`, `proactive`, `fallbackForId`; `NotisWake` += `events`; `NotisSubscription` −= `cities` (the snapshot is gone — wake assembly, fan-out audience selection, and the panel all read the live views; the panel renders without city chips when the main database is absent); a partial unique index for batch coalescing. Every statement is replay-safe.

## Testing

- Unit (notis, 194 tests): active-hours boundaries and both DST transitions; Bird payload shapes incl. the 409 recovery and SMS; the full rails matrix (pause pre-model defer, cap edges and counting rule, reply-continuation exemption, unsubscribed race, cold-send bootstrap); multi-event prompt render with a single-event byte-regression guard; the poller's four phases against fakes (never-resurrect, released gate with later-release pickup, taskId dedup, editorial-failure continuation, per-tick cap, seedOnly, paused-skips-enrollment).
- Integration (main repo, real Postgres, 20 notis tests): batch coalescing under 5 concurrent enqueues, enqueue-vs-claim, lane priority, the defer fence, and the whole migration history replayed twice.
- Golden replays: all fixtures green, unchanged on disk.
- Live dev E2E against the dev databases: see the PR comment thread.

## Deploy checklist (on merge)

1. Meta status: confirm the revised `demos_*` templates are re-approved in Bird — blocks unpausing only; the paused system needs nothing.
2. Set the five `BIRD_WHATSAPP_TEMPLATE_DEMOS_*` project-id env vars on the notis app; verify the settings-page badges.
3. `prisma migrate deploy` for `services/notis` (no main-DB migration).
4. Post-deploy: the settings switch shows **closed** (paused is the default with no row); poller last-tick populated within ~6 minutes.
5. Backlog decision: `POST /api/admin/poll {"seedOnly": true}` marks the pre-deploy backlog consumed. Events released during the paused gate coalesce into at most ONE batch wake per (inbound-enrolled) reader; on unpause the agent sees each event's age and the weekly cap still applies.
6. Unpausing is the launch: the PRD gate first (two weeks of inbound-only operation), then flip the switch, and drill it closed once.

## Known limitations (deliberate)

- Anthropic Batch API for batch-lane wakes is a later cost lever; wakes drain synchronously.
- No heartbeat event: scheduled wakes cover due notes; nothing produces heartbeats.
- Multi-realm MCP origin selection stays hardcoded to Greece (the views already carry realm/timezone for the seam).
- PostHog per-wake analytics lands with PR 5. The PRD's review queue is cut (2026-08-18): the send-filtered wakes feed is the review surface.

<h3>Greptile Summary</h3>

The PR adds proactive Notis polling, coalesced batch wakes, template delivery, send-time rails, SMS fallback, database-backed controls, and operational dashboards.

- Introduces enrollment, reconciliation, scheduled-wake, and meeting-event polling.
- Adds batch coalescing and multi-event agent execution.
- Adds proactive delivery templates, quiet-hour and weekly-cap rails, a kill switch, and SMS fallback.
- Extends the Notis schema and admin observability surfaces.

The PR is not yet safe to merge because an enrollment can still commit an active subscription while permanently losing the reader’s required transition intro.

The transition intro is delivered only after enrollment commits, while pause or terminal delivery failures make the message ineligible for the pending-only retry path and the existing subscription prevents later enrollment ticks from recreating it.

**Files Needing Attention:** services/notis/src/lib/poller.ts, services/notis/src/lib/queue.ts

| Filename | Overview |
|----------|----------|
| services/notis/src/lib/poller.ts | Adds the four-phase proactive poller, but enrollment still permits a committed subscription to permanently lose its required transition intro. |
| services/notis/src/lib/queue.ts | Adds coalescing, lane-aware claiming, proactive rails, template delivery, retries, and fallback handling. |
| services/notis/src/agent/runWake.ts | Extends wake execution to process coalesced event arrays while preserving single-event behavior. |
| services/notis/prisma/schema.prisma | Adds proactive settings, delivery metadata, scheduled-wake origins, event arrays, and editorial cache fields. |
| services/notis/src/lib/bird.ts | Adds WhatsApp template and SMS delivery operations, including existing-conversation recovery. |
| services/notis/src/app/admin/(panel)/system/page.tsx | Adds operational views for queue state, scheduled wakes, capped readers, and processed meeting events. |

<sub>Reviews (15): Last reviewed commit: ["Merge feat/notis-inbound into feat/notis..."](https://github.com/schemalabz/opencouncil/commit/307b5dd21e6b6d0ff6af47ea733f50821abcb1dc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54367688)</sub>


christosporios:
Live E2E in shadow mode against the dev databases, on the real running service (the instrumentation poller, not a test harness):

1. **Enrollment**: flagged a test user (`notisEnabledAt = now()`) in the main dev DB. The next 5-minute tick enrolled them — subscription `active`/`transition`, phone normalized, profile seeded — and recorded the `demos_transition` intro as **suppressed («shadow mode»)** with zero Bird calls. The earlier tick had also correctly refused to resurrect an unsubscribed subscription and refreshed one stale cities snapshot.
2. **Scheduled fire → real wake**: inserted a due `NotisScheduledWake` (origin `proactive`). The next tick fired it (fenced — `firedAt` set once) onto the **batch lane**; the sweep claimed it and ran a **real model wake**: 16s, $0.20, grounded MCP research, decision **silence** with an operator-voice rationale («Checked recent Athens subjects across administration/budget topics… everything found is routine»). Journal seq 1 = enrollment, seq 2 = scheduled/silence. Queue item `done` after 1 attempt.

Suites: notis 193 ✓, root tsc (jest scope) ✓, integration 89/91 — the 2 failures are the documented pre-existing ones that also fail on `main`; all 20 notis integration tests green, including batch coalescing under 5 concurrent enqueues and the migration history replayed twice. Golden replays green with fixtures unchanged on disk.

christosporios:
Pushed bb706e03 — observability tier 1, per discussion:

- **Countdown component**: live remaining-time text over a draining progress bar; SSR-safe, second-resolution.
- **«Σύστημα» page**: poller heartbeat with a real countdown to the next tick; quiet-hours phase and its release countdown; mode/pause state; the wake queue worst-first (running, failed with errors, pending with per-item deadlines); «Τι σκοπεύει να κάνει» — the agent's un-fired scheduled notes with origin badges and «κρατείται ως τις 09:00» notes; and the editorial ledger, where each digested meeting expands into its brief (subjects with the five scores, cost, fan-out width).
- **Overview rails strip**: mode, phase countdown, held/at-cap counts, and the period's suppression breakdown in one line, linking to the system page.
- **Conversation threads** show upcoming scheduled notes as system chips with countdowns, and a proactive-origin scheduled wake now captions «προγραμματισμένο» instead of «follow-up».

Lists are capped with honest «+N» counts, pages self-refresh while visible, and the at-cap panel reuses the exact countable rule the send boundary applies. Verified in the browser against the live dev databases: real poller countdown, the E2E conversation's chip, and the strip's «2 shadow» suppressions all render.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/bb706e030f6e505533329ef8e8a6d02b09f86ed0..5b59fcce42e7f309e9ab1522b7daa6588b0ed00d) to fix panel wording (squashed into the panel and observability commits): «προωθητικά» → «proactive» (no good Greek term — the settings page already used the English word), «εκδοτικό πέρασμα» → «editorial pass», and the queue header's cryptic «0 live · 0 batch» now reads «ενεργά ανά lane: …».

christosporios:
Pushed four commits (batched since the observability push):

- `d647e035` **demo seed for the dashboard** — `scripts/seed-dashboard-demo.sql` + cleanup: `fake-` prefixed rows that light up every dashboard state (all four queue states, at-cap, holds, briefs).
- `ed4ca366` **editorial ledger polish** — briefs render as headline + ranked subjects with per-dimension score bars (hover for name/value), cost, and a deep link to the meeting.
- `00d2f686` **meeting metadata on the ledger** — the poller records admin body, meeting date, and meeting name on `NotisProcessedEvent` (second migration); ledger rows show «Δημοτικό Συμβούλιο · 18 Αυγ · 18η Τακτική Συνεδρίαση».
- `8d9dbd9a` **shadow mode removed** — pause-by-default replaces it: `proactivePaused` reads true when the row is absent, so a deployment lands dark and unpausing IS the launch. One switch instead of two, one enrollment path, one cap-counting rule. Also folds in: countdown bars drain across their true window (phase start → end), fixed-width lane badges, wider name columns.

The PR body and deploy checklist are rewritten for the pause-by-default story.

christosporios:
Pushed `6a71b3c3`: the `NotisSubscription.cities` snapshot is removed — everything reads the live views now.

- Wake assembly already read the live view first; the snapshot was only an outage fallback, and during a main-DB outage the MCP server is down too, so the fallback protected an un-grounded wake.
- Fan-out audience selection asks `notis_fanout_targets` directly. Two behavior changes, both improvements: a city removed from the preferences stops producing wakes the same tick, and a user whose rollout flag is cleared drops out of fan-out immediately.
- The panel fetches city names with one bulk view read per page and renders without city chips when the main database is absent.
- Gone with it: the poller's snapshot-refresh sub-phase, the `citiesRefreshed` counter, and the fallback branch. Migration `20260818130000` drops the column.

Verified on dev: the conversation list and detail now show «Αθήνα, Ζωγράφου» for a reader whose stale snapshot held only Athens. The PR body's reconciliation and schema sections are updated.

christosporios:
Pushed three commits fixing everything a max-effort review of this branch turned up (10 finder angles, 26 adversarial verifications, 24 confirmed).

**`467abde3` — three queue-core races.** The worst could stall the whole queue permanently: a worker dying mid-batch-wake leaves a running row, the reader then texts in, and live-first ordering floated the new live row above the stale one — promoting it collided with `one_running_per_sub` on every pass, and a null claim reads as "queue empty", so every *other* reader's work stranded behind it. Also: `failItem`/`deferItem` collided with the coalescing index when a fresh pending row had taken the slot (a paused system burned attempts to terminal failure and dropped the briefs it was holding) — they merge now; and `enqueueBatchWake`'s catch-and-retry could not work for either production caller, since a 23505 inside their interactive transaction aborts it and the retry died with 25P02, taking the tick's fan-out down. An advisory lock serializes it instead.

**`f7ad47eb` — rails, clocks, Bird payloads.** The rails now ride with the row at the delivery choke point: the sweeper was re-sending transiently failed proactive rows up to an hour later with no re-check, so a reader who texted ΣΤΟΠ (or a closed kill switch) still got the template. The 24h window and the model's `<current_time>` are both judged at the send moment rather than the event's timestamp — a quiet-clamped wake was picking freeform for a window that had closed (silently lost: freeform rows get no SMS fallback) and writing «σήμερα» about yesterday. The weekly cap moved before the model, and suppression now appends a journal correction, because the wake's entry commits the message texts before any rail runs — the agent's own memory claimed the reader had been told. Quiet hours now cover the enrollment intro and the SMS fallback. Plus: a missing template project id holds enrollment instead of burning the cohort, concurrent ticks can't double-send an intro, template sends name their recipient, a cold send never stores the conversation id as the message id, the 409 recovery no longer scans the raw body, and a main-DB outage retries instead of completing a wake that told the agent this reader follows no cities.

**`faf4f19d` — panel, seed, docs.** Enrollment entries desynced the conversation viewer's delivery pairing. The demo seed was live ammunition — pending rows on real subscriptions that the real drainer claims — and is now inert by construction. The CLI runner still promised shadow-mode safety this branch had deleted.

Verification: 203 unit tests and 17 notis integration tests against real Postgres (the stale-row reclaim and both merge paths are covered there), both typecheck scopes clean, panel re-verified in the browser.

Two notes for review: `sendTemplate`'s missing `recipients` array was the one finding I could not confirm from code alone — Bird documents the field as optional, but no template has ever been sent without it from this workspace, so I added it rather than waiting for a production 4xx. And the cap now drops a saturated reader's unprompted events before the model instead of paying for a wake whose message the boundary would suppress; the journal records what went unexamined.

christosporios:
Pushed two commits and rebased onto the updated stack:

- `7f64b045` — meeting events dedup by `(cityId, meetingId, type)` rather than by task id, so a re-processed meeting is not fresh news. Tests cover both directions: a re-run does not notify, and a meeting whose first task *failed* still does. The fake database now enforces the same unique identity as the index.
- `7bbef69e` — the migration that dropped the `cities` column is deleted; the column no longer exists to remove.

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/7bbef69e0b1587974359cf4821976538e49bbaf8..340f877a046c8a121e66c9b1b54352dc260c96c7>) to rebase onto `main` after #642 merged. Same diff; the only resolution was a duplicate `.gitignore` rule for `services/notis/generated/` that #642 added first — main's wording is kept.

christosporios:
Pushed `76e452b5`: the fixes from #634's review, re-established on the shapes this PR rewrites — persist-transaction lock ordering and budget, the sweeper's database-level send claim, the fenced give-up pass, retry backoff, bounded Bird calls, 408/429 retryability, monotonic delivery status, the inbound phone seam, per-message thread ordering, and the fake database's journal scoping.

One of those was only visible through a test: the fake's where engine treated a column a row never set as `undefined` rather than NULL, so `{ sendingAt: null }` matched nothing and the sweeper's own tests went quiet rather than failing.

212 unit tests, 33 notis integration tests, both typechecks clean.

christosporios:
Merged `feat/notis-inbound` forward (`0db0d7dd`, plain merge, no conflicts) — current main plus the paired-preview wiring, see [the #632 comment](https://github.com/schemalabz/opencouncil/pull/632#issuecomment-5385384474). Verified on this branch: notis 25 suites / 212 tests, root units 1578, integration 11 suites / 99 tests, tsc clean.

christosporios:
Pushed `8a118879` — the AI Act (Article 50) disclosure in every template footer: `«Μήνυμα με τεχνητή νοημοσύνη. ΣΤΟΠ για διακοπή.»` (the transition cohort keeps its email-narrowing variant). A new test asserts every footer carries the disclosure and fits WhatsApp's 60-character cap.

The Bird side is already done and verified today: all six `demos_*` templates have the disclosure version **Active** (Meta approved each within minutes on 19/08). This also resolves the deploy-checklist item about template re-approval — and it un-bricked `demos_intro`/`demos_transition`, which had been stuck Rejected since 15/08 on Meta's once-per-24h edit rule.

christosporios:
Merged `feat/notis-inbound` forward (`315a40cf`) — brings the RSC-bypass guards (#1) and migrate-on-start (#2) up the stack; see the [#632 comment](https://github.com/schemalabz/opencouncil/pull/632#issuecomment-5385784756). The merge also guards `system/page.tsx` (the panel page this PR adds) and marks its `_lib` reader `server-only`, so the incoming auth-guard scan test stays green.

Plus one fix specific to this PR:

- `3a272bd5` **fix(notis): give the meeting fan-out transaction a budget for the audience** — review finding #3. The fan-out wrapped the processed-event row plus one `enqueueBatchWake` per subscriber in a single transaction on Prisma's default 5s budget; it rolls back at a few hundred subscribers, and since the paid editorial pass runs before the transaction, a rollback makes the next tick re-run it and time out again — a permanent no-delivery loop that keeps re-billing the model. Raised to 120s (well above the queue's 30s persist budget, since this one scales with the audience) with a wider `maxWait`. Noted in the code that a set-based enqueue is the real fix once a single city outgrows one transaction.

Verified: notis 26 suites / 215 tests, root units 1578, integration 99, tsc clean.

christosporios:
Pushed the consent cluster (review findings #4–#7) as one choke-point refactor:

- `9dee33fe` **rail and claim every delivery path, not just the first send** — #4, #5, #6. The rails and the send claim lived only at the first-send boundary, so three later paths could reach a reader who had withdrawn consent:
  - **#6** the sweeper's retry ran the rail only when `proactive` was true, but a reply-continuation follow-up is a template send with `proactive: false` — a transient failure plus a ΣΤΟΠ re-sent it to an unsubscribed reader. The rail now keys on *unprompted-at-delivery* = `proactive` OR any template send, so a free-form reactive reply (the ΣΤΟΠ confirmation) still goes out and every template respects the rails.
  - **#5** the 09:00 held-SMS release checked only the hold marker; it now re-checks the same rails and suppresses instead of sending.
  - **#4** only the sweeper claimed a row (`sendingAt`) before sending, and its 30s window was shorter than the cold path's two sequential Bird calls. The claim now lives in `deliverPendingMessage` (a shared `claimForSend`) so the boundary and the sweeper fence against each other, with a TTL above the worst-case send.
- `2b550af7` **stop firing scheduled wakes for a rolled-back reader** — #7. A panel Disable clears `notisEnabledAt` but the poller never removes the subscription, so a scheduled wake kept firing and double-served. `fireScheduledWakes` now consumes a de-flagged reader's note without a wake (fenced), reading the flag from the main views and failing open when the main DB is unreachable — the same stance as the inbound webhook and the fan-out audience query. Subscription lookups are batched into one query.

Nine new tests drive the retry/release/scheduled paths directly. Each is mutation-checked — reverting the fix makes its test fail. Verified: notis 27 suites / 224 tests, root units 1578, integration 99, tsc clean.

christosporios:
Merged the build fix forward (`8cd58357`). See the [#632 note](https://github.com/schemalabz/opencouncil/pull/632) — `server-only` removed from the two mixed `_lib` modules that client components import, so `next build` passes. Verified locally this time (build + tsc + 224 tests); CI re-running.

christosporios:
Pushed `07942541` — a redesign of the agent's memory, closing review finding #9 at the root instead of patching it:

**The journal table is gone.** It stored three things — what the agent said, what the reader wrote, why the agent acted — and the first two were copies written *before* the send outcome was known. When a rail suppressed a message, a separate correction step had to walk the copy back; losing that step (a restart between its two statements) left the agent believing it had sent a message that never went out. The copies also duplicated `NotisMessage` and `NotisWake` row for row.

**Two sources, each authoritative:**
- `<conversation>` ← `NotisMessage`: inbound rows + outbound rows whose own delivery status says they reached the reader. A suppressed/failed send simply never appears — no correction step exists to lose.
- `<decisions>` ← `NotisWake`: trigger → send/silence → why. The shell's model-less decisions are wake rows now too (ΣΤΟΠ pre-step, cap skip, phone-gone unsubscribe), with `model`/`trace` nullable as the "no model ran" marker. The ΣΤΟΠ reply carries its wake's id, which deletes the panel's index-alignment cursor for wake-less exchanges.

Also gone: the per-subscription `seq` allocation and its row-lock discipline — the exact machinery finding #9's race lived in. Migration `20260823130000` (replay-safe; drops the table, relaxes the two columns).

Simulation surfaces follow: `applyOutcome(state, events, outcome)` evolves a simulated state the way the real records would, the playground bumps its store version, fixtures transformed mechanically (no re-record — replays assert outcomes, not prompt bytes).

Verified: notis 27 suites / 226 tests + production build, root 1578 units, integration 99, tsc clean. Two of the new tests drive the redesign directly: the model's conversation excludes a suppressed row, and its decision log arrives from the wake rows.

christosporios:
Pushed the remaining review fixes (plus the merge bringing #13/#14 up the stack):

- `d0eeee6e` **409 conversation id from details only, matched whole** — finding #8. The extractor adopted root-level ids and substring UUIDs; either poisons the subscription (every later send targets a conversation that does not exist until the reader writes in). Now mirrors the main app's details-only, whole-value shape, keeping the one deliberate improvement (extracting from the two resource-path keys). Tests pin the root-id and substring rejections.
- `434bad46` **never treat an old meeting as news** — finding #11. `completedAt` is `TaskStatus.updatedAt` underneath, so a `batchRerun --force` or a bulk touch of old rows re-surfaces years-old meetings as fresh events — an editorial pass each plus WhatsApp to whole cohorts, unrecallably. Events whose meeting is older than 30 days are now consumed without a wake (recorded like seedOnly, no spend, nobody messaged); agenda events for upcoming meetings are untouched.
- `550bcec9` **a reader cannot type their way out of the message fence** — finding #15. A literal `</reader_message>` in a WhatsApp message ended the data fence and let the rest pose as shell-authored prompt text. Reader-typed fence delimiters are now visibly neutralized (`[reader_message]`) at both render sites.

All three mutation-checked — each test fails against the pre-fix code. Finding #10 (failed-intro orphan) is deferred by decision to #671. Verified: notis 27 suites / 231 tests + production build, root 1586 units, integration 99, tsc clean.

christosporios:
Pushed `08464269` — **the agent reads links the reader shares** (user-requested addition).

One tool entry: Anthropic's server-side `web_fetch` (basic `20250910` variant — no code-execution involvement), capped at 3 fetches / ~15k content tokens per wake. No fetching code in this repo; it runs inside the same API call, like the MCP connector. The API's own URL validation only permits URLs that already appeared in the conversation — the reader's message — never model-constructed ones, which closes the arbitrary-fetch/exfiltration vector by design.

`system.md` sets the product boundaries: a personal site feeds the taste profile as distilled facts; an article is the reader's question in longer form, answered from the municipal record; fetched pages are data, never instructions, and never a source for what the council said.

**Live-verified** with a real wake: the reader's link produced a `server_tool_use` fetch in the trace and the reply reasoned about the page's actual content («Άνοιξα τον σύνδεσμο, αλλά είναι απλώς μια δοκιμαστική σελίδα…»), $0.17. The trace inspector renders the fetch blocks; the golden request-invariant pins the tool and its caps. notis 27 suites / 231 tests + production build green.

christosporios:
Merged the [#632 review-response commits](https://github.com/schemalabz/opencouncil/pull/632) up the stack (`29eff403`; one trivial package.json union conflict), plus `913adef9`: the last admin-surface emoji (the simulated thread header's clock) becomes a lucide icon — the in-thread ✓✓ ticks stay, they are the WhatsApp mock's own delivery marks. Verified green: notis 27 suites / 232 tests + build, root 1586, integration 100.

christosporios:
Merged the stack forward onto current `main` (via #632/#634; no conflicts anywhere). Full sweep on the tip: root 1639 units, notis 232 + production build, integration 100, tsc clean.

PR #641, opened 2026-08-18
feat(transcribe): store utterance confidence scores from task results

christosporios:
## Summary

schemalabz/opencouncil-tasks#89 adds two confidence scores to each utterance in the transcribe result: `minWordConfidence` and `totalConfidence`. Transcribe task version 4 sends them next to the existing mean `confidence`. The import handler dropped all three scores. This PR stores them.

## Changes

- Add three nullable `Float` columns to the `Utterance` model: `confidence` (mean of word confidences), `minWordConfidence` (least confident word), `totalConfidence` (product of word confidences).
- Add migration `20260818000000_add_utterance_confidence_scores`.
- `handleTranscribeResult` writes the scores when it creates utterances.
- Mirror the two new fields on the `Utterance` interface in `apiTypes.ts`. They are optional there because `processTaskResponse` can replay stored results from task versions before 4. The handler stores `null` for missing scores.
- Add two tests for the handler: one for a v4 result, one for a replayed pre-v4 result.

## Design notes

- The columns are nullable and have no default. A default of 1 would claim full confidence for old utterances. Null keeps "unknown" distinct from a real score.
- No backfill is possible: the pipeline does not store `Word` rows, so the scores for existing utterances are not recoverable. A re-transcribe stores them.
- The mean `confidence` is included because the handler dropped it before this change. Without it, the stored trio is incomplete.

## Testing

- `npm test` — 105 suites pass, including the two new handler tests.
- `npx tsc --noEmit` — clean.
- `npm run build` with the same dummy env as the Nix build — passes.
- This container has no local database, so `prisma migrate dev --create-only` cannot run. The migration SQL comes from `prisma migrate diff --from-schema-datamodel --to-schema-datamodel --script` with the repo-pinned Prisma 5.22.0. The diff engine is the same one `migrate dev` uses.

https://claude.ai/code/session_01AQr6jhRWwMFtuv2oVqPj75

---
_Generated by [Claude Code](https://claude.ai/code/session_01AQr6jhRWwMFtuv2oVqPj75)_

<h3>Greptile Summary</h3>

Adds nullable confidence-score fields to utterances and persists rounded values from transcription task results while retaining compatibility with older results.
- Adds the corresponding Prisma schema fields and database migration.
- Stores mean, minimum-word, and total confidence during transcript import.
- Covers current and pre-v4 task results with handler tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| prisma/migrations/20260818000000_add_utterance_confidence_scores/migration.sql | Adds three nullable double-precision columns matching the Prisma model. |
| prisma/schema.prisma | Defines nullable utterance confidence fields without assigning misleading defaults to historical records. |
| src/lib/apiTypes.ts | Extends the transcription utterance result shape with optional v4 confidence fields. |
| src/lib/tasks/transcribe.ts | Rounds and persists confidence values while mapping missing legacy fields to null. |
| src/lib/tasks/__tests__/handleTranscribeResult.test.ts | Tests v4 score persistence, significant-figure rounding, and replay of older task results. |

<sub>Reviews (2): Last reviewed commit: ["feat(transcribe): store utterance confid..."](https://github.com/schemalabz/opencouncil/commit/d4f910123a95a711309ab351875b8ea89a4cb86e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54331190)</sub>

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -350,6 +350,15 @@ model Utterance {
   createdAt       DateTime @default(now())
   updatedAt       DateTime @updatedAt
 
+  // Transcription confidence scores from the transcribe task (version 4+).
+  // Null for utterances imported from older task results.
+  // The scores describe the text as the ASR transcribed it. fixTranscript and
+  // user edits rewrite text and keep the scores, so join on lastModifiedBy to
+  // tell scored-as-transcribed utterances from corrected ones.
+  confidence        Float? // arithmetic mean of word confidences
+  minWordConfidence Float? // confidence of the least confident word
+  totalConfidence   Float? // product of word confidences ≈ P(every word is right)
+
   speakerSegment SpeakerSegment @relation(fields: [speakerSegmentId], references: [id], onDelete: Cascade)
   speakerSegmentId String
   uncertain Boolean @default(false)
--- a/src/lib/tasks/transcribe.ts
+++ b/src/lib/tasks/transcribe.ts
@@ -9,6 +9,10 @@ import { requestTranscribeInternal, deleteExistingSpeakerData } from "./transcri
 import { requestFixTranscript } from "./fixTranscript";
 import { sendTaskAdminAlert } from "../discord";
 
+// Full-precision doubles are near-incompressible and inflate the meeting page
+// payload; 4 significant figures is far finer than the ASR signal warrants.
+const round4 = (v: number | null | undefined) => (v == null ? null : Number(v.toPrecision(4)));
+
 /**
  * Public entry point for transcription. Authorizes the caller, then delegates to
  * requestTranscribeInternal (which the unauthenticated poll-livestreams cron also calls).
@@ -198,6 +202,12 @@ export async function handleTranscribeResult(taskId: string, response: Transcrib
                                     endTimestamp: utterance.end,
                                     text: utterance.text,
                                     drift: utterance.drift,
+                                    // Replayed results from transcribe versions before 4
+                                    // lack the scores; null keeps "unknown" distinct from
+                                    // a real value.
+                                    confidence: round4(utterance.confidence),
+                                    minWordConfidence: round4(utterance.minWordConfidence),
+                                    totalConfidence: round4(utterance.totalConfidence),
                                 }))
                             }
                         }
--- a/prisma/migrations/20260818000000_add_utterance_confidence_scores/migration.sql
+++ b/prisma/migrations/20260818000000_add_utterance_confidence_scores/migration.sql
@@ -1,0 +1,4 @@
+-- AlterTable
+ALTER TABLE "Utterance" ADD COLUMN     "confidence" DOUBLE PRECISION,
+ADD COLUMN     "minWordConfidence" DOUBLE PRECISION,
+ADD COLUMN     "totalConfidence" DOUBLE PRECISION;
--- a/src/lib/apiTypes.ts
+++ b/src/lib/apiTypes.ts
@@ -228,7 +228,11 @@ export interface Utterance {
     language: string;
     start: number;
     end: number;
-    confidence: number;
+    confidence: number; // arithmetic mean of word confidences
+    // Optional because processTaskResponse can replay results stored by
+    // transcribe task versions before 4, which lack these two scores.
+    minWordConfidence?: number; // confidence of the least confident word
+    totalConfidence?: number; // product of word confidences ≈ P(every word is right)
     channel: number;
     speaker: number;
     drift: number;
--- a/src/lib/tasks/__tests__/handleTranscribeResult.test.ts
+++ b/src/lib/tasks/__tests__/handleTranscribeResult.test.ts
@@ -68,17 +68,17 @@ const response = {
   },
 };
 
-describe('handleTranscribeResult — fixTranscript auto-chain', () => {
-  beforeEach(() => {
-    jest.clearAllMocks();
-    mockTaskFindUnique.mockResolvedValue(task);
-    mockMeetingUpdate.mockResolvedValue({ id: MEETING_ID });
-    mockSpeakerTagCreate.mockResolvedValue({ id: 'tag-1' });
-    mockSpeakerSegmentCreate.mockResolvedValue({ id: 'segment-1' });
-    mockRequestFixTranscript.mockResolvedValue({ id: 'fix-task-1' });
-    mockSendAutoFixAlert.mockResolvedValue(undefined);
-  });
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockTaskFindUnique.mockResolvedValue(task);
+  mockMeetingUpdate.mockResolvedValue({ id: MEETING_ID });
+  mockSpeakerTagCreate.mockResolvedValue({ id: 'tag-1' });
+  mockSpeakerSegmentCreate.mockResolvedValue({ id: 'segment-1' });
+  mockRequestFixTranscript.mockResolvedValue({ id: 'fix-task-1' });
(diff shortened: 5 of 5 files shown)


src/lib/tasks/transcribe.ts
```diff
@@ -198,6 +198,12 @@ export async function handleTranscribeResult(taskId: string, response: Transcrib
                                     endTimestamp: utterance.end,
                                     text: utterance.text,
                                     drift: utterance.drift,
+                                    // Replayed results from transcribe versions before 4
+                                    // lack the scores; null keeps "unknown" distinct from
+                                    // a real value.
+                                    confidence: utterance.confidence ?? null,
+                                    minWordConfidence: utterance.minWordConfidence ?? null,
```
kouloumos: These land in `Utterance` rows that `getTranscript` pulls with `include` and no `select` (`lib/db/transcript.ts:41`), so all three serialize into the meeting page payload on every load.

Measured on `nis/mar17_2026` — 11,148 real utterance rows, confidence values simulated from the benchmark distribution:

| | raw | gzip |
|---|---|---|
| today | 4.57 MB | 675 KB |
| + 3 fields, full precision | 5.71 MB | 1050 KB (**+56%**) |
| + 3 fields, rounded to 4 s.f. | 5.33 MB | 810 KB (+20%) |

The gzip cost is disproportionate to the raw one because full-precision doubles (`0.9712345678901234`, which is what `exp()` and a running product produce) are near-incompressible, while the rest of the row is highly repetitive. Rounding at write time recovers about two thirds of it and discards nothing usable — 4 significant digits is far finer than the signal warrants:

```ts
const round4 = (v: number | null | undefined) => (v == null ? null : Number(v.toPrecision(4)));

confidence: round4(utterance.confidence),
minWordConfidence: round4(utterance.minWordConfidence),
totalConfidence: round4(utterance.totalConfidence),
```

Non-blocking, but `getMeetingData.ts:113` already carries `// Transcript is NOT cached — too large for the 2MB unstable_cache limit`, so this payload is a place where bytes are already a known constraint.

prisma/schema.prisma
```diff
@@ -350,6 +350,12 @@ model Utterance {
   createdAt       DateTime @default(now())
   updatedAt       DateTime @updatedAt
 
+  // Transcription confidence scores from the transcribe task (version 4+).
+  // Null for utterances imported from older task results.
+  confidence        Float? // arithmetic mean of word confidences
+  minWordConfidence Float? // confidence of the least confident word
+  totalConfidence   Float? // product of word confidences ≈ P(every word is right)
```
kouloumos: Worth one line noting these describe the **original ASR text**. `handleFixTranscriptResult` (`lib/tasks/fixTranscript.ts:31-38`) rewrites `text` and leaves the three scores untouched, and the import auto-chains `requestFixTranscript(..., { force: true })` (`transcribe.ts:235`), so any utterance the fixer edits keeps a score describing text it no longer holds.

I would *not* null them on update — a low-confidence utterance that the fixer then corrected is exactly the case worth finding later. Just a comment, so the first consumer doesn't read them as describing current text.

src/lib/tasks/transcribe.ts
kouloumos: These land in `Utterance` rows that `getTranscript` pulls with `include` and no `select` (`lib/db/transcript.ts:41`), so all three serialize into the meeting page payload on every load.

Measured on `nis/mar17_2026` — 11,148 real utterance rows, confidence values simulated from the benchmark distribution:

| | raw | gzip |
|---|---|---|
| today | 4.57 MB | 675 KB |
| + 3 fields, full precision | 5.71 MB | 1050 KB (**+56%**) |
| + 3 fields, rounded to 4 s.f. | 5.33 MB | 810 KB (+20%) |

The gzip cost is disproportionate to the raw one because full-precision doubles (`0.9712345678901234`, which is what `exp()` and a running product produce) are near-incompressible, while the rest of the row is highly repetitive. Rounding at write time recovers about two thirds of it and discards nothing usable — 4 significant digits is far finer than the signal warrants:

```ts
const round4 = (v: number | null | undefined) => (v == null ? null : Number(v.toPrecision(4)));

confidence: round4(utterance.confidence),
minWordConfidence: round4(utterance.minWordConfidence),
totalConfidence: round4(utterance.totalConfidence),
```

Non-blocking, but `getMeetingData.ts:113` already carries `// Transcript is NOT cached — too large for the 2MB unstable_cache limit`, so this payload is a place where bytes are already a known constraint.

prisma/schema.prisma
kouloumos: These describe the **original ASR text**, and a meaningful share of utterances no longer hold that text. Worth saying so in the comment.

`handleFixTranscriptResult` (`lib/tasks/fixTranscript.ts:31-38`) rewrites `text` and leaves the scores untouched, as does the user edit path (`lib/db/utterance.ts:26`). Measured on production, Scribe-era utterances (`createdAt >= 2026-06-10`, i.e. what v4 imports will look like):

| | utterances | % |
|---|---|---|
| unmodified | 115,683 | 79.3% |
| edited by user | 21,245 | 14.6% |
| rewritten by `fixTranscript` | 8,918 | 6.1% |

**20.7% overall**, and it isn't a few outlier meetings — across the 109 Scribe-era meetings with ≥200 utterances, the average is 24.8% modified and *zero* meetings are untouched. (Not an artifact of the empty utterances `speakerTags.ts:119` creates pre-marked `'user'` — only 3 of those 21,245 have empty text.)

The reason it matters for the first consumer: the natural query is "worst N by `totalConfidence`, send for human review", and that will keep re-surfacing utterances someone already corrected. I'd expect the staleness to concentrate in exactly the low-confidence tail such a query selects, since `fixTranscript` targets utterances the ASR got wrong and low confidence predicts those well — though that last step is inference, not something I can measure until the columns hold real values.

`lastModifiedBy` already discriminates the affected rows, so no schema change is needed and I would *not* null the scores on update — a low-confidence utterance that was then corrected is worth being able to find. Just say in the comment that these describe the text as transcribed, so a consumer knows to join on `lastModifiedBy`.

kouloumos (commented):
Checked the migration three ways: it is byte-identical to what `prisma migrate diff` regenerates from the base schema (only a trailing newline differs), it sorts after `20260815000000`, and the SQL runs in ~0.5 ms against a local PG16 `Utterance` table loaded with 200k rows — metadata-only `ADD COLUMN`, no rewrite. Nullable with no default is the right call and your reasoning for it holds.

Also confirmed the import path is the only thing creating ASR utterances: the two editor paths in `lib/db/speakerSegments.ts` and the seed's `...utterance` spread are all correctly unaffected, so there is no second writer silently storing nulls.

One note inline about payload size. Non-blocking.

(I had a second comment about the scores describing pre-edit text — withdrawn. `lastModifiedBy` and `UtteranceEdit.editedBy` already identify which utterances were rewritten and by whom, so a consumer can filter them; the observation didn't earn a comment.)

kouloumos (approved):
Migration and import path verified; the payload note above is non-blocking.

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/93315c3b944eb5cc4e4c7a8502301bbe234c6cae..d4f910123a95a711309ab351875b8ea89a4cb86e>) to apply both review suggestions:

- The handler rounds each score to 4 significant figures at write time (`round4`). A new test pins the rounding.
- The schema comment now says that the scores describe the text as the ASR transcribed it, and that consumers join on `lastModifiedBy` to spot corrected utterances.

---

PR #646, opened 2026-08-19
perf(reviews): candidate-first admin review queries

### LINKED ISSUE #303: Admin panel - Reviews: Filter "older than 30 days" is broken

### LINKED ISSUE #560: Admin reviews queries crash with Postgres shared-memory errors (code 53100) as transcript tables grow
## Summary

The `/admin` dashboard intermittently 500s and the production database spikes to ~100% CPU. Root cause is the review-tracking queries in `src/lib/db/reviews.ts`: they derive per-meeting review state on every request by filtering `CouncilMeeting` through nested Prisma `.some` relations into the two largest tables in the DB (`Utterance`, `UtteranceEdit`), which Postgres executes as **parallel hash joins that exhaust dynamic shared memory** (error `53100`). The same query family is behind the already-open #303 ("older than 30 days" filter throws an error). This is a capacity cliff the feature has been walking toward as transcript data grew — not a regression from a specific deploy.

## Impact

- **Affected:** Admin users — the reviews overview widget on `/admin` and the reviews list filters (incl. #303). During episodes, unrelated pages also fail because the heavy parallel queries saturate the shared DB CPU.
- **Since:** Latent since the review feature shipped (Jan 2026, `b4e8ccb8`); first sustained production episode 2026-07-22 09:03–09:55 UTC (~66 failures). #303 shows it was already failing intermittently earlier.
- **Severity:** Degraded/unavailable admin experience + DB CPU saturation. No data loss.
- **Current mitigation:** the overview widget is temporarily disabled on `/admin` (commit `e1835ae8`) so the dashboard renders. `getMeetingsNeedingReview` / the #303 filter remain on the crashing path. **This issue is to choose the permanent fix.**

## Root Cause

Three factors interacting:

1. **Derive-on-read design.** `getReviewStats()` (two `councilMeeting.count()` with nested `NOT { speakerSegments.some.utterances.some.utteranceEdits.some }`) and `getMeetingsNeedingReview()` (nested `.some` status/reviewer filters) recompute review state from raw edit data on every render.
2. **Missing index.** `UtteranceEdit` has only its primary key — nothing on `utteranceId` (the join column) or `editedBy` (the filter). The planner must seq-scan and hash all 452k rows.
3. **Parallel execution on a memory-bounded node.** With no usable index and large inputs, the planner chooses parallel hash joins, each requesting 16→128 MB dynamic-shared-memory segments. On the 4 GB managed node, concurrent heavy queries (each `/admin` visit fires several) exhaust the shared-memory pool → `53100`.

**Why now / not a specific deploy:** the morning's deploy (PR #549) touched these counts, so it was the first suspect — but capturing the actual Prisma-generated SQL for the pre- and post-#549 shapes and running `EXPLAIN` on production showed **identical plans**. The trigger was cumulative data growth shrinking shared-memory headroom until the parallel allocations stopped fitting. No error tracker is wired to `onRequestError` and log retention is hours, so earlier transient failures went unrecorded (#303 is the visible tip).

<details><summary><h2>Evidence</h2></summary>

**Failure signature (prod app logs, 2026-07-22):**
```
ConnectorError { code: "53100", message: "could not resize shared memory segment
"/PostgreSQL.4127208410" to 134217728 bytes: No space left on device" }
Invalid `prisma.councilMeeting.count()` invocation:  (stack -> src/lib/db/reviews.ts)
[onRequestError] render /[locale]/admin (GET /admin)
```
66 occurrences 09:03–09:55 UTC, once per `/admin` visit. Staging: 0.

**Production `EXPLAIN`** of the reviews count — Parallel Hash Semi Joins with seq scans on `Utterance` and `UtteranceEdit`, hash tables built in dynamic shared memory (the failing segments). Old (pre-#549) and new SQL shapes produced identical plans.

**Table & index sizes (prod):**

| Table | Heap | Indexes | Rows | Notes |
|-------|------|---------|------|-------|
| `Utterance` | 258 MB | 234 MB | ~1.1M | `(speakerSegmentId, startTimestamp)` = 122 MB |
| `UtteranceEdit` | 158 MB | 23 MB | 452k (222k `editedBy='user'`) | **PK only — no index on `utteranceId` or `editedBy`** |
| `SpeakerSegment` | 52 MB | 15 MB | ~99k | |

**Cluster:** DO managed Postgres 16, single 4 GB node; `max_parallel_workers_per_gather = 2`. `pg_stat_statements` **not installed** (no retroactive query attribution).

</details>

## Fix Options

Two enablers (1 is emergency-only, 2 is a shared prerequisite), then three real designs (3–5) that are a progression, not independent choices.

### Option 1 — Global `max_parallel_workers_per_gather = 0` *(emergency lever)*
**What:** cluster config change so no query uses parallel workers.
**Trades:** kills the crash class instantly and reversibly, but slows every analytical query and masks the design issue. **Symptom only** — use if we need relief before a real fix ships.

### Option 2 — Add index `UtteranceEdit(utteranceId) WHERE editedBy='user'` *(shared prerequisite)*
**What:** the partial index the review queries have always needed; ~12 MB.
**Trades:** lets the joins be index-driven instead of full seq-scan + hash. Helps 3, 4, and 5. Alone it reduces but doesn't remove the anti-join, so it's a building block, not the fix. **Do this regardless of which design we pick.**

### Option 3 — Rewrite queries (Approach A)
**What:** pick candidate meetings from `CouncilMeeting ⋈ TaskStatus` (small, indexed) first, then one grouped aggregate over just those; derive status/counts in app code.
**Kills the crash by:** removing whole-table anti-joins — the one remaining aggregate is index-driven (with #2).
**Keeps / loses:** identical outputs, no schema change, **no information lost**. Hot path still reads the big tables, but once and indexed.
**Effort / risk:** Medium; needs an `EXPLAIN` check. Fixes #303 directly.
**Relates to:** this aggregate *is* the body reused by Options 4 and 5 — no wasted work.

### Option 4 — Materialized view (Approach C)
**What:** a per-meeting aggregate MV (~606 rows) that all admin queries read; `REFRESH ... CONCURRENTLY` on a schedule, run with `SET LOCAL max_parallel_workers_per_gather = 0` so the refresh itself can't crash.
**Kills the crash by:** taking the big-table scan off the request path entirely — reads hit a tiny indexed view.
**Keeps / loses:** same outputs; gains a staleness window equal to the refresh interval (fine for an admin dashboard). Refresh is one controlled scan on a schedule.
**Effort / risk:** Medium; one SQL definition + a scheduled refresh, no per-write app hooks.
**Relates to:** its `SELECT` is Option 3's aggregate, persisted.

### Option 5 — Denormalized review state (Approach B)
**What:** store `reviewStatus` / edit counts / reviewer / timestamps on `CouncilMeeting` (or a 1:1 table), maintained incrementally on edit + task events, with a reconciliation job and one-time backfill.
**Kills the crash by:** admin queries become plain indexed lookups that never touch the big tables — instant and future-proof at any data size.
**Keeps / loses:** same outputs, always fresh (no staleness window); costs the most code and carries consistency risk (the maintenance hooks are the hard part).
**Effort / risk:** High.
**Relates to:** Option 3's aggregate is its backfill + reconciliation function.

**Combination:** full resolution = **Option 2 + one of {3, 4, 5}**. 3 → 4 → 5 is a ladder of increasing effort *and* increasing durability, and each rung reuses the previous one's aggregate query, so partial progress is never thrown away.

<details><summary><h2>Design notes (deeper reasoning behind the options)</h2></summary>

**Does the rewrite (Option 3) lose any data or capability? No.** It returns the identical `ReviewListItem` / `ReviewStats` shapes with the same aggregates (`totalUtterances`, `reviewedUtterances`, user/task edits, first/last edit, reviewers, primary reviewer, durations). It only reorders *when* the work happens. In fact the efficient aggregate already exists in the codebase (`getAggregatedMeetingStatsBatch`, raw SQL `GROUP BY` keyed by meeting) — it's currently run *after* the expensive nested filter instead of being used *as* the filter.

**So was the original implementation wrong?** The data model is fine and correctly normalized. What aged badly is the *query strategy* — using Prisma nested `.some` / `NOT some` relational filters for status/reviewer filtering (they compile to whole-table anti-joins) plus shipping the feature with no `UtteranceEdit` index. Classic "correct but doesn't scale": it worked when the tables were a fraction of today's size.

**Is the `UtteranceEdit` index big?** No — smaller than intuition suggests. `UtteranceEdit(utteranceId)` ~= 23 MB (same as the PK already on that table); the partial `WHERE editedBy='user'` version ~= 11–12 MB. For comparison the existing `Utterance(speakerSegmentId, startTimestamp)` index is 122 MB and `Utterance_pkey` is 81 MB. It's a small addition that matches the query predicate exactly.

**Caching (for Option 3) and why Option 4 is cleaner.** A short-TTL or tag-invalidated cache on `getReviewStats` would help, but the cache goes stale on every user edit, so it's an extra moving part. With the rewrite + index the query may be fast enough that caching is unnecessary — and if we want stored/fast reads, the materialized view (Option 4) *is* the cache, refreshed on a schedule, with no app-side invalidation logic. So caching is an optional layer for Option 3, made moot by Option 4.

**Materialized-view mechanics (Option 4).** The MV is tiny (~606 rows, one per meeting) and fully indexable. `REFRESH MATERIALIZED VIEW CONCURRENTLY` doesn't block reads (requires a unique index on the MV). Staleness equals the refresh interval — invisible for review data that changes at human pace. The refresh still scans the big tables once, but off the request path and, with `SET LOCAL max_parallel_workers_per_gather = 0` in the refresh transaction, it can't trigger the shared-memory crash — the same lever as Option 1, but scoped to one background job instead of the whole cluster.

**Adjacent hygiene:** install `pg_stat_statements` so the next DB incident has retroactive query attribution (its absence is why this took an `EXPLAIN`-driven investigation).

</details>

<details><summary><h2>Investigation Trail</h2></summary>

1. Observed `53100` shared-memory errors + ~100% DB CPU; Prisma attributed them to `councilMeeting.count()`.
2. Hypothesized recent deploy (PR #549) -> **ruled out**: captured actual Prisma SQL pre/post-#549, `EXPLAIN` plans identical.
3. Hypothesized traffic surge / autovacuum / DO maintenance -> **ruled out**: steady request volume, no vacuum on big tables in-window, no cluster events.
4. Hypothesized connection contention from a shared dev database -> **ruled out** as a cause: the mechanism is the parallel-hash memory demand of the queries themselves, not connection pressure.
5. Followed the stack trace to `src/lib/db/reviews.ts`; `EXPLAIN` revealed Parallel Hash Semi Joins over `Utterance` + `UtteranceEdit` as the shared-memory consumers.
6. Checked indexes -> `UtteranceEdit` has PK only; confirmed the seq-scan/hash driver.
7. Connected #303 — same query family, different entry point — establishing this as latent, not new.
8. Measured table/index sizes to size the fix and compare the proposed index against existing ones.

</details>

## Resolution (2026-08-19)

Chosen: **Option 2 + Option 3**, with no extra guards — implemented in #646.

Verification against a restored production backup (655 meetings, 1.02M utterances, 461k edits) showed:

- With the index in place, no review query plans a parallel hash join, so the `53100` mechanism is gone without Option 1 or chunking. `getReviewStats` runs in ~210 ms.
- **#303's actual failure is not in this shm family.** The unfiltered list dies on result size: its include loads full `TaskStatus` rows (~1.7 GB of request/response payloads across production) and overflows the query engine's marshaling (`Failed to convert rust String into napi string`). #646 fixes it by selecting only `type`/`status`. The assumption above that #303 was "the visible tip" of the shm problem was wrong — two separate defects, one feature.
- The unfiltered list completes but stays ~9 s: the batch aggregate still walks every meeting's utterances once. Acceptable for an admin page; **Option 4 (materialized view) remains the next rung** if we want it fast.
- PR #475's bind-variable cap is real but sits at ~16,000 meetings (~25× today). Documented in #646; chunking not shipped.

kouloumos:
## Why

The admin review feature failed in two unrelated ways as transcript data grew:

1. **Shared-memory crash (#560).** `getReviewStats` filtered `CouncilMeeting` through nested relations into `Utterance` and `UtteranceEdit`. Postgres planned parallel hash joins over both tables. The hash tables exhausted dynamic shared memory on the production node (error `53100`) and saturated the CPU.
2. **Result-size overflow (#303).** The reviews list loaded full `TaskStatus` rows. Their request and response payloads total ~1.7 GB on production. On the unfiltered list the marshaled result exceeds the query engine's string limit: `Failed to convert rust String into napi string`. The page failed before any shared-memory pressure. This root cause was found while verifying this PR against a restored production backup — it is not the mechanism #560 assumed for #303.

## What

- Add the index `UtteranceEdit(utteranceId)` (option 2 of #560; 18 MB on production data). The full index also serves the task-edit aggregate, which the partial variant from the issue cannot.
- Rewrite `getReviewStats` candidate-first (option 3): pick needs-attention meetings from the indexed `TaskStatus` relation, then split them with one indexed `EXISTS` probe. No query touches the full `Utterance`/`UtteranceEdit` tables on this path anymore.
- Drop the nested reviewer prefilter. The primary-reviewer check after aggregation already enforces the filter.
- Select only `type` and `status` from task statuses in the review includes — the only fields the code reads. This fixes #303.
- Re-enable the reviews overview widget on `/admin` (reverts the temporary disable from `e1835ae8`).
- Contain widget failures: a client error boundary plus Suspense around the widget, and an admin-scoped `error.tsx`. A widget failure degrades to one card with a Retry button, and the dashboard streams while the review queries run.
- Scope the stats to past meetings (`dateTime <= now`) and state each count's time window on its tile. The dashboard link opens the list with `last30Days=false` — the same population the widget counts.
- Rebuild the widget on the dashboard's `StatsCard` idiom; it predated that component and was the only element with its own card chrome.
- Fix the `test-backup` skill to sort backups by date across Snapshooter target directories. The old whole-path sort silently fetched a dump frozen at 2026-07-11 from a renamed target. The reproduction below depends on this fix.
- Characterization tests pin the query semantics (status split, completion windows, primary-reviewer filter, show transitions, date bound).

## Measurements

Restored production backup (2026-08-19; 655 meetings, 1.02M utterances, 461k edits) on local PostgreSQL 16:

| Path | main | this PR |
|---|---|---|
| `getReviewStats()` | 53100 crash on prod; 1.9 s hash-join shape locally | **212 ms** |
| Reviews list, unfiltered (#303 repro) | `Failed to convert rust String into napi string` | **8.7 s, 587 items** |
| Reviews list, default view | worked | 2.5 s |

The stats probe uses `JOIN LATERAL ... LIMIT 1`, which cannot be flattened into a hash join, so the 53100 plan shape is structurally unreachable on that path (verified under `enable_nestloop=off`: zero parallel-hash nodes). The batch aggregate keeps the planner's choice; its plans show no parallel hash today. The local node cannot reproduce the shared-memory exhaustion itself (its `/dev/shm` is large); the plan shapes are the evidence. The unfiltered list stays slow because the aggregate still walks every meeting's utterances once — acceptable for an admin page; option 4 of #560 (materialized view) is the next rung if it matters.

## Reproduce

1. Restore the latest production backup with the `test-backup` skill (S3 fetch mode).
2. Point `DATABASE_URL` and `DIRECT_URL` at the restored database — both, and as literal values; Prisma Migrate connects through `DIRECT_URL`, and an unset override falls back to the `.env` database.
3. Run `npx prisma migrate deploy`, then this probe with `npx tsx`:

```ts
import { getReviewStats, getMeetingsNeedingReview } from './src/lib/db/reviews';

const stats = await getReviewStats();
console.log(stats);
const items = await getMeetingsNeedingReview({ show: 'all', last30Days: false });
console.log(items.length);
```

On `main`, the second call throws the napi string error. On this branch, both calls return.

The branch merges current `main`, which added auth guards to the review data layer; the characterization suite signs in a superadmin accordingly.

PR #475 correctly identified a further latent limit on this path: the aggregate's `VALUES` list hits PostgreSQL's 32,767 bind-variable cap near ~16,000 meetings. Production is at ~600, so that cliff is not part of this fix, but the finding is documented here for when scale approaches it.

Closes #560. Closes #303.

<h3>Greptile Summary</h3>

The PR restructures admin review queries to avoid loading large task payloads and expensive whole-table join plans, while restoring the review dashboard widget with localized failure handling.
- Adds an index and candidate-first query for review statistics.
- Narrows TaskStatus selections to fields consumed by review logic.
- Adds characterization coverage for review filtering and count semantics.
- Restores the dashboard widget with Suspense and error fallbacks.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/db/reviews.ts | Reworks review statistics around candidate meeting IDs and narrows TaskStatus projections; inspected consumers and database relationships remain compatible. |
| tests/integration/reviews.queries.test.ts | Adds characterization tests and now passes the Prisma client directly to resetDatabase without the previously reported any cast. |
| src/components/admin/AdminWidgetErrorBoundary.tsx | Adds widget-scoped fallback and retry behavior; no eligible blocking issue is tied to the prior test-cast thread. |
| prisma/schema.prisma | Declares the UtteranceEdit utteranceId index added by the accompanying migration. |
| src/app/[locale]/(admin)/admin/page.tsx | Restores the reviews overview widget under Suspense and a widget-level error boundary. |

<sub>Reviews (3): Last reviewed commit: ["fix(skills): pick the newest backup acro..."](https://github.com/schemalabz/opencouncil/commit/949a8e089e543b4cecdb25b454d3577ac1eb661b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54741882)</sub>

### DIFF
--- a/src/app/[locale]/(admin)/admin/page.tsx
+++ b/src/app/[locale]/(admin)/admin/page.tsx
@@ -1,6 +1,8 @@
 import { Bell, Clock, FileText, Landmark, MessageCircle, Search, Send, Users } from 'lucide-react';
 import { StatsCard } from '@/components/ui/stats-card';
-import { ReviewsOverviewWidget } from '@/components/admin/reviews/ReviewsOverviewWidget';
+import { Suspense } from 'react';
+import { ReviewsOverviewWidget, ReviewsOverviewSkeleton } from '@/components/admin/reviews/ReviewsOverviewWidget';
+import AdminWidgetErrorBoundary from '@/components/admin/AdminWidgetErrorBoundary';
 import { NotificationSubscribersChart } from '@/components/admin/NotificationSubscribersChart';
 import { getAdminDashboardStats, getNotificationSubscribersByCity } from '@/lib/db/adminStats';
 import { withUserAuthorizedToEdit } from '@/lib/auth';
@@ -126,8 +128,15 @@ export default async function Page() {
                 />
             </section>
 
+            {/* Boundary + Suspense: the widget's queries stream in without
+                blocking the dashboard, and a widget failure degrades to one
+                card instead of the whole route (issue #560's blast radius). */}
             <section>
-                <ReviewsOverviewWidget />
+                <AdminWidgetErrorBoundary label="Transcript reviews">
+                    <Suspense fallback={<ReviewsOverviewSkeleton />}>
+                        <ReviewsOverviewWidget />
+                    </Suspense>
+                </AdminWidgetErrorBoundary>
             </section>
         </div>
     );
--- a/src/app/[locale]/(admin)/admin/page.tsx
+++ b/src/app/[locale]/(admin)/admin/page.tsx
@@ -1,3 +1,4 @@
+import { Link } from '@/i18n/routing';
 import { Bell, Clock, FileText, Landmark, MessageCircle, Search, Send, Users } from 'lucide-react';
 import { StatsCard } from '@/components/ui/stats-card';
 import { Suspense } from 'react';
@@ -132,6 +133,17 @@ export default async function Page() {
                 blocking the dashboard, and a widget failure degrades to one
                 card instead of the whole route (issue #560's blast radius). */}
             <section>
+                <div className="flex items-baseline justify-between mb-3">
+                    <h2 className="w-fit">Transcript Reviews</h2>
+                    {/* last30Days=false: the list then shows the same all-time
+                        population the widget counts. */}
+                    <Link
+                        href="/admin/reviews?last30Days=false"
+                        className="text-sm text-muted-foreground hover:text-foreground"
+                    >
+                        View all →
+                    </Link>
+                </div>
                 <AdminWidgetErrorBoundary label="Transcript reviews">
                     <Suspense fallback={<ReviewsOverviewSkeleton />}>
                         <ReviewsOverviewWidget />
--- a/src/app/[locale]/(admin)/admin/page.tsx
+++ b/src/app/[locale]/(admin)/admin/page.tsx
@@ -1,9 +1,6 @@
 import { Bell, Clock, FileText, Landmark, MessageCircle, Search, Send, Users } from 'lucide-react';
 import { StatsCard } from '@/components/ui/stats-card';
-// Temporarily disabled: getReviewStats() crashes the DB with shared-memory
-// errors under data growth (see investigation issue). Re-enable once the
-// reviews queries are fixed.
-// import { ReviewsOverviewWidget } from '@/components/admin/reviews/ReviewsOverviewWidget';
+import { ReviewsOverviewWidget } from '@/components/admin/reviews/ReviewsOverviewWidget';
 import { NotificationSubscribersChart } from '@/components/admin/NotificationSubscribersChart';
 import { getAdminDashboardStats, getNotificationSubscribersByCity } from '@/lib/db/adminStats';
 import { withUserAuthorizedToEdit } from '@/lib/auth';
@@ -129,11 +126,9 @@ export default async function Page() {
                 />
             </section>
 
-            {/* Temporarily disabled — see note on the import above.
             <section>
                 <ReviewsOverviewWidget />
             </section>
-            */}
         </div>
     );
 }
--- a/src/lib/db/reviews.ts
+++ b/src/lib/db/reviews.ts
@@ -566,13 +566,13 @@ function calculateSessionData(
 //
 // STAGE 1: DATABASE FILTERING (Prisma WHERE conditions)
 //   - Reduces dataset before fetching from database
-//   - Filters by task status (transcribe, humanReview)
-//   - Filters meetings where user has made ANY edits (for reviewer filter)
+//   - Filters by task status (transcribe, humanReview) and date
 //   - More efficient: Less data transferred and processed
 //
 // STAGE 2: JAVASCRIPT FILTERING (after fetch)
 //   - Calculates detailed review progress for each meeting
-//   - Verifies user is PRIMARY reviewer (most edits), not just any contributor
+//   - Applies the reviewer filter: verifies user is PRIMARY reviewer
+//     (most edits), not just any contributor
 //   - Excludes meetings that don't meet final criteria
 //
 // FILTER OPTIONS:
@@ -643,28 +643,6 @@ function buildStatusWhereConditions(show: ReviewFilterOptions['show']): Prisma.C
       // Just needs transcribe (includes all statuses)
       return hasTranscribe;
   }
-}
-
-/**
- * Build database where conditions for reviewer filtering
- */
-function buildReviewerWhereConditions(reviewerId: string): Prisma.CouncilMeetingWhereInput {
-  return {
-    speakerSegments: {
-      some: {
-        utterances: {
-          some: {
-            utteranceEdits: {
-              some: {
-                editedBy: 'user',
-                userId: reviewerId
-              }
-            }
-          }
-        }
-      }
-    }
-  };
 }
 
 /**
@@ -963,12 +941,11 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
   // Add date filter
   conditions.push(buildDateFilter(last30Days));
 
-  // Add reviewer filter if specified
-  // Note: This finds meetings where the user has made ANY edits
-  // The "primary reviewer" check happens after fetching
-  if (reviewerId) {
-    conditions.push(buildReviewerWhereConditions(reviewerId));
-  }
+  // The reviewerId filter is applied after aggregation (primary-reviewer
+  // check below). The deleted DB prefilter cost more than the aggregate it
+  // narrowed (measured 5.4s against 2.3s at 300 candidates on
+  // production-scale data): its nested-relation shape is the one issue
+  // #560 removed. Restoring it is a loss at any candidate count.
 
   // Combine all conditions
   const whereConditions = combineWhereConditions(conditions);
--- a/src/lib/db/reviews.ts
+++ b/src/lib/db/reviews.ts
@@ -681,6 +681,42 @@ function combineWhereConditions(
 
 
 /**
+ * Count how many of the given meetings have at least one user transcript edit.
+ *
+ * Candidate-first (issue #560): the caller supplies a small candidate list,
+ * and each candidate is probed through the UtteranceEdit(utteranceId) index.
+ * The LATERAL ... LIMIT 1 probe cannot be flattened into a join, so the
+ * planner cannot fall back to the whole-table parallel hash shape that
+ * exhausts shared memory at production scale — a plain EXISTS keeps that
+ * shape in the plan search space, held off only by a cost margin.
+ */
+async function countMeetingsWithUserEdits(meetings: MeetingId[]): Promise<number> {
+  if (meetings.length === 0) return 0;
+
+  const meetingPairs = Prisma.join(
+    meetings.map((m) => Prisma.sql`(${m.cityId}::text, ${m.meetingId}::text)`)
+  );
+
+  const [row] = await prisma.$queryRaw<[{ count: number }]>`
+    WITH candidates("cityId","meetingId") AS (VALUES ${meetingPairs})
+    SELECT count(*)::int AS "count"
+    FROM candidates c
+    JOIN LATERAL (
+      SELECT 1
+      FROM "SpeakerSegment" ss
+      JOIN "Utterance" u ON u."speakerSegmentId" = ss.id
+      JOIN "UtteranceEdit" ue ON ue."utteranceId" = u.id
+      WHERE ss."cityId" = c."cityId"
+        AND ss."meetingId" = c."meetingId"
+        AND ue."editedBy" = 'user'
+      LIMIT 1
+    ) hit ON true
+  `;
+
+  return row.count;
+}
+
+/**
  * Get aggregated stats for a meeting without loading full nested data
  * Used for efficient list views
  */
@@ -998,57 +1034,30 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
  */
 export async function getReviewStats(): Promise<ReviewStats> {
   await withUserAuthorizedToEdit({});
-  // We can derive needsReview/inProgress from presence of any user edits.
-  const baseNeedsAttentionWhere: Prisma.CouncilMeetingWhereInput = buildStatusWhereConditions('needsAttention');
+  // Candidate-first (issue #560): pick the needs-attention meetings via the
+  // small, indexed TaskStatus relation, then split them by presence of user
+  // edits with an indexed probe per candidate. Expressing that split as
+  // nested relation filters on CouncilMeeting made Postgres anti-join the
+  // full Utterance/UtteranceEdit tables on every /admin visit.
+  //
+  // The counts cover the all-time backlog of past meetings: a meeting only
+  // leaves the set when its humanReview succeeds, and future meetings are
+  // excluded because they cannot need review yet. The widget states the
+  // window and links the list with last30Days=false to show the same set.
+  const candidates = await prisma.councilMeeting.findMany({
+    where: {
+      AND: [
+        { city: CUSTOMER_CITY_WHERE },
+        buildStatusWhereConditions('needsAttention'),
+        buildDateFilter(false),
+      ],
+    },
+    select: { cityId: true, id: true },
+  });
 
-  const [needsReview, inProgress] = await Promise.all([
-    // Needs review = has transcribe, no humanReview, and NO user edits
-    prisma.councilMeeting.count({
-      where: {
-        AND: [
-          { city: CUSTOMER_CITY_WHERE },
-          baseNeedsAttentionWhere,
-          {
-            NOT: {
-              speakerSegments: {
-                some: {
-                  utterances: {
-                    some: {
-                      utteranceEdits: {
-                        some: { editedBy: 'user' },
-                      },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        ],
-      },
-    }),
-    // In progress = has transcribe, no humanReview, and HAS user edits
-    prisma.councilMeeting.count({
-      where: {
-        AND: [
-          { city: CUSTOMER_CITY_WHERE },
-          baseNeedsAttentionWhere,
-          {
-            speakerSegments: {
-              some: {
-                utterances: {
-                  some: {
-                    utteranceEdits: {
-                      some: { editedBy: 'user' },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        ],
-      },
-    }),
-  ]);
+  const meetingIds: MeetingId[] = candidates.map((m) => ({ cityId: m.cityId, meetingId: m.id }));
+  const inProgress = await countMeetingsWithUserEdits(meetingIds);
+  const needsReview = meetingIds.length - inProgress;
 
   // Get completed reviews
   const now = new Date();
--- a/src/lib/db/reviews.ts
+++ b/src/lib/db/reviews.ts
@@ -63,7 +63,10 @@ const includePattern = {
     administrativeBody: { select: { id: true, name: true } },
     taskStatuses: {
       where: whereClause.reviewTaskStatuses(),
-      orderBy: { createdAt: 'desc' as const }
+      // Only the fields hasSucceededTask reads. The full rows carry the
+      // request/response payloads (~1.7 GB across production), which
+      // overflow the engine's result buffer on unfiltered lists (#303).
+      select: { type: true, status: true }
     }
   }),
 
@@ -1051,7 +1054,11 @@ export async function getReviewStats(): Promise<ReviewStats> {
       createdAt: {
         gte: startOfWeek
       }
-    }
+    },
+    // Guard, not an optimization: humanReview payloads are ~30 bytes today,
+    // but a full-row read here puts #303 back on /admin the day the task
+    // grows a real payload.
+    select: { createdAt: true }
   });
 
   const completedToday = completedReviews.filter(
--- a/tests/helpers/factories.ts
+++ b/tests/helpers/factories.ts
@@ -154,6 +154,7 @@ export async function createTaskStatus(meetingId: string, cityId: string, data?:
     status?: string
     requestBody?: string
     responseBody?: string | null
+    createdAt?: Date
 }) {
     return prisma.taskStatus.create({
         data: {
@@ -161,8 +162,26 @@ export async function createTaskStatus(meetingId: string, cityId: string, data?:
             status: data?.status ?? 'pending',
             requestBody: data?.requestBody ?? '{}',
             responseBody: data?.responseBody ?? null,
+            createdAt: data?.createdAt,
             councilMeetingId: meetingId,
             cityId,
+        },
+    })
+}
+
+export async function createUtteranceEdit(utteranceId: string, data?: {
+    editedBy?: 'user' | 'task'
+    userId?: string | null
+    beforeText?: string
+    afterText?: string
+}) {
+    return prisma.utteranceEdit.create({
+        data: {
+            utteranceId,
+            beforeText: data?.beforeText ?? 'before',
+            afterText: data?.afterText ?? 'after',
+            editedBy: data?.editedBy ?? 'user',
+            userId: data?.userId ?? null,
         },
     })
 }
--- a/tests/integration/reviews.queries.test.ts
+++ b/tests/integration/reviews.queries.test.ts
@@ -1,0 +1,183 @@
+/** @jest-environment node */
+import prisma from '@/lib/db/prisma'
+import { getReviewStats, getMeetingsNeedingReview } from '@/lib/db/reviews'
+import { ensureTestDb, resetDatabase } from '../helpers/test-db'
+import {
+    createCity,
+    createMeeting,
+    createSpeakerSegment,
+    createSpeakerTag,
+    createTaskStatus,
+    createUser,
+    createUtterance,
+    createUtteranceEdit,
+    signInAsSuperAdmin,
+} from '../helpers/factories'
+
+/**
+ * Characterization tests for the admin review queries (issues #560, #303).
+ * They pin the query semantics so the candidate-first rewrite cannot
+ * change behavior. Correctness only — the scale problem needs prod-size
+ * data and is verified separately with EXPLAIN.
+ */
+describe('review queries', () => {
+    beforeAll(async () => {
+        await ensureTestDb()
+    })
+
+    beforeEach(async () => {
+        await resetDatabase(prisma)
+        await signInAsSuperAdmin()
+    })
+
+    /** A supported-city meeting with one utterance; returns ids for edits. */
+    async function seedMeeting(params: {
+        cityId: string
+        meetingId: string
+        transcribe?: boolean
+        humanReview?: boolean
+        humanReviewCreatedAt?: Date
+        dateTime?: Date
+    }) {
+        await createMeeting(params.cityId, {
+            id: params.meetingId,
+            dateTime: params.dateTime ?? new Date(Date.now() - 24 * 60 * 60 * 1000),
+        })
+        if (params.transcribe !== false) {
+            await createTaskStatus(params.meetingId, params.cityId, {
+                type: 'transcribe', status: 'succeeded',
+            })
+        }
+        if (params.humanReview) {
+            await createTaskStatus(params.meetingId, params.cityId, {
+                type: 'humanReview', status: 'succeeded',
+                createdAt: params.humanReviewCreatedAt,
+            })
+        }
+        const tag = await createSpeakerTag()
+        const segment = await createSpeakerSegment(params.meetingId, params.cityId, {
+            speakerTagId: tag.id,
+        })
+        const utterance = await createUtterance(segment.id)
+        return { utterance }
+    }
+
+    describe('getReviewStats', () => {
+        it('splits needs-review and in-progress by presence of user edits', async () => {
+            await createCity({ id: 'city1', status: 'supported' })
+            const editor = await createUser('editor@example.com')
+
+            // No user edits -> needsReview
+            await seedMeeting({ cityId: 'city1', meetingId: 'm-untouched' })
+            // Task edits only -> still needsReview
+            const taskOnly = await seedMeeting({ cityId: 'city1', meetingId: 'm-task-only' })
+            await createUtteranceEdit(taskOnly.utterance.id, { editedBy: 'task' })
+            // User edit -> inProgress
+            const started = await seedMeeting({ cityId: 'city1', meetingId: 'm-started' })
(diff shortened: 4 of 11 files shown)


src/app/[locale]/(admin)/admin/page.tsx
```diff
@@ -127,11 +124,9 @@ export default async function Page() {
                 />
             </section>
 
-            {/* Temporarily disabled — see note on the import above.
             <section>
                 <ReviewsOverviewWidget />
```
Comment (christosporios) [inline src/app/[locale]/(admin)/admin/page.tsx]: The widget is awaited inline. There is no `<Suspense>` boundary here, and there is no `error.tsx` under `(admin)/`. The nearest boundary is the public `src/app/error.tsx`.

If `getReviewStats()` throws or hangs, the whole dashboard fails. Admins lose users, notifications, engagement and the subscriber chart. That is the blast radius the temporary disable in `e1835ae8` removed. A `<Suspense>` wrapper plus an admin-scoped `error.tsx` limits the same failure to one card.

The admin layout calls `getCurrentUser()`, so the route renders dynamically. The widget queries therefore run on every visit. They also start only after the `Promise.all` on line 8 resolves. A `<Suspense>` wrapper lets the rest of the dashboard stream while the review probe runs.

---

src/lib/db/reviews.ts
```diff
@@ -679,6 +661,41 @@ function combineWhereConditions(
...
 
+/**
+ * Of the given meetings, which have at least one user transcript edit.
+ * Returns a set of getMeetingMapKey keys.
+ *
+ * Candidate-first (issue #560): the caller supplies a small candidate
+ * list, and each candidate is probed with an indexed EXISTS. Filtering
+ * the same condition through nested Prisma relations compiles to
+ * whole-table parallel hash joins, which exhaust shared memory at
+ * production scale.
```
christosporios: The comment presents the indexed `EXISTS` per candidate as a property of the query. The SQL does not constrain the plan.

I checked this on PostgreSQL 16, with 300 candidates against 2M utterances. Postgres inlines the CTE to a `Values Scan`. Postgres also pulls the `EXISTS` up into a semi-join. Both shapes stay in the search space. With `enable_nestloop=off` the planner emits the #560 shape: `Hash Join -> HashAggregate -> Gather -> Parallel Hash Join -> Parallel Seq Scan`.

Only the cost margin selects the intended plan (3666 against 60016). That margin rests on a bad estimate. The inner index scan on `SpeakerSegment_meetingId_cityId_startTimestamp_idx` estimates `rows=1` against `actual rows=84`, because `meetingId` and `cityId` correlate. The hash plan also ran faster in wall time in my test (424 ms against 822 ms), so the cost model does not track the measured times here.

The nested loop still won at 300, 1000, 2000 and 3000 candidates. The bad plan is latent, not current.

This matters for one sentence in the PR description: "No `Parallel Hash` node appears in any post-index `EXPLAIN`, so the 53100 mechanism is gone." The measurement supports "the planner does not choose that shape today". It does not support "the planner cannot choose that shape". A `LATERAL (SELECT 1 ... LIMIT 1)` fence would pin the shape. An `EXPLAIN` check at a production candidate count would at least record the current margin.

---

src/lib/db/reviews.ts
```diff
@@ -995,57 +1009,26 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
...
  */
 export async function getReviewStats(): Promise<ReviewStats> {
-  // We can derive needsReview/inProgress from presence of any user edits.
-  const baseNeedsAttentionWhere: Prisma.CouncilMeetingWhereInput = buildStatusWhereConditions('needsAttention');
+  // Candidate-first (issue #560): pick the needs-attention meetings via the
+  // small, indexed TaskStatus relation, then split them by presence of user
+  // edits with an indexed EXISTS per candidate. Expressing that split as
+  // nested relation filters on CouncilMeeting made Postgres anti-join the
+  // full Utterance/UtteranceEdit tables on every /admin visit.
+  const candidates = await prisma.councilMeeting.findMany({
```
christosporios, line 1027: This candidate query applies the city condition and the status condition. It does not apply `buildDateFilter`.

`getMeetingsNeedingReview` always applies `buildDateFilter` (line 943). `admin/reviews/page.tsx:57` sets `last30DaysBool = last30Days !== 'false'`, so the default is `true`. The widget counts the all-time backlog. The table that the widget links to shows the last 30 days.

An admin reads "Needs Review 214", opens the list, and sees a much smaller number. The difference is the backlog older than 30 days. Those are exactly the meetings that stay in the set, because a meeting leaves it only when a `humanReview` succeeds. `buildDateFilter` also applies `lte: now`, so the widget counts future-dated meetings that the list can never show.

This is pre-existing. `fa92c53:src/lib/db/reviews.ts` has the same unbounded shape. This PR re-enables the widget, so it puts the difference in front of admins.

---

src/lib/db/reviews.ts
```diff
@@ -62,6 +62,10 @@ const includePattern = {
     administrativeBody: { select: { id: true, name: true } },
     taskStatuses: {
       where: whereClause.reviewTaskStatuses(),
+      // Only the fields hasSucceededTask reads. The full rows carry the
+      // request/response payloads (~1.7 GB across production), which
+      // overflow the engine's result buffer on unfiltered lists (#303).
+      select: { type: true, status: true },
       orderBy: { createdAt: 'desc' as const }
```
christosporios: The new `select` drops `createdAt`. This `orderBy` now sorts on a column that the query does not return.

I generated the Prisma 5.22 SQL: `SELECT id, type, status, councilMeetingId, cityId FROM "TaskStatus" WHERE (...) ORDER BY "createdAt" DESC`. Prisma permits this, and Postgres plans a real `Sort` node for it.

No consumer reads the order. The code reads `taskStatuses` only through `hasSucceededTask` (lines 976, 983, 984, and 1087-1089), which uses `.some()`. Neither `ReviewListItem` nor `ReviewDetail` exposes the field. The cost is small, but the line is dead. Please delete it.

---

src/lib/db/reviews.ts
```diff
@@ -995,57 +1009,26 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
...
-    }),
-  ]);
+  const meetingIds: MeetingId[] = candidates.map((m) => ({ cityId: m.cityId, meetingId: m.id }));
+  const withUserEdits = await getMeetingsWithUserEdits(meetingIds);
+
+  const inProgress = meetingIds.filter((m) => withUserEdits.has(getMeetingMapKey(m))).length;
+  const needsReview = meetingIds.length - inProgress;
 
   // Get completed reviews
   const now = new Date();
```
christosporios, line 1043: The `completedReviews` query just below (line 1040) has no `select`. It returns full `TaskStatus` rows, and it reads only `.length` and `t.createdAt`. That is the pattern line 68 fixes in the same file.

The cost today is small. `markHumanReviewComplete` writes about 30 bytes into `requestBody`, and it leaves `responseBody` null (`src/lib/tasks/humanReview.ts:167-177`). The value is the guard. A later field on that payload puts #303 back on `/admin`, and the commit subject "stop loading task payloads in review lists" reads as complete. `select: { createdAt: true }` closes it.

One correction for the record: this predicate does not force a sequential scan of `TaskStatus`. Postgres drives `TaskStatus_councilMeetingId_cityId_idx` from the customer-city side, at 1x and at 10x scale.

---

src/lib/db/reviews.ts
```diff
@@ -925,12 +942,9 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
...
 
-  // Add reviewer filter if specified
-  // Note: This finds meetings where the user has made ANY edits
-  // The "primary reviewer" check happens after fetching
-  if (reviewerId) {
-    conditions.push(buildReviewerWhereConditions(reviewerId));
-  }
+  // The reviewerId filter is applied after aggregation (primary-reviewer
+  // check below). A DB prefilter would repeat the nested-relation shape
+  // that issue #560 removed, to save one bounded aggregate at most.
```
christosporios: The comment says a DB prefilter saves "one bounded aggregate at most". The saving scaled with the candidate count.

I measured `getAggregatedMeetingStatsBatch` on the restored 2M-utterance database. It costs 48 ms for 5 meetings and 2332 ms for 300 meetings.

Your decision is still correct, and I recommend against re-adding the prefilter. I also measured the deleted Prisma prefilter at 5393 ms on the same data. It plans a sequential scan on `CouncilMeeting` that drives 3000 x 84 segment probes, so it cost more than the aggregate it saved.

Only the stated reason is wrong. Please reword it. The next reader will use this sentence to decide whether the prefilter is worth restoring.

---

src/lib/db/reviews.ts
```diff
@@ -995,57 +1009,26 @@ export async function getMeetingsNeedingReview(filters: ReviewFilterOptions = {}
...
-            },
-          },
-        ],
-      },
-    }),
-  ]);
+  const meetingIds: MeetingId[] = candidates.map((m) => ({ cityId: m.cityId, meetingId: m.id }));
+  const withUserEdits = await getMeetingsWithUserEdits(meetingIds);
+
+  const inProgress = meetingIds.filter((m) => withUserEdits.has(getMeetingMapKey(m))).length;
```
christosporios: `getMeetingsWithUserEdits` returns a `Set` of keys. This line converts that `Set` back into an integer.

`CouncilMeeting` uses `@@id([cityId, id])`, so `candidates` holds no duplicates. Every returned row is a candidate by construction. `inProgress` is therefore identical to `withUserEdits.size`. The filter re-serializes and re-hashes every candidate to recover a value the helper already holds.

`SELECT count(*)::int FROM candidates c WHERE EXISTS (...)` returns one integer. That removes the per-row transfer, the `Set`, the key round-trip and the `meetingIds` array. `needsReview` then reads `candidates.length - inProgress`.

---

tests/helpers/factories.ts
```diff
@@ -142,19 +142,38 @@ export async function createTaskStatus(meetingId: string, cityId: string, data?:
...
     responseBody?: string | null
+    createdAt?: Date
 }) {
     return prisma.taskStatus.create({
         data: {
             type: data?.type ?? 'processAgenda',
             status: data?.status ?? 'pending',
             requestBody: data?.requestBody ?? '{}',
             responseBody: data?.responseBody ?? null,
+            createdAt: data?.createdAt ?? undefined,
```
christosporios: `??` cannot change this value. The option type is `createdAt?: Date` with no `| null`, so the left side is never null.

Line 66 uses the same idiom for `administrativeBodyId`. There the type is `string | null`, and the coalesce does map null to undefined for Prisma. The copy here is dead. `createdAt: data?.createdAt` is enough.

---

christosporios (commented):
Reviewed the 7 commits from `fa92c53`. I checked the query semantics against a local PostgreSQL 16 with a 2M-utterance dataset, and I ran `EXPLAIN ANALYZE` on the new SQL.

**The rewrite is correct.** I found no wrong-answer defect. `needsReview = total - inProgress` reproduces the old `NOT EXISTS` count in every case, including meetings with no speaker segments. The empty-list guard prevents the `VALUES` syntax error. The `select: { type, status }` narrowing breaks no consumer, because the code reads `taskStatuses` only through `hasSucceededTask`. Dropping the reviewer prefilter does not change the result set, because the primary reviewer is a subset of any contributor.

Three findings need a decision from you:

1. **`admin/page.tsx:128`** — the re-enabled widget has no `<Suspense>` and no admin-scoped `error.tsx`. A failure in `getReviewStats` takes down the whole dashboard, which is the blast radius the temporary disable removed.
2. **`reviews.ts:668-672`** — the SQL does not pin the plan. The #560 hash shape stays in the planner's search space, and only a cost margin with a bad row estimate keeps it out. This also qualifies one sentence in the PR description.
3. **`reviews.ts:1017`** — `getReviewStats` applies no date bound, so the widget counts an all-time backlog while the linked table defaults to 30 days. This is pre-existing, but the re-enable makes it visible.

I also left 5 small comments: a dead `orderBy` (line 69), a missing `select` on `completedReviews` (line 1040), an inaccurate cost claim in the prefilter comment (lines 945-947), a `Set` that the caller only counts (line 1030), and a no-op `??` in the test factory.

Two notes on things outside the diff:

**#303 is not fully closed.** `src/app/api/admin/reviews/volume-chart/route.ts:32` still includes `taskStatuses` with no `select`, and it also loads every `speakerSegment.utterances` for 12 weeks of meetings. `transcribe` and `fixTranscript` are the types whose `responseBody` holds the serialized transcript (`src/lib/tasks/tasks.ts:201`). `ReviewVolumeChart.tsx:51` fetches that route from `/admin/reviews`, so the napi string error stays reachable one component away from the list you fixed. A shared `selectPattern.taskSummary` would cover that route and the four sites in `src/lib/db/tasks.ts`.

**The new test never runs in CI.** `nix flake check` runs `npm test -- --testPathIgnorePatterns='tests/integration'`, and `jest.config.js` `testMatch` covers only `src/**`. No workflow calls `npm run test:integration`. `tsc` does not check `$queryRaw` strings, so the new raw SQL has no automated coverage on any push.

Two claims that I want to retract before they spread, because both look plausible and both are wrong:

- The prefilter removal is not a net performance regression. I measured the deleted prefilter at 5393 ms against the 2332 ms aggregate it saved. Details are in the inline comment.
- The `completedReviews` predicate does not force a sequential scan of `TaskStatus`. Postgres uses the existing index.

---

kouloumos:
Applied in 6b6e0dca, c578bd1a and b3b8d997, then merged main in (79448ed0 — the branch picks up the new auth guards; 9cb5cae0 signs the characterization suite in as a superadmin accordingly).

**Finding 2 (plan pinning):** the probe is now `JOIN LATERAL ... LIMIT 1` and returns the count directly, which also folds in your Set→count simplification. A lateral dependent join cannot be flattened, so the hash shape is out of the search space, not just out-costed. Your `enable_nestloop=off` probe on a fresh restored backup now yields zero parallel-hash nodes. Request-path cost: 212 ms against 210 ms unfenced. The PR description's "mechanism is gone" sentence now rests on this structural argument instead of the cost margin — thanks for catching the overclaim.

**Finding 1 (blast radius):** the widget renders inside a client error boundary plus Suspense. A failure degrades to one card with a Retry button (`router.refresh()` then state reset — a bare reset would replay the errored payload). A new admin-scoped `error.tsx` backstops the pages. One scope note: the admin layout's own auth throw still reaches the public error page, because an `error.tsx` renders inside its sibling layout; left that way deliberately.

**Finding 3 (scope):** the stats candidates now exclude future meetings (`dateTime <= now`). The counts stay all-time; each tile states its window; the dashboard link passes `last30Days=false` so the list shows the counted population. The widget was also rebuilt on the dashboard's `StatsCard` idiom (it predates that component) — visible on the preview deployment.

All five small comments are applied: the dead `orderBy` is gone, `completedReviews` selects only `createdAt`, the prefilter comment states your measured rationale, `count(*)` replaces the Set round-trip, and the factory coalesce is deleted.

**Your out-of-diff notes:** the volume-chart route now selects only `type`/`status` — you were right that it kept #303 reachable from the same page. The four `src/lib/db/tasks.ts` sites need per-consumer audits (two must read `requestBody`, one feeds API responses), so the `taskSummary` sweep gets a follow-up issue. The CI note is resolved meanwhile: main's `Integration (testcontainers)` job now runs this suite on every push.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/9cb5cae09ffba3b32a05dd77247a290a874bf228..949a8e089e543b4cecdb25b454d3577ac1eb661b) to atomize the history after the review round: the feedback commits are folded into the logical commits they amend, and the branch is now linear on current main (replacing merge 79448ed0). History-only — the compare shows no file changes.

- bd74f4ff perf(db): index UtteranceEdit(utteranceId)
- aee66576 test(reviews): pin admin review query semantics
- 05ca6398 perf(reviews): derive review stats candidate-first
- c4531d22 perf(reviews): drop the nested reviewer prefilter
- 0cf88553 fix(reviews): stop loading task payloads in review surfaces
- 5829aa58 fix(admin): re-enable the reviews overview widget
- 901dfebe fix(admin): contain reviews-widget failures to one card
- 5d64af04 style(admin): align the reviews widget with the dashboard
- 949a8e08 fix(skills): pick the newest backup across snapshooter targets

tests/integration/reviews.queries.test.ts
```diff
@@ -0,0 +1,181 @@
...
+ * change behavior. Correctness only — the scale problem needs prod-size
+ * data and is verified separately with EXPLAIN.
+ */
+describe('review queries', () => {
+    beforeAll(async () => {
+        await ensureTestDb()
+    })
+
+    beforeEach(async () => {
+        await resetDatabase(prisma as any)
```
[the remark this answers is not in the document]
kouloumos: Removed in aee66576. The helper's structural parameter already accepts the client; the cast was never needed.

src/app/[locale]/(admin)/admin/page.tsx
Comment (kouloumos) [reply src/app/[locale]/(admin)/admin/page.tsx]: Done in 901dfebe: client boundary + Suspense (the dashboard streams while the review queries run), a Retry that calls `router.refresh()` before it resets, and an admin-scoped `error.tsx`. The layout's own auth throw still reaches the public page — left deliberately, per your caveat.

src/lib/db/reviews.ts
kouloumos: Pinned in 05ca6398: the probe is `JOIN LATERAL ... LIMIT 1`, which cannot flatten into a hash join — your `enable_nestloop=off` probe now yields zero parallel-hash nodes. 212 ms against 210 ms unfenced on a restored backup. The PR description is reworded on the structural basis.

src/lib/db/reviews.ts
kouloumos: Done in 05ca6398 and 5d64af04: candidates bound to `dateTime <= now`; the counts stay all-time, each tile states its window, and the dashboard link passes `last30Days=false` so the list shows the counted population.

src/lib/db/reviews.ts
kouloumos: Deleted in 0cf88553.

src/lib/db/reviews.ts
kouloumos: Added `select: { createdAt: true }` in 0cf88553, with the guard rationale in a comment. Noted your seq-scan correction for the record.

src/lib/db/reviews.ts
kouloumos: Reworded in c4531d22 with your measured numbers: the prefilter cost 5.4 s against the 2.3 s aggregate it narrowed.

src/lib/db/reviews.ts
kouloumos: Folded into the fenced probe in 05ca6398: `count(*)` over the lateral join; the Set and the key round-trip are gone.

tests/helpers/factories.ts
kouloumos: Deleted in aee66576.

PR #643, opened 2026-08-18
Redesign of the `/search` page

pkakelas:
# Redesign of the `/search` page 

Third PR solving https://github.com/schemalabz/opencouncil/issues/625.
Depends on https://github.com/schemalabz/opencouncil/pull/637.

Screenshot 2026-08-18 at 9 25 40 PM

- Replace the sidebar filters with a pill-based filter bar under the search box. Add filters for administrative body, topic, and date range.
- On mobile, a filter icon opens the same filters in a full-screen overlay.
- Show results as full-width rows instead of a grid of cards, with a larger page size. Speakers appear as a stack of small avatars.
- The landing map search and the /search query box now share the same search field component.
- Remove the aurora background from the search layout.

<h3>Greptile Summary</h3>

The PR redesigns `/search` with shared query controls, responsive pill and overlay filters, expanded search criteria, and full-width result rows.
- Adds city-scoped administrative-body, party, person, topic, and date filters.
- Wires the new filter fields through URL state and Elasticsearch query construction.
- Reuses the search input across the landing and search experiences.
- Updates result presentation, speaker avatars, localization, and responsive scroll behavior.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| src/components/search/SearchPage.tsx | Coordinates URL-backed filter state, search requests, responsive controls, and paginated result rendering. |
| src/components/search/hooks/useSearchFilterData.ts | Loads city-scoped filter options and prevents superseded asynchronous requests from committing stale data. |
| src/lib/search/query.ts | Extends Elasticsearch query generation for administrative-body, topic, and date filtering. |
| src/components/search/SearchFilters.tsx | Implements the desktop pill-based filter bar using shared filter data. |
| src/components/search/SearchFilterSections.tsx | Implements the mobile full-screen filter sections using the same filter model. |

<sub>Reviews (6): Last reviewed commit: ["fix(search): reserve scrollbar gutter on..."](https://github.com/schemalabz/opencouncil/commit/7773b723b6e50bf2171fea7082006b7cc3af2cd7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54465926)</sub>

**Context used:**

- Context used - CLAUDE.md ([source](https://github.com/schemalabz/opencouncil/blob/main/CLAUDE.md))

Issue #546, opened 2026-07-17
Rethink the landing "popular searches" chips — what should they surface?


kouloumos:
### Concept

The landing v2 search box showed a "Δημοφιλείς αναζητήσεις" chip row in its idle/empty state. The chips were meant to reflect real demand — the most-repeated queries from the `SearchQuery` log, falling back to a curated list when history was thin. In practice the feature fed on a table its own UI barely filled, and the design that would have "fixed" that (#535) exposed a deeper problem: because landing searches only resolve to **categories** or **municipalities**, and those get canonicalized to their taxonomy names before storage, the popularity feed can only ever contain topic labels and city names — a closed vocabulary already fully displayed elsewhere on the page. It can't surface the long tail or surprise anyone, which is the whole point of a "popular" module.

We shipped a hotfix (e38deefc, *"stop showing low-quality popular searches"*) to remove the chip row from production. This issue is to decide whether and how to bring the feature back.

### The open question

What job should the empty-state chip row actually do? A few genuinely different directions, each implying a different data source:

- **Onboarding** — teach people what they can type (a category, a municipality, an address). Content should be exemplary and legible, likely hand-curated. Not necessarily "popular" at all.
- **Real demand** — surface organic, possibly surprising queries. Needs a data source richer than canonicalized category/municipality names (e.g. subject-level search), plus quality filtering, and re-opens the abuse/privacy surface the canonicalization was closing.
- **Drive to real content** — shortcuts to high-value destinations that exist now (active cities, well-covered topics/subjects), curated or derived from actual content rather than search logs.
- **Drop it** — the empty state doesn't need a chip row at all.

No decision yet — that's what this issue is for.

### Context (current state after the hotfix)

- **Dead code:** `src/app/api/landing/popular-searches/route.ts` still exists but is no longer called by anything. Decide whether to remove or repurpose it.
- **Retained guard:** the curated `popularSearches` list in `messages/{el,en,fr}/landingV2.json` is now used only for one non-visual guard in `SearchBody.tsx` — it suppresses a bogus "fly to «text»" geocode option when the typed text matches a curated keyword. Worth deciding whether this guard should stay coupled to a keyword list at all, or be reimplemented off `Topic` names directly (via `detectCategoryQuery` / `classifySearchQuery` in `src/lib/landing/landingData.ts`), since it's nearly redundant with the existing category fuzzy-match.
- **PR #535** (the attempted "make the chips real" rebuild) is being closed in favor of this rethink.

PR #657, opened 2026-08-21
feat(search): wire the landing map to search, and make filter extraction correct and visible

christosporios:
Fourth PR against #625, now rebased onto `main` since #643 merged.

## Why

The landing page and `/search` ran two unrelated searches. `/search` queries Elasticsearch; the landing box filtered the subjects already in memory — the last three months, by title and location text. A question the landing could not answer locally fell through to `classifySearchQuery`, which classifies anything it cannot match as an address, and Enter handed it to Google Geocoding. A content question reached a geocoder and never reached the index.

PostHog says the box is not working: ~18 landing searches a week against ~1,750 landing views, 91% of them resolving as addresses. `/search` gets ~50 a week despite the landing linking to it only from a submenu.

Wiring the landing to `search()` means adopting its AI filter extraction — which is the right call on a map, because extraction produces map state. But extraction had to be fixed first: it was invisible, and nobody could decline it.

## Search library

- **`extractFilters` on `SearchConfig`** decides whether the query text is read at all. On by default, so no existing caller changed.
- **`derivedFilters` on the response** reports what the query text supplied that the caller had not, mirroring the merge: a field is reported exactly when the merge took the extraction's value instead of the caller's. The realm default is not reported — that is the scope of the surface, not something the query asked for.
- **The geocode follows the search's own scope.** When a query named a place but no municipality, it resolved that place against every city of the realm — a geometry read plus two Google Places calls each. One search for "πλατεία" paid for the whole realm.
- **Extraction runs on Haiku** with a small output cap. It sits in front of every text search, so its latency is the reader's latency, and the task is small: read a short query against a fixed city list, answer with a JSON object of at most three fields.
- **`isLatest` is gone.** The prompt asked the model for it and the schema validated it, but nothing had read it since `491ea716` removed the handler. Ranking now boosts recent meetings, which serves "τελευταία" better than that handler did — it clamped the search to a single meeting of one municipality, and only when the query named exactly one.
- **Retrieval split from hydration.** Every search hydrated its hits into full subject rows: the introducer with their roles, every contribution with each speaker's roles, the highlights, the decision. A map wants a pin and a card and would discard nearly all of it — and pay for it first. `searchSubjectsInRealm` answers with ids in relevance order; `searchInRealm` is that plus the hydration it always did, unchanged.

## Extraction, made visible

A query naming a municipality narrowed the results and said nothing. On `/search` the filter pills then sat next to results they did not describe: the search had scoped itself to Χανιά and every control still read "all".

A derived city or period now lands on the pill that owns it, marked, and in the URL, so a shared link carries the same narrowing the sender saw. Two rules keep it honest: new query text drops what the last one supplied, so an old answer cannot fence in the next search; and setting or clearing a filter by hand takes it off the mark, so the search stops re-deriving it and a pill the reader removed does not come back.

`/search` also stops searching twice for these queries — it now recognises the search it just ran and skips it.

## The landing map

Free text commits as a **search filter**. The map pins, the sidebar list, the municipality counts and the empty state all describe the search, rather than the default window with the search filtered out of it.

- **`GET /api/map/search`** answers in the map's own shape. Elasticsearch ranks; the ids hydrate through the same finders that back `/api/map/subjects`, so the client's existing transforms take the rows unchanged — and they carry the discussion time a search result never had. Capped at 150: above that the map degrades pins to dots, so more results buy no fidelity while the marker layer pays for them on every viewport change.
- **The list follows relevance** while a search is active. Importance ranks by how long a council discussed something, relevance by how well it answers what was typed; a list ordered by both is ordered by neither.
- **Changing a filter re-runs the search** rather than narrowing what came back — the best results for a category are not the ones that happened to survive from the best results for everything. That re-run does not read the query text again.
- **Derived filters move the landing's own chips**, for the same reason they move the pills on `/search`.
- **The dropdown gained the row that was missing**, above the address one, and Enter gained a fourth arm. The address tier stays, decided by shape rather than by whether the index recognised the words: a house number, or one of the words a Greek address opens with. Everything else is a question.
- **The chip, the counts and the handoff.** The search shows as a removable chip beside the category pills. Under the list: how many matches sit outside the current area, a way to frame them all, and a link handing the query to `/search`. That link needed a `/search` URL, which five call sites each built by hand and none agreed on — `buildSearchHref` is the one answer.

## Instrumentation

#625 names the gap plainly: no event records a click on a result, so every argument about relevance stays an opinion. The commit is now recorded — how long it took, how many results, whether the cap bit, which filters the query text supplied — and so is the click, with where in the results it sat.

## Testing

`npm test` (1,622), `npx tsc --noEmit` and `npm run lint` are clean. `npm run build` was not run — a dev server was using the tree.

`src/lib/search/core.ts` had no tests despite deciding which filters every search runs with. This adds them, covering the two rules that were already load-bearing before this PR — the realm cap, and extraction filling only what the caller left unset — alongside the new behaviour: the model is consulted by default and not at all when the caller declines; an extracted city fills an unset filter, never replaces one the caller set, and is dropped when it names another realm; an extracted empty list falls back to the realm instead of widening past it; `derivedFilters` reports a derived city or period and reports nothing for a filter the caller set; the geocode stays inside the caller's cities. Plus retrieval returning ids in relevance order and dropping hits the database no longer marks released.

**Not verified in a browser.** This session had no database or Elasticsearch, so everything above is covered by unit tests and reading, not by clicking through it. Worth a look on the preview — especially the landing commit flow, the chip, the out-of-view count, and that a period-naming query fires one search rather than two.

One thing to be aware of when reading CI: `klitiki › should handle mixed casing` is an order-dependent flake that fails on roughly one full run in three. It reproduces on `main` without this branch.

<h3>Greptile Summary</h3>

The PR unifies landing-page and full-page search behavior, exposes query-derived filters, and adds an Elasticsearch-backed map search flow.
- Adds map-oriented search retrieval, hydration, relevance ordering, result limits, and analytics.
- Synchronizes derived municipality and date filters with visible controls and shareable URLs.
- Updates landing navigation, result summaries, filter controls, translations, and search coverage messaging.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/search/SearchPage.tsx | Coordinates extraction, URL-derived filters, stale-request suppression, and completed-search deduplication; the previously reported state race and persistent interval mismatch are addressed. |
| src/components/landing/v2/LandingV2.tsx | Replaces local text filtering with committed map search state and integrates derived filters, result framing, URL restoration, and analytics. |
| src/components/landing/v2/hooks/useCommittedSearch.ts | Encapsulates committed-search request lifecycle, retries, stale-response handling, and derived-filter application. |
| src/app/api/map/search/route.ts | Adds realm-scoped Elasticsearch retrieval followed by map-compatible database hydration with relevance order and truncation metadata. |
| src/lib/search/core.ts | Separates ranked ID retrieval from full hydration and reports filters actually derived from query text. |

<sub>Reviews (13): Last reviewed commit: ["feat(landing): offer the full search fro..."](https://github.com/schemalabz/opencouncil/commit/e37d64e76330bdfa70a4a059ef7249ed0b43d8c4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55571843)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/0dff2763cdb90a1c03cec9a6f41ecb577c0bd98c..63b0f69796b610c505d6690d01eb9ef7c286e58d) to rebase onto `main` now that #643 has merged, and to fix the review finding:

- **Rebased onto `main`.** One file conflicted, `SearchPage.tsx`, against `4ee3fc7` from the #643 review. Both changes survive: the `searchRunIdRef` stale-result guards, and the `executedSearchRef` guard that skips a search already run. One ordering detail turned out to be load-bearing — the dedup check now runs **before** the run id is claimed. Claiming the id and then returning early would strip the current run from a request still in flight; that request would finish, find itself stale, discard its own results, and leave nothing to replace them. In `updateSearchParams` the derived-key stripping and the new `replace`/`push` split sit side by side and do not interact.

- **Fixed the diverging date dedup key.** The finding is right, and it defeated the guard it was about. `executedSearchRef` recorded the range the model answered with, but a derived range reaches the URL as a calendar day and comes back out bounded on local day edges, so the two were never equal and every date-deriving query searched twice. It now records what the URL will carry.

  The city key deliberately stays the whole derived list. Only its first entry reaches the URL, so a query naming two municipalities fails to match on purpose and searches again, narrowed to the one the pill shows — the results have to agree with the pill, which is the point of showing it.

Also updated the PR description: it still described only the first half of this work, from before it became a single PR.

`npm test` (1,622), `npx tsc --noEmit` and `npm run lint` are clean. The date path has no automated coverage — `executedSearchRef` lives in a component with no test — so it is worth confirming on the preview: a query naming a period should light the date pill and fire **one** search, not two.

---

christosporios:
Pushed [3a742ca](https://github.com/schemalabz/opencouncil/commit/3a742ca) — the landing search box, reworked from preview feedback. Net −169 lines.

- **The client-side matcher is gone.** The dropdown is now a fixed list of actions — filter by category, filter by municipality, search the subjects, fly to an address — walked with the arrow keys, first highlighted, Enter takes the highlighted one. That replaces the arbitration ladder that used to guess, and lets the reader overrule the guess.
- **The address option is decided by the shape of the text everywhere.** The row used to ask how many loaded subjects the text matched while Enter asked the shape, so the two could disagree about the same query.
- **The petition CTA left the dropdown.** It fired on any unrecognised "δήμος X" and claimed that municipality was not covered yet, including for names that do not exist. The one at the foot of the Δήμοι tab stays.
- **The query stays in the box after committing**, and emptying the box now ends the search as well as the address locate.
- **The chip is opaque.** It was 5% orange over a map, so place labels read through the text. Separately, `CategoryFilterBar` never re-measured when the chip mounted — `leading` was not in the effect's deps and the `ResizeObserver` watches the container, not its content — so the row widened past the container with no scroll arrow and no fade, leaving the trailing pills clipped flat against the edge. That was the "overlap" in the screenshot; both are fixed.
- **The list caption names the query** it is ranked against instead of reading "most-discussed", and "Αναζήτηση στις συζητήσεις" is now "Αναζήτηση στα θέματα".

**The slowdown after closing a search is fixed at the cause.** The matcher needed an all-time corpus, so the first keystroke refetched every subject in the realm — order 10,000 rows each carrying a paragraph of markdown, against a few hundred for the default window. A committed search hid that array behind its own results; clearing the search handed it back to the map, one Mapbox marker per subject with four listeners and a reprojection each per frame. Worse, once such a fetch resolved, typing again cancelled the refetch that would have replaced it and served the array straight from state. Nothing reads that corpus now, so it is not fetched.

Two things I found but deliberately did **not** change, both pre-existing:

- The subject marker layer is scoped to the filtered set rather than the viewport, so pin count tracks the data volume rather than what is on screen. It stops mattering once the array is bounded, but it is the reason a large array is fatal rather than merely slow.
- `getMapSubjects` has no `take` and ships full `description` for every row. Worth a cap independently of this path.

Also removed the orphaned `popular-searches` route along with the curated keyword list, which settles the "retained guard" item in #546. `getPopularSearchQueries` is left in place — #546 is still deciding whether that feature comes back.

`npm test` (1,619), `npx tsc --noEmit` and `npm run lint` are clean. Not verified in a browser — worth checking on the preview that the arrow keys read well, and that search → cancel → pan is quick now.

---

christosporios:
Pushed [e24253d](https://github.com/schemalabz/opencouncil/commit/e24253d) — fixes the filters becoming unreachable once a search is committed.

The panel shows either the actions for the typed text or the filters, and it chose between them by asking whether the box had text. That only worked because committing used to blank the box; keeping the query made the actions always win. So the filter icon was never selecting a mode — it relied on the box being empty to mean "show filters".

It now takes its half from the control that opened it, which is what `MobileSearchOverlay` has always done: the icon opens the filters, focusing the box goes back to the actions. `SearchBody` already accepted `forceFilters` for exactly this; only the desktop side never set it. The icon also toggles its own half rather than the panel — closing on a second click only reads as closing if the first click opened what you are looking at.

I swept the other conditions that key off the query text, since the same "text means the reader is typing" inference was wired in a few places. Both candidates turned out fine, and slightly more correct than before:

- `FewResultsHint` treats non-empty text as an active narrowing — which a committed search is — and its clear action calls `setQuery('')`, which now also clears the search.
- `MobileHeader`'s `searchActive` styles the box as active, which is right when it holds the committed query.

`npm test` (1,619), `npx tsc --noEmit` and `npm run lint` are clean.

One thing still worth a decision, unchanged for now: the query is visible in up to three places at once — the input, the chip in the filter bar, and the caption that names it. That was worth it when the box went blank after committing; it may not be now.

---

christosporios:
Pushed [292a9d5](https://github.com/schemalabz/opencouncil/commit/292a9d5) — both review findings were valid and are fixed.

**Stale marker skips a required search.** Correct, and the worse of the two. The marker records a search that *finished*, and nothing cleared it when another started, so a later pass could match it and skip a request the running one was about to contradict — exactly the sequence described. Starting a search now clears it, which is also all it was ever for: suppressing the single pass the URL write triggers, and nothing further out.

**Date key represents another interval.** Also correct, and it means my earlier fix for this thread's predecessor traded one bug for a worse one. Recording the URL's normalized interval made the keys match, but it suppressed a search that was doing real work: the first request runs with the model's range, the URL carries a calendar day and reads back local day edges, and those are rarely the same instants. Suppressing the follow-up left the results describing one period while the pill and the shareable link claimed another.

So the marker now records **what the search actually ran with**. When the two intervals genuinely differ, the follow-up runs and its results match the URL; when they agree, it is still skipped. A date-deriving query therefore costs two searches — that is the correction, not waste. The derived-city path is unaffected and still dedupes in one.

I traced both against the flows before committing: derived city → one search; derived date → two, then terminates; multi-city derived → two, narrowed to the city the pill shows; nothing derived → one, no URL write.

`npm test` (1,619), `npx tsc --noEmit` and `npm run lint` are clean.

Worth stating plainly: `SearchPage` has no test file, and both of these are sequencing bugs in `performSearch` that only a component harness would catch. I have not built one here, so the coverage gap that let the first fix regress is still open — the checks above do not protect this path.

---

christosporios:
I pushed [`cde67eea`](https://github.com/schemalabz/opencouncil/commit/cde67eea) with fixes from a code review of the branch. Fourteen findings, ranked by what they cost a reader.

**The landing search**

- A shared `?search=` link never showed its results. On first mount the effect that clears the search when the box is empty runs after the URL-restore effect, against the first render's empty query, and cancelled the request the restore had just started. It now skips its first run, the way the URL-writing effect above it already does.
- A one-sided date filter widened the search instead of narrowing it. `/api/map/subjects` honours either bound alone; `/api/map/search` dropped the filter unless both were set, so committing a search under "from 1 Jan 2026" returned every year while the chip said otherwise.
- Any query holding a digit went to the geocoder. `looksLikeAddress` tested `/\d/`, so "προϋπολογισμός 2026" was address-shaped and Enter geocoded it — the failure this PR opens by describing. It now looks for a house number, a whole number of at most three digits, which a year cannot be.
- A failed request replaced the results with the map's own recent subjects under the reader's query text, and ended the search — the re-run effect is gated on it, so no later filter change could retry. The last successful results now stay, a pill says the search failed, and a retry button runs it again.
- A query naming two municipalities left pins on the map from a municipality no chip named, and the settling-re-run key was recorded from the truncated single-city value so nothing reconciled it. `/search` already handles this case the other way round.
- Mobile had the search without the affordances that explain an empty strip. The outside-view count, the fit-bounds button and the `/search` handoff now ride as the last card of the strip, which is the position the desktop footer occupies.

**`/search`**

Clearing the box reset `lastLoggedQueryRef` and `executedSearchRef` but not `extractedQueryRef`, so re-typing the same query in one session skipped extraction and returned an unnarrowed set with no pill.

**The library and its callers**

- The model's `dateRange` reached the Elasticsearch range filter as two unvalidated strings. A range no date filter can parse narrows the results and never reaches a pill; both bounds must now parse.
- The MCP search tool declines extraction. It takes its filters as arguments, and each page derived independently, so two pages of one query could be computed over different city sets.
- `/api/search` returns `derivedFilters` and the OpenAPI schema describes it. A caller could not otherwise see the scope it was given.
- An empty `subjectIds` matched every subject in the realm rather than none.

Also: the map cap reads `SUBJECT_DOT_THRESHOLD` rather than restating 150; `DerivedSearchFilters` is now `Pick<DerivedFilters, ...>` instead of a second declaration; the click position recorded for a result is its rank in the search rather than its index in the viewport-filtered list; and the deleted viewport-helper tests are back, since `subjectInViewport` gained a consumer here.

`npx tsc --noEmit`, `npx eslint` and the 17 affected suites (299 tests) are clean. **Not verified in a browser** — no dev server, database or Elasticsearch in this session, so the two new UI pieces (the failure pill and the mobile results card) are covered by types and reading only.

One overlap to note: I had also cleared `executedSearchRef` in the error path, but 292a9d5f clears it at the start of every search, which covers that case more broadly. I dropped my version.

christosporios:
Rebased onto `main` for the security fixes — [`cde67ee..f4dbca1`](https://github.com/schemalabz/opencouncil/compare/cde67eeaf61de2e5462af66503887d5902c31ba1..f4dbca157a790982d739da01252a8547e253e2d9). All 19 commits replayed with no conflicts; the diff is unchanged.

The branch now sits on `ad82cdd`, picking up `b03a476` (RSC auth bypass), `f3d1804` (seedUser privilege escalation), `e69cde1` (consultation HTML sanitisation) and `ad82cdd` (onboarding location transaction).

Two things I checked rather than assumed:

- **The auth fix does not reach the search path.** It gates `src/lib/db` exports that were exposed as unauthenticated Server Actions — `apiKeys`, `meetings`, `notifications`, `tasks`, `users` and others. Everything the search work imports from that layer is `cities`, `prisma`, `searchQueries` and `subject`, none of which it touched, and it explicitly leaves public reads and the cached read path alone.
- **The toolchain moved more than the code.** `@types/node` went from `^20` to `^24.13.3` with a lockfile change, so I reinstalled before checking. `npx tsc --noEmit` is clean regardless, which was the step most likely to break.

`npm test` (1,631), `tsc` and `npm run lint` are clean.

One note on process: my first push was rejected by `--force-with-lease`, because @kouloumos's `cde67ee` had landed on the branch since my last push. Forcing past that would have destroyed it. I reset to the remote head and rebased *that* instead, so the review-pass commit is preserved and sits last.

---

christosporios:
Correction to the note above: `cde67ee` is authored by Christos Porios, not kouloumos. Apologies for the wrong attribution and the stray mention.

---

christosporios:
Pushed [2bb455c](https://github.com/schemalabz/opencouncil/commit/2bb455c) — the missing footer on `/search`, plus the coverage note.

**Why the footer was missing:** `/search` is the only page in the `(interfaces)` route group, and that group's layout renders a header and a main and nothing else. The footer comes from `(other)`, the city `(other)` layout, or consultations — `(interfaces)` was never one of them, so the page has never had one. It now renders it the same way those layouts do.

**The coverage note** sits directly above it, as asked, and takes the same gate as every other entry point into `/explain` — `hasExplainPage(realm)`, which is `greece` only. That helper's own comment makes the rule explicit: a link pointing at a 404 is worse than no link, so the Greek-realm restriction falls out of the existing mechanism rather than a new check. The link is relative (`/explain#oc-coverage`, an anchor that already exists in `ExplainFeatures`) so it stays locale-aware rather than hard-coding the domain.

The copy is a message key with the `<link>` rich-text idiom the footer already uses, translated into all four locales — the Greek realm still serves an English UI, and the note should follow the reader's language even though the gate is the realm.

`npm test` (1,631), `tsc` and `npm run lint` are clean.

---

christosporios:
Pushed [7094366](https://github.com/schemalabz/opencouncil/commit/7094366) — mobile search pills.

With a committed search the query showed three times over: the header box, the floating chip, and the caption naming the ordering. Mobile now keeps only the caption. Clearing moves to the box the query is already in — tap it, clear it, and the effect watching an empty box ends the search. I checked that path before removing the ✕, since the chip carried mobile's only one.

Desktop keeps its chip: there it sits among the category pills, reading as one active filter beside the others and carrying the same clear button. That left the chip's `floating` variant with no caller, so it is gone, along with the now-unused `onClearSearch` on the mobile layout.

The caption also stops borrowing the flame. A flame means hottest-and-most-recent, which is the map's own ranking; a search orders by relevance. Sharing one glyph made the two orderings indistinguishable at a glance, which is the single thing that caption exists to prevent — it now shows a search glyph in that mode.

`npm test` (1,631), `tsc` and `npm run lint` are clean.

---

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/709436638f88890f22b59a4c6371b2507ca68fe9..c24200670d3b4e65beefa764e31a0a27b9182ed9>) to add the PostHog changes. Two of them are fixups, so they squash into the commits that caused them.

**New commit — `/search` instrumentation**

- `search_performed` no longer rides `skipQueryLog`. That flag stops the database log from storing one query two times, and it was also gating the event. A filter change and a page turn are both real searches that return different results, and neither recorded anything. The filter flags therefore described only the first run of a query. The event now carries the page, the derived filters, and the duration.
- `has_date_filter` also reads `dateTo`. A search with only an end date now counts.
- `search_failed` is new. A failed search and a search with no matches were both silent, so they were the same absence.
- `captureException` moves after the `isCurrentRun` guard. A superseded run no longer reports an error that no reader saw.
- `search_result_opened` records the rank of a clicked result across the whole result set, not the page. `SubjectRow` calls `onOpen` before the new-tab branch, so a row that opens in a new tab counts too.

**Fixups — landing events**

- `results` becomes `results_shown`. The landing count is capped at 150 and counts rows that survive hydration. `/search` reports the true total under `results_count`. The two names collided in one property, so a chart over both surfaces averaged a capped count with an uncapped one.
- `landing_search.kind` becomes `option`. It named what the box decided to do with the text. It now names which row of a fixed list the reader picked. The values are the same, but the question changed.
- `landing_search_failed` is new, for the same reason as on `/search`.
- The subject matcher was the only caller that selected a subject with the source `search`. That arm of the union is gone.

One change of plan: I said I would drop `landing_search` because `landing_search_committed` duplicates it. It does not. `landing_search` is the only search signal that reaches `landing_first_action`, and the pair separates intent from outcome — which rows people pick against what those searches return. Both events stay, and the rename makes the break in meaning explicit.

`npx tsc --noEmit`, `npm run lint` and `npm test` pass. `klitiki › should handle mixed casing` failed once and passed on a re-run. It fails on `main` too, so it is not a regression.

---

christosporios:
Pushed three commits ([`637b80a`](https://github.com/schemalabz/opencouncil/commit/637b80a), [`c1b6a84`](https://github.com/schemalabz/opencouncil/commit/c1b6a84), [`e37d64e`](https://github.com/schemalabz/opencouncil/commit/e37d64e)) that surface `/search` and `/mcp`:

- **Rename**: "OpenCouncil AI" is now "OpenCouncil MCP" in the site footer, the landing menus, the `/mcp` breadcrumb, and the SEO title, in all four locales.
- **New entry points**: the site header gets a robot icon next to the search icon, linking to `/mcp`. The desktop landing rail gets one row for each page, above the "…" menu, in the sign-in link's shape. The rail row shows the short label "MCP" and carries "OpenCouncil MCP" as its accessible name. The mobile drawer gets one top-level row for each page. Both landing surfaces fire a `nav_link` event with `{ target, surface }`.
- **Explainer handoff**: the relevance popover on the landing now ends with the "Αναζήτηση παντού" link. It carries the committed query and filters to `/search`. The three handoff sources now set `surface` on `search_handoff` (`explainer`, `list_footer`, `strip_card`).

`npx tsc --noEmit`, `npm run lint` and `npm test` pass (the known `klitiki` flake failed once; it reproduces on `main`).

---

Issue #547, opened 2026-07-18
Recurring TransformStream render alerts persist after #537 — deployed runtime is still Node 22


kouloumos:
## Summary
Production emits recurring `App Router render` alerts caused by a Node runtime bug (`TypeError: controller[kState].transformAlgorithm is not a function`) that is fixed only in Node ≥ 24.15.0. PR #537 attempted to fix this by moving to Node 24, but it doesn't change the deployed runtime: App Platform builds with the Node buildpack (not the Dockerfile), and the buildpack resolved the widened `engines.node` range back to Node 22. The alerts will keep recurring until the deployed runtime is ≥ 24.15.0.

## Impact
- **Affected**: alert noise only — the crash fires when a client has already disconnected mid-stream (closed tab, LB timeout), so no user-visible failure
- **Since**: recurring for as long as we've been on Node 22.x; unchanged by #537 (verified on the post-merge staging build)
- **Severity**: cosmetic/operational (alert fatigue, masks real errors)

## Background
Production and staging deploy via DO App Platform using the **Node buildpack**, not the repo's Dockerfile — both app specs have `dockerfile_path: null` and `environment_slug: node-js`, and their build logs open with buildpack detection. The Dockerfile only serves the local Docker dev setup.

The buildpack selects the Node version from `engines.node` in `package.json` — but not as "highest version matching the range". When given a *wide* range (one spanning multiple release lines), it deliberately limits its choice to the release line its metadata considers the **Active LTS** — currently a 22.x pick — rather than the newest matching version. This is a stability guard: a range like `>=20 <25` would otherwise silently jump majors on rebuild. The documented opt-out is the `NODEJS_ALLOW_WIDE_RANGE=true` config var, printed in the build log itself.

Note: the two apps currently run **different buildpack generations** — staging uses the newer CNB (`heroku/nodejs v0.342.5`, which printed the Active-LTS message), production the older `heroku/nodejs-engine v1.5.8` shim. Production runs Node **22.23.1**, staging **22.22.2**.

## Root Cause
Three factors compound:
1. **#537's premise was inverted**: it bumped `Dockerfile`/`Dockerfile.dev` to `node:24-alpine`, but App Platform never reads them. The only deployment-relevant change was `engines.node`.
[...]

PR #662, opened 2026-08-22
Align the Nix toolchain with the deployed runtime and automate flake.lock

kouloumos:
## Why

Three problems made the Dependabot backlog impossible to triage.

**The Nix toolchain had drifted from `package.json`.** `a6f55ac0` raised `engines.node`, because production hit a TransformStream race that Node 24.15.0 fixes. It changed `Dockerfile`, `Dockerfile.dev` and `package.json`, but not `flake.nix`. App Platform does not read the Dockerfile, so the only change that reached a deployment was `engines.node`. Meanwhile `nix flake check` kept testing on Node 20.19.1 — a runtime the application declares it does not support. See #547.

**Every Dependabot pull request carried a red check that meant nothing.** The preview deploy routes bots to the reviewer-gated `preview` environment. Nobody approves it, and GitHub marks the request failed after 30 days. The failure says nothing about the dependency.

**`flake.lock` had no owner.** `package-lock.json` was automated and `flake.lock` was not, so the two halves of the toolchain drift apart by construction. The Node gap is the first symptom, not the last.

## What changes

`build(nix): track nixpkgs-unstable, pin Prisma and Postgres separately`

The `nixos-24.11` pin existed only to hold two packages back. Invert it: track `nixpkgs-unstable` for the toolchain, and keep one input, `nixpkgs-pinned`, on the old rev for those packages alone.

`nixpkgs-pinned` supplies:
- `prisma-engines` / `prisma` 5.22.0, which must correspond to `@prisma/client` 5.22.x. Current nixpkgs carries only 6.x/7.x.
- The `postgresql_16` + PostGIS 3.3.5 stack. PostGIS 3.3.5 cannot build against the channel's GEOS 3.14, whose configure no longer installs `geos-config`. Postgres comes from the same rev, because an extension must be built against the server it loads into.

Current nixpkgs removed the `nodePackages` set and dropped top-level `npm`. The Prisma CLI now comes from the pinned rev as top-level `prisma`. `nodePackages.npm` is gone, because Node 24 ships npm itself.

`build(nix): drop notis Node 24 override`

`notis` pinned its own runtime because the main toolchain was on Node 20. The channel move makes that redundant. Both attributes resolve to the same store path, so the output does not change.

`ci(preview): skip bot-authored pull requests`

A lockfile bump has nothing to preview, and `pull_request_target` grants secrets — a dependency bump must not be built with them. The gate uses `pull_request.user.type`, not a list of bot names and not `github.actor`. `github.actor` becomes the maintainer when a maintainer pushes to a bot branch, and bots push branches to this repository, so a same-repo check does not exclude them either.

`build(deps): let Dependabot maintain flake.lock`

Dependabot has supported the Nix ecosystem since April 2026. `nixpkgs-pinned` is ignored, because its purpose is to stay still. Dependabot does no version comparison for Nix inputs; it moves an input to the newest upstream commit, which would silently swap both the Prisma engine and the Postgres stack.

## Effect

| | before | after |
|---|---|---|
| Node in Nix CI and dev shell | 20.19.1 (EOL) | 24.19.0 |
| npm | 10.x | 11.17.0 |
| Prisma CLI / engines | 5.22.0 | 5.22.0 (unchanged) |
| PostGIS / Postgres | 3.3.5 / 16.9 | 3.3.5 / 16.9 (unchanged) |

The npm generation matters beyond version numbers. npm 11 enforces the platform-package lockfile check that npm 10 skips — the check that let a malformed lockfile reach production in #547. Dependabot pull requests are lockfile changes, so the gate that judges them can now detect that class of defect.

## Verification

`nix flake check` passes on the committed tree: lint, types, and 110 suites / 3075 tests.

`nix flake check` builds only 3 of the 14 flake outputs, so every package was built separately. That found the PostGIS failure, which CI would not have caught — the gate would have passed and `nix run .#dev-db` would have failed for every developer. All 11 packages build after the fix.

The Postgres change is byte-identical, not merely equivalent: `mkPostgresCompat` resolves to the same store path on `main` and on this branch.

`opencouncil-prod` builds, and its `start.sh` references `nodejs-24.19.0`, so previews run the application on this toolchain. The built server starts Next.js 16.2.12 on Node 24 and loads every native module.

## After merge

Two things need watching, because neither can be proven before merge.

- The first `nix` Dependabot pull request. If it proposes moving `nixpkgs-pinned`, the `ignore` key needs a different form for this ecosystem, and the fallback is to fetch that nixpkgs inside `flake.nix` so it never enters `flake.lock`.
- The bot skip. It should show as skipped rather than waiting on the next Dependabot pull request.

Relates to #547. That issue asks to stay open until the alerts are confirmed quiet, so this does not close it.

<h3>Greptile Summary</h3>

The PR aligns the Nix development and build toolchain with Node 24 while retaining pinned Prisma and PostgreSQL/PostGIS packages.
- Tracks `nixpkgs-unstable` and preserves compatibility-sensitive packages through `nixpkgs-pinned`.
- Adds weekly Dependabot maintenance for Nix inputs while excluding the compatibility pin.
- Skips secret-bearing preview workflows for bot-authored pull requests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| flake.nix | Moves the main toolchain to unstable nixpkgs, overlays pinned Prisma packages, preserves the compatible PostgreSQL/PostGIS stack, and removes the redundant Notis Node override. |
| flake.lock | Locks the unstable toolchain revision and records the separate compatibility pin. |
| .github/dependabot.yml | Adds weekly Nix input updates while excluding the intentionally fixed `nixpkgs-pinned` input. |
| .github/workflows/preview-deploy.yml | Prevents bot-authored pull requests from entering the secret-bearing preview workflow. |

<sub>Reviews (2): Last reviewed commit: ["build(deps): let Dependabot maintain fla..."](https://github.com/schemalabz/opencouncil/commit/2edc4692b1f135acf1ad91536ce96e10608fd124) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55831709)</sub>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -42,3 +42,23 @@ updates:
         update-types:
           - "minor"
           - "patch"
+
+  # Nix flake inputs (flake.lock). Without this, package-lock.json is automated
+  # and flake.lock is not, so the two halves of the toolchain drift apart — that
+  # is how the Nix build ended up on EOL Node 20 while package.json required
+  # Node 24. Version updates only; Dependabot does not do Nix security updates.
+  - package-ecosystem: "nix"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "nix"
+    ignore:
+      # nixpkgs-pinned is pinned to one rev on purpose. It supplies
+      # prisma-engines 5.22.0 (must match package.json) and the PostGIS 3.3.5
+      # Postgres stack (PostGIS 3.3.5 cannot build against current GEOS).
+      # Dependabot's Nix support does no version comparison — it moves an input
+      # to the newest upstream commit — so left alone it would silently swap
+      # both. Bump it only alongside the matching upgrade. See flake.nix.
+      - dependency-name: "nixpkgs-pinned"
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -29,7 +29,14 @@ jobs:
   # with write access deploy automatically via 'preview-auto', everyone
   # else (forks, dependabot) goes through the reviewer-gated 'preview'.
   gate:
-    if: github.event.action != 'labeled' || github.event.label.name == 'preview-notis'
+    # Skip bot-authored PRs (Dependabot, Renovate). Two reasons: a lockfile
+    # bump has nothing to preview, and pull_request_target grants secrets — a
+    # dependency-bump PR must not be built with them. `user.type` catches every
+    # app-authored PR, unlike a list of bot login names. Bots push branches to
+    # this repo, so a same-repo/fork check does not exclude them.
+    if: >-
+      (github.event.action != 'labeled' || github.event.label.name == 'preview-notis')
+      && github.event.pull_request.user.type != 'Bot'
     name: Determine Deploy Environment
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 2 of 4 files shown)


kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/2d593135fd3c29a6a282f5d4664169f11b06c683..2edc4692b1f135acf1ad91536ce96e10608fd124) to rebase onto main's `notisNote` fix (0bebdc44), so the preview deploy can run.

PR #669, opened 2026-08-22
Keep the Nix Node in step with engines.node, and align @types/node

kouloumos:
## Why

Four places declare a Node version, and nothing reconciles them:

| Where | Value | Who reads it |
|---|---|---|
| `package.json` → `engines.node` | `>=24.15.0 <25` | DO App Platform's Node buildpack — this selects production's runtime |
| `flake.nix` → `pkgs.nodejs` | 24.19.0 | Nix CI, dev shell, preview runtime |
| `Dockerfile` / `Dockerfile.dev` | `node:24-alpine` | local Docker development only |
| `ci.yml` notis job | `node-version: '24'` | that job |

`engines.node` is the one that decides production. The Nix side drifted from it for a month — `a6f55ac0` raised the range without touching `flake.nix`, so CI kept testing on EOL Node 20. Nothing failed, because npm only warns on an engine mismatch and this repository does not set `engine-strict`. See #547.

## What changes

**`build(nix): assert nodejs satisfies engines.node`**

`flake.nix` now reads `engines.node` from `package.json` and throws when the nixpkgs Node does not satisfy it. The check runs inside `forAllSystems`, so it guards the `pkgs` that every devShell, package, check and app receives.

Two failure paths, both verified to fire:
- Node outside the range names both versions and says which to move.
- A range the check cannot parse throws a distinct error naming `assertNodeSatisfiesEngines`, so a change of range shape fails loudly instead of passing silently.

The check understands ranges shaped `>=X <Y`, which is the shape the range has today. A different shape (for example `^24`) throws the parse error rather than being ignored.

**`build(deps): align @types/node with the Node runtime`**

`@types/node` describes the runtime, so its major belongs to `engines.node`. It sat at `^20` while the runtime moved to Node 24, so the types described a runtime we no longer run. Dependabot then offered `^26` in #666, which overshoots the other way: `tsc` would accept Node 26 APIs that do not exist in production.

This sets `^24.13.3` to match `engines.node`, and adds a Dependabot `ignore` for `@types/node` **major** bumps. Dependabot cannot see the coupling — no metadata links a types package to a runtime, and it does not read `engines.node` ([dependabot-core#7426](https://github.com/dependabot/dependabot-core/issues/7426)). It only ever offers the newest major, so no configuration of it produces the right answer. Security updates bypass `ignore` rules.

The lockfile also drops a stale `@posthog/cli/node_modules/prettier` entry that npm had marked `extraneous`. npm removed it while regenerating; it is cruft, not a version change.

## Scope

This does not touch the `Dockerfile` or `setup-node` declarations, which remain unreconciled — a four-way consistency check is a separate change. It adds no `ignore` rule for `prisma` or `@elastic/elasticsearch`: Prisma bumps stay visible deliberately, and the Elasticsearch coupling is to the deployed cluster, which needs a decision this PR should not presume.

## Verification

`nix flake check` passes: lint, types, and the full test suite. `@types/node@^24.13.3` typechecks clean. Both throw paths in the new check were exercised by temporarily setting an unsatisfiable range and an unparseable one.

Supersedes #666.

<h3>Greptile Summary</h3>

The PR establishes `engines.node` as the shared Node-version source for Nix and the Notis CI job, while aligning Node type declarations with the Node 24 runtime.
- Adds Nix checks for parseable, identical workspace engine ranges and compatibility with nixpkgs.
- Makes the Notis CI job read its Node version from `package.json`.
- Updates root and Notis `@types/node` dependencies and lockfile entries to Node 24.
- Prevents Dependabot from independently proposing major `@types/node` upgrades.

The PR appears safe to merge.

No blocking failure remains; the current changes address both previously reported issues.

| Filename | Overview |
|----------|----------|
| flake.nix | Adds an anchored range parser, workspace engine-consistency validation, and a nixpkgs Node compatibility assertion; the previously reported compound-range bypass is closed. |
| services/notis/package.json | Aligns the Notis workspace’s Node declarations with its declared Node 24 runtime, resolving the prior nested Node 20 type selection. |
| package-lock.json | Resolves the root and Notis workspace declarations to `@types/node` 24.13.3 while retaining dependency-specific Node 20 declarations only under `apache-arrow`. |
| .github/workflows/ci.yml | Replaces the duplicated Node major with `node-version-file: package.json` for the Notis job. |
| .github/dependabot.yml | Ignores independent major updates for `@types/node` so its major remains coupled to the runtime declaration. |
| package.json | Updates the root Node type declarations from major 20 to major 24. |

<sub>Reviews (7): Last reviewed commit: ["ci: derive the notis job&#39;s Node from eng..."](https://github.com/schemalabz/opencouncil/commit/c60545f507382e437afb805cfc8935534abc0b99) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55867301)</sub>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,6 +26,14 @@ updates:
           - "patch"
     # Major bumps (e.g. Next.js 14 -> 16) are intentionally left as
     # individual, ungrouped PRs so they can be reviewed and tested on their own.
+    ignore:
+      # @types/node describes the runtime, so its major is set by
+      # engines.node — not by what npm has published. Dependabot cannot see
+      # that coupling and will always offer the newest major, which is wrong
+      # whenever it differs from the Node we run. Move this deliberately,
+      # together with engines.node and the nixpkgs input.
+      - dependency-name: "@types/node"
+        update-types: ["version-update:semver-major"]
 
   # GitHub Actions used in workflows
   - package-ecosystem: "github-actions"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,11 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '24'
+          # Derived from engines.node rather than repeated here: setup-node
+          # reads volta.node, then devEngines.runtime, then engines.node from
+          # the manifest, and this repo declares only the last. One fewer place
+          # for the Node version to drift.
+          node-version-file: package.json
           cache: npm
 
       # Workspace install at the root covers packages/ui and services/notis.
--- a/flake.nix
+++ b/flake.nix
@@ -41,16 +41,108 @@
         prisma = prismaPkgs.nodePackages.prisma;
       };
 
+      # package.json's engines.node is the single source of truth for the Node
+      # major: DO App Platform's Node buildpack selects the production runtime
+      # from it. Nothing reconciled it with the Nix toolchain, so the two drifted
+      # for a month — CI tested on EOL Node 20 while production ran Node 24 (see
+      # issue #547). Fail the build instead of drifting again.
+      # Node version consistency. package.json's engines.node is the source of
+      # truth: DO App Platform's Node buildpack selects production's runtime
+      # from it. Nothing reconciled it with the Nix toolchain, so the two
+      # drifted for a month — CI tested on EOL Node 20 while production ran
+      # Node 24 (see issue #547).
+      #
+      # Workspaces share one lockfile and one runtime, so every manifest that
+      # declares engines.node must declare the SAME range; a workspace cannot
+      # actually run a different Node. Checking only "nixpkgs satisfies each"
+      # would let root ">=24 <25" and a workspace ">=22 <23" both pass while
+      # disagreeing. The workspace list comes from the root's workspaces globs,
+      # so a new workspace is covered without editing this file.
+      nodeConsistency =
+        let
+          inherit (nixpkgs) lib;
+          rootPkg = lib.importJSON ./package.json;
+          dirsOf = glob:
+            let
+              base = lib.removeSuffix "/*" glob;
+              dir = ./. + "/${base}";
+            in
+            if lib.hasSuffix "/*" glob && builtins.pathExists dir then
+              map (n: "${base}/${n}")
+                (builtins.attrNames
+                  (lib.filterAttrs (_: t: t == "directory") (builtins.readDir dir)))
+            else [ glob ];
+          candidates = [ "." ] ++ lib.concatMap dirsOf (rootPkg.workspaces or [ ]);
+          present = lib.filter (d: builtins.pathExists (./. + "/${d}/package.json")) candidates;
+          declaring = lib.filter
+            (d: ((lib.importJSON (./. + "/${d}/package.json")).engines or { }) ? node)
+            present;
+          rangeOf = d: (lib.importJSON (./. + "/${d}/package.json")).engines.node;
+          nameOf = d: if d == "." then "package.json" else "${d}/package.json";
+          ranges = map (d: { manifest = nameOf d; range = rangeOf d; }) declaring;
+          rootRange = rootPkg.engines.node or null;
+          disagreeing = lib.filter (r: r.range != rootRange) ranges;
+          # builtins.match anchors to the whole string, so this accepts only a
+          # bare ">=X <Y". A compound range does not match and falls through to
+          # the parse error, rather than being compared against whichever bound
+          # happened to be captured.
+          parsed = builtins.match " *>=([0-9]+(\\.[0-9]+)*) +<([0-9]+(\\.[0-9]+)*) *" rootRange;
+        in
+        { inherit disagreeing rootRange parsed; };
+
+      assertNodeSatisfiesEngines = pkgs:
+        let
+          inherit (nixpkgs) lib;
+          inherit (nodeConsistency) disagreeing rootRange parsed;
+          have = pkgs.nodejs.version;
+        in
+        if rootRange == null then
+          throw ''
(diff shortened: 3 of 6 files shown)


kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/ec8a889cdf70c97c23761ced017fc49753633db3..6c6e77aac12361fb57efefa95a0eec3473d521d5) with both review findings fixed, folded into their commits:

- **Notis retains Node 20 types** — valid, and caused by this PR. `services/notis` declares the same `engines.node` but pinned `@types/node: ^20`; once the root moved to `^24`, npm could no longer hoist a shared copy and installed a nested `services/notis/node_modules/@types/node@20.19.43`. CI typechecks notis on Node 24, so it would have been checking against the wrong declarations. Both are now `^24.13.3` and the nested copy is gone. Verified `npx tsc --noEmit` in `services/notis` passes.
- **Range parser accepts extra syntax** — valid. `builtins.match` anchors to the whole string, and the `.*` wrappers defeated that. A compound range was greedily matched on its *last* `>=` and `<`, so `">=20 <21 || >=24.15.0 <25"` would have silently compared against the wrong bounds instead of throwing. Now a single anchored expression: compound ranges fall through to the documented parse error in both orderings, verified.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/6c6e77aac12361fb57efefa95a0eec3473d521d5..a8e9d0c397965f231cb5128bfbe4233516e455ee) to make the assertion workspace-aware.

It read only the root `package.json`, but `services/notis` declares its own `engines.node` and CI typechecks it separately — so the same drift could reappear one level down. The check now covers every manifest declaring `engines.node`, discovered from the root's `workspaces` globs rather than a hardcoded list, so a new workspace is covered without touching `flake.nix`.

Errors name the offending manifest. Verified by breaking each in turn: `engines.node (">=26.0.0 <27") in services/notis/package.json` and `... in package.json`.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/a8e9d0c397965f231cb5128bfbe4233516e455ee..3230f3b524283bedbf5bd9309076b89322c76320) to strengthen the check, after looking at what the workspace split actually implies.

`services/notis` has no Dockerfile and no App Platform spec — it is built by `notis-prod` and deployed through pr-previews, so nothing reads its `engines.node` to select a runtime. It shares the root lockfile and the root's Node.

The previous version asserted that each manifest's range was *satisfiable* by the Nix Node. That is too weak: root `">=24.15.0 <25"` and notis `">=24.0.0 <25"` are both satisfied by Node 24.19.0, so they could disagree while passing. Since workspaces cannot actually run different Nodes, the check now requires every manifest that declares `engines.node` to declare the **same** range, then checks the Nix Node against it.

Verified across six cases: agreement passes; a workspace differing by a satisfiable range is caught; by an unsatisfiable one is caught; an unsatisfiable common range is caught; `^24` reports the parse error; and a workspace declaring nothing (`packages/ui`) is ignored.

kouloumos:
Correcting my previous comment: I said `services/notis` has no App Platform spec and is deployed only through pr-previews. That is wrong. Notis **is** deployed on DO — as a second service component inside both `opencouncil-prod` and `opencouncil-stag`.

The conclusion holds, for a better reason than I gave. Both components have `source_dir: "/"` and `environment_slug: node-js`, so the buildpack resolves Node from the **root** `package.json` for both. Production's notis build log:

```
[INFO] Reusing Node v24.19.0
Using npm v11.17.0
Installing node_modules using npm (from package-lock.json)
```

So `services/notis/package.json`'s `engines.node` is never consulted by a deployment — a divergent value there would document a requirement nothing honours. That makes the agreement check more justified, not less.

[Force-pushed](https://github.com/schemalabz/opencouncil/compare/3230f3b524283bedbf5bd9309076b89322c76320..50fec175be01b979ab105b4f33864718bfaa3cdf) with the commit message corrected to state this. The tree is unchanged; only the reasoning was wrong.

kouloumos:
Added c2c87d52: the notis job now uses `node-version-file: package.json` instead of `node-version: '24'`.

On the four-way consistency check we discussed — I dropped it. Regex-parsing YAML and Dockerfiles to guard drift in a local-dev container and one CI job costs more than it returns, and breaks on any legitimate refactor of those files. Deriving the value removes the declaration instead of policing it.

setup-node resolves a manifest by checking `volta.node`, then `devEngines.runtime`, then `engines.node`; this repo declares only the last. Confirmed in the v4 source (`src/util.ts:29`), and confirmed on this run:

```
node-version-file: package.json
Found in cache @ /opt/hostedtoolcache/node/24.19.0/x64
node: v24.19.0
```

Same version as the Nix toolchain and as production's buildpack.

The two Dockerfiles still pin `node:24-alpine`. Docker cannot read package.json at `FROM` time, so deriving them needs a build arg threaded through docker-compose plus a fallback default that reintroduces a literal. They are local-development only, so they keep the literal.

Unrelated but worth noting from the same log: `setup-node@v4` itself runs on Node 20, which GitHub now warns is deprecated — a concrete reason to land #650.

kouloumos:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/c2c87d52d75d58118ad1b52517d8e1a68e370190..c60545f507382e437afb805cfc8935534abc0b99) with two findings from a self-review of the final state:

- `nodeConsistency` exported `ranges`, which nothing consumed. Removed.
- Removing `engines.node` from the root produced `error: attribute 'engines' missing` — the one failure path with no guidance, while every other gives an actionable message. It now says so directly and explains why the field matters.

Verification matrix is now seven cases, all passing: agreement passes; a workspace differing by a satisfiable range is caught; by an unsatisfiable one is caught; an unsatisfiable common range is caught; `^24` reports the parse error; a workspace declaring nothing is ignored; and a root with no `engines.node` reports that directly.

PR #674, opened 2026-08-23
Remove podcast generation and drop its tables

christosporios:
OpenCouncil no longer supports podcasts. This removes the feature end to end and drops its tables. It also removes `splitMediaFile`, which the podcast button was the last thing able to start.

## Why

What remained was a half-dead feature:

- Nothing in the repository called `requestGeneratePodcastSpec`. No UI, route or script could create a `PodcastSpec`. Specs could only arrive from an externally started task, from the seed dump, or from `copy_db.sh`.
- The one podcast UI could only list specs that already existed and start audio splitting for them. It carried hard-coded English strings.
- `handleSplitMediaFileResult` ran a `podcastPart.findUnique` on **every** `splitMediaFile` callback, highlight-only ones included.
- `docs/guides/meeting-lifecycle.md` documented a user-triggered podcast flow that never had an implementation.

## What changes

**Schema and migration**
- Delete the `PodcastSpec`, `PodcastPart` and `PodcastPartAudioUtterance` models, the `PodcastPartType` enum, and the back-relations on `CouncilMeeting` and `Utterance`.
- `20260824120000_drop_podcast_tables` drops the four foreign keys, the three tables and the enum. That is all it does. See **Task rows** below.

**Code**
- Delete `src/lib/tasks/generatePodcastSpec.ts`, `src/lib/db/podcasts.ts` and `src/components/meetings/admin/PodcastSpecs.tsx`.
- Remove the registry entry and the `TASK_CONFIG` key. That also removes the type from the derived `MeetingTaskType` union and from the admin task filter.
- Delete `src/lib/tasks/splitMediaFile.ts` and everything that referenced it, including `SplitMediaFileRequest`, `SplitMediaFileResult` and `MediaType` in `apiTypes.ts`. See **splitMediaFile** below.
- Delete `GeneratePodcastSpecRequest`, `GeneratePodcastSpecResult` and the `PodcastPart` union from `apiTypes.ts`.
- Drop `podcastSpecs` from `MeetingWithAllData`, and delete the `isMeetingWithAllData` guard, which never had a caller.
- Remove the admin section, the two i18n keys in all four locales, and the unused `Podcast` icon import in `PilotPage.tsx`.
- `seed.ts`, the city reset route, `copy_db.sh` and `generate_seed_dump.ts` no longer create, delete, copy or export the tables.

**Docs**: the lifecycle guide, `task-architecture.md`, `database-seeding.md`, `CLAUDE.md`, `scripts/README.md` and the outreach copy in `email_municipality.ts`.

The `scripts/README.md` table count goes from 26 to **27**, not down. Removing the three podcast tables took it to 27, and the same edit added four table names the list had been missing: `Decision`, `SubjectAttendance`, `SubjectVote` and `MeetingAttendance`. The list now matches the `TABLES` array in `copy_db.sh` exactly.

The last commit corrects three pre-existing documentation errors in the sections this PR already edits: the flowchart and sequence diagram drew a Search Sync task that does not exist, the stage list did not match `CORE_PROCESSING_TASKS`, and `CLAUDE.md` named three of the eight registered task handlers.

## splitMediaFile

An earlier revision kept this file on the grounds that highlights share it. That was wrong.

- `requestSplitMediaFileForHighlight` has no callers, on this branch or on `main`. Highlight rendering moved to `generateHighlight`.
- The only two `startTask('splitMediaFile', ...)` sites were both inside `splitMediaFile.ts`, and this PR removes the podcast one.
- `src/app/api/tasks/` holds only `generate-highlight`, so no generic route can start a task by type.

Nothing could start a `splitMediaFile` task after the podcast removal, so keeping the client would leave exactly the surface this PR set out to delete. The task server still implements the task. This drops our unused client, not the capability.

## Task rows

An earlier revision also had the migration delete `TaskStatus` rows of both types, on the grounds that a callback for one would hit `Unsupported task type` and return 500. A production count settled it:

| type | status | rows | newest |
|---|---|---|---|
| `splitMediaFile` | `succeeded` | 15 | 2025-07-23 |
| `generatePodcastSpec` | — | 0 | — |

No row of either type is `pending` or `started`. The 500 needs an in-flight row, so the risk those statements addressed does not exist, and the podcast statement was a no-op in production. Both statements are gone. The migration now only drops the tables and the enum.

The 15 rows that remain are inert: `getMeetingTaskStatus` iterates `Object.keys(TASK_CONFIG)` so an unknown type is ignored, `TaskFilters` builds its options from the same keys so the type is not selectable, and no callback can arrive. That 2025-07-23 date is also independent evidence for the section above — nothing has created a `splitMediaFile` task in 13 months.

## Compatibility

- An older `seed_data.json` keeps its `podcastSpecs` keys. The seeder now ignores them, so old dumps still seed.
- `generateVoiceprint` and `VoicePrint` stay. Voiceprints never depended on podcasts.
- No Elasticsearch change: no PGSync view read these tables.

## Verification

| Check | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npm run lint` | clean |
| `npm test` | 1647 passed, 117/117 suites |
| `npm run build` | passed |
| `tests/integration/task-handlers.test.ts` | passed |
| Full integration suite, serial | matches the clean-tree baseline exactly |
| `grep -ri "podcast\|splitmediafile\|MediaType"` | the four historical migrations, the new drop migration, and the unrelated `getMeetingMediaType` in `utils.ts` |

The branch is rebased onto `main`, and the migration is renamed forward to `20260824120000` so its folder order matches the applied order alongside `20260823214500_notis_perf_indexes`.

The migration was applied to a real PostgreSQL 16 + PostGIS database three ways: a full `migrate deploy` replay on an empty database, a folder-order replay, and a run over populated tables. In the last case it dropped the three tables and the enum, left **every** `TaskStatus` row intact including `pending` rows of both removed types, and left the referenced `Utterance` untouched. `prisma migrate diff` reports no drift between the migrated database and `schema.prisma`.

Note on the integration suite: this environment has no Docker daemon, so testcontainers cannot start. The runs above used the harness's `TEST_DATABASE_URL` path against a local PostgreSQL. Two pre-existing problems showed up there, both reproduced on an unmodified tree and neither related to this change: `server-only` is absent from `node_modules` and from `package-lock.json` while `jest.integration.config.js` has no moduleNameMapper for it, and `notis-queue.test.ts` and `cache.test.ts` fail identically on `main`.

Separately, `utils.test.ts` `klitiki('ΝΙΚΟΣ')` failed twice across many full-suite runs and passed on every other one, on this branch and before the rebase. It is a pure string function unrelated to this change, so this looks like a pre-existing intermittent failure rather than anything introduced here.

> [!WARNING]
> The migration drops three tables and does not roll back. It no longer deletes any row outside them.

<h3>Greptile Summary</h3>

This PR removes the unsupported podcast feature and its obsolete media-splitting client.
- Drops the podcast tables, foreign keys, and enum.
- Removes podcast and `splitMediaFile` task handlers, contracts, triggers, and UI.
- Updates seed, reset, database-copy, localization, and documentation paths.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| prisma/migrations/20260824120000_drop_podcast_tables/migration.sql | Drops the three podcast tables, their foreign keys, and the podcast-part enum without modifying unrelated records. |
| prisma/schema.prisma | Removes the podcast models and their relations from meetings and utterances. |
| src/lib/tasks/registry.ts | Removes handlers for the retired podcast-generation and media-splitting task types. |
| src/lib/tasks/types.ts | Removes the retired task types from the central task configuration and derived task union. |
| src/lib/apiTypes.ts | Removes the request and result contracts used only by podcast generation and media splitting. |
| prisma/seed.ts | Stops importing podcast records while remaining compatible with old seed files that contain unused podcast keys. |
| scripts/copy_db.sh | Removes the dropped podcast tables from the database-copy table list. |

<sub>Reviews (5): Last reviewed commit: ["docs: correct the task pipeline descript..."](https://github.com/schemalabz/opencouncil/commit/0b220aaf359315c69195bdfa22a69505523c491c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56027786)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/f49c5b9516bca30747eacbb136dc4cee26a3cd49..3ce552724f1275b803701a2a4b2134bb96675f11) to apply self-review findings:

- `splitMediaFile.ts`: drop the `getCouncilMeeting` import, which lost its only caller when `requestSplitMediaFileForPodcast` went
- `splitMediaFile.ts`: replace the `findUnique` plus `update` pair with one `updateMany`, and drive the warning from its `count`. That halves the round trips the interactive transaction holds open
- `scripts/README.md`: the count is 27 tables, not 23. The list was also missing `Decision`, `SubjectAttendance`, `SubjectVote` and `MeetingAttendance`, so it now matches the `TABLES` array in `copy_db.sh`
- migration: correct the comment above the `DELETE`. A callback whose row this statement removes gets a 404 from the route, not a 500. The 500 was what a kept row would produce
- `meeting-lifecycle.md`: the Media Processing Layer is video only now. The audio half described the podcast path

Three findings I did not apply:

- **`isMeetingWithAllData` does not check `taskStatuses`.** True, but pre-existing: the guard on `main` checked `subjects`, `speakerSegments`, `highlights` and `podcastSpecs`, never `taskStatuses`. The function also has no callers anywhere in the repository, so the unsound narrowing cannot fire today. Out of scope for this PR.
- **`SplitMediaFileResult.duration` / `startTimestamp` / `endTimestamp` are unread, and `MediaType` can only be `video`.** These types describe the wire contract with the task server at `TASK_API_URL`, not our usage of it. The server still sends those fields and still accepts audio. Narrowing the types would make them an incomplete description of the payload we actually receive.
- **The `DELETE` has no status filter, so it also removes pending rows.** Deliberate. Filtering to terminal rows would leave `pending` rows that can never complete, since the handler is gone. Nothing in the repository can start a `generatePodcastSpec` task, so an in-flight one would have to come from outside this app.

`tsc --noEmit`, `eslint` and `npm test` (1631 passed) are clean on the squashed branch.

---

christosporios:
Pushed 04700c7 to delete the unused `isMeetingWithAllData` guard.

An earlier review round flagged that the guard does not check `taskStatuses`. I reported that as pre-existing. The history confirms it, and says more: the guard has never had a caller. It entered in #539 and no commit since references it outside its own file.

It also does not do what its signature promises. It takes `any`, which `CLAUDE.md` forbids, and it tests for key presence instead of type, so `{ subjects: "hello", speakerSegments: null, highlights: 42 }` satisfies it and narrows to `MeetingWithAllData`. A `is` predicate is an unchecked assertion to the compiler, and this one asserts close to nothing.

`MeetingWithAllData` itself stays. It is the element type of `SeedData.meetings`.

This leaves a real gap open, which the guard was never wired to close: `loadSeedData` in `src/lib/seed-data.ts` returns `JSON.parse(...)` as `SeedData` with no validation, and that file is downloaded from GitHub at runtime when it is absent. `prisma/seed.ts` then works on `any` throughout. The fix is a zod schema at the `loadSeedData` boundary. The repository already has zod and `src/lib/zod-schemas/`. That is separate work and out of scope here.

`tsc --noEmit`, `eslint` and `npm test` (1631 passed, 113/113 suites) are clean.

---

src/lib/tasks/splitMediaFile.ts
```diff
@@ -103,44 +63,21 @@ export async function handleSplitMediaFileResult(taskId: string, response: Split
         throw new Error('Invalid response format: parts should be an array');
     }
 
-    // Update either PodcastParts or Highlights based on which ID exists
+    // Each part id is a highlight id, as set by requestSplitMediaFileForHighlight
```
kouloumos: The description keeps this file because `splitMediaFile` is shared with highlights. That stopped being true in 39fc310b (2025-08-26), which swapped `HighlightModeBar` from `requestSplitMediaFileForHighlight` to `requestGenerateHighlight`. `requestSplitMediaFileForHighlight` has had no callers since — not on this branch, not on main. The only two `startTask('splitMediaFile', ...)` call sites are both in this file, and this PR deletes the other one along with `requestSplitMediaFileForPodcast`. There is no generic task-start route, so after merge nothing can create a `splitMediaFile` task.

That makes the whole file dead, not just the podcast branch of it. Either keep it and say in the description that the handler stays for in-flight rows rather than for highlights, or remove it: this file, the `registry.ts` import and entry, the `TASK_CONFIG` key in `types.ts`, `SplitMediaFileRequest`/`SplitMediaFileResult` in `apiTypes.ts` plus `MediaType` (whose only users they are), and the `splitMediaFile` example in `checkTaskIdempotency.test.ts:240`, which can use `generateHighlight` as line 229 already does. The backend still implements the task, so this drops our client rather than the capability.

If it stays, one note on the handler as rewritten: when `count` is 0 the callback warns and the task still settles as succeeded. The invariant that every part id is a highlight id now lives only in this comment, and `count === 0` is exactly the signal that it was violated — throwing there routes through the existing catch in the callback route and marks the task failed. `handleSplitMediaFileResult` also has no test today; it needs a mocked prisma, not an integration test. If the file goes, neither matters.

prisma/migrations/20260824120000_drop_podcast_tables/migration.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- DropEnum
+DROP TYPE "PodcastPartType";
+
+-- TaskStatus.type is a plain string with no foreign key to the tables above, so
+-- Prisma does not generate this statement and the DROP statements do not need it.
+-- The rows are dead: the generatePodcastSpec handler is gone, so a callback that
+-- kept its row would fail with "Unsupported task type", and the admin task filter
+-- no longer lists the type. After this DELETE such a callback gets a 404 instead.
+DELETE FROM "TaskStatus" WHERE "type" = 'generatePodcastSpec';
```
kouloumos: If you take the removal option, these rows need the same treatment as the podcast ones — a surviving `splitMediaFile` row whose callback arrives after the registry entry is gone gets exactly the 500 this comment describes:

```sql
DELETE FROM "TaskStatus" WHERE "type" = 'splitMediaFile';
```

Worth a count first. Staging has none, of any status.

docs/guides/meeting-lifecycle.md
```diff
@@ -34,15 +34,12 @@ flowchart TD
     H --> I[✅ Meeting Ready<br/>Fully processed & searchable]
```
kouloumos: Node H is a task that does not exist — there is no search-sync type in `TASK_CONFIG` or `taskHandlers`, and the Elasticsearch admin page says the sync runs through pgsync. The sequence diagram spells it out as a queued task at lines 129-132, and line 180 lists it under Post-Processing. Pre-existing, but this diagram is already being edited.

Two more in the same file. Line 180 also lists "Media File Splitting", which the comment on `splitMediaFile.ts` may make untrue. And the stage list at lines 54-60 claims to be derived from the task config, but includes Process Agenda, which is `requiredForPipeline: false`, and omits `transcriptSent`, which is not. The two errors cancel in the count, so it survives a glance.

CLAUDE.md
```diff
@@ -161,7 +161,6 @@ OpenCouncil uses a **decoupled async job processing system**:
 - `transcribe.ts` - Audio transcription
 - `summarize.ts` - AI-generated summaries
```
kouloumos, line 163: With the podcast entry gone this list is 3 of the 8 registered handlers — `processAgenda`, `fixTranscript`, `generateHighlight`, `splitMediaFile` and `pollDecisions` are all missing. (`tasks.ts` is correctly labelled orchestration.) Pre-existing, but the list was already being edited.

kouloumos (commented):
The removal itself is complete. I merged the branch onto main locally and applied the migration to a database carrying main's full history plus podcast rows and task rows of four types — the three tables and the enum go, exactly the `generatePodcastSpec` rows are deleted, and nothing else is touched.

The branch is 131 commits behind, so the migration is back-dated relative to `20260823214500_notis_perf_indexes`, which main already carries. That is fine here: the two both touch `TaskStatus`, but their effects commute — the notis index is partial on `processAgenda`/`summarize` rows, and this deletes `generatePodcastSpec` ones. `migrate deploy` on a populated database, a full folder-order replay, and `migrate dev` against a database behind by exactly this migration all come back clean. Renaming the folder forward when you rebase would still be tidier, since nothing in Prisma reconciles the recorded order with the directory order.

The one thing worth deciding before merge is what happens to `splitMediaFile`. The description keeps it on the grounds that it is shared with highlights, and that was true until the highlight caller moved to `generateHighlight`. Since then the podcast button has been its only producer, and this PR removes that — so the task type survives with nothing able to start one. That decision also settles the outstanding bot comment, because the in-flight case it describes concerns the only kind of `splitMediaFile` task that has existed for the past year.

One smaller thing: the description says the copy_db table list goes to 23, and the commit body repeats it, but the diff writes 27. It went up rather than down because the edit also folds in four table names the list was missing — a good fix, just invisible in the history as written.

src/lib/tasks/splitMediaFile.ts
christosporios: You are right, and I verified every step before acting on it. `requestSplitMediaFileForHighlight` has one occurrence in the repository, its own definition, on this branch and on `main`. The only two `startTask('splitMediaFile', ...)` sites were both in this file. `src/app/api/tasks/` holds only `generate-highlight`, so there is no generic route either.

Taken the removal option, in 29230bb: the file, the `registry.ts` import and entry, the `TASK_CONFIG` key, `SplitMediaFileRequest`/`SplitMediaFileResult` and `MediaType`. `MediaType` had no other user — `getMeetingMediaType` in `utils.ts` is a different function that returns `{ label, icon }` and never referenced the type.

The `checkTaskIdempotency.test.ts` case is deleted rather than re-pointed. It duplicated the `generateVoiceprint` case directly above it, and line 229 already covers `generateHighlight`, so moving it would have left three tests asserting the same thing.

Your note on the handler is moot now that the file is gone, but it was the right call: `count === 0` was the signal the invariant had broken, and warning there let the task settle as succeeded.

I have corrected the description. It now states the real reason and no longer claims highlights share this code.

---

prisma/migrations/20260824120000_drop_podcast_tables/migration.sql
christosporios: Added in 29230bb, with a comment giving the reason.

I cannot run the production count myself — `CLAUDE.md` forbids connecting to production, and this environment has no `.env`. Your staging count of zero is the only figure I have. **Please confirm the production count before merge.** The statement is a no-op if the count is zero, and every row it can match is dead either way, but the number is worth seeing first.

I did test the behaviour: on a database carrying the full migration history I inserted `TaskStatus` rows of five types, including `splitMediaFile` and `generatePodcastSpec` in both `succeeded` and `pending`. The migration reported `DELETE 2` and `DELETE 2`, and left the `transcribe`, `summarize` and `generateHighlight` rows alone.

---

docs/guides/meeting-lifecycle.md
christosporios: All three confirmed and fixed in de8d1fd.

**Node H.** No `searchSync` key exists in `TASK_CONFIG` or `taskHandlers`. Removed the node, joined `G --> I`, and deleted the `Step 5: Search Synchronization` block from the sequence diagram. The Search Layer line now states the real mechanism, so the information is corrected rather than dropped:

> **Search Layer**: Elasticsearch integration for content discovery and search. pgsync replicates the database into the index continuously. There is no search-sync task.

**Line 180.** Now reads `Highlight Generation, Voiceprint Generation, Decision Polling`. Search Sync and Media File Splitting are both gone, and `pollDecisions` was missing.

**The stage list.** Confirmed: `MEETING_STAGES` is `['scheduled', ...CORE_PROCESSING_TASKS, 'ready']`, and `CORE_PROCESSING_TASKS` holds every key with `requiredForPipeline: true` — `transcribe`, `fixTranscript`, `humanReview`, `transcriptSent`, `summarize`. Dropped Process Agenda, added Transcript Sent, and added a note naming the derivation so the next edit has something to check against.

---

CLAUDE.md
christosporios: Fixed in de8d1fd. The list now names every registered handler, plus `registry.ts` itself so the map is findable:

```
- transcribe.ts        - Audio transcription
- fixTranscript.ts     - Automatic transcript correction
- summarize.ts         - AI-generated summaries
- processAgenda.ts     - Subject extraction from the agenda PDF
- generateHighlight.ts - Highlight video rendering
- generateVoiceprint.ts- Speaker voice recognition
- pollDecisions.ts     - Diavgeia decision polling
- registry.ts          - Task type to result handler map
- tasks.ts             - Task orchestration
```

That is all seven remaining handlers after `splitMediaFile` goes, so the list and `taskHandlers` now agree.

---

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/04700c7a3f8d9337d59296410d8e1f06dc272347..de8d1fd35805a44bb0c720c385b70eecf6f88c2d) to address @kouloumos's review. Details are on each thread; the summary:

- **Rebased onto `main`** (was 136 behind, no conflicts) and renamed the migration forward to `20260824120000`, so its folder order matches the applied order beside `20260823214500_notis_perf_indexes`. The rename is a fixup on the commit that creates the migration, so the file carries its final name from the start rather than moving later in the branch.
- **Removed `splitMediaFile` entirely** (29230bb). You were right that it is dead, and I verified each step. The description's claim that highlights share it was wrong and is now corrected.
- **`DELETE FROM "TaskStatus" WHERE "type" = 'splitMediaFile'`** added beside the podcast one.
- **Corrected the three pre-existing doc errors** (de8d1fd): the Search Sync task that does not exist, the stage list that did not match `CORE_PROCESSING_TASKS`, and the `CLAUDE.md` handler list.
- **Fixed the 23 versus 27 discrepancy** you caught. The description and commit body said 23, the diff wrote 27. 27 is right: removing three podcast tables took the list to 27, and the same edit folded in four names it had been missing. The description now explains why the number went up.

Two threads left open for you: the `splitMediaFile` decision, so you can check the removal, and the `TaskStatus` count, which needs a production number I cannot get — `CLAUDE.md` forbids connecting to production. Your staging count of zero is all I have.

Verification on the rebased branch: `tsc --noEmit`, `eslint`, `npm test` (1647 passed, 117/117) and `npm run build` all clean. The integration suite matches the `main` baseline exactly, serial — `notis-queue.test.ts` and `cache.test.ts` fail identically on both.

One thing worth flagging that is not mine: `utils.test.ts` `klitiki('ΝΙΚΟΣ')` failed twice across many full-suite runs and passed every other time, including once before the rebase. `main` passes it consistently in the runs I did. It is a pure string function with no connection to this diff, so it looks like a pre-existing intermittent rather than anything here — but you may want to know it exists.

---

prisma/migrations/20260824120000_drop_podcast_tables/migration.sql
christosporios: Counted, and the answer changed the decision. Production:

| type | status | rows | newest |
|---|---|---|---|
| `splitMediaFile` | `succeeded` | 15 | 2025-07-23 |
| `generatePodcastSpec` | — | 0 | — |

Nothing is `pending` or `started`. That removes the premise I wrote both statements on: `Unsupported task type` needs a callback, a callback needs an in-flight row, and there is none. The podcast statement was also a no-op in production.

So **both `DELETE` statements are gone** as of 0b220aa. The migration drops the tables and the enum and nothing else. The 15 rows stay: `getMeetingTaskStatus` iterates `Object.keys(TASK_CONFIG)` so an unknown type is ignored, `TaskFilters` builds its options from the same keys so the type is not selectable, and no callback can reach them. Deleting them would have destroyed the record of work that really ran, irreversibly, to buy nothing. If they should ever go, that is a one-line query rather than a migration.

The statements were removed by amending the two commits that introduced them, not by a follow-up commit, so no commit in the branch ever carries a `DELETE`. Both commit bodies are reworded to match; they previously asserted the 500.

Your 2025-07-23 figure is also the best evidence for the `splitMediaFile` removal itself: 13 months with nothing able to create one, which lines up with the highlight caller having moved to `generateHighlight`.

Re-verified after the change: the migration replayed over a populated database drops the three tables and the enum and leaves all five seeded `TaskStatus` rows intact, including `pending` rows of both removed types. `prisma migrate diff` reports no drift. `tsc`, `eslint` and `npm test` (1647 passed) are clean.

---

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/de8d1fd35805a44bb0c720c385b70eecf6f88c2d..0b220aaf359315c69195bdfa22a69505523c491c) to remove both `DELETE FROM "TaskStatus"` statements from the migration.

@kouloumos asked for a production count before merge. It came back: 15 `splitMediaFile` rows, every one `succeeded`, newest 2025-07-23, and zero `generatePodcastSpec` rows. Nothing `pending` or `started`.

That invalidates the reason I gave for the statements. `Unsupported task type` needs a callback, a callback needs an in-flight row, and there is none — so the statements only destroyed history, irreversibly, and the podcast one was a no-op in production. The migration now drops the three tables and the enum and nothing else.

I removed them by amending the two commits that introduced them rather than adding a follow-up commit, so no commit on the branch ever carries a `DELETE`. Both commit bodies are reworded, since each previously asserted the 500.

Re-verified after the change:

- The migration replayed over a populated database drops the tables and the enum and leaves **all five** seeded `TaskStatus` rows intact, including `pending` rows of both removed types.
- `prisma migrate deploy` on a clean database applies all 116 migrations, with `20260824120000_drop_podcast_tables` last, after `20260823214500_notis_perf_indexes`.
- `prisma migrate diff` reports no drift against `schema.prisma`.
- `tsc --noEmit`, `eslint` and `npm test` (1647 passed, 117/117) are clean.

The migration thread is resolved. The `splitMediaFile` thread stays open for you to check the removal.

---

PR #675, opened 2026-08-23
notis: launch-week — incremental delivery, message supersede, SMS channel, review fixes

christosporios:
This PR contains the work of the first production days: fixes from live use, two reviewed feature sets, and the fixes from a max-effort code review.

## Reader-facing behavior

**Incremental delivery.** A reply used to reach the reader only after the whole wake finished: 30–60 seconds of silence, then every message at once. Reactive wakes now deliver each `send_message` the moment the model emits it. The shell injects `deps.deliver` for reactive wakes only. Rows are created wakeId-less mid-loop and adopted by the wake row at persistence. After the first delivery attempt, an error finalizes the wake instead of retrying it — a duplicate answer is worse than a truncated one. The batch lane keeps boundary delivery and all rails.

**A newer message supersedes the one being answered.** Two mechanisms cover the two windows. Before the wake starts: the live lane coalesces per subscription — a second inbound appends to the pending row (compare-and-swap; a new partial unique index enforces one pending live row per reader, and the append gives a fresh message a fresh budget). While the wake runs: mid-run absorption — the loop asks for newer messages at each turn start and before each turn's sends. Absorbed messages are consumed from the queue, injected into the model's conversation, and become part of the wake. Held sends get honest tool acks; a held turn commits nothing (no send, no schedule, no profile write, no unsubscribe) and grants a bonus turn; a turn-start update clears a latched unsubscribe. The model — not the shell — judges whether the new message cancels the old request or adds to it. Nothing is lost on failure: absorbed messages are re-enqueued when a wake fails before any delivery.

**SMS is a conversation channel.** When a WhatsApp send fails terminally, the same content goes out once as SMS: templates in their rendered shell, replies as the raw body. One function serves every trigger path (send boundary, incremental closure, sweeper give-up, delivery-status webhook), replay-proof via the unique `fallbackForId`. A railed row's fallback obeys pause and quiet hours; a reply's fallback bypasses them — only the unsubscribed check stands. Inbound SMS is served for enrolled readers: ΣΤΟΠ works over SMS (with an SMS confirmation), other messages wake the agent, and the answer falls back to SMS when WhatsApp cannot carry it. Nothing enrolls over SMS; unknown phones stay with the main app.

**Consent boundary.** Every unsubscribe path (bare ΣΤΟΠ, the agent tool, the poller's phone-gone) suppresses the reader's pending outbound in its transaction. The deliver closure re-reads the subscription at the send instant, so a ΣΤΟΠ that lands mid-wake stops the remaining sends. Queued non-reactive wakes for unsubscribed readers are consumed without a model run. `unsubscribe_user` is honored only on wakes that carry a reader message.

**Send spacing.** Consecutive sends go out at least one second apart (one keyed pacer for all four send paths). Same-second sends were observed arriving inverted on the handset.

**Prompt.** Links always carry the `https://` prefix (WhatsApp does not reliably make bare domains tappable), and a stated interest is read one notch wider than its words.

## Queue correctness

The claim machinery gained: a per-turn heartbeat so long wakes are not reclaimed mid-run (a lost claim aborts the loop); a lane-agnostic 23505 recovery (the live lane's new index could strand a failing row in `running`); fail-forward on the persist step after real deliveries; and deterministic, paced sweeper re-sends. The 24h window check counts WhatsApp inbound only. The SMS path enforces the same rollback and phone-ownership gates as WhatsApp.

## Admin panel

The thread shows in-flight and failed wakes (queue-backed records with state chips), every reader message of a coalesced or absorbing wake, orphaned deliveries, and reader/outgoing bubbles in one chronological stream. The system page's queue table shows live work only, with a 7-day failures footer; the digested-meetings ledger pages by 20. The overview gets a red cell when wakes failed or wait for a retry, «Τι λένε οι χρήστες» shows each reader once, the chart tooltip resolves repeated hour labels correctly, and suppressed sends have their own label, bar segment, and are excluded from the fail rate. A conversation exports as one JSON file (`?traces=0` for the light version) for offline evaluation. The settings roadmap card is gone.

## Infrastructure

Six indexes across both databases (`User.phone` was seq-scanned on every webhook event). CI runs the testcontainer contract suites, serially. Two notis migrations (`railed` flag with backfill; live-lane dedupe plus unique index) apply on deploy.

## Verification

260 unit tests and 102 integration tests green, plus the production build. A max-effort review (10 finder angles, per-candidate adversarial verification, gap sweep) ran against this branch; 15 of its 16 findings are fixed here and the 16th (SMS replies) became the SMS-channel feature. Regression tests cover the supersede flows, both claim-loss timings, the mid-wake ΣΤΟΠ, the retracted unsubscribe, the absorbed-message restore, and the SMS second leg end-to-end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,3 +54,29 @@ jobs:
       - run: npm test -w notis
 
       - run: SKIP_ENV_VALIDATION=1 npm run build -w notis
+
+  integration:
+    name: Integration (testcontainers)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version-file: package.json
+          cache: npm
+
+      - run: npm ci
+
+      # Both generated clients: the contract suites exercise the main views
+      # AND the notis queue against real PostgreSQL containers.
+      - run: npm run prisma:generate
+      - run: npm run prisma:generate -w notis
+
+      # The suites CI could not previously see: view/grant contracts
+      # (notis-views, notis-rollout, notis-exclusion), queue claims
+      # (notis-queue), migration replay, and the notification paths.
+      # Serially: parallel testcontainers flake under contention (observed
+      # locally — 9 spurious failures hot, 0 after warmup); CI runners are
+      # slower still, and a deterministic 2-minute job beats a fast red one.
+      - run: npm run test:integration -- --runInBand
--- a/prisma/migrations/20260823214500_notis_perf_indexes/migration.sql
+++ b/prisma/migrations/20260823214500_notis_perf_indexes/migration.sql
@@ -1,0 +1,14 @@
+-- Perf indexes for the notis integration paths (cross-session review,
+-- finding 13). Replay-safe: IF NOT EXISTS throughout.
+
+-- Every WhatsApp webhook event resolves its sender by phone on BOTH apps'
+-- gates (the main app's notis-served check and notis's own lookup) —
+-- without an index each event seq-scans "User".
+CREATE INDEX IF NOT EXISTS "User_phone_idx" ON "User"("phone");
+
+-- The notis poller's meeting-event feed reads succeeded pipeline tasks by
+-- recency through the notis_meeting_events view; the partial keeps the
+-- per-tick scan off the full TaskStatus history.
+CREATE INDEX IF NOT EXISTS "TaskStatus_notis_events_idx"
+  ON "TaskStatus" ("updatedAt")
+  WHERE type IN ('processAgenda', 'summarize') AND status = 'succeeded';
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -848,6 +848,9 @@ model User {
   serviceApiKeys ServiceApiKey[]
   mcpTokens UserMcpToken[]
   searchQueries SearchQuery[]
+
+  // Webhook gates resolve senders by phone on every WhatsApp event.
+  @@index([phone])
 }
  
 model Account {
--- a/services/notis/prisma/migrations/20260823210000_railed_flag/migration.sql
+++ b/services/notis/prisma/migrations/20260823210000_railed_flag/migration.sql
@@ -1,0 +1,6 @@
+-- The delivery-instant rail flag, stamped per row instead of derived per
+-- call site. Backfill approximates history: every proactive or template
+-- row was railed; reply-continuation freeform rows before this flag are
+-- all delivered already, so their value no longer matters.
+ALTER TABLE "NotisMessage" ADD COLUMN IF NOT EXISTS "railed" BOOLEAN NOT NULL DEFAULT false;
+UPDATE "NotisMessage" SET "railed" = true WHERE "proactive" = true OR "deliveryMode" = 'template';
--- a/services/notis/prisma/migrations/20260823214600_perf_indexes/migration.sql
+++ b/services/notis/prisma/migrations/20260823214600_perf_indexes/migration.sql
@@ -1,0 +1,23 @@
+-- Perf indexes (cross-session review, finding 13). Replay-safe: the
+-- integration harness applies migrations twice.
+
+-- Panel aggregates, the overview series, and cap counting all scan
+-- messages by time across subscriptions.
+CREATE INDEX IF NOT EXISTS "NotisMessage_createdAt_idx" ON "NotisMessage"("createdAt");
+
+-- The sweeper polls for stale pending outbound every minute. The partial
+-- stays tiny (pending rows drain within seconds) while the table grows
+-- with every message ever sent.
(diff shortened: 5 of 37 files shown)


christosporios:
Pushed b337b8de: the panel now surfaces in-flight and failed wakes. The production launch exposed the gap: a wake failed 3 times and then failed terminally, and the panel showed nothing — the thread hid the reader's message (no NotisWake row exists before completion), and the overview stayed green.

- The conversation thread synthesizes records from live queue rows: the reader's bubble renders with a state chip (waiting / running / retrying with error and next attempt / terminally failed).
- The system page queue table shows live work only. Terminal failures move to a footer: "N οριστικές αποτυχίες τις τελευταίες 7 ημέρες · τελευταία πριν X".
- The overview rails strip gains a red cell when a wake failed in the last 7 days or waits for a retry.

christosporios:
Pushed b647d4be: the conversation view gains an «Εξαγωγή JSON» button. It downloads the subscription, all messages, all wakes with full traces, live queue rows, and scheduled wakes as one file, for offline evaluation. The route calls requireAdmin() like every admin route.

christosporios:
Pushed e14bb4f7: incremental delivery for reactive wakes, and the https:// link rule in the prompt. Each send_message now goes out the moment the model emits it (reactive lane only); after the first delivery attempt a wake finalizes on error instead of retrying, so the model never re-runs after the reader received messages. Details and the changed transaction guarantee are in the updated PR description.

christosporios:
Pushed 902f52f2: the overview chart tooltip resolved repeated hour labels to the first matching bucket (a 24h window starts and ends in the same wall-clock hour), so the current hour showed 0 while messages arrived minutes ago. The x axis now uses the unique bucket key; the label is display-only. The SQL bucketing was verified correct against production data.

christosporios:
Pushed b7b93c5b: both unsubscribe paths now suppress the subscription's pending outbound rows inside their transaction, so a sweeper retry can never deliver after a ΣΤΟΠ (the goodbye/confirmation survives — created after the suppression, or exempted by id). This is the consent half of cross-session review finding 1; the full rail-class stamp follows in the findings PR.

christosporios:
Pushed 906ed2e5: review findings 1 (full rail-class stamp, with migration), 2 (SMS replies and receipts served), 5 (unsubscribe only on reader wakes), and the finding-12 panel batch — plus send spacing after an observed same-second message inversion on WhatsApp, and a clock-pin for two SMS tests that failed after 23:00 Athens. 250 tests, tsc, production build green.

christosporios:
Pushed 6acbc86f: finding 13 (six indexes across both databases — User.phone was seq-scanned per webhook event) and finding 14 (CI now runs the testcontainer contract suites, serially). Integration suites verified locally: 100/100 both parallel-after-warmup and serial.

christosporios:
Pushed 33f09670: a newer reader message now supersedes the one being answered. Live-lane coalescing (append to the pending row, new partial unique index + dedupe migration) covers corrections that arrive before the wake starts; mid-run absorption (turn-start and pre-send probes, consume-and-inject) covers corrections that arrive while the model works — the model judges correction vs addition, the shell only guarantees no message is answered twice and none is lost. Thread renders every reader message as its own bubble. 254 unit + 102 integration tests green.

christosporios:
Pushed 5f87a34e: the digested-meetings ledger pages by 20, «Τι λένε οι χρήστες» dedupes to one row per reader, and the settings roadmap card is gone. The wakes feed was already paged.

christosporios:
Pushed 4a8245a9: 15 of the 16 max-effort review findings fixed, grouped by mechanism — consent boundary (send-instant status re-read, held-turn gates for schedule/profile/unsubscribe, latch clearing, poller suppression via a shared helper, pre-model skip for unsubscribed non-reactive wakes), queue failure paths (lane-agnostic 23505 merge, fresh budget on live append, absorbed-event restore on pre-delivery failure, persist fail-forward, per-turn claim heartbeat), SMS (shared main-DB gate applied to the SMS path, WhatsApp-only window check), and panel/ops (orphaned deliveries render in the thread, ordered sweeper re-sends, suppressed bar segment, compact export with ?traces=0). Finding 8 (routing replies to SMS-originated questions over SMS) is deferred as a product decision. 259 unit + 102 integration tests, tsc, production build green; regression tests added for the restore, mid-wake ΣΤΟΠ, held-turn, latch-clear, and both claim-loss timings.

christosporios:
Pushed 9d9216b0: SMS becomes a real conversation channel — a terminal WhatsApp failure sends the same content once over SMS from one shared function on every trigger path (send boundary, incremental closure, sweeper give-up, status webhook); replies go raw, templates keep their shell, and the rails follow the row's class. Plus the polish batch: chronological bubble interleave in the thread, one keyed send pacer for all four send paths (the sweeper had none), send outcomes returned instead of re-read, a single-scan overview query, and the schema / status-set / alert-helper consolidations. The PR description now covers the whole PR. 260 unit + 102 integration tests and the production build are green.

PR #681, opened 2026-08-24
feat(mcp): filter meetings by administrative body

christosporios:
## Why

Notis told a reader that the 3rd Municipal Community of Athens last met on **28 January**. The body had met six more times since — 18/02, 18/03, 22/04, 20/05, 17/06, 14/07 — and had another session **three days after the conversation**.

The data was complete and correct. The MCP had no way to reach it.

`list_meetings` accepted a city and a date range only. `search` carries no body filter, and it ranks subjects by relevance rather than by date. So Notis searched narrow date windows, found no κοινότητα meeting, fell back to `search("3η Δημοτική Κοινότητα Πετράλωνα Γκάζι συνεδρίαση")`, and read the one `3η Δημοτική Κοινότητα` row on page 1 as the most recent session.

I reproduced the query against production. The ranking is stable: that January row scores 44.4 at rank 1, the other nine rows on page 1 are Δημοτικό Συμβούλιο items that only *mention* the 3rd community, and the real 14 July session sits at **rank 18** — outside the default `pageSize: 10`. Its subjects discuss parking in Petralona and never say "3η Δημοτική Κοινότητα".

One call answers the question correctly today: `list_meetings(cityId: "athens", timeFilter: "past", pageSize: 50)` returns the July session at item 16. Notis never made it, because no tool advertised that a body filter was possible.

## What changed

**`list_meetings` takes a body filter.** New `administrativeBodyIds` and `administrativeBodyTypes` parameters. `getCouncilMeetingsForCity` already supported both — the embed widget uses them. The MCP layer dropped them on the way through.

**`get_city` returns the city's administrative bodies.** Id, name, `name_en` and type, so a caller can resolve «3η Δημοτική Κοινότητα» to the id `list_meetings` takes instead of guessing. `get_city` becomes identity-aware for this: a body whose meetings are all drafts is a dead filter option for a caller who cannot see drafts, so only a caller who can see them receives it.

**`search` documents its ranking.** The tool description and the server instructions now say a query ranks subjects by relevance, that the top hit is not the newest match, and that a row missing from page 1 is not proof of absence. Both point at `list_meetings` for "when did this body last meet".

## Testing

- **Integration** (`tests/integration/mcp-meetings.test.ts`, 6 tests, real Postgres): rebuilds the incident's shape — a community that met in January and July, a committee that kept meeting through August — and asserts the filter returns the July session, that types separate κοινότητες from επιτροπές, that ids win over types, that the filter composes with a date window, that a `get_city` body id round-trips into `list_meetings`, and that a draft-only body stays hidden from anonymous callers.
- **Unit** (`src/lib/mcp/__tests__/server.test.ts`, 4 tests): the schema accepts both filters and rejects an unknown body type; the handler forwards both filters and the caller identity to the data layer. Verified these fail when the pass-through is removed.
- Full suite: 1652 unit tests pass, 114 integration tests pass, `npm run lint` and `npm run build` clean.

`next-intl/server` ships ESM that jest does not transform, so any module importing `src/lib/mcp/data.ts` failed to load under the integration config. A mock maps it to its keys.

## Notes for review

- This does not fix everything behind the incident. Notis re-researches from scratch on every wake — `runWake` builds a fresh message array and only delivered texts carry over — so it had already retrieved the 27 August session two minutes earlier, in the previous wake, and could not remember it. That is a `services/notis` concern, not an MCP one, and is left alone here.
- No database migration. The filter uses columns and a code path that already exist.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N2pyLFuMDSdhLZ19U5gg2t)_

<h3>Greptile Summary</h3>

The PR adds administrative-body filtering to MCP meeting listings and exposes usable body identifiers through `get_city`.
- Adds ID- and type-based administrative-body filters with validation and caller-aware visibility.
- Makes date boundaries inclusive for the full specified calendar day.
- Clarifies relevance ranking and directs body-specific queries to `list_meetings`.
- Adds unit and integration coverage for filtering, visibility, validation, and date bounds.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/mcp/server.ts | Adds non-empty administrative-body filter schemas, forwards filters and identity, and updates MCP tool guidance; the previously reported empty-filter issue is fixed. |
| src/lib/mcp/data.ts | Validates body IDs, applies body filters and inclusive UTC date bounds, and returns identity-aware administrative-body metadata. |
| tests/integration/mcp-meetings.test.ts | Covers body filtering, filter precedence, date composition, visibility, and identifier round-tripping without the previously reported unsafe cast. |
| src/lib/mcp/__tests__/server.test.ts | Verifies schema rejection of empty filters and confirms filters and caller identity reach the data layer. |
| src/lib/mcp/instructions.ts | Documents relevance-ranked search behavior and the correct workflow for finding a body's latest meeting. |

<sub>Reviews (2): Last reviewed commit: ["fix(mcp): make the list\_meetings date bo..."](https://github.com/schemalabz/opencouncil/commit/399f4dc7f57945bf4940dc9454d7961d31f97757) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56224989)</sub>

### DIFF
--- a/jest.integration.config.js
+++ b/jest.integration.config.js
@@ -18,6 +18,7 @@ module.exports = {
         '^@/lib/notifications/welcome$': '<rootDir>/tests/mocks/notificationsWelcome.ts',
         '^@/lib/discord$': '<rootDir>/tests/mocks/discord.ts',
         '^@/lib/cities$': '<rootDir>/tests/mocks/cities.ts',
+        '^next-intl/server$': '<rootDir>/tests/mocks/nextIntlServer.ts',
         '^@/(.*)$': '<rootDir>/src/$1',
     },
     transform: {
--- a/src/lib/mcp/__tests__/server.test.ts
+++ b/src/lib/mcp/__tests__/server.test.ts
@@ -6,7 +6,9 @@
 // goes stale every time a tool is added.
 jest.mock('../data', () =>
     new Proxy({ __esModule: true } as Record<string, unknown>, {
-        get: (target, prop: string) => (prop in target ? target[prop] : jest.fn()),
+        // Memoized, not a fresh mock per access: the forwarding tests below
+        // assert on the same function object the tool callback closed over.
+        get: (target, prop: string) => (target[prop] ??= jest.fn()),
     })
 );
 
@@ -15,10 +17,11 @@ jest.mock('../data', () =>
 jest.mock('../../db/prisma', () => ({ __esModule: true, default: {} }));
 
 import { Realm } from '@prisma/client';
-import type { McpServer } from '@modelcontextprotocol/server';
+import type { McpServer, ServerContext } from '@modelcontextprotocol/server';
 import { registerOpenCouncilServer } from '../server';
 import { mcpRealmStore, requestContext } from '../realm-context';
 import type { McpIdentity } from '../auth';
+import * as data from '../data';
 
 const HIGHLIGHT_TOOLS = [
     'create_highlight', 'generate_highlight_video', 'list_highlights',
@@ -32,17 +35,25 @@ const CATEGORIES = ['discovery', 'directory', 'meetings', 'highlights'];
 type RecordedToolConfig = {
     annotations?: { readOnlyHint?: boolean };
     _meta?: { category?: string };
+    inputSchema?: { parse: (value: unknown) => unknown };
 };
+type ToolHandler = (args: never, ctx: ServerContext) => Promise<unknown>;
+
+/** A tool context carrying the identity verifyMcpToken would have attached. */
+const ctxFor = (identity: McpIdentity) =>
+    ({ http: { authInfo: { extra: { identity } } } }) as unknown as ServerContext;
 
 /** What a connection with this identity would see in tools/list & prompts/list. */
 function advertised(identity: McpIdentity, { inRequestScope = true } = {}) {
     const tools: string[] = [];
     const prompts: string[] = [];
     const meta: Record<string, RecordedToolConfig> = {};
+    const handlers: Record<string, ToolHandler> = {};
     const recorder = {
-        registerTool: (name: string, config: RecordedToolConfig) => {
+        registerTool: (name: string, config: RecordedToolConfig, handler: ToolHandler) => {
             tools.push(name);
             meta[name] = config;
+            handlers[name] = handler;
         },
         registerPrompt: (name: string) => { prompts.push(name); },
     } as unknown as McpServer;
@@ -53,7 +64,7 @@ function advertised(identity: McpIdentity, { inRequestScope = true } = {}) {
     } else {
         register();
     }
-    return { tools, prompts, meta };
+    return { tools, prompts, meta, handlers };
 }
 
 describe('highlight tools are advertised only on authenticated connections', () => {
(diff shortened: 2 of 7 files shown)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/5cb606a64d9caace6c318469002f1dfe648792bd..399f4dc7f57945bf4940dc9454d7961d31f97757) to act on a self-review. Two fixups squashed into their targets, plus one new commit.

**The body filter could still answer with a future date.** `list_meetings` applies no `dateTime` constraint unless `timeFilter` or `from`/`to` is set, and it orders `dateTime: desc` — so a scheduled meeting sorts ahead of every held one. The description told a caller to use `administrativeBodyIds` for "when did this body last meet" without naming `timeFilter`. The 3rd κοινότητα of Athens had a session pending three days out, so the promoted flow would have returned 27 August as its last meeting. The description and the server instructions now pair the body filter with `timeFilter`, and a test asserts the bare filter puts the scheduled meeting first while `past` and `upcoming` separate them.

**An empty array disabled the filter.** `getCouncilMeetingsForCity` gates on `.length > 0`, so `administrativeBodyIds: []` fell through to no body filter and returned every meeting in the city. Both arrays now carry `.min(1)`.

**An unknown body id returned an empty list.** A hallucinated id, a stale one, or one from another city produced `{ meetings: [] }` — indistinguishable from a body that never met, which is the inference this PR exists to prevent. `cityId` was validated; body ids were not. They now raise a `NotFoundError` naming `get_city`, mirroring how an unknown city points at `list_cities`.

**The body-type gloss covered municipalities only.** `AuthorityType` also has `region`, where the bodies are Περιφερειακό Συμβούλιο and Περιφερειακή Επιτροπή and κοινότητες do not exist.

**`to` was documented inclusive but behaved exclusive.** This one is pre-existing, and it is its own commit rather than a fixup. `mcpListMeetings` parsed each bound with `new Date(value)`, which reads a bare `YYYY-MM-DD` as UTC midnight, and the data layer applies `lte` to a timestamp. So `to: "2026-07-14"` excluded the 14 July 15:30 session — the exact meeting this branch is about. `to` now binds to the end of the named day.

Smaller items: the draft-visible arm of `get_city` had no test (every test ran anonymous, so `getAdministrativeBodiesForCity` was never called by the suite); `get_city` awaited the realm check, the city, the identity check and the parties in series when the last three are independent; the `next-intl/server` mock omitted three exports `src` imports and added two nothing uses; and `resetDatabase(prisma as any)` did not need the cast, which `CLAUDE.md` forbids.

Integration coverage went from 6 tests to 11. I verified the two new correctness tests fail when their fix is reverted, so they are not vacuous. Full suite: 1653 unit, 119 integration, lint and build clean.

Three findings I did **not** act on, as follow-ups rather than scope for this PR:

- No tool returns an administrative body id except `get_city`. `list_meetings`, `get_meeting` and `get_subject` expose the body as a display name, so a caller landing on a subject cannot pivot to that body's other meetings without matching Greek text.
- `list_meetings` returns no `total` or has-more signal, so an exactly-full page is indistinguishable from the end of the list. `search` returns `total`.
- The guidance now appears in three places (server instructions, `search` description, `list_meetings` description) and nothing detects drift between them. I cut the third copy where I could — `list_meetings` no longer restates how `search` ranks — but the cross-references remain, since MCP clients may show tool descriptions without the instructions block.

---

PR #682, opened 2026-08-24
notis: write about the big story when nothing is theirs, and fix the template 422

christosporios:
Two changes. The second one repairs a live production failure.

## 1. The reader hears about the big story

A resident enrolled with three interests: cleanliness, bar noise and parking near her street. The next Athens council meeting spent 103 minutes on the Metro Line 4 tunnel damage in Κυψέλη. She received nothing. Measured against her real profile: **0 of 4 runs sent a message.**

The agent gave the reason itself:

> «η αναδιοργάνωση Δημοτικής Αστυνομίας είναι **γενικό διοικητικό θέμα χωρίς hyperlocal βαρύτητα**»

That subject carries the topic Δημόσια Ασφάλεια, which is in her enrollment list. The topic matched. The agent rejected it because it is not near her street.

The prompt tells the agent to stay quiet when something "is far from where they live **and** does not affect the whole municipality". The model reads that "and" as an "or". No instruction anywhere says that citywide weight is a reason to write.

### What changed

The prompt now says it, and sets the bar by body. The δημοτικό συμβούλιο gives most items one or two minutes; half an hour or more marks what the city decided that night. A δημοτική επιτροπή and a δημοτική κοινότητα decide narrower things, so their bar does not move.

The brief already prints `discussion: Nmin` per subject. The council item is 103 min. Nothing in either guard case is above 5 min.

| Case | Body | Before | After |
|---|---|---|---|
| `general-interest-fallback` | ΔΣ — metro damage, 103 min | **0/4** | **4/4** |
| `committee-routine-silence` | ΔΕ — tender extension, procurement | 3/3 | **4/4** |
| `community-offarea-silence` | ΔΚ — bar music hours, Αμπελόκηποι | 3/3 | **4/4** |

All three come from real Athens meetings. The two guards matter more than the change: `community-offarea-silence` is the hardest false positive available — six **ωράριο μουσικής** decisions, her exact interest, 2.5km away in another community. It stays silent.

No regressions. `profile-learning` 3/3, `promise-durable` 3/3, `scoping-intent` 0/3 (known-failing, documented in the fixture).

## 2. Bird rejected every proactive update

```
Bird send failed for message cmt7bsccd002o5p50p1rxqij0: API returned 422:
{"code":"InvalidPayload", ... "missing value for variable 'link_path'"}
```

Three shells hold their URL button as `https://opencouncil.gr/{{link_path}}`. Bird rejects the send when that variable is absent. `templateBody` sent only `demos_text`.

Checked in the Bird console:

| Shell | Variables |
|---|---|
| `demos_update_agenda` | `demos_text` + `link_path` |
| `demos_update_news` | `demos_text` + `link_path` |
| `demos_followup` | `demos_text` + `link_path` |
| `demos_intro`, `demos_transition`, `demos_checkin` | none |

This is why launch looked healthy. The intro and transition shells declare no variables, so the enrollment cohort went out normally. The first proactive update was the first send that could fail, and every one after it would have failed too.

**The failure is permanent.** A 422 is not in `isRetryableStatus`, so the row goes terminal instead of staying pending for the sweeper. The SMS fallback fires, so those readers get the text without the buttons.

### What changed

The link path comes from the wake's event, which names the meeting. Every example value in Bird is a meeting path. A scheduled follow-up names no meeting, so there the path comes from the link the agent wrote into the body. If both are empty, the send uses the explainer page instead of failing.

`TemplateDef` gains `hasLinkPath`, which mirrors Bird the way this file already mirrors the shells.

## Also here

- The dangling-tool-call backstop from #679 checked only the last message. The end_turn repair paths orphan tool calls in the middle of the list. It now scans every assistant turn.
- The fixture scrubber left the reader's name in message bodies, because Greek accents defeated the match. It now strips combining marks, and it picks a pseudonym that cannot equal the reader's own name.
- **The fixtures no longer quote anyone.** Two fixtures already on main held a postal address and a workplace street. All reader turns are now paraphrased: same intent, same register, different words. Every keyword the cases assert on is unchanged, and the suite scores the same 18/24 as before the paraphrase.

## Verification

- `npx tsc --noEmit` clean. `npm test` passes 303 tests. `SKIP_ENV_VALIDATION=1 npm run build` succeeds.
- Every behavioural claim is n≥3 against the live model.
- No migration. No schema change.

## Known limits

- **The Bird mirror has no drift detector.** `hasLinkPath` is copied by hand from the console. The tests compare the file to itself, so a variable added in Bird stays invisible until a terminal 422 in production. This is the same detection lag as the bug above.
- `scoping-intent` stays 0/3. Four prompt wordings and 11 live replays produced one success. Prompt text is not the lever there.
- The three new fixtures repeat the same persona in full.

<h3>Greptile Summary</h3>

The PR improves Notis proactive updates, repairs Bird template delivery, strengthens dangling tool-call recovery, and anonymizes real-conversation evaluation fixtures.
- Adds a prompt fallback for major citywide council stories while preserving silence for routine committee and off-area community matters.
- Supplies Bird’s required `link_path` template variable with an explainer-page fallback.
- Repairs orphaned tool results across the complete message history.
- Generalizes fixture locations and scrubs resident-name variants from exported profiles, conversations, and decisions.

The PR appears safe to merge.

No blocking failure remains; the previously reported resident-name leak, identifying street-level fixture data, and non-terminating replacement loop are resolved at the current head.

| Filename | Overview |
|----------|----------|
| services/notis/scripts/eval-case-from-export.ts | The replacement cursor now advances after each substitution, resolving the previously reported non-terminating short-name case, while exported resident fields receive accent-insensitive name scrubbing. |
| services/notis/fixtures/evals/general-interest-fallback.json | The fixture now uses a public-square-level location and pseudonymized conversation while exercising the citywide-story fallback. |
| services/notis/fixtures/evals/committee-routine-silence.json | Adds a generalized guard fixture ensuring routine committee business remains silent. |
| services/notis/fixtures/evals/community-offarea-silence.json | Adds a generalized guard fixture ensuring another community’s hyperlocal licensing decisions remain silent. |
| services/notis/src/agent/runWake.ts | Reconciles unanswered tool calls throughout the message history rather than checking only the final turn. |
| services/notis/src/agent/templates.ts | Models which Bird templates require `link_path` and derives a template-safe path from message links. |
| services/notis/src/lib/bird.ts | Includes the required link variable when sending proactive Bird templates. |
| services/notis/src/lib/queue.ts | Passes the extracted or fallback link path through the proactive delivery boundary. |

<sub>Reviews (4): Last reviewed commit: ["fix(notis): close the review findings on..."](https://github.com/schemalabz/opencouncil/commit/ebc1de126743dc3223bede81cc21534ea00c6b9a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56267432)</sub>

**Context used:**

- Knowledge Base — [Notis notification service](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/notis-notification-service.md)
- Knowledge Base — [Resident notifications](https://app.greptile.com/schema-labs/-/custom-context/knowledge-base/schemalabz/opencouncil/-/docs/resident-notifications.md)


christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/02665ea051cb861f1f9f881f1e446add8665d844..9b3334d0ae8e1143d908e355dbe43ff3c0cbd179>) to correct a commit message that described only half its contents. Same diff.

**Both Greptile findings were right, and I acted on both.**

**The infinite loop is real.** I reproduced it before fixing: the scrubber restarted its search at position 0 after every replacement, so any reader whose name matched the pseudonym looped for ever and hung the export. «Εύα» and «Ευα» both hang; «Ευφη» and «Ευαγγελια» terminate. The search now advances past each replacement, and every one of those names terminates.

**The identifiability finding was right too, and worse than the comment knew** — it flagged the three fixtures on this branch, but the same sweep found two fixtures **already on main and released**:

- `scoping-intent` carried a full postal address in the profile: street, number and postcode.
- `promise-durable` and `profile-learning` carried the reader's workplace street, both in his own message and in the reply that echoed it back.

All are now neighbourhood level. The three fixtures here name a public square instead of a residential street, which moves every distance the cases turn on by at most 130m — Κυψέλη 2.23km → 2.11km, Πετράλωνα 2.37 → 2.40, Αμπελόκηποι 2.56 → 2.58. `profile-learning` asserted on the street name, so it now asserts on the neighbourhood.

Nothing regressed: `profile-learning` 3/3, and the three new cases 3/3 each. The scenario fixtures were checked and are clean. Street names elsewhere in the fixtures are council record — the addresses of municipal works — and stay.

**Note for the record:** commit `687ce51f`, already on main and in release 2026.8.16, names that workplace street in its message. Git history cannot be corrected without rewriting a public branch, so I have left it and am flagging it rather than quietly ignoring it. Your call whether it is worth acting on.

**Also in this push,** the `link_path` source changed after @christosporios asked what happens with two links, an outside link, or none. The answers were: first link wins (arbitrary), falls back to the explainer, falls back to the explainer. The wake's event now takes precedence — it names the meeting deterministically, and every example value in Bird is a meeting path. The body link is now only the fallback for a scheduled follow-up, whose event names no meeting. Precedence is pinned by tests.

298 tests pass, build clean.

christosporios:
Pushed `4023f765` and `ebc1de12`: the paraphrase, plus a full review pass. 303 tests, build clean.

**Fixtures no longer quote anyone.** Every reader-authored turn is paraphrased — same intent, same register (lowercase, dropped accents, the original typos), different words. The keywords each case asserts on are untouched, so «κηπους», «σαμοθρακη», «προηγουμενη», «μπιφ», «κλιπακι», «εξαρχεια» and «θυμηθεις» all stand. «Ας γνωριστούμε» stays verbatim because it is a quick-reply label from our own templates, not something a reader wrote.

Re-measured at n=3 across the whole suite: **18/24, identical to the run before the paraphrase.** Five cases hold at 3/3, scoping-intent stays 0/3 as documented, and the two that moved are both inside their historical range with unchanged inputs.

**The review's most useful finding was against my own tests.** Nothing exercised the code path whose absence caused the 422 — the wire tests called `realBird` directly with an explicit `linkPath`, and the resolvers were tested in isolation. Deleting the `linkPath` argument from either send call site left the suite green at 298/298. `FakeBird`'s type omitted the field too, so no test could have asserted it. Three tests now drive `deliverPendingMessage` end to end, and I verified they fail when the argument is removed.

Also fixed: a query string could ride into `link_path` and cause the same 422 (dropped at the first `?` or `#`, ids now encoded); both silent paths now log — a lookup broken for everyone was indistinguishable from one that found nothing, and the explainer fallback recorded nothing at all; `bird.test.ts` leaked `global.fetch` past its tests because `restoreAllMocks` does not undo a plain assignment; and `dangling-tool-calls` is recorded once per wake rather than once per turn.

**Two doc comments were attached to the wrong things** — inserting the new helper left `deliverPendingMessage`'s comment about the delivery choke point and ΣΤΟΠ rails describing a wake lookup instead. That is the same defect as the janitor.ts hunk in the previous review, which I had already been told about and then repeated twice. Both moved back.

**Three findings left open, none of which can lose a message:**

1. **No drift detection on the Bird template mirror.** `hasLinkPath` is hand-copied from the console and pinned by tests that compare it to itself. Add a variable in Bird and nothing here notices until a terminal 422 in production — the same detection lag as the bug this PR fixes. The real fix reads the template definition from Bird's API in CI.
2. **The three new fixtures duplicate ~150 lines of identical persona.** A shared persona file would make divergence impossible.
3. **The dangling-call backstop unshifts into a message that may carry a cache breakpoint**, invalidating that prefix. Correctness is fine; it costs a full-price re-read on a rare path.

Happy to take (1) here or as a follow-up — it is the one with real operational value.

PR #683, opened 2026-08-24
notis: detect drift between the template mirror and Bird

christosporios:
Follow-up to #682, which fixed the missing `link_path` but left the gap that allowed it.

## The problem

The shells in `src/agent/templates.ts` are a hand-copy of the Bird console. Every variable a shell declares must be supplied, or Bird rejects the send with a 422. A 422 is terminal: the row fails, the SMS fallback covers it, and the reader gets the text without buttons. Nothing else notices.

That is how `link_path` shipped broken. The console changed on 23 August. The repository did not. The first proactive update failed days later, and it surfaced as one alert line rather than as "every send of this shell is failing".

**The mirror's unit tests cannot catch this.** They compare the file to constants typed from the same reading, and to itself. Both pass forever while Bird moves on. Only Bird knows.

### Before the deploy

`npm run check:templates` reads each shell's declared variables from Bird and compares them to the mirror.

```
  ok       demos_intro
  ok       demos_transition
  DRIFT    demos_checkin
           Bird declares link_path and we never send it — every send of this shell fails with a 422.
```

It runs in the notis CI job. Without credentials it skips with a visible log line, so forks and Dependabot are unaffected. Missing credentials exit 78, not 0 — "cannot ask" must never read as "no drift".

### After it

Bird's 422 body already names the variable it wanted. The send error now lifts that name to the front and says which file is stale:

> Bird requires the template variable 'link_path', which this send did not supply — the shell's entry in src/agent/templates.ts is out of date with the Bird console.

One alert now carries its own diagnosis.

## One list, two callers

`declaredVariables()` builds the wire parameters **and** is what the check compares. The thing verified is the thing sent. A mirror checked against a different list from the one the sender uses would pass while the sends still failed.

## Setup needed

The CI step needs these repository secrets. Until they exist the step skips cleanly, so this PR is safe to merge first:

- `BIRD_API_KEY`, `BIRD_WORKSPACE_ID`
- `BIRD_WHATSAPP_TEMPLATE_DEMOS_{INTRO,TRANSITION,UPDATE_AGENDA,UPDATE_NEWS,FOLLOWUP}`

## Known limits

- `demos_checkin` has no project id and no send path, so it cannot be checked. The script reports it as not checked rather than passing over it silently.
- The check reads the union of variables across a project's template versions. A variable declared by any version can reach a send, so the union is the safe comparison.

## Verification

- Against the console as it stands: no drift.
- A simulated variable added to a shell: reported by name.
- Without credentials: exits 78.
- `npx tsc --noEmit` clean, `npm test` 309 passing, build succeeds.

<h3>Greptile Summary</h3>

The PR adds a CI drift check between Notis template declarations and Bird, centralizes the variable list used by both checking and sending, and improves diagnostics for missing-variable responses.
- Adds the `check:templates` command and wires it into the Notis CI job.
- Compares locally declared template variables with Bird project metadata.
- Builds outgoing Bird parameters from the shared declaration helper.
- Adds targeted tests for variable comparison and terminal 422 diagnostics.

The PR should not merge until missing project IDs for sendable templates fail the drift check instead of allowing CI to pass.

A sendable template whose optional project secret is absent is recorded as not checked, but skipped templates do not affect the process exit status, allowing the exact template drift this change is intended to prevent to reach deployment.

**Files Needing Attention:** services/notis/scripts/check-template-drift.ts, .github/workflows/ci.yml

| Filename | Overview |
|----------|----------|
| .github/workflows/ci.yml | Adds the credential-gated Bird template drift check to the Notis CI job. |
| services/notis/scripts/check-template-drift.ts | Implements Bird metadata retrieval, template-variable comparison, reporting, and process exit handling. |
| services/notis/src/agent/templates.ts | Adds shared helpers for declaring and comparing each template’s wire variables. |
| services/notis/src/lib/bird.ts | Uses shared variable declarations to construct template parameters and enriches missing-variable errors. |
| services/notis/src/lib/__tests__/bird.test.ts | Covers shared variable declarations, drift comparison, and missing-variable diagnostics. |
| services/notis/package.json | Exposes the template drift checker as a workspace script. |

<sub>Reviews (2): Last reviewed commit: ["feat(notis): detect drift between the te..."](https://github.com/schemalabz/opencouncil/commit/e6e7d17263b17fb1c432381a20f3277a3712b6db) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56285754)</sub>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,35 @@ jobs:
 
       - run: SKIP_ENV_VALIDATION=1 npm run build -w notis
 
+      # The template mirror in src/agent/templates.ts is a hand-copy of the
+      # Bird console, and only Bird knows when it goes stale. A variable added
+      # there makes every send of that shell fail with a terminal 422 that
+      # degrades quietly to SMS, so the check belongs before the deploy rather
+      # than in a reader's inbox.
+      #
+      # Skipped without credentials — forks and Dependabot cannot read secrets,
+      # and a missing key must not fail their builds. It still runs on every
+      # push to this repository, which is where a template edit lands.
+      - name: Check the Bird template mirror for drift
+        # The skip is a shell test rather than a step `if`: a step's own env
+        # block is not visible to its condition, and `secrets` is not available
+        # there either. This way the skip is also visible in the log.
+        run: |
+          if [ -z "$BIRD_API_KEY" ]; then
+            echo "No Bird credentials in this context — skipping the drift check."
+            exit 0
+          fi
+          npm run check:templates -w notis
+        env:
+          BIRD_API_KEY: ${{ secrets.BIRD_API_KEY }}
+          BIRD_WORKSPACE_ID: ${{ secrets.BIRD_WORKSPACE_ID }}
+          BIRD_WHATSAPP_TEMPLATE_DEMOS_INTRO: ${{ secrets.BIRD_WHATSAPP_TEMPLATE_DEMOS_INTRO }}
+          BIRD_WHATSAPP_TEMPLATE_DEMOS_TRANSITION: ${{ secrets.BIRD_WHATSAPP_TEMPLATE_DEMOS_TRANSITION }}
+          BIRD_WHATSAPP_TEMPLATE_DEMOS_UPDATE_AGENDA: ${{ secrets.BIRD_WHATSAPP_TEMPLATE_DEMOS_UPDATE_AGENDA }}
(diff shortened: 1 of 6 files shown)


kouloumos:
Rebased onto current main and force-pushed. One conflict in `services/notis/src/lib/bird.ts`: main added the `bird-extract` import beside `TEMPLATES`, this branch replaced `TEMPLATES` with `declaredVariables`. Resolved as the union — `TEMPLATES` has no remaining use in the file, since `declaredVariables()` does that lookup itself.

Typecheck clean, 354 tests pass. The pre-rebase state is at [`backup/pr-683-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-683-original).

PR #690, opened 2026-08-25
chore(landing): remove the /old-landing route

christosporios:
## What this changes

The immersive landing page at `/` replaced the old landing page. The `/old-landing` route kept the previous version reachable. Nothing links to it any more, so this PR removes the route and the code that only it used:

- the `Landing`, `Hero`, `HeaderBar`, `CityOverview` and `CityMiniCard` components
- `src/lib/db/landing.ts`, which supplied the Substack feed and the `LandingCity` / `LandingPageData` types
- `fetchLatestSubstackPostCached` in `src/lib/cache/queries.ts`
- `HIRING_CONFIG` in `src/lib/features/config.ts`, the source of the hiring badge in the header bar
- the `Landing` translation keys for the removed components, in all four catalogs
- `docs/landing-page.prd.md`, which specifies only the removed page

`OpenCouncilDescription` stays, because the `/petition` page uses it. Its unused `animate` prop goes, because only `Hero` passed it.

## No redirect

An earlier revision of this PR added a `/old-landing` redirect pair. Review asked for evidence that the pair guards a real index entry. The evidence says it does not, so the second commit removes it again.

- **Search Console holds no impression and no click for the route.** The synced by-page table covers 2025-02-17 to 2026-08-23, with 64832 rows over 7197 pages. The route existed from 2026-07-09 to 2026-08-25, inside that window. A search for any page that contains `old-landing` returns no row.
- **Analytics holds 3 pageviews**, from 2 people, all direct, all on 2026-08-25 between 20:17 and 20:20 UTC. The 47 days before hold none. The time matches the work that removed the route, so these are most probably manual checks.
- **No commit ever added a link to the route**, and `sitemap.ts` never listed it.

A permanent redirect is a permanent entry in a table that nobody prunes, so this PR leaves the redirect table as it is. `next.config.mjs` and `next-config-redirects.test.ts` are now identical to `main`.

## Code that the deletion made unreachable

The old landing was the last caller of several things. This PR removes them in the same sweep:

- **`MunicipalitySelector`** — `hero.tsx` was the only caller that let `hideQuickSelection` keep its default of `false`, so the `!hideQuickSelection` branch could no longer render. The prop and the block go, and the file's last `framer-motion` import goes with them.
- **`MeetingCard`** — `city-overview.tsx` was the only caller that passed `mostRecent` and `headingLevel`. The card now renders only through the generic `List` component, which passes neither. The "most recent meeting" badge and its `MeetingCard.mostRecent` key in four catalogs go. The heading is now a plain `h2`, which was already the default.
- **`packages/ui/tailwind-preset.ts`** — the `fade-in` keyframe and the `animate-fade-in` utility lost their only user, `landing.tsx`.

## Verification

- `npx tsc --noEmit` — exit 0
- `npm run lint` — clean
- `npm test` — 1690 tests. All suites pass, except an intermittent failure in `src/lib/__tests__/utils.test.ts` that also occurs on `main`; see the note below.
- `next build` — compiles, TypeScript passes, 223 static pages generate
- A repo-wide search for `old-landing`, `HIRING_CONFIG`, `fetchLatestSubstackPost`, `@/lib/db/landing` and `@/lib/features/config` returns no hits.
- The four message catalogs stay key-parallel. Each keeps the same eight `Landing.desc*` keys and the same twelve `MeetingCard` keys. `OpenCouncilDescription` is the only consumer of the `Landing` namespace, and it reads exactly those eight keys.

## Notes for the reviewer

These came up in review. This PR does not change them, because each needs a decision or a change outside this scope:

- `getAllCitiesMinimalCached` still computes the `_count` aggregates per city. `/old-landing` was the only reader. Its one remaining caller, `/petition`, ignores them. A slim `select` belongs in a separate change, because `/api/cities/all` publishes `_count` from the **uncached** `getAllCitiesMinimal` under an OpenAPI schema.
- The `Landing` namespace and `src/components/landing/OpenCouncilDescription.tsx` now serve only `/petition`. The live landing uses the `landingV2` namespace under `src/components/landing/v2/`. Moving the file and folding the namespace belongs in a follow-up.
- Six `MeetingCard` keys have no reader: `beingProcessed`, `noVideo`, `subjects`, `withAgenda`, `withAudio` and `withVideo`. They predate this PR.
- The catalogs also hold four namespaces with no consumer: `AboutPage`, `Chat`, `PersonCard` and `Offer`, about 19 KB across the four files. This staleness predates this PR.
- The redirect test pins the `source` patterns only, not the destination or the status code. The gap applies to every rule in the table. @kouloumos has a `resolveRedirect` helper for a follow-up.
- The site no longer reads the Substack RSS feed anywhere. The `/explain` carousel uses the static list in `src/lib/explain/substackPosts.ts`, so the channel stays reachable, but the home page has no link to it. Restoring a link is a product decision.
- `src/lib/__tests__/utils.test.ts` holds an order-dependent `klitiki` test that fails in a full run about half the time and passes in isolation. It predates this change: a full run of the suite at `origin/main` in a separate worktree reproduces the same failure in 2 of 4 runs, and this branch changes neither `src/lib/utils.ts` nor that test.

https://claude.ai/code/session_014JPqRfTTvvpvgkHGHEGSMK

<h3>Greptile Summary</h3>

The PR removes the obsolete `/old-landing` page and code used only by that page.
- Deletes the old landing components, data access, cache query, feature configuration, and product specification.
- Removes unused translation keys, animation tokens, component props, and unreachable selector UI.
- Keeps the petition landing description and simplifies its remaining component contracts.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/app/[locale]/(other)/old-landing/page.tsx | Deletes the obsolete localized route and its landing-page data-fetching path. |
| src/components/meetings/MeetingCard.tsx | Removes props and badge behavior used only by the deleted landing-page caller. |
| src/components/onboarding/selectors/MunicipalitySelector.tsx | Removes quick-selection UI that no remaining caller displayed. |
| src/lib/cache/queries.ts | Removes the cached Substack query after deletion of its final caller. |
| packages/ui/tailwind-preset.ts | Removes the unused fade-in keyframe and animation utility. |

<sub>Reviews (4): Last reviewed commit: ["chore(landing): drop the /old-landing re..."](https://github.com/schemalabz/opencouncil/commit/0f2cfa6ad77c579682fbcbc513729a69614c2e3f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56878512)</sub>


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/1aed3e9daa4a2cbcd59a7c86f914657087e08713..bc83bdc905f52698521a47d90d9ff5f6922eb0a1) to sweep the code that the deletion made unreachable. The earlier description listed these as out of scope. They are the same defect class as the deletion itself, so this change removes them:

- `MunicipalitySelector`: the `!hideQuickSelection` branch could no longer render, because `hero.tsx` was the only caller that let the prop keep its default of `false`. The prop, the block, and the file's last `framer-motion` import go.
- `MeetingCard`: `city-overview.tsx` was the only caller that passed `mostRecent` and `headingLevel`. The card now renders only through the generic `List` component, which passes `item`, `editable` and `cityTimezone`. Both props, the "most recent meeting" badge, and the `MeetingCard.mostRecent` key in four catalogs go. The heading becomes a plain `h2`, which was already the default.
- `packages/ui/tailwind-preset.ts`: the `fade-in` keyframe and the `animate-fade-in` utility lost their only user, `landing.tsx`.
- `docs/landing-page.prd.md`: this PRD specifies only the removed page.

The commit body and the PR description now record all of this.

Verification after the change: `npx tsc --noEmit` exits 0, `npm run lint` is clean, and `npm test` passes 124/124 suites and 1690 tests. The four catalogs stay key-parallel, at eight `Landing` keys and twelve `MeetingCard` keys each.

I left four items alone and listed them under "Notes for the reviewer", because each needs a decision or a change outside this scope. The `getAllCitiesMinimalCached` `_count` fetch is the one worth a look: `/old-landing` was its only reader, but the counts cannot simply be dropped, because `/api/cities/all` publishes them from the uncached function under an OpenAPI schema.

---

kouloumos (approved):
Approved.

Two notes. Neither of them blocks this PR.

**Six `MeetingCard` keys have no reader.** They are `beingProcessed`, `noVideo`, `subjects`, `withAgenda`, `withAudio` and `withVideo`. `MeetingCard.tsx:52` is the only file that binds the namespace, and it reads six other keys. These six predate this PR. The base component read the same live set plus `mostRecent`. Your sweep of `mostRecent` is exact, so this is only adjacent.

**The redirect test pins the paths, not the destinations.** `matchesAnyRedirect` builds its patterns from a regex that captures `source:` only. I changed `destination` to `/nonexistent` and the suite still passed. I changed `permanent` to `false` and the suite still passed. The gap is pre-existing. It applies to `/map`, `/chat`, `/petitions` and the four `/meetings/` rules. It is worth naming here because your rationale is SEO, and the target URL and the 308 are the two things a search engine reads. Deleting either new rule does fail the new case, so the case itself is sound. I have a `resolveRedirect` helper that closes the gap. Tell me if you want it in a follow-up.

christosporios:
Thanks. I verified both notes. Both hold.

**The six `MeetingCard` keys.** `MeetingCard.tsx:52` is the only binder of the namespace, and its six `t()` calls read `loading`, `upcoming`, `today`, `notPublic`, `moreSubjects` and `noSubjects`. The file constructs no key dynamically, so `beingProcessed`, `noVideo`, `subjects`, `withAgenda`, `withAudio` and `withVideo` have no reader. They predate this PR, so I leave them here. I can remove them in a follow-up if you want them gone.

**The redirect test.** Confirmed: the helper builds its patterns from `/source:\s*'([^']+)'/g`, so it reads `source` only. The destination and the status code stay unpinned for all seven rules.

Yes to `resolveRedirect` in a follow-up. I do not add a partial destination check for the two new rules here, because your helper replaces the same extraction step and covers every rule. Two mechanisms for one job would only need untangling later.

---

kouloumos:
One question about the redirect. It does not change my approval.

The redirect guards an index entry that cannot exist. `/old-landing` had no discovery path:

- The route appeared on 2026-07-09, as a rename of `page.tsx` in `25e4c312`. The URL did not exist before that date, so no earlier index entry points at it. The indexed URL is `/`, which still resolves.
- Nothing ever linked to it. Five commits in the whole history contain the string `old-landing`, and none of them adds a link. The only occurrence is the page's own `buildHreflangAlternates('/old-landing', locale)`.
- `sitemap.ts` never listed it, in any revision.
- `robots.ts` is a generic `allow: '/'`. It permits a crawl. It does not advertise a URL.

A self-referential canonical and a permissive robots rule apply only after a crawler reaches the page. Nothing sent one there. That is the difference from `/map` and `/chat`, which were linked public pages.

So we add a `permanent: true` — a promise we keep forever — for a URL nobody holds. Redirect tables never lose entries, because nobody can prove an entry is safe to remove.

Please check Search Console for `/old-landing`. If it shows no impressions and no coverage entry, drop both rules and the test case: 23 deletions, and the remaining redirect tests still pass. If Search Console does show it, keep the rule and correct the comment instead — it currently states a reason that never applied.

christosporios:
Checked. You are right, and [`0f2cfa6a`](https://github.com/schemalabz/opencouncil/commit/0f2cfa6a) drops both rules and the test case. `next.config.mjs` and `next-config-redirects.test.ts` now match `main` exactly.

**Search Console.** The project syncs Search Console into the warehouse, so I queried it rather than the console UI. The by-page table covers 2025-02-17 to 2026-08-23, and holds 64832 rows over 7197 distinct pages. The route existed from 2026-07-09 to 2026-08-25, inside that window.

```sql
SELECT page, sum(impressions), sum(clicks)
FROM googlesearchconsole_search_analytics_by_page
WHERE page ILIKE '%old-landing%'
GROUP BY page
```

No row. No impression, no click, at any time.

One limit, so the evidence is not read wider than it is: this table is search performance. It is not the index-coverage report. A URL that Google crawled but never showed in a result would not appear here. So I confirm the impressions half of your test, not the coverage half.

**Analytics closes that gap from the other side.** Over the route's whole life, `/old-landing` has 3 pageviews, from 2 people, all `$direct`, all on 2026-08-25 between 20:17:22 and 20:20:27 UTC, on `opencouncil.gr`. The 47 days before hold none. The time matches the work that removed the route, so these are most probably manual checks. A crawl that no impression follows would leave nothing here, but a person who holds the URL in a bookmark would. Nobody does.

**Your history claims check out.** `25e4c312` is dated 2026-07-09. Three commits in the history contain the string: that one, `d35ac126` (which swapped `buildHreflangAlternates` for `buildCanonicalAlternates`), and the removal. None of them adds a link. `git log -S "old-landing" -- src/app/sitemap.ts` is empty for every revision.

Two process notes:

- The removal is a second commit, not an amend. A force-push is blocked in my environment right now. Squash at merge if you want the single commit.
- I could not update the comment you flagged, because the change deletes it. For the record, its claim was narrower than the PR body's: it said the route stayed reachable and is now removed. The body was the part that argued from indexability, and it now carries the evidence above instead.

---

PR #701, opened 2026-08-31
Improve city selector UX on mobile with full-height dialog

christosporios:
## Summary
Refactors the city selector (Combobox) to provide a better mobile experience by replacing the centered dialog with a full-height sheet anchored to the top of the screen. Also extracts the city item rendering into a reusable component and groups cities by notification support status.

## Key Changes

- **Mobile-optimized Combobox dialog**: Converts the centered dialog to a full-height sheet (`h-[100dvh]`) that anchors to the top, preventing the software keyboard from covering results. Removes the 300px height cap on the results list on mobile.

- **Responsive input sizing**: Sets the search input to `text-base` on mobile to prevent iOS Safari from zooming the page when the field takes focus.

- **Extracted CityComboboxItem component**: Creates a reusable `CityComboboxItem` component that displays cities with visual indicators (bell icon for notification-supporting municipalities, map pin for others). Replaces inline item rendering in both `MunicipalitySelector` and `NotificationPreferencesSection`.

- **City grouping by notification support**: Groups cities into "available" (supports notifications) and "unavailable" (petition-only) sections in the city selector, reducing cognitive load when browsing ~375 municipalities. Only a small subset can send notifications.

- **Improved dialog accessibility**: Adds `aria-describedby={undefined}` to the dialog since the title names the task and the field has its own placeholder. Increases the close button touch target to 40px with `-m-3 p-3` padding.

- **UI refinements**:
  - Changes the city selector trigger button icon from Bell to Plus
  - Adjusts button sizing to be full-width on mobile (`w-full`) and auto-width on desktop (`sm:w-auto`)
  - Updates the combobox width from `w-64` to `w-80` for better visibility
  - Removes the unused `size` prop from `CitySelector`

- **Test coverage**: Adds tests for mobile viewport behavior, including dialog presentation, search filtering consistency, and verification that the results list is not capped at 300px.

- **Internationalization**: Adds translation keys for city grouping labels (`groupAvailable`, `groupUnavailable`) in all supported languages (English, Greek, French, Serbian).

## Implementation Details

The Combobox now detects viewport width on mount to choose between a Popover (desktop) and full-height Dialog (mobile). The dialog uses `max-h-full` and `flex-1` on mobile to fill available space after the software keyboard appears, ensuring results remain visible. The `CommandList` is wrapped with responsive classes to maintain proper scrolling behavior.

The `CityComboboxItem` component uses consistent icon sizing (`!h-3 !w-3`) to override the Combobox's default icon sizing, and includes visual grouping with colored circular backgrounds (orange for notification-supporting, gray for others).

https://claude.ai/code/session_01Wbuvnit1VtAvSYNxuvqaek

<h3>Greptile Summary</h3>

The PR refactors the mobile city selector into a responsive drawer and introduces shared, grouped municipality rendering.
- Replaces the mobile dialog with a flex-based drawer and uncapped scrolling results.
- Groups municipalities by notification availability and adds localized group labels.
- Extracts shared city-row rendering and adds mobile and grouping tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/components/Combobox.tsx | Replaces the mobile dialog with a responsive drawer, adjusts list sizing and focus behavior, and omits empty groups. |
| src/components/__tests__/Combobox.test.tsx | Adds coverage for mobile presentation, filtering, focus behavior, result-list sizing, and empty groups. |
| src/components/cities/CityComboboxItem.tsx | Introduces a reusable municipality row with notification-support indicators. |
| src/components/onboarding/selectors/MunicipalitySelector.tsx | Reuses the extracted city item component without changing selection behavior. |
| src/components/profile/NotificationPreferencesSection.tsx | Groups municipalities by notification support and updates the responsive selector trigger. |

<sub>Reviews (4): Last reviewed commit: ["fix(ui): open the mobile combobox as a d..."](https://github.com/schemalabz/opencouncil/commit/9f8835da5612a2e90a153f13f259ed1eb2195b97) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58589599)</sub>

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -9,3 +9,20 @@ require('@testing-library/jest-dom');
 const { TextEncoder, TextDecoder } = require('util');
 if (typeof global.TextEncoder === 'undefined') global.TextEncoder = TextEncoder;
 if (typeof global.TextDecoder === 'undefined') global.TextDecoder = TextDecoder;
+
+// jsdom doesn't implement matchMedia, which vaul (the mobile drawer) queries on
+// mount. A stub that matches nothing is enough: the components under test fall
+// back to their default branch, and the tests drive the behaviour they care
+// about themselves.
+if (typeof window !== 'undefined' && typeof window.matchMedia !== 'function') {
+    window.matchMedia = (query) => ({
+        matches: false,
+        media: query,
+        onchange: null,
+        addListener: () => { },
+        removeListener: () => { },
+        addEventListener: () => { },
+        removeEventListener: () => { },
+        dispatchEvent: () => false,
+    });
+}
--- a/messages/el.json
+++ b/messages/el.json
@@ -674,6 +674,8 @@
         "selectCityPlaceholder": "Επιλέξτε Δήμο...",
         "searchCityPlaceholder": "Αναζητήστε Δήμο...",
         "cityNotFound": "Δεν βρέθηκε Δήμος.",
+        "groupAvailable": "Δήμοι με ειδοποιήσεις",
+        "groupUnavailable": "Δεν υποστηρίζονται ακόμα",
         "selectCityButton": "Επιλέξτε Πόλη",
         "preferencesTitle": "Προτιμήσεις ειδοποιήσεων",
         "sendTiming": "Στέλνουμε μια ενημέρωση πριν από μια συνεδρίαση και μια ενημέρωση μετά από αυτήν.",
--- a/messages/en.json
+++ b/messages/en.json
@@ -684,6 +684,8 @@
         "selectCityPlaceholder": "Select Municipality...",
         "searchCityPlaceholder": "Search Municipality...",
         "cityNotFound": "No municipality found.",
+        "groupAvailable": "Municipalities with notifications",
+        "groupUnavailable": "Not available yet",
         "selectCityButton": "Select City",
         "preferencesTitle": "Notification preferences",
         "sendTiming": "We send a heads-up before a meeting and a follow-up after it.",
--- a/messages/fr.json
+++ b/messages/fr.json
@@ -674,6 +674,8 @@
     "selectCityPlaceholder": "Sélectionnez une commune...",
     "searchCityPlaceholder": "Rechercher une commune...",
     "cityNotFound": "Aucune commune trouvée.",
+    "groupAvailable": "Communes avec notifications",
+    "groupUnavailable": "Pas encore disponibles",
     "selectCityButton": "Sélectionnez une ville",
     "preferencesTitle": "Préférences de notification",
     "sendTiming": "Nous envoyons une information avant une séance et une autre après celle-ci.",
--- a/messages/sr.json
+++ b/messages/sr.json
@@ -684,6 +684,8 @@
         "selectCityPlaceholder": "Изаберите општину...",
         "searchCityPlaceholder": "Претражите општину...",
         "cityNotFound": "Није пронађена ниједна општина.",
+        "groupAvailable": "Општине са обавештењима",
+        "groupUnavailable": "Још нису доступне",
         "selectCityButton": "Изаберите град",
         "preferencesTitle": "Подешавања обавештења",
         "sendTiming": "Шаљемо најаву пре седнице и извештај након ње.",
--- a/packages/ui/src/dialog.tsx
+++ b/packages/ui/src/dialog.tsx
@@ -33,8 +33,15 @@ const DialogContent = React.forwardRef<
   React.ElementRef<typeof DialogPrimitive.Content>,
   React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & {
     align?: "start" | "center"
+    /**
+     * Extra classes for the close button. The default 16px box is the size of
+     * its icon, which is small for a touch target, but it sits over the corner
+     * where a `p-0` dialog puts its content. Grow it only where the content
+     * leaves the corner free.
+     */
+    closeClassName?: string
   }
->(({ className, children, align = "center", ...props }, ref) => (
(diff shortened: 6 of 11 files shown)


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/1e3c8c1b93bbae2a2d5d852ee010f4bbf13c9d59..89aeda155b20aadc596892dbc264a85f18a3c1e1) to fix two review findings, and squashed the fixups:

- **The larger close button is now opt-in.** The first version grew the close button of every dialog to 40px. That button sits over the top right corner, which is where a dialog with `p-0` puts its own content. `DialogContent` now takes a `closeClassName`, and only the mobile combobox sheet asks for the larger target.

  Measured with `elementFromPoint` over the calendar in the `RolesList` date dialogs: the 40px box covered **96%** of the next-month button, so a click there closed the dialog. Every other dialog now renders the same markup as `main` again (`cn(<original>, undefined)` returns the original string unchanged).

  One thing the measurement turned up: the **16px** box already covers **31%** of that same next-month button, the centre included, so a click in the middle of it closes the dialog on `main` today. That bug is older than this branch, so I left it alone rather than widen this PR.

- **Empty group headings.** cmdk hides a group whose items the search filtered out, but not one that never had any. The city picker fetches its list on the first click of the trigger, so both headings drew above "no municipality found" each time it opened, and stayed there if the request failed. `Combobox` now drops groups with no items, which covers every caller. Added a test.

Rebased into 5 commits. The first one changed the most: it adds the `closeClassName` option rather than growing the button everywhere, so it carries a new message.

`tsc`, `eslint` and `npm run build` are clean, and each commit passes the test suite on its own. One unrelated pre-existing flake shows up on about one run in three: `klitiki › should handle mixed casing`, where `klitiki('ΝΙΚΟΣ')` returns `Νικος` instead of `Νίκο`. It is non-determinism inside the third-party declension library, and it is not touched by this branch.

---

christosporios:
Pushed 3d8e0b4 and 33f99f5. On an iPhone the results list did not scroll: a drag over it moved the page instead.

**The cause.** The sheet is `position: fixed` at `100dvh`, which measures the layout viewport. iOS shrinks only the *visual* viewport for the software keyboard, and the sheet takes focus when it opens, so the keyboard is up straight away and the lower part of the sheet sits behind it. Safari pans the page when the user drags inside a fixed element taller than the visual viewport, rather than scroll a list inside it.

The sheet now measures the visual viewport and takes that height, plus the offset Safari applies when it scrolls the visual viewport to keep the focused field in sight. With a keyboard 384px tall on a 844px screen, the sheet becomes 460px and the whole list stays above it, where before 384px of the list was behind the keyboard. A drag then scrolls the list 241px and moves the page by 0. The `100dvh` class stays as the fallback for a browser without the API, which I checked separately: the sheet keeps its full height and the list still scrolls.

The second commit fixes a related gap I found while measuring. Radix locks scrolling with `react-remove-scroll`, which sets `overflow: hidden` on `body`. This app scrolls on `html`, so the page behind the sheet never locked at all. It now uses the same hook as the search filter panel, which locks both elements. The list also gets `overscroll-contain`, and `min-h-0` against the flex `min-height: auto` behaviour that would let the list grow past the sheet and be clipped, with nothing left to scroll.

**What I could not check.** There is no WebKit build in this environment, so I could not reproduce the original fault or confirm the fix on Safari itself. Chromium with touch emulation scrolled the list correctly both before and after, so it never showed the bug. What I did verify there is the mechanism this fix changes: the sheet resizes to the visual viewport, the list ends above the keyboard line, and a drag scrolls the list and not the page. The reasoning about Safari's panning is from the behaviour of a fixed element taller than the visual viewport, not from a run on the device, so it is worth a retest on the preview.

`tsc`, `eslint` and `npm run build` are clean, and 1732 tests pass.

---

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/33f99f5...9f8835d) to replace the mobile dialog with the vaul drawer, and dropped the two commits before it, which did not work.

I installed WebKit in this environment this round, so the fix is measured on the engine rather than reasoned about. With an iPhone profile, opening the sheet and then calling `window.scrollTo`:

| | dialog (previous head) | drawer (now) |
| --- | --- | --- |
| page moved while the sheet is open | **300px** | **0px** |
| `body.position` | `relative` | `fixed` |
| `body.top` | unset | `-300px` |
| `html` overflow | `hidden` | `visible` |
| focused on open | `INPUT` | `BUTTON` |

The second column is the fix. The first column is the bug: `overflow: hidden` was applied to `html` and WebKit scrolled the page anyway. That is what vaul's own source says, in the comment above `preventScrollMobileSafari`: Safari keeps scrolling the page once its toolbars collapse, and again whenever the keyboard covers part of the viewport. vaul pins the body with `position: fixed` at the current offset, cancels the touch moves that would scroll the page, and marks the one element that may scroll instead. This repo already uses the same drawer for the mobile sheets of the consultation views.

The drawer no longer takes focus when it opens, so the keyboard stays down and the whole list is on screen. The field is one tap away.

Dropped, because measurement showed they do nothing on WebKit: the `html` scroll lock, and the sheet sized to `window.visualViewport`. `DialogContent` gives back the `closeClassName` that only the dialog sheet used, so `packages/ui/src/dialog.tsx` is untouched by this PR now.

Chromium still behaves: a drag over the list scrolls it 313px and moves the page by 0, at either end of the list and over the header. `tsc`, `eslint` and `npm run build` are clean, and 1733 tests pass.

What I still cannot test is a real iPhone. Playwright's WebKit on Linux is the same engine, but it has no software keyboard, so the keyboard cases in that vaul comment are covered by vaul rather than by anything I measured.

---

PR #711, opened 2026-09-03
Implement AI-generated illustration for every subject

### LINKED ISSUE #705: AI-generated illustration for every subject
# Feature

Currently, subjects are text only. The landing list is a column of grey cards, and the subject page opens on a paragraph. We would like to improve the UI by generating an image for each subject. These images can be shown in multiple places:

- The landing page:
- The specific subject page
- The council page

Image

# Proposal

Every subject discussed in a council meeting gets its own illustration: a small, friendly scene in a consistent visual style that shows what the subject is about — a broken water main, a new pedestrian street, a school renovation. The images are generated automatically as subjects are published, an admin can swap out any that miss the mark for a real photo, and subjects that don't have an image yet show a simple colored placeholder. The goal is to make the landing page scannable at a glance and give each subject page a proper header, so a resident browsing what their municipality discussed sees a lively, readable feed instead of a wall of text — at a cost of about three cents per image.

We need to implement three components:

- A library that turns a subject into an image.
- One API route in the main app that reads, creates and replaces images.
- A `<SubjectImage>` component.

## The library

Lives in `packages/subject-images`, published as `@opencouncil/subject-images` (same setup as `@opencouncil/ui`, so add it to `transpilePackages`). It imports no Next, no Prisma and no env.mjs; the caller passes the API key in.

```ts
export function buildPrompt(input: { title: string; description: string }): string;

export async function generate(prompt: string, deps: { apiKey: string }): Promise<Buffer>;  // WebP, 1344×768

// Look up a subject's image. null means "no image" — the caller decides what fallback to show.
export async function resolve(subjectId: string): Promise<{ url: string } | null>;

export async function store(subjectId: string, image: Buffer): Promise<void>;
```

`resolve` does the existence check against S3 and builds the CDN URL, ETag included as a cache-buster. `store` writes the object. The object key lives inside the library, so no other code can disagree with it.

`buildPrompt` fills a constant template with the subject's title and description. The template holds the style wording, plus negative instructions that block readable text, real signage, logos and recognisable faces.

The image is wide because both places we show it are wide, and a 1K image costs the same at any aspect ratio. `generate` calls Gemini (`gemini-3.1-flash-lite-image`) and converts the result to WebP with sharp, which is already a dependency.

## The bucket

One object per subject:
`subject-images/{subjectId}.webp`

## The route

`src/app/api/subject/[subjectId]/image/route.ts`, following the existing `api/subject/[subjectId]/...` routes.

**GET** calls `resolve()` on the library and returns the result.

**POST** is superadmin only, with two modes:

- *Generate* — `buildPrompt` → `generate` → `store`
- *Upload* — validate the file, then `store`

Generating over a manual upload replaces it, which is how you undo one. The generate mode is a thin wrapper around an internal `generateImageForSubject()` function, because the pipeline calls that function directly — it has no session and could not pass the superadmin check.

**Fallback**: a flat SVG in the topic's color with the topic's icon, built from `topicStyle`. Subjects without a topic get a neutral one. No broken images, no layout shift.

## When images get generated

**From the pipeline.** The `processAgenda` and `summarize` result handlers call `generateImageForSubject()` for each of the meeting's subjects, fire-and-forget, with failures logged and alerted on Discord. The first covers agenda subjects days before the meeting; the second covers non-agenda subjects, which only exist once there is a transcript. Subjects that already have an object are skipped, so the second run costs nothing for the first run's work. Meetings with no agenda PDF are covered by the second trigger alone.

**From the read route.** A GET that misses can start a generation in the background. Old subjects get images as people look at them, so there is no backfill job to run or babysit.

## Frontend

`<SubjectImage subjectId />` renders a plain `` pointing at `/api/subject/{id}/image`, with `loading="lazy"`. We skip `next/image`: the route redirects, and the stored file is already WebP.

## Admin UI

The moment an admin knows an image misses the mark is the moment they are looking at it, so the controls live on the image.

**On the subject page.** The `<SubjectImage>` header gets a hover overlay, visible only to authorized users, with two buttons:

- *Regenerate* — calls the POST route in generate mode.
- *Replace…* — opens a dialog with a file input, reusing the `ImageCropDialog` flow from `CityForm` locked to 7:4, then calls the POST route in upload mode.

## Cost

$0.0336 per image with `gemini-3.1-flash-lite-image`.

| Scenario | Cost |
|---|---|
| All 11,734 existing subjects | ~$394 |
| 800–1,300 new subjects per month | $27–44 |

Note that a backfill through the batch API costs about a third of this (~$130).

## Backfill

The read route already fills in old subjects lazily, so a backfill is not strictly required. But given the low cost — ~$130 with the batch API — I think we should backfill all existing subjects anyway. Two reasons:

- Every subject page and landing card gets an image from day one, instead of the fallback placeholder lingering on rarely-visited subjects.
- It doubles as a backtest of the prompt: running it across all 11,734 subjects shows whether the style holds up for every kind of subject — infrastructure, budgets, zoning, personnel — before we commit to it, rather than discovering weak spots one subject at a time.

## Suggested style
Retro pixel art, like a calm scene from a 1990s console game. This is the prompt template's style wording:

> Clean pixel art in a retro 16-bit video-game aesthetic: crisp hard-edged sprites on a visible pixel grid, no anti-aliasing, no smooth gradients. Rich, vibrant, saturated palette of around 16–24 tones — deep teal and sky blue, warm terracotta and orange, bright leafy greens, magenta and coral accents, sunlit cream and sand. Bold flat shapes with minimal dithering, large simple cloud forms, big areas of open sky or plain wall, and detail concentrated only on the focal element. Side-view or slight isometric perspective, like a calm, uncluttered scene from a 1990s console game. Cheerful and lively in colour, minimal in content.

Example outputs:

Image
Image

## Future work

- **Styles.** Turn the prompt constant into a `STYLES` map and record the style on the object, so a later restyle can be scoped to the old images.
- **Size variants.** One file serves both surfaces today, so a 400px landing card downloads the full image. Store a smaller variant and add a `?w=` parameter.

pkakelas:
Closes #705

Every subject gets one AI-generated illustration. The landing cards, the subject page
and the city page's top subject show it. A subject with no image shows a placeholder
in its topic colour, which the browser draws.

## How it works

**The library** — `@opencouncil/subject-images` in `packages/subject-images`. It turns
a subject into an image and stores it: it builds the prompt, calls Gemini, converts the
reply to WebP 1344×768 with sharp, and reads and writes the object in DigitalOcean
Spaces. It imports no Next, no Prisma and no `env.mjs`. The caller passes the Spaces
client, the bucket, the folder, the public origin and the Gemini key. `toWebp` reads the
real format from the bytes and refuses anything that is not a JPEG, PNG, WebP, GIF or
AVIF, or that would decode past 16 times the canonical size. The object key accepts only
`[A-Za-z0-9_-]+` as an id. The size constants live in a `./constants` subpath, so a
client component can import them without sharp.

**One route** — `GET`/`POST /api/subject/{subjectId}/image`.

- `GET` redirects to the stored image on the CDN with a 5 minute public cache. The
  target URL carries the object's ETag, so a replaced image never serves stale.
- A `GET` that finds no object answers 404 with a 1 minute cache and schedules a
  generation with `after()`; the `` hides itself and the topic placeholder the
  browser drew under it stays. A bucket lookup error answers 404 without a generation
  and alerts Discord once per 10 minutes, so a bad preview config does not break every
  card or bill anything.
- `POST` is superadmin only and answers 404 for an unknown subject. `{"mode": "generate"}`
  redraws the subject over whatever is stored, and answers 409 when nothing was drawn. A
  multipart `file` replaces it: at most 5 MB, the type read from the bytes, a 400 with
  the reason when sharp cannot read it.

**The guards** live in Valkey (`CACHE_URL`), so they hold across containers and expire on
their own: one generation per subject at a time, a 10 minute back-off after a failure, a
subject given up on for 30 days after three failures in a row, and a 10 minute pause of
every generation after the bucket refused a write. A forced regenerate waits for a run in
progress, so its image lands last. A write that lands while the model draws, such as an
admin upload, is kept. Without `CACHE_URL` an in-process map guards one container.

There is no database column. The object key is `{SUBJECT_IMAGES_PREFIX}/{subjectId}.webp`,
`subject-images/8bit` by default. The folder names the style, so a later restyle writes a
sibling folder and leaves these objects in place.

## Where the images come from

1. **The pipeline.** `processAgenda` and `summarize` schedule the meeting's images with
   `after()` once they save the subjects, so the work outlives the callback response.
   The first run draws the agenda subjects days before the meeting. The second run draws the non-agenda subjects, which exist only
   once there is a transcript. A subject that already has an object is skipped, so the
   second run costs nothing for the first run's work.
2. **The read route.** Old subjects get an image as people open them.
3. **`scripts/backfill-subject-images.ts`**, which is optional. It takes the newest
   `--window` subjects the landing can show (released meetings, public cities, discussed
   subjects), skips the ones that have an image (one listing of the bucket folder says
   which), and ranks the rest in the landing page's order of importance through the
   shared ranker, so `--limit 50` draws the 50 undrawn subjects a visitor is most likely
   to see and a second run draws the next 50. Flags: `--city`, `--limit`, `--months`,
   `--window` (default 2000), `--concurrency` (default 3), `--force`.

## Surfaces

- **Landing card** follows the design in the issue. The image is the hero. The topic pill
  and the city logo sit on it. The location and the title sit on a gradient at its foot.
  The facts, the description, the link and an "AI-generated image" label come below. On
  the map preview the logo sits left of the close button.
- **Subject page** gets a wide header image. A superadmin sees a hover overlay with
  Regenerate and Replace; while hidden it takes no pointer events.
- **Search rows and the city page's hot topics** show the image behind a scrim in the
  card's own colour. Those rows no longer draw speakers, so the search results fetch no
  council rosters and the city page ships none. The `person_opened` event with
  `surface: 'avatar_stack'` no longer fires from the city page; nothing replaces it.
- New strings in all four locales.

## Configuration

| Variable | Meaning | Required | Default |
|---|---|---|---|
| `GEMINI_API_KEY` | Gemini key. When unset, no image is generated and every subject shows the placeholder. | No | - |
| `SUBJECT_IMAGES_PREFIX` | Folder inside `DO_SPACES_BUCKET`. The folder names the style. | No | `subject-images/8bit` |
| `CACHE_URL` | Valkey, for the guards above. When unset, one container guards its own runs and the back-off does not apply. | No | - |

Each environment writes to its own `DO_SPACES_BUCKET` and reads through its own
`CDN_URL`. A preview with no key spends nothing. Cost is about $0.034 per image.

## Tests

Seven new test files cover the library's prompt, generation and store, the route, the
service, the client placeholder and the backfill ranker. Two existing task tests assert
that the triggers are scheduled with `after()`.

## Docs

`docs/guides/subject-images.md` follows `docs/guides/framework.md`, with a sequence
diagram and the component pointers. `docs/environment-variables.md` and `.env.example`
list the two variables.

<h2><a href="https://app.greptile.com/api/retrigger?id=59998465"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Flib%2FsubjectImages.ts%3Aundefined-224%0AIf%20Valkey%20is%20unavailable%20while%20unauthenticated%20requests%20for%20an%20undrawn%20subject%20reach%20separate%20application%20processes%2C%20%60cacheAcquire%60%20reports%20success%20to%20each%20process%20and%20the%20process-local%20guard%20cannot%20deduplicate%20them%2C%20causing%20duplicate%20billed%20Gemini%20calls%20and%20storage%20writes.%20**How%20this%20was%20verified%3A**%20The%20public%20GET%20reaches%20%60claimAndGenerate%60%2C%20while%20%60cacheAcquire%60%20returns%20true%20without%20a%20Valkey%20client%20or%20after%20a%20Valkey%20error.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=711&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;Security&nbsp;**Shared generation claim fails open** <a href="https://github.com/schemalabz/opencouncil/pull/711#discussion_r3990222672">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
src/lib/subjectImages.ts:undefined-224
If Valkey is unavailable while unauthenticated requests for an undrawn subject reach separate application processes, `cacheAcquire` reports success to each process and the process-local guard cannot deduplicate them, causing duplicate billed Gemini calls and storage writes. **How this was verified:** The public GET reaches `claimAndGenerate`, while `cacheAcquire` returns true without a Valkey client or after a Valkey error.

---

`````

</details>

<h3>Summary</h3>

- Introduces the reusable subject-image generation and storage package.
- Adds API, Valkey coordination, backfill, and meeting-pipeline integration.
- Adds responsive placeholders, administrative controls, localized disclosures, documentation, and tests.

<sub>Reviews (10) · Last reviewed commit: ["feat(nav): draw the MCP link with a plug..."](https://github.com/schemalabz/opencouncil/commit/84569817ecb1d01479c837db275082b31db48a07)</sub>

### DIFF
--- a/packages/subject-images/package.json
+++ b/packages/subject-images/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "@opencouncil/subject-images",
+  "version": "0.0.0",
+  "private": true,
+  "exports": {
+    ".": "./src/index.ts",
+    "./constants": "./src/constants.ts",
+    "./store": "./src/store.ts",
+    "./image": "./src/image.ts"
+  },
+  "dependencies": {
+    "@aws-sdk/client-s3": "^3.1060.0",
+    "@google/genai": "^2.20.0",
+    "sharp": "^0.35.4"
+  }
+}
--- a/packages/subject-images/src/store.ts
+++ b/packages/subject-images/src/store.ts
@@ -1,0 +1,116 @@
+import { HeadObjectCommand, ListObjectsV2Command, PutObjectCommand, type S3Client } from '@aws-sdk/client-s3';
+
+/** What may name an object: cuids, and nothing that could leave the folder. */
+const SUBJECT_ID = /^[A-Za-z0-9_-]+$/;
+
+export function isSubjectImageId(subjectId: string): boolean {
+    return SUBJECT_ID.test(subjectId);
+}
+
+/**
+ * Only this module knows where a subject's image lives. The prefix names the
+ * style (`subject-images/8bit`): a later restyle writes a sibling folder and
+ * leaves these.
+ *
+ * The id is checked here and not only at the route, because this key is also
+ * what an upload writes: `../../logo` must never become one.
+ */
+export function objectKey(subjectId: string, prefix: string): string {
+    if (!isSubjectImageId(subjectId)) {
+        throw new Error(`Not a subject image id: ${JSON.stringify(subjectId)}`);
+    }
+    return `${folder(prefix)}${subjectId}.webp`;
+}
+
+/** The prefix as the folder its objects sit in: no leading slash, one trailing slash. */
+function folder(prefix: string): string {
+    return `${prefix.replace(/^\/+|\/+$/g, '')}/`;
+}
+
+export interface StoreDeps {
+    client: S3Client;
+    bucket: string;
+    /** Folder inside the bucket, e.g. `subject-images/8bit`. */
+    prefix: string;
+}
+
+export interface ResolveDeps extends StoreDeps {
+    /** Public origin the objects are served from, without a trailing slash (the CDN, or the bucket itself). */
+    publicBaseUrl: string;
+}
+
+export interface ResolvedImage {
+    url: string;
+}
+
+type S3Error = { name?: string; $metadata?: { httpStatusCode?: number } };
+
+function asS3Error(error: unknown): S3Error | null {
+    return typeof error === 'object' && error !== null ? (error as S3Error) : null;
+}
+
+/**
+ * The S3 errors that mean "no such object". HeadObject answers with no body,
+ * so on that path the status is all there is to read; the names cover the
+ * commands that do send one. A missing bucket also answers 404 to HeadObject
+ * and cannot be told apart here; it surfaces at the PutObject, as NoSuchBucket.
+ */
+export function isS3NotFound(error: unknown): boolean {
+    const e = asS3Error(error);
+    return e?.name === 'NotFound' || e?.name === 'NoSuchKey' || e?.$metadata?.httpStatusCode === 404;
+}
+
+/**
+ * Look up a subject's image. `null` means "no image"; the caller picks the
+ * fallback. The ETag rides along as a cache-buster, so a replaced image gets
+ * a new URL and no CDN or browser cache keeps serving the old one.
+ */
+export async function resolve(subjectId: string, deps: ResolveDeps): Promise<ResolvedImage | null> {
+    try {
+        const key = objectKey(subjectId, deps.prefix);
+        const head = await deps.client.send(new HeadObjectCommand({ Bucket: deps.bucket, Key: key }));
+        const version = head.ETag?.replace(/"/g, '');
+        // A no-op for the ids the guard admits; it keeps the URL right if that pattern ever widens.
+        const path = key.split('/').map(encodeURIComponent).join('/');
+        const url = `${deps.publicBaseUrl}/${path}${version ? `?v=${version}` : ''}`;
+        return { url };
+    } catch (error) {
+        if (isS3NotFound(error)) return null;
+        // A key that may read and write objects but not list the bucket gets
+        // 403, not 404, for an object that is not there. Read it as a miss, so
+        // the feature works on such a key; a key that is wrong altogether then
+        // fails at the PutObject, where the failure is alerted. Everything
+        // else (301 for a wrong region, 5xx, no status for a network error) is
+        // a bucket that cannot be read, and the caller must not draw against it.
+        if (asS3Error(error)?.$metadata?.httpStatusCode === 403) return null;
+        throw error;
+    }
+}
+
+/** The subjects that have an object under the prefix: one request per 1,000 keys. */
+export async function listStoredSubjectIds(deps: StoreDeps): Promise<Set<string>> {
+    const prefix = folder(deps.prefix);
+    const ids = new Set<string>();
+    let token: string | undefined;
+    do {
+        const page = await deps.client.send(new ListObjectsV2Command({ Bucket: deps.bucket, Prefix: prefix, ContinuationToken: token }));
+        for (const object of page.Contents ?? []) {
+            const id = object.Key?.slice(prefix.length).replace(/\.webp$/, '');
+            if (id && isSubjectImageId(id) && object.Key !== `${prefix}${id}`) ids.add(id);
+        }
+        token = page.IsTruncated ? page.NextContinuationToken : undefined;
+    } while (token);
+    return ids;
+}
+
+/** Write the subject's image. Overwrites whatever was there. */
+export async function store(subjectId: string, image: Buffer, deps: StoreDeps): Promise<void> {
+    await deps.client.send(new PutObjectCommand({
+        Bucket: deps.bucket,
+        Key: objectKey(subjectId, deps.prefix),
+        Body: image,
+        ContentType: 'image/webp',
+        ACL: 'public-read',
+        CacheControl: 'public, max-age=31536000, immutable',
+    }));
+}
--- a/src/app/api/subject/[subjectId]/image/route.ts
+++ b/src/app/api/subject/[subjectId]/image/route.ts
@@ -1,0 +1,125 @@
+import { NextRequest, NextResponse, after } from 'next/server';
+import { UnreadableImageError, isSubjectImageId } from '@opencouncil/subject-images';
+import { subjectExists } from '@/lib/db/subject';
+import { getCurrentUser } from '@/lib/auth';
+import { MAX_IMAGE_BYTES } from '@/lib/utils/imageUpload';
+import {
+    generateImageForSubject,
+    generateImageForSubjectInBackground,
+    isSubjectImageGenerationEnabled,
+    reportLookupFailure,
+    resolveSubjectImage,
+    storeSubjectImage,
+} from '@/lib/subjectImages';
+
+/**
+ * A subject's illustration.
+ *
+ * GET redirects to the stored image, or answers 404 and asks for the image in
+ * the background, so old subjects get theirs as people look at them. The
+ * <img> that got the 404 hides itself and leaves the topic placeholder the
+ * browser drew under it. The redirect is cacheable for a few minutes: Cloudflare
+ * then absorbs the repeat loads a landing page of cards makes, and the ETag in
+ * the target URL keeps a replaced image from ever being served stale. Admins
+ * bust the browser copy with a query parameter after they change an image.
+ */
+
+const REDIRECT_CACHE = 'public, max-age=300, s-maxage=300';
+const MISS_CACHE = 'public, max-age=60, s-maxage=60';
+
+type RouteContext = { params: Promise<{ subjectId: string }> };
+
+/** The id names an object in the bucket before it is looked up, so it is checked before anything else. */
+function invalidId(subjectId: string) {
+    return isSubjectImageId(subjectId) ? null : NextResponse.json({ error: 'Invalid subject id' }, { status: 400 });
+}
+
+export async function GET(_request: NextRequest, context: RouteContext) {
+    const { subjectId } = await context.params;
+    const invalid = invalidId(subjectId);
+    if (invalid) return invalid;
+
+    // A bucket that cannot be reached (bad credentials on a preview, an outage)
+    // must not break every card on the page: answer as a miss, and do not
+    // queue a generation that would fail at the same store.
+    let lookupFailed = false;
+    let resolved = null;
+    try {
+        resolved = await resolveSubjectImage(subjectId);
+    } catch (error) {
+        lookupFailed = true;
+        await reportLookupFailure(subjectId, error);
+    }
+    if (resolved) {
+        return NextResponse.redirect(resolved.url, { status: 302, headers: { 'Cache-Control': REDIRECT_CACHE } });
+    }
+
+    if (!(await subjectExists(subjectId))) {
+        return NextResponse.json({ error: 'Subject not found' }, { status: 404 });
+    }
+
+    if (!lookupFailed && isSubjectImageGenerationEnabled()) {
+        after(() => generateImageForSubjectInBackground(subjectId));
+    }
+
+    return NextResponse.json({ error: 'No image yet' }, { status: 404, headers: { 'Cache-Control': MISS_CACHE } });
+}
+
+export async function POST(request: NextRequest, context: RouteContext) {
+    const user = await getCurrentUser();
+    if (!user?.isSuperAdmin) {
+        return NextResponse.json({ error: 'Unauthorized' }, { status: 403 });
+    }
+
+    const { subjectId } = await context.params;
+    const invalid = invalidId(subjectId);
+    if (invalid) return invalid;
+    // Either branch writes a permanent, public object under this id.
+    if (!(await subjectExists(subjectId))) {
+        return NextResponse.json({ error: 'Subject not found' }, { status: 404 });
+    }
+    const contentType = request.headers.get('content-type') ?? '';
+
+    if (contentType.includes('multipart/form-data')) {
+        const file = (await request.formData()).get('file');
+        if (!(file instanceof File)) {
+            return NextResponse.json({ error: 'No file provided' }, { status: 400 });
+        }
+        if (file.size > MAX_IMAGE_BYTES) {
+            return NextResponse.json({ error: `Image exceeds ${MAX_IMAGE_BYTES / (1024 * 1024)} MB` }, { status: 400 });
+        }
+        // The declared type is whatever the client set. The library reads the
+        // real one from the bytes, so nothing that is not a raster reaches sharp.
+        try {
+            await storeSubjectImage(subjectId, Buffer.from(await file.arrayBuffer()));
+        } catch (error) {
+            if (error instanceof UnreadableImageError) {
+                return NextResponse.json({ error: error.message }, { status: 400 });
+            }
+            const message = error instanceof Error ? error.message : 'Image upload failed';
+            return NextResponse.json({ error: message }, { status: 500 });
+        }
+        return NextResponse.json({ ok: true, mode: 'upload' });
+    }
+
+    const body = await request.json().catch(() => ({}));
+    if (body?.mode !== 'generate') {
+        return NextResponse.json({ error: 'Expected { mode: "generate" } or a multipart file upload' }, { status: 400 });
+    }
+    if (!isSubjectImageGenerationEnabled()) {
+        return NextResponse.json({ error: 'Image generation is not configured (GEMINI_API_KEY)' }, { status: 503 });
+    }
+
+    try {
+        const outcome = await generateImageForSubject(subjectId, { force: true });
+        // A forced run only ever reports 'generated' or throws; anything else
+        // means nothing was drawn, and the admin must not be told otherwise.
+        if (outcome !== 'generated') {
+            return NextResponse.json({ error: `Image was not regenerated (${outcome})`, outcome }, { status: 409 });
+        }
+        return NextResponse.json({ ok: true, mode: 'generate', outcome });
+    } catch (error) {
+        const message = error instanceof Error ? error.message : 'Image generation failed';
+        return NextResponse.json({ error: message }, { status: 500 });
+    }
+}
--- a/src/lib/db/subject.ts
+++ b/src/lib/db/subject.ts
@@ -718,3 +718,35 @@ export async function getUtterancesForSubject(subjectId: string) {
 
     return utterances;
 }
+
+/** What the illustration prompt is built from. `null` when the subject does not exist. */
+export async function getSubjectPromptInput(subjectId: string) {
+    return prisma.subject.findUnique({
+        where: { id: subjectId },
+        select: { name: true, description: true },
+    });
+}
+
+export async function subjectExists(subjectId: string): Promise<boolean> {
+    const row = await prisma.subject.findUnique({ where: { id: subjectId }, select: { id: true } });
+    return row !== null;
+}
+
+export async function getSubjectIdsForMeeting(cityId: string, councilMeetingId: string): Promise<string[]> {
+    const rows = await prisma.subject.findMany({
+        where: { cityId, councilMeetingId },
+        select: { id: true },
+    });
+    return rows.map((row) => row.id);
+}
+
+/** Newest first, optionally one city, optionally capped — the order the image backfill walks in. */
+export async function getSubjectIdsNewestFirst(options: { cityId?: string; limit?: number } = {}): Promise<string[]> {
+    const rows = await prisma.subject.findMany({
+        where: options.cityId ? { cityId: options.cityId } : undefined,
+        select: { id: true },
+        orderBy: { createdAt: 'desc' },
+        take: options.limit,
+    });
+    return rows.map((row) => row.id);
+}
--- a/src/lib/subjectImages.ts
+++ b/src/lib/subjectImages.ts
@@ -1,0 +1,265 @@
+import 'server-only';
+import { buildPrompt, generate, listStoredSubjectIds, resolve, store, toWebp, type ResolvedImage } from '@opencouncil/subject-images';
+import { env } from '@/env.mjs';
+import { cacheAcquire, cacheDelete, cacheGetJSON, cacheSetJSON } from '@/lib/cache/valkey';
+import { getSubjectIdsForMeeting, getSubjectPromptInput } from '@/lib/db/subject';
+import { stripMarkdown } from '@/lib/formatters/markdown';
+import { s3Client } from '@/lib/s3';
+import { sendErrorAdminAlert } from '@/lib/discord-core';
+
+/**
+ * The app-side face of `@opencouncil/subject-images`: wires the library to the
+ * Spaces client, the bucket and the Gemini key, and owns the guards a
+ * fire-and-forget pipeline needs — no duplicate generations for one subject,
+ * no retry storm for a subject Gemini keeps refusing, and no generation that
+ * paints over what an admin uploaded while the model drew.
+ *
+ * The guards are markers in Valkey, so they hold across containers and expire
+ * on their own: the claim of a container that died mid-run releases with its
+ * TTL, and a refused subject is left alone for a month rather than retried
+ * every ten minutes for the life of a process. Without CACHE_URL (dev,
+ * previews) there are no markers, and the in-process map below is the only
+ * in-flight guard — enough for one container.
+ */
+
+/** How long a failed subject waits before a page view may retry it. */
+const RETRY_AFTER_MS = 10 * 60 * 1000;
+/** Failures in a row before a subject is given up on; the prompt refuses some subjects for good. */
+const GIVE_UP_AFTER = 3;
+/** How long the failure count is kept, so the retries can add up to a give-up. */
+const FAILURE_TTL_S = 24 * 60 * 60;
+/** How long a given-up subject is left alone. */
+const GIVE_UP_TTL_S = 30 * 24 * 60 * 60;
+/** Longer than a generation takes; releases the claim of a container that died mid-run. */
+const IN_FLIGHT_TTL_S = 3 * 60;
+/** How often a forced run asks again for a claim another container holds. */
+const CLAIM_POLL_MS = 1000;
+/** One Discord alert per window for a bucket that cannot be read, not one per card per page view. */
+const LOOKUP_ALERT_TTL_S = 10 * 60;
+/** How long every generation pauses after the bucket refused a write: one billed call and one alert, not one per subject. */
+const STORE_BROKEN_TTL_S = 10 * 60;
+/** Generations run at once when a meeting's subjects are processed together. */
+const MEETING_CONCURRENCY = 3;
+
+/** This process's runs; the Valkey claim is what the other containers see. */
+const inFlight = new Map<string, Promise<GenerateOutcome>>();
+
+type FailureRecord = { count: number; at: number };
+
+const inFlightKey = (subjectId: string) => `subject-images:in-flight:${subjectId}`;
+const failureKey = (subjectId: string) => `subject-images:failed:${subjectId}`;
+const LOOKUP_ALERT_KEY = 'subject-images:lookup-alerted';
+const STORE_BROKEN_KEY = 'subject-images:store-broken';
+
+export type GenerateOutcome =
+    | 'generated'
+    /** an object was already there, and the run was not forced */
+    | 'exists'
+    /** something else wrote the object while the model drew; that write is kept */
+    | 'superseded'
+    | 'in-flight'
+    | 'recent-failure'
+    /** failed GIVE_UP_AFTER times in a row; left alone until the record expires */
+    | 'refused'
+    /** the bucket refused a write minutes ago; nothing is drawn until that pause ends */
+    | 'store-unavailable'
+    | 'disabled';
+
+function storeDeps() {
+    return { client: s3Client, bucket: env.DO_SPACES_BUCKET, prefix: env.SUBJECT_IMAGES_PREFIX };
+}
+
+function resolveDeps() {
+    return { ...storeDeps(), publicBaseUrl: env.CDN_URL.replace(/\/+$/, '') };
+}
+
+function sleep(ms: number): Promise<void> {
+    return new Promise((done) => setTimeout(done, ms));
+}
+
+/** Log and alert Discord; the alert's own failure must not surface. */
+function reportFailure(message: string, error: unknown, context: Record<string, string>): void {
+    console.error(`${message}:`, error);
+    const detail = error instanceof Error ? error.message : String(error);
+    sendErrorAdminAlert({ source: 'subject-images', error: detail, context })
+        .catch((alertError) => console.error('Failed to send subject image alert:', alertError));
+}
+
+/** Generation is on wherever a Gemini key is set; previews and dev without one serve only misses. */
+export function isSubjectImageGenerationEnabled(): boolean {
+    return Boolean(env.GEMINI_API_KEY);
+}
+
+export function resolveSubjectImage(subjectId: string): Promise<ResolvedImage | null> {
+    return resolve(subjectId, resolveDeps());
+}
+
+/**
+ * Store an admin-supplied image, normalised to the canonical WebP. Throws
+ * UnreadableImageError for bytes that are not a raster or that would decode
+ * past the pixel limit; the route answers 400 for that.
+ */
+export async function storeSubjectImage(subjectId: string, image: Buffer): Promise<void> {
+    await store(subjectId, await toWebp(image), storeDeps());
+}
+
+/** Every subject that has an image, for the backfill to skip. One request per 1,000 objects. */
+export function listSubjectsWithImages(): Promise<Set<string>> {
+    return listStoredSubjectIds(storeDeps());
+}
+
+/**
+ * The bucket could not be read (bad credentials, a wrong region, an outage).
+ * The read route serves misses meanwhile; this makes sure someone hears about
+ * it — once per window, not once per card per page view.
+ */
+export async function reportLookupFailure(subjectId: string, error: unknown): Promise<void> {
+    if (await cacheAcquire(LOOKUP_ALERT_KEY, LOOKUP_ALERT_TTL_S)) {
+        reportFailure(`Subject image lookup failed for ${subjectId}`, error, { subjectId });
+    } else {
+        console.error(`Subject image lookup failed for ${subjectId}:`, error);
+    }
+}
+
+/** Why a subject must not be drawn now, or null when it may be. */
+async function heldBack(subjectId: string): Promise<'recent-failure' | 'refused' | 'store-unavailable' | null> {
+    if (await cacheGetJSON(STORE_BROKEN_KEY)) return 'store-unavailable';
+    const failure = await cacheGetJSON<FailureRecord>(failureKey(subjectId));
+    if (!failure) return null;
+    if (failure.count >= GIVE_UP_AFTER) return 'refused';
+    return Date.now() - failure.at < RETRY_AFTER_MS ? 'recent-failure' : null;
+}
+
+async function recordFailure(subjectId: string, error: unknown): Promise<void> {
+    const previous = await cacheGetJSON<FailureRecord>(failureKey(subjectId));
+    const count = (previous?.count ?? 0) + 1;
+    const gaveUp = count >= GIVE_UP_AFTER;
+    await cacheSetJSON(failureKey(subjectId), { count, at: Date.now() }, gaveUp ? GIVE_UP_TTL_S : FAILURE_TTL_S);
+    const message = gaveUp
+        ? `Subject image generation failed for ${subjectId}; giving up after ${count} attempts`
+        : `Subject image generation failed for ${subjectId} (attempt ${count} of ${GIVE_UP_AFTER})`;
+    reportFailure(message, error, { subjectId, attempt: String(count) });
+}
+
+async function generateAndStore(subjectId: string, apiKey: string, force: boolean): Promise<GenerateOutcome> {
+    try {
+        const before = await resolveSubjectImage(subjectId);
+        if (before && !force) return 'exists';
+
+        const subject = await getSubjectPromptInput(subjectId);
+        if (!subject) throw new Error(`Subject ${subjectId} not found`);
+
+        // Descriptions carry markdown and REF:UTTERANCE links; the model should
+        // read the sentence, not the markup.
+        const prompt = buildPrompt({ title: stripMarkdown(subject.name), description: stripMarkdown(subject.description) });
+        const image = await generate(prompt, { apiKey });
+
+        // The model takes tens of seconds. An admin who uploaded meanwhile keeps
+        // their file: the ETag in the URL says whether the object changed. The
+        // window between this check and the write is milliseconds, not seconds.
+        const after = await resolveSubjectImage(subjectId);
+        if (after?.url !== before?.url) return 'superseded';
+
+        try {
+            await store(subjectId, image, storeDeps());
+        } catch (error) {
+            // A bucket that refuses one write (missing, wrong key) refuses them
+            // all. HeadObject cannot tell a missing bucket from a missing object,
+            // so this is where it shows; pausing here keeps a misconfiguration
+            // at one billed call and one alert instead of one per subject.
+            await cacheSetJSON(STORE_BROKEN_KEY, { at: Date.now() }, STORE_BROKEN_TTL_S);
+            throw error;
+        }
+        await cacheDelete(failureKey(subjectId));
+        return 'generated';
+    } catch (error) {
+        await recordFailure(subjectId, error);
+        throw error;
+    }
+}
+
+/**
+ * Generate and store the image for one subject.
+ *
+ * Without `force` an existing object is kept, so the summarize run costs
+ * nothing for what the agenda run already drew; a subject already being drawn,
+ * one that failed in the last ten minutes, and one given up on are reported
+ * rather than drawn. With `force` the object is replaced, which is also how an
+ * admin undoes a manual upload; a forced run waits for any run in progress,
+ * here or in another container, so the forced image is the one that lands
+ * last. Either way a write that landed while the model drew is kept. Throws on
+ * failure after alerting Discord; background callers catch, the admin route
+ * turns it into a 500.
+ */
+export async function generateImageForSubject(subjectId: string, options: { force?: boolean } = {}): Promise<GenerateOutcome> {
+    const apiKey = env.GEMINI_API_KEY;
+    if (!apiKey) return 'disabled';
+    const force = Boolean(options.force);
+
+    let running = inFlight.get(subjectId);
+    if (running && !force) return 'in-flight';
+    while (running) {
+        await running.catch(() => undefined);
+        running = inFlight.get(subjectId);
+    }
+
+    // Registered before the first await, so two calls in one tick cannot both
+    // pass the check above — the only guard there is without CACHE_URL.
+    const run = claimAndGenerate(subjectId, apiKey, force);
+    inFlight.set(subjectId, run);
+    try {
+        return await run;
+    } finally {
+        if (inFlight.get(subjectId) === run) inFlight.delete(subjectId);
+    }
+}
+
+/** The back-off check and the cross-container claim, then the run itself. */
+async function claimAndGenerate(subjectId: string, apiKey: string, force: boolean): Promise<GenerateOutcome> {
+    if (!force) {
+        const held = await heldBack(subjectId);
+        if (held) return held;
+    }
+
+    while (!(await cacheAcquire(inFlightKey(subjectId), IN_FLIGHT_TTL_S))) {
+        if (!force) return 'in-flight';
+        await sleep(CLAIM_POLL_MS);
+    }
+
+    try {
+        return await generateAndStore(subjectId, apiKey, force);
+    } finally {
+        await cacheDelete(inFlightKey(subjectId));
+    }
+}
+
+/** Fire-and-forget variant for the read route: never rejects. */
+export function generateImageForSubjectInBackground(subjectId: string): Promise<void> {
+    return generateImageForSubject(subjectId).then(() => undefined, () => undefined);
+}
+
+/**
+ * Generate images for every subject of a meeting, a few at a time. Subjects
+ * that already have an object are skipped inside generateImageForSubject.
+ * Never rejects: the callers do not await it, so a rejection here would be
+ * an unhandled one. Each failure, the subject lookup included, is alerted on
+ * its own.
+ */
+export async function generateImagesForMeeting(cityId: string, councilMeetingId: string): Promise<void> {
+    if (!isSubjectImageGenerationEnabled()) return;
+
+    let queue: string[];
+    try {
+        queue = await getSubjectIdsForMeeting(cityId, councilMeetingId);
+    } catch (error) {
+        reportFailure(`Subject lookup for images failed for meeting ${councilMeetingId}`, error, { cityId, councilMeetingId });
+        return;
+    }
+
+    const worker = async () => {
+        for (let id = queue.shift(); id !== undefined; id = queue.shift()) {
+            await generateImageForSubjectInBackground(id);
+        }
+    };
+    await Promise.all(Array.from({ length: Math.min(MEETING_CONCURRENCY, queue.length) }, worker));
+}
--- a/src/lib/tasks/processAgenda.ts
+++ b/src/lib/tasks/processAgenda.ts
@@ -5,6 +5,8 @@ import prisma from "../db/prisma";
 import { revalidateMeeting } from "../cache";
 import { saveSubjectsForMeeting } from "../db/utils";
 import { withUserAuthorizedToEdit } from "../auth";
+import { after } from "next/server";
+import { generateImagesForMeeting } from "../subjectImages";
 import { requestProcessAgendaInternal } from "./processAgendaInternal";
 
 /**
@@ -87,6 +89,12 @@ export async function handleProcessAgendaResult(taskId: string, response: Proces
     // (~500ms/recipient), so revalidating only after it finishes would let early
     // recipients open the meeting and see stale content.
     revalidateMeeting(task.councilMeeting.cityId, task.councilMeeting.id);
+
+    // Illustrations for the agenda subjects, days before the meeting. Through
+    // after(): the callback must not wait on Gemini, and minutes of Gemini
+    // calls must outlive the response rather than escape it. Each failure
+    // alerts on its own.
+    after(() => generateImagesForMeeting(task.councilMeeting.cityId, task.councilMeeting.id));
 
     // Create notifications if administrative body allows it
     const adminBody = task.councilMeeting.administrativeBody;
--- a/.env.example
+++ b/.env.example
@@ -33,6 +33,10 @@ NEXTAUTH_URL=http://localhost:3000
 RESEND_API_KEY=your_resend_api_key
 ANTHROPIC_API_KEY=your_anthropic_api_key
 GOOGLE_API_KEY=your_google_api_key
+# Optional: Gemini key for subject illustrations. Unset = no generation, fallback images only.
+GEMINI_API_KEY=
+# Folder inside DO_SPACES_BUCKET for the subject illustrations. The folder names the style.
+SUBJECT_IMAGES_PREFIX=subject-images/8bit
 TASK_API_URL=http://localhost:3005
 TASK_API_KEY=your_task_api_key
 ELASTICSEARCH_URL=http://localhost:9200
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN apk add --no-cache openssl openssl-dev
 # Workspace member manifests must be present before npm install, or npm
 # resolves the workspace-shaped lockfile against missing directories.
 COPY package*.json .npmrc ./
+COPY packages/subject-images/package.json packages/subject-images/
 COPY packages/ui/package.json packages/ui/
 COPY services/notis/package.json services/notis/
 
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -12,6 +12,7 @@ RUN apk add --no-cache openssl openssl-dev
 # Workspace member manifests must be present before npm install, or npm
 # resolves the workspace-shaped lockfile against missing directories.
 COPY package*.json .npmrc ./
+COPY packages/subject-images/package.json packages/subject-images/
 COPY packages/ui/package.json packages/ui/
 COPY services/notis/package.json services/notis/
 
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -102,6 +102,8 @@ Copy the output and set it as your `NEXTAUTH_SECRET` in your `.env` file.
 | Variable | Description | Required | Default |
 |----------|-------------|----------|---------|
 | `ANTHROPIC_API_KEY` | API key for Claude. | Yes | - |
+| `GEMINI_API_KEY` | Gemini key for the subject illustrations. When unset, no image is generated and subjects show the topic-coloured fallback. | No | - |
+| `SUBJECT_IMAGES_PREFIX` | Folder inside `DO_SPACES_BUCKET` that holds the subject illustrations. The folder names the style. | No | `subject-images/8bit` |
 
 ### Search Configuration
 | Variable | Description | Required | Default |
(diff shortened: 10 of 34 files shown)

### CODE REVIEW

**Comment by kouloumos** [comment]
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/5071d1d75af410f6e5c324703618a73b69506974..db547a585cd86408a2306ddbb8822a919a271c63) to rebase this onto `main`, and I moved the base from `feature/city-page-redesign` to `main`.

The branch carried its own rebased copy of the city page redesign, not the copy that `main` merged. It forked at `f685c505` and it never picked up the last commits of `feature/city-page-redesign`. I reset to `main` and cherry-picked the four `subject-images` commits.

Two of the missing commits rewrite `src/components/meetings/subject/subject.tsx`, so `feat(subject-images): show the illustration on cards and pages` conflicted there. I resolved it in favour of main's direction. That changes your version in three places:

- **The `<nav>` context row is gone.** `cb0dd86c` deleted it. The date it carried now sits in the banner meta row.
- **The administrative body is out of the banner.** `c932a20e` removed it as the third copy inside the first 116px of a phone screen. The header breadcrumb still names it.
- **The phone decision strip uses `DecisionCard`.** `cb0dd86c` replaced the hand-rolled markup with the shared component, `collapsible` and `lg:hidden`. The rail's decision card changed the same way, and that one applied without a conflict.

The illustration work itself is unchanged: the `SubjectImage` banner, `SubjectImageAdminControls`, the `imageVersion` cache-buster, the extracted `actions` const, and the two-column restructure.

`package.json` also conflicted. `main` bumped `@mux/mux-video` and `@mux/mux-video-react` to `^0.31.2`. I kept those versions and added `@opencouncil/subject-images`.

`npx tsc --noEmit` passes, and `npm test` passes with 172 suites and 2153 tests.

Please check the banner on a preview. Its meta row now reads agenda number, date, then speakers and minutes. I did not verify that in a browser.

The branch state before the rebase is at [`backup/pr-711-original`](https://github.com/kouloumos/opencouncil/tree/backup/pr-711-original).

File: src/lib/topicStyle.ts [hunk 1]
```diff
@@ -68,3 +68,28 @@ export function topicStyle(color?: string | null, variant: 'soft' | 'solid' = 's
         icon: `color-mix(in srgb, ${c} 65%, black)`,
     };
 }
+
+/** `color-mix(in srgb, hex weight, other)` as a literal hex, for surfaces that cannot run CSS. */
+export function mixHex(hex: string, other: string, weight: number): string {
+    const parse = (h: string) => {
+        const c = h.replace('#', '');
+        return [0, 2, 4].map((i) => parseInt(c.slice(i, i + 2), 16));
```
kouloumos: `parse` slices at fixed offsets 0/2/4 and never expands the 3-digit form, so byte 4 of `"abc"` is `""` and `parseInt("", 16)` is `NaN`:

```
mixHex("#abc", "#ffffff", 0.24) => "#ebc5NaN"
mixHex("#f00", "#ffffff", 0.24) => "#fbc2NaN"
```

`subjectImageFallbackSvg` puts that straight into the served SVG. I checked what a browser does with the result: `fill="#ebc5NaN"` computes to `rgb(0, 0, 0)` and `stroke="#6f08NaN"` computes to `none`. So a subject under such a topic serves a **solid black 1344×768 placeholder with no icon**, on the landing card, the subject header and every row surface.

3-digit hex is an accepted topic colour rather than a hypothetical. `HEX_REGEX` in `src/lib/utils/colorSuggestion.ts:6` allows `{3}`, `src/lib/zod-schemas/topic.ts:10` is the only write-side validation, and `src/lib/zod-schemas/__tests__/topic.test.ts:55` is titled "should accept 3-character hex colors".

**It is not live today** — I read all 15 topics from `/api/topics` and every one is 6-digit. So this is one topic edit away rather than currently broken.

`contrastText`, forty lines above in this file, already guards `c.length < 6`. `hexToRgb` in `colorSuggestion.ts:10` is better still: it expands the shorthand and returns `null` on invalid input. It is module-private today, so this means exporting it — which also removes the fourth copy of hex parsing in the repo.

src/components/subject/SubjectRow.tsx
```diff
@@ -134,7 +132,19 @@ export function SubjectRow({ subject, city, meeting, persons, showContext = true
                     <div className="absolute inset-y-0 left-0 w-1" style={{ backgroundColor: rail }} />
                 </div>
 
-                <div className="flex flex-col gap-3 py-3 pl-4 pr-3 sm:flex-row sm:items-stretch sm:gap-6 sm:py-4 sm:pl-5 sm:pr-4">
+                {/* The illustration fills the row, faded by a scrim in the card's own
+                    colour, as on the city page's hot topics. The scrim, rather than an
+                    opacity on the image, is what gives the text one contrast ratio
+                    against every picture the model draws: an opacity bleaches the light
+                    images and leaves the dark ones heavy. It thins on hover. */}
+                <span className="absolute inset-0 z-0" aria-hidden>
```
kouloumos: This span and the scrim below it paint over the topic rail, so the rail is invisible on every row.

The rail wrapper is the first child and carries no `z-index`. This span and the scrim are later siblings at `z-0`. All three are positioned, so they paint in the same step in tree order, and the later two win.

Measured on `/en/athens/aug26_2026`:

```
rail        rgb(96,125,139)  x=394  w=4
image span                   x=394  w=894
railWrap z-index = auto   (child 0)
imgSpan  z-index = 0      (child 1)
scrim    z-index = 0      (child 2)
```

The illustration is an opaque WebP, so once it loads the rail is gone entirely. Before it loads the 80% white scrim already reduces it to a fifth of its strength.

To see it: clone a row, delete the image span and the scrim span, and compare the left 30px magnified. The rail appears as a clear vertical bar in the second copy and not in the first.

The comment at lines 120-126 explains why the rail has to reach the card outline. That work is currently unreachable. `z-10` on the rail wrapper fixes it.

src/components/cities/overview/HotTopicRow.tsx
```diff
@@ -92,28 +79,40 @@ export function HotTopicFacts({ card, timezone, locale, size }: {
...
+            <span className="absolute inset-0 z-0" aria-hidden>
+                <SubjectImage subjectId={subject.id} alt="" />
+            </span>
+            {/* The scrim is what keeps the text legible against every image the model
+                draws, bright or dark: one wash in the card's own colour, rather than
+                an opacity on the image, which would wash the light images out and
+                leave the dark ones opaque. It thins on hover, which is the row's only
+                affordance now that it carries no hover tint of its own. */}
+            <span
+                className="absolute inset-0 z-0 bg-card/80 transition-colors duration-200 group-hover:bg-card/70"
```
kouloumos: `bg-card/80` over an arbitrary illustration puts the row's secondary text below WCAG AA. The same scrim is used in `SubjectRow.tsx:144`, so both surfaces are affected.

I took the seven images the Athens city page currently serves, extracted the exact strip `object-cover` shows for a 934×65 row (12.2% of the image height, rows 337-431), composited it under `rgba(255,255,255,0.8)`, and measured against `text-muted-foreground` `rgb(120,113,108)`:

| | contrast |
|---|---|
| on the opaque card, before this change | **4.80:1** |
| through the scrim, mean pixel | **3.73 - 3.91:1**, 7 of 7 below AA |
| on hover, when the scrim thins to `bg-card/70` | 3.28 - 3.49:1 |
| worst individual pixel | 3.04 - 3.16:1 |

AA for small text is 4.5:1. This is the mean on every image on the page, not a worst case I constructed. The title is unaffected at 12.5:1 and above.

There is no theme caveat: the repo has no `darkMode` config, no `.dark` token block and no theme provider, so this is the only theme that renders.

`bg-card/90` restores the margin, or the meta line moves to `text-foreground/70`.

src/lib/subjectImages.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Without `force` an existing object is kept, so the summarize run costs
+ * nothing for what the agenda run already drew. With `force` the object is
+ * replaced, which is also how an admin undoes a manual upload. Throws on
+ * failure after alerting Discord; background callers catch, the admin route
+ * turns it into a 500.
+ */
+export async function generateImageForSubject(subjectId: string, options: { force?: boolean } = {}): Promise<GenerateOutcome> {
+    const apiKey = env.GEMINI_API_KEY;
+    if (!apiKey) return 'disabled';
+    if (inFlight.has(subjectId)) return 'in-flight';
```
kouloumos: This check sits above the `force` handling, so a forced regenerate is dropped whenever any generation for the same subject is already running. Lines 76 and 82 both honour `force`; this one does not.

The sequence is reachable through the normal admin path:

1. A superadmin opens a subject that has no image yet. `subject.tsx` renders `SubjectImage` with `loading="eager"`, so the `GET` fires immediately.
2. The route's `after()` callback starts a **non-force** generation, which holds `inFlight[X]` for the whole Gemini round trip.
3. The admin sees the placeholder and clicks Regenerate, which is the natural first action on a placeholder.
4. This line returns `in-flight` and nothing is generated.
5. The route still answers `200 {"ok":true,"mode":"generate","outcome":"in-flight"}`, and `SubjectImageAdminControls.tsx:34` only checks `res.ok`, so line 38 shows "Image updated" over an unchanged image.

The force is discarded rather than deferred: the run holding the lock is the non-force one, which returns `exists` and does nothing when an object is already stored.

Moving the `force` check above this line costs one duplicate generation in a race the module's own design note already accepts. The client should also stop treating any 2xx as "the image changed".

src/app/api/subject/[subjectId]/image/route.ts
```diff
@@ -0,0 +1,104 @@
...
+        if (!(file instanceof File)) {
+            return NextResponse.json({ error: 'No file provided' }, { status: 400 });
+        }
+        if (!UPLOAD_TYPES.has(file.type)) {
+            return NextResponse.json({ error: `Unsupported image type ${file.type}` }, { status: 400 });
+        }
+        if (file.size > UPLOAD_MAX_BYTES) {
+            return NextResponse.json({ error: 'Image exceeds 10 MB' }, { status: 400 });
+        }
+        await storeSubjectImage(subjectId, Buffer.from(await file.arrayBuffer()));
```
Comment (kouloumos) [inline src/app/api/subject/[subjectId]/image/route.ts]: This is the only awaited call in `POST` that is not inside a try/catch, and it can reject two ways: sharp cannot decode the buffer, or the `PutObject` fails. The generate branch below is wrapped; this one is not.

Confirmed on the preview, signed in as a superadmin:

| request | result |
|---|---|
| `{"mode":"nonsense"}` | 400, clean JSON |
| multipart, `image/tiff` | 400, clean JSON |
| multipart, `image/png` with 2 KB of random bytes | **500, zero-byte body, no content-type** |

The client cannot recover a message from that. `SubjectImageAdminControls.tsx:35` does `res.json().catch(() => ({}))`, then falls back to `res.statusText`. The preview serves HTTP/2, which carries no reason phrase, so I measured `res.statusText === ""` on this origin. The admin gets a red toast with an empty description.

The realistic trigger is a Spaces outage during a replace, not a malformed file. That case also raises no Discord alert, while every generate failure does through `reportFailure`.

Wrapping these two lines in the same shape as lines 97-103, and routing the failure through `reportFailure`, fixes both halves.

src/app/api/subject/[subjectId]/image/route.ts
```diff
@@ -0,0 +1,104 @@
...
+        lookupFailed = true;
+        console.error(`Subject image lookup failed for ${subjectId}:`, error);
+    }
+    if (resolved) {
+        return NextResponse.redirect(resolved.url, { status: 302, headers: { 'Cache-Control': REDIRECT_CACHE } });
+    }
+
+    const subject = await getSubjectTopicForImage(subjectId);
+    if (!subject) {
+        return NextResponse.json({ error: 'Subject not found' }, { status: 404 });
```
Comment (kouloumos) [inline src/app/api/subject/[subjectId]/image/route.ts]: This 404 carries no `Cache-Control`, so no edge can absorb a repeat. Verified on the preview:

```
$ curl -I https://pr-711.opencouncil.dev/api/subject/zzz-not-real/image
HTTP/2 404
content-type: application/json
```

It matters because of the ordering above: `resolveSubjectImage` runs before this existence check, so an unauthenticated request with any id spends one S3 `HeadObject` and one Prisma `findUnique` before the 404, and can repeat without limit. `src/lib/s3.ts` uses a single process-wide `S3Client`, and the Smithy node handler defaults to 50 sockets shared with every other Spaces call in the app.

The amplification is 1:1 and I measured about 181 ms per request, so this is a small finding rather than an availability problem. `public, max-age=300` on the 404 closes it.

Separately, the JSDoc above and `docs/guides/subject-images.md:70` both state that Cloudflare absorbs the repeat loads a landing page makes. That needs a Cache Rule, because Cloudflare's Standard cache level selects by file extension and this path has none. I could not confirm such a rule exists, so the claim may want softening to what the origin does.

src/lib/db/subject.ts
```diff
@@ -184,7 +184,7 @@ export async function getSubjectCountsByCityCached(realm: Realm): Promise<Record
                     councilMeeting: {
                         released: true,
                         dateTime: { lte: new Date() },
-                        city: { ...PUBLIC_CITY_WHERE, realm },
+                        city: realm ? { ...PUBLIC_CITY_WHERE, realm } : PUBLIC_CITY_WHERE,
```
kouloumos: This function's signature is still `(realm: Realm)`, and `Realm` is a Prisma enum of `greece | france | cyprus | serbia`. No member is falsy, so the else branch is unreachable.

The widening on line 366 is correct and deliberate, because `getSubjectsForImageBackfill` passes `null`. This one is not needed. Leaving it in suggests both functions accept a cross-realm mode when only one does, and `realm` is also the cache key and the revalidation tag two lines below, so a null here would produce the tag `realm:null:landing-subjects`, which nothing ever busts.

Reverting this hunk keeps the compiler guard where it is doing work.

src/lib/db/subject.ts
```diff
@@ -718,3 +719,68 @@ export async function getUtterancesForSubject(subjectId: string) {
...
+
+/** The topic colour and icon the image placeholder is drawn from. `null` when the subject does not exist. */
+export async function getSubjectTopicForImage(subjectId: string) {
+    return prisma.subject.findUnique({
+        where: { id: subjectId },
+        select: { topic: { select: { colorHex: true, icon: true } } },
+    });
+}
+
+export async function getSubjectIdsForMeeting(cityId: string, councilMeetingId: string): Promise<string[]> {
```
kouloumos: No `released` or city-status filter here, so the pipeline draws images for meetings that may never become public.

`handleProcessAgendaResult` calls `generateImagesForMeeting` when the agenda PDF is parsed, days before the meeting sits. `released` is `@default(false)` and nothing in that handler sets it. So a meeting that is cancelled, or simply never released, still gets its whole agenda illustrated at $0.034 per subject. A 40-item agenda is about $1.36 spent on images nothing renders.

Every sibling public read in this file gates on both conditions — `getSubjectCountForCity`, `getSubjectCountsByCityCached`, and the backfill this PR adds through `buildMapSubjectWhere` and `PUBLIC_CITY_WHERE`. Against those four, the omission reads as an oversight rather than a decision.

Adding `councilMeeting: { released: true, city: PUBLIC_CITY_WHERE }` puts the gate where it cannot be forgotten. If drawing early is deliberate, so a meeting is illustrated the moment it is released, then keep the `city` half at least and say in the JSDoc that the `released` gate is intentionally absent.

scripts/backfill-subject-images.ts
```diff
@@ -0,0 +1,73 @@
...
+            } catch {
+                counts.failed++;
+            }
+            done++;
+            if (done % 25 === 0 || done === subjects.length) {
+                console.log(`${done}/${subjects.length}`, counts);
+            }
+        }
+    };
+    await Promise.all(Array.from({ length: Math.min(argv.concurrency, queue.length) }, worker));
```
kouloumos: `argv.concurrency` reaches `Math.min` unvalidated, so a bad value makes the script do nothing and still exit 0:

| `--concurrency` | workers started |
|---|---|
| `0` | 0 |
| `-1` | 0 |
| `abc` (NaN) | 0 |
| `2.5` | 2 |

With zero workers `Promise.all([])` resolves at once and the script prints `done { generated: 0, exists: 0, … }`, which is indistinguishable from a completed run over an empty candidate set.

`.strict()` is already on, so yargs is the right place for the constraint:

```js
.option('concurrency', { type: 'number', default: 3, coerce: n => {
  if (!Number.isInteger(n) || n < 1) throw new Error('--concurrency must be a positive integer');
  return n;
}})
```

That turns it into a usage error before any work starts, and a future flag cannot route around it.

src/components/landing/v2/SubjectCard.tsx
```diff
@@ -144,70 +145,65 @@ export function SubjectCard({
...
+            <div className="relative aspect-[7/4] w-full overflow-hidden bg-muted">
+                <SubjectImage subjectId={subject.id} alt="" />
+                <span
+                    className="absolute left-3 top-3 inline-flex max-w-[calc(100%-4.5rem)] items-center gap-1.5 rounded-full px-2.5 py-1 text-xs font-bold shadow-sm"
+                    style={{ backgroundColor: topicBar.background, color: topicBar.icon }}
+                >
+                    <Icon name={subject.topic.icon || 'hash'} color={topicBar.icon} size={14} />
+                    <span className="truncate">{subject.topic.name}</span>
+                </span>
+                {subject.cityLogo && !onClose && (
```
kouloumos: The `!onClose` term removes the municipality logo from every map popup. `mapMarkers.tsx:42` always passes `onClose`, and the prop is required there, so the preview variant never shows a logo. On `main` the gate is `{subject.cityLogo && (`, with no variant term.

On the landing map and the city map the popup is now the only thing a reader sees when they click a pin, and pins on the landing map span many municipalities. I could not find another element in the popup that names the city.

If the intent was to keep the logo from colliding with the close button, moving it rather than dropping it keeps the fact on the card.

src/components/landing/v2/SubjectCard.tsx
```diff
@@ -144,70 +145,65 @@ export function SubjectCard({
...
-                    'flex items-center gap-2 border-b border-border font-bold text-xs',
-                    preview ? 'px-3 py-2' : 'px-4 py-2.5',
+            {/* hero — the illustration carries the topic, the place and the title; the card
+                below it holds the facts. The image box is reserved at 7:4 before the bytes
+                arrive, so the list does not jump as images load. */}
+            <div className="relative aspect-[7/4] w-full overflow-hidden bg-muted">
+                <SubjectImage subjectId={subject.id} alt="" />
+                <span
+                    className="absolute left-3 top-3 inline-flex max-w-[calc(100%-4.5rem)] items-center gap-1.5 rounded-full px-2.5 py-1 text-xs font-bold shadow-sm"
+                    style={{ backgroundColor: topicBar.background, color: topicBar.icon }}
```
kouloumos: This hand-rolls a topic pill that `TopicPill` already provides, and this PR uses the real one twice — `subject.tsx:287` and `HotTopicLead.tsx:57`.

`TopicPill`'s own doc calls it "the static form the subject header and the hot-topic lead share", and `topicSurfaceStyle` exists because the three-colour mapping "was repeated at seven call sites". This is the eighth.

The copy has drifted already: it sets `backgroundColor` and `color` but not `borderColor`, which `topicSurfaceStyle` sets, and it has no border class. Its glyph is 14px against `TopicPill`'s 13px. So one PR renders the same element three ways on three surfaces.

```jsx
<TopicPill
    label={subject.topic.name}
    icon={subject.topic.icon}
    colorHex={subject.topic.color}
    className="absolute left-3 top-3 max-w-[calc(100%-4.5rem)] shadow-sm"
/>
```

If the truncation is load-bearing, it belongs inside `TopicPill` rather than around it.

src/components/subject/SubjectImage.tsx
```diff
@@ -0,0 +1,42 @@
+import { IMAGE_HEIGHT, IMAGE_WIDTH } from '@opencouncil/subject-images/constants';
+import { cn } from '@/lib/utils';
+
+/** The image route, with an optional cache-buster for a copy the viewer just changed. */
+export function subjectImageUrl(subjectId: string, version?: number | string): string {
+    return `/api/subject/${subjectId}/image${version ? `?v=${version}` : ''}`;
```
kouloumos: `version` is typed `number | string` but tested for truthiness, so `0` and `""` both drop the parameter and return the un-busted URL.

No caller passes either today — `subject.tsx` uses `Date.now()` — so this is latent. The `| string` branch also has no caller. `version?: number` with `version !== undefined` makes the type and the behaviour agree.

src/lib/topicStyle.ts
christosporios: `mixHex` has no short-hex guard, so a 3-digit topic colour makes the placeholder a solid black box.

`HEX_REGEX` in `src/lib/utils/colorSuggestion.ts:6` is `/^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/`, and `src/lib/zod-schemas/topic.ts` validates `Topic.colorHex` against it. An admin can save `#abc` today.

`parse('#abc')` returns `[170, 12, NaN]`, so `mixHex('#abc', '#ffffff', 0.24)` returns `#ebc5NaN`. `subjectImageFallbackSvg` writes that value into `fill` and into `stroke`. Both are invalid, so the browser applies `fill: black` and `stroke: none`. Every subject of that topic that has no image shows a full-bleed black placeholder, on the landing cards, the hot-topic rows, the search rows and the subject page banner.

`contrastText` at line 29 in this same file has the guard that is missing here: `if (c.length < 6) return '#ffffff'`.

`hexToRgb` in `src/lib/utils/colorSuggestion.ts:10` already expands 3-digit hex before it parses, and `src/lib/utils/embedTheme.ts:9` has a second copy. Export one of them and build `mixHex` on it, rather than add a fourth hex parser.

The new tests pass only `#2a9d8f`, `#000000` and `null`.

---

src/components/subject/SubjectRow.tsx
```diff
@@ -134,7 +132,19 @@ export function SubjectRow({ subject, city, meeting, persons, showContext = true
...
+                    <SubjectImage subjectId={subject.id} alt="" />
+                </span>
+                <span
+                    className="absolute inset-0 z-0 bg-card/80 transition-colors duration-300 group-hover/row:bg-card/70"
```
christosporios, line 141: These two spans paint over the topic colour rail, so the rail is invisible on every subject row.

The rail wrapper at line 128 is positioned and has no `z-index`, so it is `z-index: auto`. These spans are positioned with `z-index: 0` and come later in tree order. CSS paints positioned descendants with `z-index: auto` and `z-index: 0` together in one step, in tree order. The opaque `object-cover` image and then the `bg-card/80` scrim therefore cover the 4px bar at `left-0` (line 132).

Before this change the rail was the only positioned overlay, so it painted above the static content.

The rail code and the comment that explains it are unchanged, so this looks unintended rather than a deliberate removal. `z-10` on the rail wrapper restores it.

`HotTopicRow` is not affected, because it has no rail.

---

src/components/subject/SubjectRow.tsx
```diff
@@ -157,7 +167,7 @@ export function SubjectRow({ subject, city, meeting, persons, showContext = true
                                 )}
                             </div>
 
-                            <h3 className="line-clamp-2 text-[15px] font-semibold leading-snug transition-colors duration-300 group-hover/row:text-[hsl(var(--orange))] sm:text-base">
+                            <h3 className="line-clamp-2 text-[15px] font-semibold leading-snug group-hover/row:font-bold sm:text-base">
```
christosporios: Hover now swaps the font weight on a `line-clamp-2` title, which reflows the text under the pointer.

A title that fills both clamped lines at `font-semibold` does not fit at `font-bold`. The last words move past the clamp and the ellipsis replaces them as soon as the pointer enters the row. Moving down a list of results makes each title change in turn.

`font-weight` is also not interpolatable, so the swap is instant. The `transition-colors` that this replaced did animate.

The previous colour hover was layout-neutral.

---

src/components/subject/SubjectRow.tsx
```diff
@@ -28,7 +27,8 @@ interface SubjectRowProps {
     subject: SubjectWithRelations & { statistics?: Statistics; matches?: SearchMatches };
     city: City;
     meeting: CouncilMeeting & { administrativeBody?: AdministrativeBody | null };
-    persons: PersonWithRelations[];
+    /** Kept for the callers; the row no longer draws its speakers. */
+    persons?: PersonWithRelations[];
```
christosporios: The row no longer draws speakers, but the callers still pay for the data that fed them.

`SubjectListContainer.tsx:66` fetches `/api/cities/{cityId}/people` for every city in the result set and gates rendering on `isDataLoading`. `SearchPage.tsx:510` renders that container with `variant="row"`. A result set that spans 8 municipalities still makes 8 roster requests, holds 8 full council rosters in state, and delays the rows behind them, for a prop this component ignores.

Keeping the prop hides that the fetch can now go. Drop both, or skip the fetch when `variant === 'row'`.

---

src/lib/subjectImageFallback.ts
```diff
@@ -0,0 +1,45 @@
...
+function toPascalCase(kebab: string): string {
+    return kebab.split('-').map((part) => part.charAt(0).toUpperCase() + part.slice(1)).join('');
+}
+
+function escapeAttribute(value: unknown): string {
+    return String(value).replace(/&/g, '&amp;').replace(/"/g, '&quot;').replace(/</g, '&lt;');
+}
+
+function iconMarkup(name: string | null | undefined, color: string): string {
+    const node = (name && icons[toPascalCase(name) as keyof typeof icons]) || icons.Hash;
```
christosporios: This indexes a plain object, so it reads the prototype chain. A topic icon named `__proto__` returns 500 instead of the placeholder.

`toPascalCase('__proto__')` returns `__proto__`, and `icons['__proto__']` returns `Object.prototype`. That value is truthy, so `|| icons.Hash` does not fire. `Object.prototype` is not iterable, so the array destructure on the next line throws `TypeError: node is not iterable`. The route then returns 500, and every card of that topic shows a broken image. That is the failure the placeholder exists to prevent.

An admin, a seed or an import can store that value; `Topic.icon` is a free string.

`Object.hasOwn(icons, key)` before the destructure fixes it.

---

src/lib/subjectImageFallback.ts
```diff
@@ -0,0 +1,45 @@
+import { icons } from 'lucide';
```
christosporios: This puts the whole lucide icon table into the server chunk of the route that every card hits.

`lucide@0.439.0` declares no `exports` map, so this resolves to `dist/esm/lucide.js`, whose `icons` is a namespace over all 1,526 icon modules. Line 24 indexes it dynamically, so no bundler can remove a single icon. About 1 MB of JS is parsed on the first request to this route. The package unpacks to about 19.7 MB and 3,072 files in `node_modules` and in the container image.

To be clear: this does not reach any client bundle. The only importer is the route handler.

There is a second cost. `src/components/icon.tsx:7` builds `ICON_NAMES` from `lucide-react@0.436.0`, and the topic admin validates icon names against that list. This file resolves the same names against `lucide@0.439.0`. Two packages at two versions now decide which icon names are valid.

`icon.tsx:20` imports one icon at a time through `next/dynamic`, with a comment that explains why. This is the opposite of that choice. The set of icons that can be reached is bounded by the `Topic` table, which holds about 20.

Two ways out:

1. Keep the server SVG and drop the package. Deep-import the icons you need (`lucide/dist/esm/icons/hash.js` resolves, because there is no `exports` map), or generate a small name-to-node map from the `Topic` table.
2. Do not render the placeholder on the server at all. Render `SubjectImage` only when an image exists, and otherwise render the topic wash plus the existing `Icon` on the client, the way `ImageOrInitials` already does for people, cities and parties.

Option 2 also removes `mixHex`, `topicStyleHex`, `toPascalCase`, `escapeAttribute` and the `__proto__` crash below, and the black-box bug in `topicStyle.ts`.

---

src/lib/subjectImages.ts
christosporios: The `inFlight` check runs before `force`, so an admin Regenerate can do nothing and still report success.

Line 76 gates `failedAt` on `!options.force`. This line does not gate `inFlight`.

The sequence: a superadmin opens a subject that has no image. The GET schedules `after(() => generateImageForSubjectInBackground(subjectId))`, which holds the id in `inFlight` for the 10 to 30 seconds the Gemini call takes. The admin clicks Regenerate inside that window. `generateImageForSubject(id, { force: true })` returns `'in-flight'` here, before the force path runs. `route.ts:99` answers `200 {ok: true, mode: 'generate', outcome: 'in-flight'}`. `SubjectImageAdminControls.submit` reads only `res.ok`, so it shows `imageUpdated` and bumps `imageVersion`. Nothing was regenerated, and the image that eventually lands is the un-forced one.

A double-click on Regenerate does the same, and so does a `generateImagesForMeeting` run that holds the subject.

The route also answers `ok: true` for `'recent-failure'` and `'disabled'`, so three outcomes report success to the admin.

Move the `inFlight` check inside `if (!options.force)`, or make force wait on the in-flight promise. The client should read `outcome` rather than only `res.ok`.

---

src/lib/subjectImages.ts
```diff
@@ -0,0 +1,132 @@
...
+        if (!options.force && await resolveSubjectImage(subjectId)) return 'exists';
+
+        const subject = await getSubjectPromptInput(subjectId);
+        if (!subject) throw new Error(`Subject ${subjectId} not found`);
+
+        // Descriptions carry markdown and REF:UTTERANCE links; the model should
+        // read the sentence, not the markup.
+        const prompt = buildPrompt({ title: stripMarkdown(subject.name), description: stripMarkdown(subject.description) });
+        const image = await generate(prompt, { apiKey });
+        await store(subjectId, image, storeDeps());
```
christosporios: An admin upload and a running generation write the same key, and the generation wins.

This is the normal admin flow, not an edge case. The admin opens a subject that has no image. The GET serves the placeholder and starts a generation, which passes its existence check at line 82 and then waits on Gemini. The admin clicks Replace, crops and uploads. `storeSubjectImage` writes the file in about a second and the client shows `imageUpdated`. Seconds later this line replaces that upload.

`storeSubjectImage` (line 57) neither joins nor reads `inFlight`, and this function never re-checks existence after the model returns.

Guarding only the upload path cannot fix this, because generation calls the library `store` directly and bypasses `storeSubjectImage`.

The admin sees their upload until the 300 second redirect cache expires. Then it is gone, and nothing is logged.

---

src/lib/subjectImages.ts
```diff
@@ -0,0 +1,132 @@
...
+ * a shared lock.
+ */
+
+/** How long a failed subject waits before a page view may retry it. */
+const RETRY_AFTER_MS = 10 * 60 * 1000;
+/** Generations run at once when a meeting's subjects are processed together. */
+const MEETING_CONCURRENCY = 3;
+
+const inFlight = new Set<string>();
+const failedAt = new Map<string, number>();
```
christosporios: `failedAt` entries expire only on a later success, and there is no terminal state for a permanent refusal.

The app runs as a long-lived `next start` process. An entry is added on every failure at line 95 and removed only at line 92. Nothing expires an entry after `RETRY_AFTER_MS`, so the map grows for the life of the process.

The retry behaviour matters more than the memory. The `CONSTRAINTS` block in `SYSTEM_PROMPT` is written to make the model refuse fires, accidents and injury, so subjects that always fail do exist. Such a subject is retried every 10 minutes forever, per container. Each retry is a billed Gemini call plus one `sendErrorAdminAlert`, and `src/lib/discord.ts` has no throttle or dedup.

`src/lib/cache/valkey.ts` already exists for exactly this. Its header says it is for "application-level caching and dedup markers", and the `cacheHas` doc comment names the one-shot marker case. `cacheSetJSON(key, value, 600)` gives the back-off with expiry for free. A separate marker with a long TTL gives the terminal "refused" state that is missing, and the same marker throttles the Discord alert.

The same store also covers `inFlight` on line 26. The header comment at lines 14 to 18 says the process-local bound "does not justify a shared lock", but the deployment already runs Valkey, so the trade-off is not the one the comment assumes. `cacheSetJSON` has no `NX` option today, so this needs one helper of about ten lines that calls `SET key value NX EX ttl`. The TTL then releases the marker if a container dies mid-generation, which the in-memory `Set` cannot do across containers.

Every helper in that module degrades to `null` or `false` when `CACHE_URL` is unset, so dev and previews keep working with no new requirement.

---

src/lib/subjectImages.ts
```diff
@@ -0,0 +1,132 @@
+import { buildPrompt, generate, resolve, store, toWebp, type ResolvedImage } from '@opencouncil/subject-images';
```
christosporios: This module reads the Gemini key and holds the S3 client, and it has no `import 'server-only'`.

`CLAUDE.md` says: "For new code, keep private data access and service functions in `server-only` modules." 37 modules under `src/lib/` carry the marker, including ones this file imports.

Nothing stops a client component importing `isSubjectImageGenerationEnabled` today.

---

packages/subject-images/src/store.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export interface ResolvedImage {
+    url: string;
+}
+
+function isNotFound(error: unknown): boolean {
+    if (typeof error !== 'object' || error === null) return false;
+    const e = error as { name?: string; $metadata?: { httpStatusCode?: number } };
+    return e.name === 'NotFound' || e.name === 'NoSuchKey' || e.$metadata?.httpStatusCode === 404;
```
christosporios: `isNotFound` gets both directions wrong for the errors Spaces returns.

`HeadObject` has no response body, so the status code is all there is to read.

**A wrong bucket reads as "no image".** A missing or misnamed bucket also answers 404, as `NoSuchBucket`. `resolve` then returns `null`, `route.ts` leaves `lookupFailed = false`, `after()` queues a generation, Gemini draws and bills, and the `PutObject` fails against the same missing bucket. The result is a Discord alert and a 10 minute lockout per subject, per container. That is the case `route.ts:36-38` says it guards against: "do not queue a generation that would fail at the same store."

**A least-privilege key kills the feature silently.** A Spaces key with `GetObject` and `PutObject` but no `s3:ListBucket` answers 403 for a *missing* object, not 404. `resolve` then throws for every subject, `lookupFailed` becomes `true`, and line 56 suppresses generation. Every subject serves the placeholder forever, no image is ever drawn, and no alert fires, because the route only calls `console.error`. The single symptom is one log line per card per page view.

A wrong region gives `301 PermanentRedirect`. Only a network error, where `$metadata.httpStatusCode` is undefined, deserves the rethrow this gives everything.

`src/lib/s3.ts:95` holds the same predicate inside `fileExists`, and it does not recognise `NoSuchKey`, so the two copies already disagree. Lift one shared `isS3NotFound` out of `fileExists`, and handle 403 explicitly.

---

packages/subject-images/src/store.ts
```diff
@@ -0,0 +1,62 @@
+import { HeadObjectCommand, PutObjectCommand, type S3Client } from '@aws-sdk/client-s3';
+
+/**
+ * Only this module knows where a subject's image lives. The prefix names the
+ * style (`subject-images/8bit`): a later restyle writes a sibling folder and
+ * leaves these.
+ */
+export function objectKey(subjectId: string, prefix: string): string {
+    return `${prefix.replace(/\/+$/, '')}/${subjectId}.webp`;
```
christosporios: `subjectId` reaches the object key with no encoding and no character check.

`route.ts:42` calls `resolveSubjectImage` before the subject-exists check at line 51, so the `HeadObject` is unauthenticated and can address an arbitrary key. `GET /api/subject/..%2F..%2Flogo/image` builds the key `subject-images/8bit/../../logo.webp`. The 302 against 404 split is then an existence oracle for any `.webp` key in the bucket.

The superadmin POST writes through this same key, which is the worse half: an upload can be aimed outside the prefix.

Real subject ids are cuids, so this needs a crafted request.

Three smaller defects sit on this line and on line 44:

- `replace(/\/+$/, '')` strips only *trailing* slashes. `SUBJECT_IMAGES_PREFIX=/subject-images/8bit` gives the key `/subject-images/8bit/x.webp` and the URL `https://cdn//subject-images/...`.
- A `#` in the id truncates the URL at line 44 at the fragment, so `?v=` is never applied and the redirect points at the wrong object.
- `env.CDN_URL.replace(/\/$/, '')` in `src/lib/subjectImages.ts:36` strips one trailing slash only.

`encodeURIComponent(subjectId)`, a `/^[A-Za-z0-9_-]+$/` guard, and a leading-slash strip on the prefix close all four.

---

src/app/api/subject/[subjectId]/image/route.ts
Comment (christosporios) [inline src/app/api/subject/[subjectId]/image/route.ts]: Three problems in the upload branch.

**The type check does not constrain what sharp decodes.** `file.type` is metadata the client sets. sharp reads the real format from the magic bytes. A multipart part that declares `image/png` but carries SVG bytes passes this check and reaches librsvg, which is an external-reference surface and can make outbound requests from the server process. A crafted 40 KB PNG that decodes to 250 megapixels also passes `UPLOAD_MAX_BYTES` and sits under sharp's 268 MP default, which allocates about 1 GB inside the request process. The route is superadmin only, so the blast radius is bounded, but the check that is written does not do the job it appears to do. `sharp(image, { limitInputPixels: IMAGE_WIDTH * IMAGE_HEIGHT * 16, failOn: 'error' })` plus a magic-byte check would.

**No error handling.** The generate branch at lines 95 to 101 wraps its call and returns a 500 with a message. This line does not. A file sharp cannot decode throws out of the handler, so `SubjectImageAdminControls.submit` shows `res.statusText` instead of a reason.

**No subject-exists check.** The GET path checks at line 51. This path does not, so a mistyped id writes a permanent public-read orphan that nothing serves and nothing cleans up.

Separately, `UPLOAD_TYPES` and `UPLOAD_MAX_BYTES` make a third accepted-types list and a second size cap. `src/components/admin/product-updates/imageTag.ts:22` already exports `validateImageFile({ type, size })` over `MAX_IMAGE_BYTES` and `ACCEPTED_IMAGE_TYPES`, and its comment says it exists because `POST /api/upload` validates neither. `src/types/upload.ts:18` holds `ALLOWED_LOGO_CONTENT_TYPES`. The app now answers "which image formats do we accept" three different ways.

---

src/components/subject/SubjectImageAdminControls.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const upload = (file: File) => {
+        setCropFile(null);
+        const form = new FormData();
+        form.append("file", file);
+        return submit({ body: form });
+    };
+
+    return (
+        <>
+            <div className={cn("absolute inset-x-0 bottom-0 flex justify-end gap-2 bg-gradient-to-t from-black/60 to-transparent p-3 opacity-0 transition-opacity focus-within:opacity-100 group-hover:opacity-100", className)}>
```
christosporios: `opacity-0` without `pointer-events-none` leaves the hidden buttons clickable.

Visibility comes only from `group-hover` and `focus-within`, and an element at `opacity: 0` still hit-tests.

`subject.tsx` mounts this with `className="inset-x-0 top-0 bottom-auto ..."`, so the invisible strip sits across the top of the hero image, about 56px tall. On a touch device `:hover` never resolves, so a superadmin who taps the top right of the illustration hits Regenerate. The image is replaced and no control was ever visible.

---

src/lib/db/subject.ts
```diff
@@ -718,3 +719,68 @@ export async function getUtterancesForSubject(subjectId: string) {
...
+            allTime: options.monthsBack === undefined,
+        }),
+        select: {
+            id: true,
+            cityId: true,
+            locationId: true,
+            councilMeeting: { select: { dateTime: true, administrativeBody: { select: { type: true } } } },
+        },
+    });
+    const discussionSeconds = await getDiscussionSecondsForSubjects(rows.map((row) => row.id));
```
christosporios, line 784: The backfill loads every discussed subject on the platform, then passes every id to one raw `IN (...)`.

`getSubjectsForImageBackfill` calls `buildMapSubjectWhere(null, { ..., allTime: options.monthsBack === undefined })`. `realm: null` spans every realm, `allTime` drops the lower date bound, and the `findMany` at line 764 has no `take`. With the documented invocation `npx tsx scripts/backfill-subject-images.ts --limit 50`, `rows` is every released, discussed subject ever, platform-wide.

This line hands all of those ids to `getDiscussionSecondsForSubjects`, whose `$queryRaw` at line 217 is `WHERE u."discussionSubjectId" IN (${Prisma.join(subjectIds)})`. That is one bind parameter per id. PostgreSQL's wire protocol caps a statement at 65535 parameters, so past that row count the script fails with a bind-message error before it draws anything. Below that it is a slow scan over the largest join in the schema. Every existing caller of that helper (lines 419, 457, 476) is narrowed by one realm plus the 6 month default window.

`scripts/backfill-subject-images.ts:41` then slices to `--limit` after the fetch and the ranking, so `--limit 50` still loads and ranks the whole corpus into memory first.

There is a second problem in the same design. The script ranks by importance, not by "missing an image". Run 2 with `--limit 50` repeats the whole scan, then spends 50 `HeadObject` calls to learn the top 50 already have images, and draws nothing. To reach 500 images you must raise `--limit` and re-scan each time. One `ListObjectsV2` over `SUBJECT_IMAGES_PREFIX` returns 1,000 keys per request and gives the done-set to exclude first, and `take` in SQL bounds the candidate window.

---

src/lib/tasks/processAgenda.ts
```diff
@@ -88,6 +89,11 @@ export async function handleProcessAgendaResult(taskId: string, response: Proces
     // recipients open the meeting and see stale content.
     revalidateMeeting(task.councilMeeting.cityId, task.councilMeeting.id);
 
+    // Illustrations for the agenda subjects, days before the meeting. Not
+    // awaited: the callback must not wait on Gemini, and each failure alerts
+    // on its own.
+    generateImagesForMeeting(task.councilMeeting.cityId, task.councilMeeting.id);
```
christosporios: This promise escapes the request without `after()`, and the same PR uses `after()` correctly in the read route.

`handleProcessAgendaResult` is reached from the taskStatuses route handler, which awaits the handler and returns its JSON at once. `generateImagesForMeeting` then runs three workers, each awaiting a Gemini call and an S3 put, for every subject of the meeting. For a 40-subject meeting that is minutes of work outside the request's lifetime.

A rolling deploy, a container restart, or any runtime that freezes the instance after the response kills the loop part-way. The meeting keeps images only for the subjects drawn so far. No `TaskStatus` row records the attempt, nothing retries, and the function never rejects, so nothing is logged either.

`src/app/api/subject/[subjectId]/image/route.ts:57` does the same kind of work with `after(...)`. `src/lib/tasks/summarize.ts:169` needs the same change.

The two added tests assert only that the function was called, and the call order, so they pass either way.

---

packages/subject-images/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "exports": {
+    ".": "./src/index.ts",
+    "./constants": "./src/constants.ts"
+  },
+  "dependencies": {
+    "@aws-sdk/client-s3": "^3.1060.0",
+    "@google/genai": "^2.20.0",
+    "sharp": "^0.33.4"
```
christosporios: This range does not overlap the root's, so the lockfile installs a second sharp and a second libvips.

The root declares `"sharp": "^0.35.4"`. A caret on a 0.x version pins the minor, so the two ranges cannot resolve to one install. `package-lock.json` on this head carries `packages/subject-images/node_modules/sharp@0.33.5` at line 25193, with 18 nested `@img/sharp-*` platform packages, next to the root's `sharp@0.35.4`.

Every `npm ci`, every container image and the Nix build now carry a duplicate native image library that the app never calls, and a sharp upgrade has to be applied in two manifests.

Nothing in `generate.ts` uses a 0.33-only API. `"sharp": "^0.35.4"`, or a `peerDependency` on it, removes the nested tree.

---

src/components/cities/overview/HotTopicLead.tsx
```diff
@@ -30,36 +23,31 @@ interface HotTopicLeadProps {
...
+ * container would read as two lists that happen to be adjacent. It carries its
+ * illustration the same way they do, filling the entry. Its scrim is dark where
+ * theirs are light: the entry the page is claiming matters reads as the primary
+ * block of the list, and inverting it is what says "this one is open".
  */
-export function HotTopicLead({ card, rank, maxSeconds, cityId, timezone, locale }: HotTopicLeadProps) {
+export function HotTopicLead({ card, rank, cityId, timezone, locale }: HotTopicLeadProps) {
     const t = useTranslations('cityOverview');
-    const { subject, meeting, stats, speakers } = card;
+    const { subject, meeting, stats } = card;
```
christosporios, line 33: The avatar rows are gone, so nothing reads `card.speakers`, but the producer still builds it and ships it.

`HotTopicRow` never drew speakers, and this component no longer does. `src/lib/hotSubjectCards.ts:113` still calls `displayedSpeakers(...)`, and the field at line 48 still carries the comment "for the avatar row".

Every city overview page therefore serializes up to 6 full `PersonWithRelations` records per card, with their role and party relations, into the RSC payload and into the `getHotSubjectCardsCached` entry. `HOT_CARDS_CACHE_VERSION` was not bumped, so existing cached entries keep the field as well.

Separately, the `person_opened` event with `surface: 'avatar_stack'` now never fires from the city overview, and nothing replaces it. Any dashboard segmented on that surface reports zero from here on.

---

src/components/landing/v2/SubjectCard.tsx
christosporios: This guard hides the city logo on the map preview, which is the one surface that spans several municipalities.

`onClose` is passed only by the map-preview variant, so `!onClose` reads as "hide the city whenever this is a preview". The reason appears to be that the new logo slot at `right-3 top-3` collides with the close button at `right-2 top-1`.

The same rewrite replaced the facts block that carried the city. The hero overlay now shows the topic pill, the location line and the title, and the row below shows `subject.bodyName`, which names the administrative body rather than the municipality.

So a user who clicks a pin on the national map gets a card that names no municipality at all.

Moving the logo to the left of the close button restores it.

---

christosporios (commented):
Inline comments on the defects that have a clear fix. Design-level points follow in a separate comment.

One finding cannot be inline, because the file is not in the diff, and it is the one that blocks a deploy.

## `Dockerfile` and `Dockerfile.dev` do not copy the new workspace manifest

Both files copy the workspace manifests before `RUN npm install`:

```dockerfile
COPY package*.json .npmrc ./
COPY packages/ui/package.json packages/ui/
COPY services/notis/package.json services/notis/
```

directly under this comment:

> Workspace member manifests must be present before npm install, or npm resolves the workspace-shaped lockfile against missing directories.

`package.json:66` now declares `"@opencouncil/subject-images": "*"`, and the `packages/*` workspace glob matches the new member, but nothing copies `packages/subject-images/package.json`. That directory arrives only at `COPY . .` in stage 2, which runs after the install.

`@google/genai` is declared only in the new member's manifest, so it is never installed. The image builds and starts. The first generation then throws `Cannot find module '@google/genai'`.

Verified on `db547a5`.

Add one line to both files:

```dockerfile
COPY packages/subject-images/package.json packages/subject-images/
```

---

christosporios:
One more thing: Let's move the two guards to Valkey so that they're shared across instances. inFlight and failedAt live in one process.

I'll follow up with design feedback once these are resolved.

pkakelas:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/db547a585cd86408a2306ddbb8822a919a271c63..3a8f66d16c69574fcdad0f396c0443024a9e286f>) to apply the review, every inline thread and the Valkey follow-up, squashed into the four commits:

- Both Dockerfiles copy `packages/subject-images/package.json` before `npm install`.
- The placeholder moved to the client. `SubjectImage` draws the topic wash and glyph under the ``, which hides itself when the route answers 404. The server SVG, the hex helpers and the `lucide` package are gone.
- Rows and cards: the topic rail sits above the image, the title hover is a colour change again, the hidden admin strip takes no pointer events, the city logo stays on the map preview, the row variant fetches no rosters, and the city page ships no speakers. The embed widget still asks for them with `withSpeakers`.
- The generation guards live in Valkey: one run per subject across containers, a forced run that waits and lands last, `superseded` when a write landed mid-draw, a 10 minute back-off, a 30 day give-up after three failures, and a 10 minute pause of every generation after the bucket refuses a write. The route answers 409 when nothing was drawn. Both pipeline handlers schedule through `after()`.
- Route and library: `server-only`; the id must match `^[A-Za-z0-9_-]+$` before any lookup and in the key builder; one shared `isS3NotFound`, 403 read as a miss, other lookup failures alerted once per 10 minutes; uploads sniffed from the bytes with a pixel limit in sharp, 400/404/500 with reasons, one size cap and one module for the three format lists.
- The backfill takes a bounded newest-first window, lists the bucket once to skip drawn subjects, then ranks.
- `sharp` on the root's range, so one copy in the lockfile. Imports above the mocks in the new tests. Tests for each change.

`npm run build`, the full Jest suite and `eslint .` pass.

src/components/cities/overview/HotTopicLead.tsx
pkakelas: One correction to this: the embed widget still draws the avatar row, so it does read `card.speakers`. I kept the field for that surface only. The embed asks for it with `withSpeakers`; the city page and its cache never get it.

pkakelas:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/3a8f66d16c69574fcdad0f396c0443024a9e286f..31fde92bdd645827f46470c8b700b176d38b3e3d>) to rebase onto `main` and fix the two CI failures:

- `main` moved `sendErrorAdminAlert` to `src/lib/discord-core.ts` in 61b8737a; the service now imports it from there.
- The integration tests call the task handlers with no request scope, so `after()` threw. That test file now stubs `after`; the unit tests keep asserting the scheduling.

christosporios:
Spent some time thinking about performance. The site serves about 1,400 pageviews a day today, peaking at ~60 in a single minute, and the illustrations add roughly 6,000 image requests a day on top of that. At ten times the traffic the peak is still only 60 to 100 image requests a second, each one a short HeadObject that waits on the network rather than the CPU, so two instances carry it with room to spare. The images are lazy on every surface except the subject page hero, so nothing waits on them to render — the one thing worth doing first is running the backfill, so a cold start does not turn traffic into a burst of generation.

I'll spend some time on the UI / design, and then merge this!

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/31fde92bdd645827f46470c8b700b176d38b3e3d..84569817ecb1d01479c837db275082b31db48a07) to apply the outcome of the design review:

- The subject page's illustration runs edge to edge under the header bar. The title block keeps to the reading column.
- The superadmin tools sit inside the image as a hazard-striped `AdminStrip`. The strip shows on hover and on keyboard focus. The metadata button joins it.
- The landing card drops the "Εικόνα από ΤΝ" footer label. A robot chip on the picture opens a popover that says both the picture and the summary are AI-made, with a link to the about page. The same chip sits on the subject page header.
- The MCP link in the navigation uses a plug instead of the robot, so the robot means AI-made content only. This is a new commit.
- New: the `AIDisclosure` component, the `Common.aiDisclosure` strings in four locales, and a render test for the card.
- The city page stays as it was.

The changes to existing commits are squashed into them.

christosporios:
Thank you! This was merged as part of #748.

PR #694, opened 2026-08-28
feat(city): rebuild the city page, the app header, and add a map tab

christosporios:
Rebuilds the city page around what a reader comes for, and rebuilds the app header that sits above it.

## The city page

The hero becomes a dense identity band: the municipality's name, its support status, and its counts on one line. The rail moves to a column beside the tabs instead of a band above them, and it sticks while the tab scrolls. On a phone the rail's cards start closed, so the tabs sit near the top.

The overview leads with the ranking of what the council has been debating. The reader can re-scope that ranking by administrative body and by period. Below it, a meeting module switches between the council and all bodies, and a council band closes the page with the parties that hold it.

The meeting list gets its own route. Meeting cards are rebuilt around the date. People and parties now share one card component.

A new **map tab** shows the δήμος's subjects on the map (see below).

## The header

Page actions leave the header and take a row of their own, so the header holds the path and the account control only. The account control is now always an avatar; a pencil badge on it marks edit rights over the thing on screen, and the menu says in words what the badge means — the old shield icon is gone.

The nav owns the left column, so the page carries one vertical rule that the second row aligns to. MCP and the guide return to the bar where it has room.

## The map tab

The landing's subject map is reused rather than reimplemented. `useSubjectMarkers` was already subject-only, so the δήμοι donuts and the petition layer are simply never called here.

Three pieces move to `src/components/map/subjects/`:

- `SubjectStrip` and `SubjectExpandedCard`, lifted out of `MobileLayout.tsx` where they were private functions
- `useSubjectMapState`, which owns the state every subject map keeps — selection, strip preview, viewport, the open cluster box — with the moveend capture that feeds them

`LandingV2` consumes the same hook and loses nine declarations. Its behaviour does not change. `useMapViewCapture` now takes `setCenterMunicipality` as optional, so a map whose δήμος is fixed makes no `/api/cities/at` call.

The tab is one card: a header band that opens the full map, then the map. On a phone the map is full-bleed with the subject strip over it; from `xl` the subject list is a column beside it.

The window is fixed at the last 3 months, the same window the landing opens on. There is no period picker, no topic chips and no widening fallback: every other period and every other filter belongs to the full map, which the header band links to. A δήμος that has not met in that window says so on a card naming the window.

## Data and caching

- Meeting lists get a projection that fits a card, instead of the full row.
- Hot subject cards are cached, and carry less.
- Time-filtered meeting queries carry a TTL. Such a result is a function of `now`, and no tag can express "a meeting has since happened".
[...]

PR #724, opened 2026-09-06
feat(people): order everyone by elected order

christosporios:
Follows #694, which has merged. This branch now targets `main` and merges on its own.

## Summary

One field now decides how people are listed: `Role.electedOrder`, the elected order of a seat on an administrative body. The party rank (`Role.rank`) goes away. Three cities set it by hand as a stand-in for the elected order, and the minutes could not use it, because they need the order of the whole body.

The order is the same on every page:

- **In one body**: the mayor, then the chair, then the parties in turn, then elected order inside the party, then surname.
- **In one party**: the mayor, then the party head, then the members by the body they sit on (council, committee, community, none), then elected order in that body, then surname.
- **Across a city**: the mayor, then each body in turn, then everyone else by surname. A person appears once, with the first body they sit on.

The people page shows all people by default. When the reader picks one body, the list follows the order of that body. When the reader picks one body type, only the bodies of that type place people.

A body's list groups its members by party, because that is how a municipality publishes its council. The grouping is a rule in the sort, not a property of the numbers. A Greek municipal election awards seats across the parties, so the order of election interleaves them: on Orestiada the council reads Νέα Ξανά 1, an independent 2, Νέα Ξανά 3–8, ΠΡΩΤΗ ΓΡΑΜΜΗ 9. Every number is right, and a list that followed them alone would still be wrong. A party's block sits where its first-elected member sits, so the blocks follow the election without a seat count. Members who hold no party close the list.

`electedOrder` therefore keeps its own meaning: the seat's order of election.

## Changes

- `src/lib/sorting/people.ts` holds the three rules. `List` gets a `sortItems` hook, so the people page can order by the selected body.
- The migration writes the party rank of Chalandri, Orestiada and Papagos-Cholargos into the elected order of the council role, in the order those cities show today, and then drops `Role.rank`.
- `createPerson` and `editPerson` now carry `electedOrder`. They dropped it before, so a person edit erased the order of the whole body. That is how Vrilissia and Zografou lost most of theirs.
- The party ranking sheet, its API route, and `updateRoleRankings` are removed.
- `City.peopleOrdering` stays as a column with its default. The form field, the schema field, the API handling and the sort branch are removed. The column can go in a later change once the new order has run in production.
- The public API and the OpenAPI schema lose `Role.rank`.
- Three helpers move to their homes: `isMayor` to `src/lib/utils/roles.ts`, `getSurname` to `src/lib/formatters/name.ts`, and the body-type comparator next to `ADMIN_BODY_TYPE_ORDER`.

## Verification

- Type check, lint, 165 test suites and the production build pass.
- The migration ran on a local PostgreSQL 16 against the real roles of four cities. Counts and ranges match, with no duplicates. A synthetic case covers two parties that share a name.
- The new tests were checked by mutation against every case the review named, and against the party-block key.

## Visible effects

- Chalandri and Orestiada: the party blocks hold, and the members inside a block follow the elected order.
- Papagos-Cholargos: two 3-member parties swap, and the party whose head was ranked third reorders. The city gets its full order from a data script that follows.
- Default cities: the mayor moves from the end of their party list to the top. The party page previously placed a party head with no council seat after every councillor.
- Vrilissia and Zografou: the few people that still carry an elected order lead their party block until the order is restored.

## Open

- The minutes sort on `electedOrder` alone, so a πρακτικό prints one order of election while a page prints party blocks. Whether the πρακτικά should group by party is a separate decision, and this PR does not touch them.
- Pinning an Αντιπρόεδρος or a Γραμματέας above their party block needs a signal on `Role`. `isHead` marks the chair, and we do not match on role names. Only the mayor and the chair are pinned today.

## Follow-ups

- Run the Papagos-Cholargos data script for the council and committee officers, and for the mayor's missing head flag.
- Fill the elected order for the supported municipalities from the election results and the committee decisions.
- Move the `electedOrder` carry-over into `editPerson`, so a caller of the PUT route that is not this form keeps the order. See the open thread on `RolesList.tsx`.

## Deploy note

Migrations run at build time. Between the column drop and the container swap, the old client still selects `rank` on every role read. A low-traffic deploy, or a split into a code release and a later drop, avoids that window.

https://claude.ai/code/session_01LbuxmSdnZVX23fJQELSi33

<h2><a href="https://app.greptile.com/api/retrigger?id=60924703"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/schema-labs/-/pull-requests/schemalabz/opencouncil/724"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until body lists follow the saved body-wide elected order.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Flib%2Fsorting%2Fpeople.ts%3A186%0AWhen%20members%20belong%20to%20different%20parties%2C%20the%20comparator%20checks%20the%20party%20block%20before%20%60electedOrder%60.%20A%20later-elected%20member%20in%20the%20first%20party%20can%20therefore%20appear%20before%20an%20earlier-elected%20member%20in%20the%20next%20party.%20This%20breaks%20the%20stated%20one-body%20order%20and%20differs%20from%20the%20body-wide%20order%20that%20%60ElectedOrderSheet%60%20saves.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=724&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Party Blocks Override Election Order** <a href="https://github.com/schemalabz/opencouncil/pull/724#discussion_r3960979269">▶</a>

<details><summary><h3>Summary</h3></summary>

- It centralizes city, party, and body sorting in `src/lib/sorting/people.ts`.
- It carries `electedOrder` through person edits and clears it when a role moves to another body.
- It migrates existing party ranks and removes the former ranking API and interface.
- The latest revision also groups body members into party blocks, which conflicts with the body-wide elected order.
</details>


christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/e0e8c6a83c067d1b24368f47e40be3637f52301d..dec3876998ac5855e18be0478467997da712e592>) to fold the review fixes into the two commits:

- `sortBodyMembers` now keys on seat membership before elected order. A person listed with a body but without a seat on it (a deputy mayor under the council filter) sorts after every member, numbered or not. The test covers the unnumbered case.
- A role edit keeps `electedOrder` only while the role keeps the same administrative body. A role moved to another body or type starts without one.
- The migration numbers the backfilled council roles after the highest `electedOrder` the city already has, so a value set by hand before the migration keeps its place. Verified on a local copy of the production roles with a pre-set value.
- `OrderedRole.cityId` is required, so a caller cannot lose the mayor-first rule by omitting it.
- The sorting module uses `filterActiveRoles` and `isActivePartyRole` instead of its own copies.
- `src/lib/db/roles.ts` imports through the `@/` alias.
- The city schema tests are rebuilt from the base with only the `peopleOrdering` cases removed. The first version also dropped the `authorityType` tests and the `Enum validation` wrapper by mistake.
- The last `rank: null` in `tests/helpers/builders.ts` is gone.

---

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/dec3876998ac5855e18be0478467997da712e592..33477f41789e812299a58fc42a80c08763597c44>) to apply the body rule when one body type is selected:

- `sortPeople` takes an optional body type. With it, only the bodies of that type place people, so a list filtered to the committees follows committee order even for people who also sit on the council.
- The party page uses it when one body type is picked. Otherwise it keeps the party's own order.
- The people page uses it when one type is picked and no single body is chosen. Before, a type with several bodies fell back to the city order, where councillors led.
- Test added for the type-restricted order.

---

src/components/parties/Party.tsx
```diff
@@ -87,6 +85,14 @@ function PartyMembersTab({
         people.filter(person => isActivePartyMember(person, party.id) && filterByAdminBodyType(person)),
         [people, party.id, filterByAdminBodyType]);
 
+    // One body type picked: that type's order, as on the people page. Otherwise
+    // the party's own order.
+    const orderedPeople = useMemo(() =>
+        selectedTypes.length === 1
+            ? sortPeople(activePeople, selectedTypes[0])
+            : sortPartyMembers(activePeople, party.id),
```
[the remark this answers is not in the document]
christosporios: Intended. With one body type selected, the party page follows the order of that body, the same as the people page. The party head has no special place in a body's order. In a city with the election-derived order the head is already first among the party's councillors, so the council view does not change.

---

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/33477f41789e812299a58fc42a80c08763597c44..505a95d64e39b9e83e91e520a814bb4229daae31>) to rebase onto the current head of `feature/city-page-redesign`, which now carries `main`. The diff of the PR is unchanged.

---

kouloumos:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/505a95d64e39b9e83e91e520a814bb4229daae31..10bead12fc70606e668f539ed4f04ea68e4456f5>) to rebase onto the current head of `feature/city-page-redesign` and link the two PRs into a GitHub stack (#731). The diff of the PR is unchanged.

prisma/migrations/20260906120000_replace_party_rank_with_elected_order/migration.sql
```diff
@@ -0,0 +1,66 @@
...
+    JOIN "AdministrativeBody" body
+      ON body.id = council."administrativeBodyId" AND body.type = 'council'
+    JOIN "Person" person ON person.id = council."personId"
+    GROUP BY person."cityId"
+),
+numbered AS (
+    SELECT ranked.role_id,
+           ROW_NUMBER() OVER (
+               PARTITION BY ranked."cityId"
+               ORDER BY ranked.members DESC, ranked."partyId", ranked.rank, ranked.name
```
kouloumos: Equal-size parties tie on `partyId`, a cuid. The page today uses the fetch order of `getPartiesForCity`, which has no `orderBy`. Both are arbitrary, so they need not agree, and "Papagos — one 3-member party changes" is this case. A tie-break a reader can predict, such as the party name, and one line in the comment would make it predictable.

src/lib/sorting/people.ts
```diff
@@ -1,14 +1,51 @@
...
+ *   sit on. With a body type selected, only the bodies of that type place
+ *   people, so a page filtered to the committees lists them in committee order.
+ *
+ * The minutes read `electedOrder` through {@link getElectedOrderForBody} and
+ * {@link compareRanks} and nothing else, so the order a page shows and the
+ * order the minutes print come from the same numbers.
+ */
+
+/** The role fields the ordering reads. Every role shape in the app carries them. */
+export interface OrderedRole {
```
kouloumos: `OrderedRole` re-spells Prisma `Role` fields by hand, and the two optional members widen the contract for no caller. Every call site passes people built with `roleWithRelationsInclude`, so `administrativeBody` is always loaded. If a caller ever omits it, every body ranks as `NO_BODY` and the city order degrades to surname without an error.

```ts
export type OrderedRole = Pick<Role, 'isHead' | 'cityId' | 'partyId' | 'administrativeBodyId' | 'electedOrder' | 'startDate' | 'endDate'> & {
    administrativeBody: Pick<AdministrativeBody, 'id' | 'name' | 'type'> | null;
};
```

Type check and all 16 tests pass with this.

src/lib/sorting/people.ts
```diff
@@ -39,80 +65,160 @@ export const compareRanks = (aRank: number | null, bRank: number | null): number
...
+export const sortPersonsByLastName = <T extends OrderedPerson>(persons: T[]): T[] => {
+    return [...persons].sort(compareByLastName);
+};
 
-        // Finally sort by name
-        return a.name.localeCompare(b.name);
-    });
+const activeRoles = (person: OrderedPerson): OrderedRole[] => filterActiveRoles(person.roles);
+
+const isMayor = (person: OrderedPerson): boolean => activeRoles(person).some(isMayorRole);
```
kouloumos: Three of the private helpers here already exist, or have a home, elsewhere:

- `isMayor` is the sixth inline "holds an active mayor role" (`VotingSection.tsx:77`, `MeetingDecisionsPage.tsx:97`, `composition.ts:51`, `getMinutesData.ts:104`, `pollDecisions.ts:108`). `src/lib/utils/roles.ts` has the person-level pattern in `isActivePartyMember`; an exported `isMayor(person)` there lets the other five follow later.
- `lastName` is the fourth "last whitespace word" split. `src/lib/formatters/name.ts` does it three times and exports no surname getter.
- `bodyTypeRank` is the fourth private body-type ranking (`brochure.ts:19`, `db/coverage.ts:22`, the widget page). `ADMIN_BODY_TYPE_ORDER` lives in `administrativeBodies.ts`; a comparator next to it is the unit.

src/lib/sorting/people.ts
```diff
@@ -39,80 +65,160 @@ export const compareRanks = (aRank: number | null, bRank: number | null): number
...
+
+const NO_BODY = ADMIN_BODY_TYPE_ORDER.length;
+
+const bodyTypeRank = (type: AdministrativeBodyType | null | undefined): number => {
+    const index = type ? ADMIN_BODY_TYPE_ORDER.indexOf(type) : -1;
+    return index === -1 ? NO_BODY : index;
 };
 
+/** The seat that places a person: the first body type they sit on, and their elected order there. */
+function seat(person: OrderedPerson): { typeRank: number; electedOrder: number | null } {
```
kouloumos: `seat()` and `bodyRole()` read a person's seat by different rules. A person on two bodies of one type gets the lowest order across them here, but `sortPeople` places them under the first body by name with that body's order. Concrete case: A holds community-1 order 5 and community-2 order 1, B holds community-1 order 2. The party page gives `[A, B]`, the people page `[B, A]`. Staging has no such person today, and the second-seat branch on line 108 has no test, so a regression here produces no failure. One reader for both functions, and a test for this case, closes it.

src/lib/sorting/__tests__/people.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    });
+
+    it('does not mutate the input', () => {
+        const people = [person('b', 'B', [seat(COUNCIL, 2)]), person('a', 'A', [seat(COUNCIL, 1)])];
+        sortBodyMembers(people, COUNCIL.id);
+        expect(names(people)).toEqual(['b', 'a']);
+    });
+});
+
+describe('sortPartyMembers', () => {
```
kouloumos: Two mutations survive this suite: setting the surname comparison in `sortPartyMembers` to `0`, and deleting the second-seat branch in `seat()`. Both pass 16/16. The party rule's surname step is never decisive here, because every case has an order or a body difference above it. These two cases kill both mutants:

```ts
it('falls back to surname between members of the same body without an elected order', () => {
    const people = [
        person('omega', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COUNCIL)]),
        person('alpha', 'Άννα Άλφα', [partyRole(PARTY), seat(COUNCIL)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['alpha', 'omega']);
});

it('reads the lowest elected order across seats of the same body type', () => {
    const people = [
        person('two-committees', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COMMITTEE, 5), seat(COMMITTEE_B, 1)]),
        person('one-committee', 'Άννα Άλφα', [partyRole(PARTY), seat(COMMITTEE, 2)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['two-committees', 'one-committee']);
});
```

Also: the title at line 90 says "numbered or not", but a non-member cannot carry a number for that body.

src/components/persons/RolesList.tsx
```diff
@@ -85,19 +85,24 @@ export default function RolesList({ personId, cityId, roles, parties, administra
     const handleSubmit = (values: z.infer<typeof formSchema>, e: React.FormEvent) => {
         e.preventDefault(); // Prevent form submission
 
+        const administrativeBodyId = values.type === 'administrativeBody' ? values.administrativeBodyId || null : null;
+        // The elected order belongs to a seat on one body. It stays only while
+        // the role keeps that seat; a role moved to another body or type starts
+        // without one.
+        const keepsSeat = administrativeBodyId !== null && administrativeBodyId === editingRole?.administrativeBodyId;
```
kouloumos: This keeps the order alive on the client because `editPerson` deletes every role and recreates them from the form. Any other caller of the PUT route, which parses roles with `JSON.parse … as Role[]` and no schema, still wipes the field, as it wiped `rank`. The seat is already a unique key on `Role`. If `editPerson` carries `electedOrder` over by `(cityId, partyId, administrativeBodyId)` from the existing rows, the form can stop sending the field, `keepsSeat` disappears, and the invariant lives next to the write. That also stops role ids from churning on every edit, which the elected-order sheet keys on.

src/components/cities/CityPeople.tsx
```diff
@@ -93,9 +81,19 @@ export default function CityPeople({
         return null;
     }, [searchParams, allPeople, typeOptions]);
 
+    // The order follows the selection the same way the filter reads it: one
+    // body picked under a type, that body's order; one type picked, the bodies
+    // of that type in turn; otherwise the city order.
+    const sortItems = useCallback((items: PersonWithRelations[], selectedValues: AdministrativeBodyType[]) => {
+        const selectedType = selectedValues.length === 1 ? selectedValues[0] : null;
+        if (selectedType && selectedType !== 'council' && resolvedBodyId) return sortBodyMembers(items, resolvedBodyId);
```
kouloumos: `selectedType !== 'council'` now appears four times in this file (76, 89, 107, 128), and this condition is the filter's condition at 105–108 written again. `resolvedBodyId` is already null unless a non-council body matched. One `selectedBodyId(selectedValues)` computed once and used by both `filter` and `sortItems` reduces this to `bodyId ? sortBodyMembers(items, bodyId) : sortPeople(items, selectedType)`.

src/components/List.tsx
```diff
@@ -203,6 +211,7 @@ export default function List<T extends { id: string }, P = {}, F = string | unde
 
         return true;
     });
+    const filteredItems = sortItems ? sortItems(matchingItems, selectedFilters) : matchingItems;
```
kouloumos: `List.test.tsx` has no `sortItems` case. Three wiring mutants pass it: ignoring the result, sorting `items` instead of the filtered rows, and passing an empty selection. A two-test addition in the existing harness fails on all three: rendered order equals the returned order, and the hook receives the rows that passed search and filter together with the selection. I can paste it if you want it.

prisma/migrations/20260906120000_replace_party_rank_with_elected_order/migration.sql
kouloumos: Equal-size parties tie on `partyId`, a cuid. The page today uses the fetch order of `getPartiesForCity`, which has no `orderBy`. Both are arbitrary, so they need not agree, and "Papagos — one 3-member party changes" is this case. A tie-break a reader can predict, such as the party name, and one line in the comment would make it predictable.

src/lib/sorting/people.ts
kouloumos: `OrderedRole` re-spells Prisma `Role` fields by hand, and the two optional members widen the contract for no caller. Every call site passes people built with `roleWithRelationsInclude`, so `administrativeBody` is always loaded. If a caller ever omits it, every body ranks as `NO_BODY` and the city order degrades to surname without an error.

```ts
export type OrderedRole = Pick<Role, 'isHead' | 'cityId' | 'partyId' | 'administrativeBodyId' | 'electedOrder' | 'startDate' | 'endDate'> & {
    administrativeBody: Pick<AdministrativeBody, 'id' | 'name' | 'type'> | null;
};
```

Type check and all 16 tests pass with this.

src/lib/sorting/people.ts
kouloumos: Three of the private helpers here already exist, or have a home, elsewhere:

- `isMayor` is the sixth inline "holds an active mayor role" (`VotingSection.tsx:77`, `MeetingDecisionsPage.tsx:97`, `composition.ts:51`, `getMinutesData.ts:104`, `pollDecisions.ts:108`). `src/lib/utils/roles.ts` has the person-level pattern in `isActivePartyMember`; an exported `isMayor(person)` there lets the other five follow later.
- `lastName` is the fourth "last whitespace word" split. `src/lib/formatters/name.ts` does it three times and exports no surname getter.
- `bodyTypeRank` is the fourth private body-type ranking (`brochure.ts:19`, `db/coverage.ts:22`, the widget page). `ADMIN_BODY_TYPE_ORDER` lives in `administrativeBodies.ts`; a comparator next to it is the unit.

src/lib/sorting/people.ts
kouloumos: `seat()` and `bodyRole()` read a person's seat by different rules. A person on two bodies of one type gets the lowest order across them here, but `sortPeople` places them under the first body by name with that body's order. Concrete case: A holds community-1 order 5 and community-2 order 1, B holds community-1 order 2. The party page gives `[A, B]`, the people page `[B, A]`. Staging has no such person today, and the second-seat branch on line 108 has no test, so a regression here produces no failure. One reader for both functions, and a test for this case, closes it.

src/lib/sorting/__tests__/people.test.ts
kouloumos: Two mutations survive this suite: setting the surname comparison in `sortPartyMembers` to `0`, and deleting the second-seat branch in `seat()`. Both pass 16/16. The party rule's surname step is never decisive here, because every case has an order or a body difference above it. These two cases kill both mutants:

```ts
it('falls back to surname between members of the same body without an elected order', () => {
    const people = [
        person('omega', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COUNCIL)]),
        person('alpha', 'Άννα Άλφα', [partyRole(PARTY), seat(COUNCIL)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['alpha', 'omega']);
});

it('reads the lowest elected order across seats of the same body type', () => {
    const people = [
        person('two-committees', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COMMITTEE, 5), seat(COMMITTEE_B, 1)]),
        person('one-committee', 'Άννα Άλφα', [partyRole(PARTY), seat(COMMITTEE, 2)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['two-committees', 'one-committee']);
});
```

Also: the title at line 90 says "numbered or not", but a non-member cannot carry a number for that body.

src/components/persons/RolesList.tsx
kouloumos: This keeps the order alive on the client because `editPerson` deletes every role and recreates them from the form. Any other caller of the PUT route, which parses roles with `JSON.parse … as Role[]` and no schema, still wipes the field, as it wiped `rank`. The seat is already a unique key on `Role`. If `editPerson` carries `electedOrder` over by `(cityId, partyId, administrativeBodyId)` from the existing rows, the form can stop sending the field, `keepsSeat` disappears, and the invariant lives next to the write. That also stops role ids from churning on every edit, which the elected-order sheet keys on.

src/components/cities/CityPeople.tsx
```diff
@@ -67,9 +81,19 @@ export default function CityPeople({
...
```
kouloumos: `selectedType !== 'council'` now appears four times in this file (76, 89, 107, 128), and this condition is the filter's condition at 105–108 written again. `resolvedBodyId` is already null unless a non-council body matched. One `selectedBodyId(selectedValues)` computed once and used by both `filter` and `sortItems` reduces this to `bodyId ? sortBodyMembers(items, bodyId) : sortPeople(items, selectedType)`.

src/components/List.tsx
```diff
@@ -148,10 +211,16 @@ export default function List<T extends { id: string }, P = {}, F = string | unde
...
```
kouloumos: `List.test.tsx` has no `sortItems` case. Three wiring mutants pass it: ignoring the result, sorting `items` instead of the filtered rows, and passing an empty selection. A two-test addition in the existing harness fails on all three: rendered order equals the returned order, and the hook receives the rows that passed search and filter together with the selection. I can paste it if you want it.

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/10bead12fc70606e668f539ed4f04ea68e4456f5..866a12b6f802655e1d5080f5f0024b821808cb6d>) to put the branch back on the current head of `feature/city-page-redesign`.

`feature/city-page-redesign` was force-pushed after the last rebase, so this branch held its own copy of about 160 redesign commits under new SHAs. That is why the PR showed 278 changed files and a conflict. Replaying the two commits onto the current head brings it back to 33 files.

One conflict, resolved in favour of the deletion: `512c4d7` routed the membership check in `PartyMemberRankingSheet.tsx` through `isActivePartyMember`, and this PR deletes that file with the rest of the party-ranking feature.

No content of the two commits changed. Type check, lint, 165 test suites and the production build pass on the new base.

---

prisma/migrations/20260906120000_replace_party_rank_with_elected_order/migration.sql
kouloumos: Equal-size parties tie on `partyId`, a cuid. The page today uses the fetch order of `getPartiesForCity`, which has no `orderBy`. Both are arbitrary, so they need not agree, and "Papagos — one 3-member party changes" is this case. A tie-break a reader can predict, such as the party name, and one line in the comment would make it predictable.

src/lib/sorting/people.ts
kouloumos: `OrderedRole` re-spells Prisma `Role` fields by hand, and the two optional members widen the contract for no caller. Every call site passes people built with `roleWithRelationsInclude`, so `administrativeBody` is always loaded. If a caller ever omits it, every body ranks as `NO_BODY` and the city order degrades to surname without an error.

```ts
export type OrderedRole = Pick<Role, 'isHead' | 'cityId' | 'partyId' | 'administrativeBodyId' | 'electedOrder' | 'startDate' | 'endDate'> & {
    administrativeBody: Pick<AdministrativeBody, 'id' | 'name' | 'type'> | null;
};
```

Type check and all 16 tests pass with this.

src/lib/sorting/people.ts
kouloumos: Three of the private helpers here already exist, or have a home, elsewhere:

- `isMayor` is the sixth inline "holds an active mayor role" (`VotingSection.tsx:77`, `MeetingDecisionsPage.tsx:97`, `composition.ts:51`, `getMinutesData.ts:104`, `pollDecisions.ts:108`). `src/lib/utils/roles.ts` has the person-level pattern in `isActivePartyMember`; an exported `isMayor(person)` there lets the other five follow later.
- `lastName` is the fourth "last whitespace word" split. `src/lib/formatters/name.ts` does it three times and exports no surname getter.
- `bodyTypeRank` is the fourth private body-type ranking (`brochure.ts:19`, `db/coverage.ts:22`, the widget page). `ADMIN_BODY_TYPE_ORDER` lives in `administrativeBodies.ts`; a comparator next to it is the unit.

src/lib/sorting/people.ts
kouloumos, line 135: `seat()` and `bodyRole()` read a person's seat by different rules. A person on two bodies of one type gets the lowest order across them here, but `sortPeople` places them under the first body by name with that body's order. Concrete case: A holds community-1 order 5 and community-2 order 1, B holds community-1 order 2. The party page gives `[A, B]`, the people page `[B, A]`. Staging has no such person today, and the second-seat branch on line 108 has no test, so a regression here produces no failure. One reader for both functions, and a test for this case, closes it.

src/lib/sorting/__tests__/people.test.ts
kouloumos, line 161: Two mutations survive this suite: setting the surname comparison in `sortPartyMembers` to `0`, and deleting the second-seat branch in `seat()`. Both pass 16/16. The party rule's surname step is never decisive here, because every case has an order or a body difference above it. These two cases kill both mutants:

```ts
it('falls back to surname between members of the same body without an elected order', () => {
    const people = [
        person('omega', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COUNCIL)]),
        person('alpha', 'Άννα Άλφα', [partyRole(PARTY), seat(COUNCIL)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['alpha', 'omega']);
});

it('reads the lowest elected order across seats of the same body type', () => {
    const people = [
        person('two-committees', 'Ζήσης Ωμέγα', [partyRole(PARTY), seat(COMMITTEE, 5), seat(COMMITTEE_B, 1)]),
        person('one-committee', 'Άννα Άλφα', [partyRole(PARTY), seat(COMMITTEE, 2)]),
    ];
    expect(names(sortPartyMembers(people, PARTY))).toEqual(['two-committees', 'one-committee']);
});
```

Also: the title at line 90 says "numbered or not", but a non-member cannot carry a number for that body.

src/components/persons/RolesList.tsx
kouloumos, line 92: This keeps the order alive on the client because `editPerson` deletes every role and recreates them from the form. Any other caller of the PUT route, which parses roles with `JSON.parse … as Role[]` and no schema, still wipes the field, as it wiped `rank`. The seat is already a unique key on `Role`. If `editPerson` carries `electedOrder` over by `(cityId, partyId, administrativeBodyId)` from the existing rows, the form can stop sending the field, `keepsSeat` disappears, and the invariant lives next to the write. That also stops role ids from churning on every edit, which the elected-order sheet keys on.

src/components/cities/CityPeople.tsx
kouloumos: `selectedType !== 'council'` now appears four times in this file (76, 89, 107, 128), and this condition is the filter's condition at 105–108 written again. `resolvedBodyId` is already null unless a non-council body matched. One `selectedBodyId(selectedValues)` computed once and used by both `filter` and `sortItems` reduces this to `bodyId ? sortBodyMembers(items, bodyId) : sortPeople(items, selectedType)`.

src/components/List.tsx
```diff
@@ -211,6 +219,7 @@ export default function List<T extends { id: string }, P = {}, F = string | unde
...
```
kouloumos, line 222: `List.test.tsx` has no `sortItems` case. Three wiring mutants pass it: ignoring the result, sorting `items` instead of the filtered rows, and passing an empty selection. A two-test addition in the existing harness fails on all three: rendered order equals the returned order, and the hook receives the rows that passed search and filter together with the selection. I can paste it if you want it.

kouloumos (commented):
I ran the migration on a copy of the staging rows and replayed the base and head sort code on the same rows. The migration is careful: it wrote 58 council orders, touched no ended or non-council role, and left every pre-existing order alone.

**What changes versus production.** I ran today's sort code and the PR's on production's own rows for the 12 supported municipalities, with the migration replayed first for the three party-rank cities. The side-by-side per city, with every moved row highlighted, is here: [Elected Order Delta](https://claude.ai/code/artifact/5edec38e-dfa1-43f4-9455-e454a3b6d1d1).

The pattern is the same everywhere. Council view: mayor first, president second, otherwise today's surname order, with deputy mayors who hold no seat at the end. Party pages of default cities change from first-name order to surname order behind the mayor and the head. The party-rank cities keep their party pages, except two cases in Papagos: the 3-member party whose head was ranked third, which the description already names, and one ranked member without a council seat who moves from 11th to last. Seven current councillors across the three cities carry no party rank, so they get no number and sort after the numbered ones. The default landing changes from the council view to all people, which only matters for Athens, 45 to 155 cards. In Papagos-Cholargos the mayor's role lacks the head flag in production data, so the president sorts first there until that is fixed; a data fix, not this PR.

`role.rank` disappears from every public people and parties payload. OpenAPI is updated; nothing else tells an API consumer.

**Deploy note.** Migrations run at build time. Between the column drop and the container swap, the old client still selects `rank` on every role read. A low-traffic deploy, or a split into a code release and a later drop, avoids the window.

Verified on the preview: the rendered order on the Zografou and Papagos-Cholargos pages equals the sort code on every surface, with orders set through the API and the two scripts. A person edit through the real form recreated the role row and kept the order.

kouloumos:
The elected order for all 12 supported municipalities is now on this PR's preview: council orders from the 2023 election results, committee orders from each municipality's July 2026 committee election decision on Diavgeia. Per city, with match rates, decision ids and the names left by hand: [Elected Order Fill](https://claude.ai/code/artifact/d04fdf97-6985-4e46-a7b6-cb0f9395e103). Example: https://pr-724.opencouncil.dev/athens/people with the council chip.

The same run against production is a per-city runbook we hold ready for after the merge.

prisma/migrations/20260906120000_replace_party_rank_with_elected_order/migration.sql
christosporios: Fixed. The backfill now orders equal-size parties by party name, and the comment says why.

Replayed on a copy of the production rows: this is exactly the Papagos case. The two 3-member parties swap, so «8 Προτάσεις» now takes 17–19 and «Σύγχρονη Πολιτεία» 23–25. Counts, ranges and the absence of duplicates are unchanged for all four cities.

---

src/lib/sorting/people.ts
christosporios: Applied as written, with `administrativeBody` required. Type check and the suite pass.

---

src/lib/sorting/people.ts
christosporios: All three moved to the homes you name:

- `isMayor(person)` is exported from `src/lib/utils/roles.ts`, beside `isMayorRole`.
- `getSurname(name)` is exported from `src/lib/formatters/name.ts`.
- `administrativeBodyTypeRank` and `compareAdministrativeBodies` sit next to `ADMIN_BODY_TYPE_ORDER` in `administrativeBodies.ts`.

The sorting module reads all three. I left the other five mayor call sites, the other name splits and the two body-type copies alone, so this PR stays about ordering.

---

src/lib/sorting/people.ts
christosporios: Fixed. `seat()` now picks the person's placing body with `compareAdministrativeBodies`, the same comparator `sortPeople` walks the bodies with, and reads that body's order. So both surfaces read one rule, and your A/B case gives `[B, A]` on each.

The choice is the city order's, not the lowest number: a person's rank in one committee should not decide their place in another committee's list.

Confirmed no municipality has anyone holding two active seats on bodies of one type, so nothing visible changes today. A test now covers it, and fails if `seat()` goes back to taking whichever seat comes first.

---

src/lib/sorting/__tests__/people.test.ts
christosporios: Both cases added, and the title now reads «puts every member, numbered or not, before someone with no seat on the body».

I changed your second case, because the behaviour it asserts is the one the comment above reports as the inconsistency. `seat()` now takes the placing body rather than the lowest number, so the member of two committees is placed by their 5 in «Δημοτική Επιτροπή», not by their 1 in «Επιτροπή Ποιότητας Ζωής». The test asserts `['one-committee', 'two-committees']` and asserts `sortPeople` gives the same, so the two surfaces cannot drift apart again.

One detail worth keeping: I had to list the later body first in the fixture. With the roles in body order, "take whichever seat comes first" happens to pick the right one and the mutant survives.

Verified by mutation. Surname step to `0`: 1 failure. `seat()` taking the first seat found: 1 failure. Both pass without the new cases.

---

src/components/List.tsx
christosporios: Added, three cases in the existing harness: the rendered order equals the returned order, the hook receives only the rows that passed the search, and it receives the current selection. The rows now carry a `data-testid`, so document order is readable.

Verified against the three mutants you name: ignoring the result fails 3, sorting `items` fails 1, passing an empty selection fails 1.

---

src/components/cities/CityPeople.tsx
christosporios: Fixed. One `selectedBodyId(selectedValues)`, used by both `filter` and `sortItems`, so the rows on screen and the body that orders them cannot disagree. The condition is down from four occurrences to one; the remaining `=== 'council'` is in `renderAfterFilters`, where it answers a different question — whether to draw the sub-picker at all.

---

src/components/persons/RolesList.tsx
christosporios: Agreed on the diagnosis, and I am leaving this one open for the author to call, because it is the only change here that moves an invariant rather than tidying one.

You are right that `keepsSeat` guards the form, not the write. Any other caller of the PUT route still wipes the field, exactly as it wiped `rank`, and that is the bug this PR set out to end. Carrying the order over inside `editPerson`, keyed on `(cityId, partyId, administrativeBodyId)`, puts the invariant next to the write, lets the form stop sending the field, and stops role ids churning on every edit, which the elected-order sheet depends on.

What holds me back is that it changes `editPerson`'s contract for every caller, not only this form, so it wants its own diff and its own reading. I would rather ship the ordering here and do that next, unless the author wants it in this PR.

---

christosporios:
Thanks for the verification run and the delta — the two artifacts made this easy to check against.

I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/866a12b6f802655e1d5080f5f0024b821808cb6d..59701c82c50e3de1c988433b635b0f60f872c2eb>) with seven of the eight findings folded into the first commit:

- The backfill orders equal-size parties by party name. On the production rows this is the Papagos case: the two 3-member parties swap.
- `OrderedRole` is a `Pick` of the Prisma row, with `administrativeBody` required.
- `isMayor(person)`, `getSurname(name)` and `administrativeBodyTypeRank` / `compareAdministrativeBodies` moved to `roles.ts`, `name.ts` and `administrativeBodies.ts`. The other call sites stay for a later change.
- `seat()` reads the placing body through the same comparator `sortPeople` walks the bodies with, so the party list and the city list cannot order a person differently.
- Three sorting cases and three `List` cases added, each verified against the mutants you named.
- One `selectedBodyId` on the people page, used by both the filter and the order.

Left open: the `editPerson` carry-over. I agree with it and replied on the thread — it changes that function's contract for every caller, so I would rather it had its own diff.

Two things from your report I am not acting on here, both correct:

- The Papagos mayor's missing head flag is a data fix. It is in the SQL we prepared for that municipality.
- The deploy window between the column drop and the container swap is a release decision, not a code one.

Type check, lint, 165 test suites and the production build pass. The migration was replayed on a copy of the production rows for the four affected cities: same counts and ranges, no duplicates.

---

prisma/migrations/20260906120000_replace_party_rank_with_elected_order/migration.sql
```diff
@@ -0,0 +1,72 @@
...
+    JOIN "AdministrativeBody" body
+      ON body.id = council."administrativeBodyId" AND body.type = 'council'
+    JOIN "Person" person ON person.id = council."personId"
+    GROUP BY person."cityId"
+),
+numbered AS (
+    SELECT ranked.role_id,
+           ROW_NUMBER() OVER (
+               PARTITION BY ranked."cityId"
+               ORDER BY ranked.members DESC, ranked.party_name, ranked.rank, ranked.name
```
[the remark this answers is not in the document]
christosporios: Correct, and a regression from the previous push. The party name replaced the id rather than sitting in front of it, so two parties sharing a name fell through to the member keys. The id is back as the tie-break after the name, and the comment says why.

Reproduced before fixing, on two same-sized parties named «Ιδιο Ονομα» with two members each. Before: `Α Ενα, Β Ενα, Α Δυο, Β Δυο` — interleaved by rank. After: `Α Ενα, Α Δυο, Β Ενα, Β Δυο` — one block each.

Replayed on the production rows for the four affected cities as well: identical to the previous run, since no municipality has two parties sharing a name.

---

src/lib/sorting/__tests__/people.test.ts
```diff
@@ -0,0 +1,234 @@
+import {
+    compareRanks,
+    getElectedOrderForBody,
+    sortBodyMembers,
+    sortPartyMembers,
+    sortPeople,
+    sortPersonsByLastName,
+    type OrderedPerson,
+    type OrderedRole,
+} from '../people';
```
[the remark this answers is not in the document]
christosporios: Not changing this one. A test reaching its own subject through `../` is the repository's convention, and this file follows it.

The clearest evidence is the file beside it. `src/lib/sorting/__tests__/parties.test.ts` opens with:

```ts
import { sortParties } from '../parties';
import type { PartyWithPersons } from '@/lib/db/parties';
```

That is the same shape as here: the module under test relatively, everything else through `@/`. Across the test suite the count is 166 relative imports to 49 aliased ones.

I read the CLAUDE.md rule as being about reaching across the tree, which is where a relative path costs a reader. A test and its subject are one unit in one directory, and the `../` says so.

---

christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/f6ebcff858786215747bd37532ea6e4276d21c9b..b3151adc747019b01fe246012708cc96052da7fa>) to group a body's members by party in the code, instead of expecting the numbers to carry the grouping.

The preview showed the reason. A Greek municipal election awards seats across the parties, so the order of election interleaves them. On Orestiada the council read: Νέα Ξανά 1, an independent 2, Νέα Ξανά 3–8, ΠΡΩΤΗ ΓΡΑΜΜΗ 9, Νέα Ξανά 10. Every number was right, and the page was still wrong.

The rule for one body is now: the mayor, then the chair, then the parties in turn, then elected order inside the party, then surname. A party's block sits where its first-elected member sits, so the blocks follow the election and no count is needed. Members who hold no party close the list. `electedOrder` keeps its own meaning: the seat's order of election.

On the preview's Orestiada rows this gives the mayor, the πρόεδρος, the 14 remaining Νέα Ξανά members, the 7 of ΠΡΩΤΗ ΓΡΑΜΜΗ, then Δημοτική Επαναφορά, Αυτόνομη Κίνηση, Τόπος να ζω, then the three independents.

`OrderedRole` now carries `party`, so a block can be named rather than identified by id. Every caller already loads it through `roleWithRelationsInclude`.

Four cases cover the grouping, and all four fail when the block key is removed. Type check, lint, 165 test suites and the build pass.

One consequence to note: the minutes still sort on `electedOrder` alone, so a πρακτικό lists one order of election while the page lists party blocks. That is a separate decision, and I have not touched the minutes.

---

src/lib/sorting/people.ts
```diff
@@ -39,80 +73,192 @@ export const compareRanks = (aRank: number | null, bRank: number | null): number
...
+                // No party sorts after every party, so the independents close the list.
+                block: block ?? blocks.size,
+                electedOrder: role?.electedOrder ?? null,
+            };
+        },
+        (a, b) =>
+            compareFlags(a.mayor, b.mayor)
+            || compareFlags(a.member, b.member)
+            || compareFlags(a.chair, b.chair)
+            || a.block - b.block
```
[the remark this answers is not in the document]
christosporios: Intended, and it is the point of this push. A later-elected member of the first party appearing before an earlier-elected member of the next party is the grouping working.

A Greek municipal election awards seats across the parties, so the order of election interleaves them. On Orestiada the council reads Νέα Ξανά 1, an independent 2, Νέα Ξανά 3–8, ΠΡΩΤΗ ΓΡΑΜΜΗ 9. Every number is correct, and a list that followed them alone put an independent second and split the largest party around a rival. A municipality publishes its council by παράταξη, and Papagos-Cholargos asked for «τα μέλη ανά παράταξη σύμφωνα με τη σειρά εκλογής».

The stated order is not broken: the doc comment for this function changed in the same commit, and now reads mayor, chair, parties in turn, elected order inside the party, surname.

Your second point is the useful one. `ElectedOrderSheet` does list a body flat, and that stays right: the sheet edits `electedOrder`, which is the seat's own order of election and is body-wide by nature. The page groups that fact for a reader. So the two differ on purpose, and an admin who reorders in the sheet sees a different sequence on the page. I am raising that with the author as a question about the sheet, not changing the sort.

---

PR #733, opened 2026-09-09
perf(explain): pin the table of contents, defer the embeds

angelospk:
## What this changes

Two things on `/explain`, both about how the page behaves while you read it.

The table of contents no longer jitters. It held its position with a `transform` written from a rAF scroll handler. Scrolling runs on the compositor thread, so the column lagged the page and snapped back on almost every frame. Sampling its position on each frame of a scripted scroll from 1200px to 8000px, it was off its pin on 78 of 80 frames, by up to 678px. `position: sticky` pins it on the compositor instead, which measures 0 of 84 frames off the pin on a production build. It also drops a layout read and a style write from every scroll frame.

The old comment said an `overflow-hidden` ancestor broke sticky. That wrapper went away with the aurora background, so the workaround was no longer needed.

The agenda PDF and the TikTok embed now load near the viewport instead of on mount. They sit far below the fold, so their third-party weight was landing in the initial load for nothing.

## Notes for review

The TikTok embed owns its script tag. Each mount appends a tag, and it removes that same tag on unmount. `embed.js` scans the document when it runs, and it does not watch for blockquotes added later, so a client-side navigation back to this page needs a fresh element to re-run the script.

Where the browser has no `IntersectionObserver`, the script loads at mount instead. A reader who never gets the player is worse off than a reader who pays for it at load.

The blockquote stays in the server HTML as the accessible, indexable fallback.

## Testing

`TikTokEmbed.test.tsx`: the script stays away while the video is far from the viewport, it loads as the reader approaches, a mount removes only its own tag, a remount re-runs the script, the path without `IntersectionObserver` loads at mount, and the caption stays visible.

`AgendaViewer.test.tsx`: the agenda iframe carries `loading="lazy"`.

`npm run build` passes.

Checked in a browser against the production build. Stepping the scroll from 1500px to 7000px in 250px increments and reading the pinned column two frames later, all 23 samples sat at 96px.

<h2><a href="https://app.greptile.com/api/retrigger?id=62029145"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The pull request appears safe to merge, with no outstanding correctness or repository-rule issues.

<h3>Summary</h3>

- Makes the desktop table of contents sticky and independently scrollable on short screens.
- Keeps the active table-of-contents entry visible by adjusting only the column’s scroll position.
- Lazily loads agenda PDF iframes and TikTok’s embed script.
- Adds coverage for embed loading, cleanup, remounting, fallback behavior, and markup.

<sub>Reviews (7) · Last reviewed commit: ["fix(explain): keep active table of conte..."](https://github.com/schemalabz/opencouncil/commit/56fcaad4876c2edd302c39944c4f6080651fe46c)</sub>

### DIFF
--- a/src/app/[locale]/(other)/explain/ExplainReader.tsx
+++ b/src/app/[locale]/(other)/explain/ExplainReader.tsx
@@ -87,6 +87,30 @@ export function ExplainReader({
                 ?.closest("[data-toc-group]")
                 ?.querySelector("[data-toc-grouphead]")
                 ?.setAttribute("aria-current", "true");
+
+            // The pinned column scrolls its own content on a short screen, so the
+            // active entry can sit outside it. Move the column to the entry, and
+            // move nothing else: scrollIntoView walks every scrollable ancestor,
+            // and on this sticky box it drags the page instead of the column.
+            //
+            // Stop at the neighbour past the active entry, so the entry the reader
+            // wants never sits flush against the edge of the column. Entries wrap
+            // to two lines, so their heights differ and a fixed margin would cut a
+            // wrapped neighbour in half.
+            const aside = activeLink?.closest<HTMLElement>("[data-toc]");
+            if (aside && activeLink) {
+                const entries = [...aside.querySelectorAll<HTMLElement>("a")];
+                const i = entries.indexOf(activeLink);
+                const box = aside.getBoundingClientRect();
+                const link = activeLink.getBoundingClientRect();
+                if (link.top < box.top) {
+                    const lead = (i > 0 ? entries[i - 1] : null) ?? activeLink;
+                    aside.scrollTop += lead.getBoundingClientRect().top - box.top;
+                } else if (link.bottom > box.bottom) {
+                    const lead = (i >= 0 ? entries[i + 1] : null) ?? activeLink;
+                    aside.scrollTop += lead.getBoundingClientRect().bottom - box.bottom;
+                }
+            }
         }
 
         if (active >= 0) {
--- a/src/app/[locale]/(other)/explain/ExplainReader.tsx
+++ b/src/app/[locale]/(other)/explain/ExplainReader.tsx
@@ -33,14 +33,10 @@ export function ExplainReader({
     // on mobile so the current subtitle stays visible while reading the content.
     const [stickyIdx, setStickyIdx] = useState(-1);
 
-    // Track the section in view and pin the desktop table of contents.
+    // Track the section in view. The ToC itself pins with CSS position: sticky.
     useEffect(() => {
         const els = sections.map((s) => document.getElementById(s.id));
-        const aside = document.querySelector<HTMLElement>("[data-toc]");
-        const grid = aside?.parentElement ?? null;
         const OFFSET = 120; // spy threshold, below the fixed header
-        const PIN_TOP = 96; // where the ToC pins (matches the header offset)
-        const LG = 1024;
         let raf = 0;
 
         const compute = () => {
@@ -60,19 +56,6 @@ export function ExplainReader({
             // the Substack "further reading" section shouldn't pin its title
             const show = idx >= 0 && passed && sections[idx].id !== "substack";
             setStickyIdx(show ? idx : -1);
-
-            // pin the ToC: position: sticky is broken by an overflow-hidden
-            // ancestor, so translate it to hold PIN_TOP while it's in range.
-            if (aside && grid) {
-                if (window.innerWidth < LG) {
-                    aside.style.transform = "";
-                } else {
-                    const gridTop = grid.getBoundingClientRect().top;
-                    const max = Math.max(0, grid.offsetHeight - aside.offsetHeight);
-                    const t = Math.min(Math.max(0, PIN_TOP - gridTop), max);
-                    aside.style.transform = `translateY(${t}px)`;
-                }
-            }
         };
         const onScroll = () => {
             if (!raf) raf = requestAnimationFrame(compute);
@@ -85,7 +68,6 @@ export function ExplainReader({
             window.removeEventListener("scroll", onScroll);
             window.removeEventListener("resize", onScroll);
             if (raf) cancelAnimationFrame(raf);
-            if (aside) aside.style.transform = "";
         };
     }, [sections]);
 
--- a/src/components/embeds/TikTokEmbed.tsx
+++ b/src/components/embeds/TikTokEmbed.tsx
@@ -1,11 +1,20 @@
 "use client";
 
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
+
+const EMBED_SCRIPT = "https://www.tiktok.com/embed.js";
 
 /**
  * Embeds a TikTok video via TikTok's official embed.js, which replaces the
  * blockquote with an iframe once it loads. The blockquote (caption + links)
  * stays as accessible, indexable fallback until the script processes it.
+ *
+ * On a long page like /explain the video sits thousands of pixels below the
+ * fold, and booting the player at page load blocks the main thread while the
+ * reader is still reading the first sections. The script therefore loads once
+ * the reader comes within four viewports of the video. The player needs about
+ * two viewports of scrolling to appear on a slow device, so a shorter lead
+ * leaves the reader looking at the caption instead of the video.
  *
  * Reusable: pass the video id, its canonical URL (`cite`) and the caption
  * markup as children.
@@ -19,18 +28,53 @@ export function TikTokEmbed({
     cite: string;
     children: React.ReactNode;
 }) {
+    const holder = useRef<HTMLQuoteElement>(null);
+
     useEffect(() => {
-        const script = document.createElement("script");
-        script.src = "https://www.tiktok.com/embed.js";
-        script.async = true;
-        document.body.appendChild(script);
+        const el = holder.current;
+        if (!el) return;
+
+        // embed.js scans the document when it runs, and it does not watch for
+        // blockquotes added later. A client-side navigation back to this page
+        // mounts a blockquote the earlier run never saw, so each mount carries
+        // its own tag and takes it away again: appending a fresh element is
+        // what re-runs the script, and the file itself comes from the cache.
+        let script: HTMLScriptElement | null = null;
+        const load = () => {
+            script = document.createElement("script");
+            script.src = EMBED_SCRIPT;
+            script.async = true;
+            document.body.appendChild(script);
+        };
+
+        // Without an observer there is no way to tell how far the video is, and
+        // a reader who never gets the player is worse off than one who pays for
+        // it at load.
+        if (typeof IntersectionObserver === "undefined") {
+            load();
+            return () => script?.remove();
+        }
+
+        const observer = new IntersectionObserver(
+            (entries) => {
+                if (entries.some((e) => e.isIntersecting)) {
+                    load();
+                    observer.disconnect();
+                }
+            },
+            { rootMargin: "400% 0px" },
+        );
+        observer.observe(el);
+
         return () => {
-            script.remove();
+            observer.disconnect();
+            script?.remove();
         };
     }, []);
 
     return (
         <blockquote
+            ref={holder}
             className="tiktok-embed not-prose"
             cite={cite}
             data-video-id={videoId}
--- a/src/components/embeds/__tests__/AgendaViewer.test.tsx
+++ b/src/components/embeds/__tests__/AgendaViewer.test.tsx
@@ -1,0 +1,18 @@
+import { render } from '@testing-library/react';
+import { AgendaViewer } from '../AgendaViewer';
+
+// The lightbox and the localized Link are not under test here.
+jest.mock('../ZoomLightbox', () => ({ ZoomLightbox: () => null }));
+jest.mock('@/i18n/routing', () => ({
+    Link: ({ children }: { children: React.ReactNode }) => <a href="/">{children}</a>,
+}));
+
+describe('AgendaViewer', () => {
+    it('leaves the PDF to load as the reader approaches it', () => {
+        const { container } = render(
+            <AgendaViewer title="Ημερήσια Διάταξη" pdf="https://example.org/agenda.pdf" />,
+        );
+
+        expect(container.querySelector('iframe')).toHaveAttribute('loading', 'lazy');
+    });
+});
--- a/src/components/embeds/__tests__/TikTokEmbed.test.tsx
+++ b/src/components/embeds/__tests__/TikTokEmbed.test.tsx
@@ -1,0 +1,125 @@
+import { render, act, screen } from '@testing-library/react';
+import { TikTokEmbed } from '../TikTokEmbed';
+
+// jsdom has no IntersectionObserver. This stub keeps the observed callback, so
+// a test decides when the reader scrolls near the video.
+let triggers: IntersectionObserverCallback[] = [];
+const disconnect = jest.fn();
+
+class IntersectionObserverStub implements IntersectionObserver {
+    readonly root = null;
+    readonly rootMargin = "";
+    readonly thresholds: ReadonlyArray<number> = [];
+    private readonly callback: IntersectionObserverCallback;
+
+    constructor(callback: IntersectionObserverCallback) {
+        this.callback = callback;
+        triggers.push(callback);
+    }
+
+    observe() {}
+    unobserve() {}
+    takeRecords(): IntersectionObserverEntry[] {
+        return [];
+    }
+    // A disconnected observer hears nothing more. Without this, an unmounted
+    // component still answers a later scroll and loads a tag it cannot remove.
+    disconnect = () => {
+        triggers = triggers.filter((t) => t !== this.callback);
+        disconnect();
+    };
+}
+
+beforeEach(() => {
+    triggers = [];
+    disconnect.mockClear();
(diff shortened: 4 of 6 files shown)


angelospk:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/c030a01ce8ba44ede3a7ed84048f69b10eab84d5..a5fbd10483073d1ae7b583afbc76006524c34e92) to fix a bug a review found in the TikTok change:

- `embed.js` scans the document when it runs and does not watch for blockquotes added later. The old code loaded the tag at most once per document, so a client-side navigation back to `/explain` mounted a blockquote the earlier run never saw. I reproduced it against a production build: script tag present, blockquote never processed, no player. The tag is now replaced instead of kept, which re-runs the script; the file itself comes from cache. Verified the same way, the player now appears after navigating away and back.
- Updated the remount test, which asserted the old behaviour.

angelospk:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/a5fbd10483073d1ae7b583afbc76006524c34e92..3e04d5f35ffc7aa17ce4c31682adf840d0bf6328) to fix the `any` cast in the test.

`window.IntersectionObserver` now takes a named class that implements the interface, so the assignment type-checks on its own. The callback list is typed `IntersectionObserverCallback[]`, and the single entry the component reads is asserted once instead of per call.

On the `400%` rootMargin: I checked this in Chrome rather than reasoning from the CSS rule, because the same point came up in two reviews. Percentage values in `rootMargin` resolve per axis, so the vertical margins go against the root's height, not its width. Two measurements, both against a production build:

- Element root, 1200px wide by 300px tall, `rootMargin: "100% 0px"`. The observer fired when the target was 300px below the root, which is the height.
- Viewport root, 1440px by 902px, same margin. It fired at 900px below the viewport.

If it resolved against width, the first would have fired at 1200px and the second at 1440px. So `400% 0px` does give four viewport heights, and the comment above it is accurate. Happy to be shown otherwise if you have a case where it behaves differently.

src/components/embeds/TikTokEmbed.tsx
```diff
@@ -19,18 +28,39 @@ export function TikTokEmbed({
...
         };
+
+        const observer = new IntersectionObserver(
+            (entries) => {
+                if (entries.some((e) => e.isIntersecting)) {
+                    load();
+                    observer.disconnect();
+                }
+            },
+            { rootMargin: "400% 0px" },
```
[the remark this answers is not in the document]
angelospk: Checked this in Chrome rather than reasoning from the CSS percentage rule, since it came up in two separate reviews.

Percentage values in `rootMargin` resolve per axis: the vertical margins go against the root's height, the horizontal ones against its width. Two measurements against a production build:

| root | rootMargin | fired at | width | height |
|---|---|---|---|---|
| element, 1200x300 | `100% 0px` | 300px below | 1200 | 300 |
| viewport, 1440x902 | `100% 0px` | 900px below | 1440 | 902 |

If it resolved against width, the first would have fired at 1200px and the second at 1440px. So `400% 0px` does give four viewport heights on both a wide desktop and a tall phone, and the comment above it holds.

Leaving the code as is. Happy to reopen if you have a case where it measures differently.

src/components/embeds/TikTokEmbed.tsx
```diff
@@ -19,18 +28,42 @@ export function TikTokEmbed({
...
+        // and the file itself comes from the cache.
+        const load = () => {
+            document.querySelector(`script[src="${EMBED_SCRIPT}"]`)?.remove();
+            const script = document.createElement("script");
+            script.src = EMBED_SCRIPT;
+            script.async = true;
+            document.body.appendChild(script);
         };
+
+        const observer = new IntersectionObserver(
```
kouloumos, line 58: This observer drops the guard that `src/components/MountOnVisible.tsx` carries:

```ts
if (typeof IntersectionObserver === 'undefined') { setVisible(true); return; }
```

Without it, a browser with no `IntersectionObserver` never loads the player. The current code always loads it, so this is a regression. Please add the fallback and a test for it.

src/components/embeds/TikTokEmbed.tsx
```diff
@@ -19,18 +28,42 @@ export function TikTokEmbed({
...
+        const el = holder.current;
+        if (!el) return;
+
+        // embed.js scans the document when it runs, and it does not watch for
+        // blockquotes added later. A client-side navigation back to this page
+        // mounts a blockquote the earlier run never saw, so the tag is replaced
+        // rather than kept: inserting a new element is what re-runs the script,
+        // and the file itself comes from the cache.
+        const load = () => {
+            document.querySelector(`script[src="${EMBED_SCRIPT}"]`)?.remove();
```
kouloumos: This removes a script tag that another owner may have added, and the cleanup on line 61 no longer removes the tag this effect created. So the component deletes an element it does not own, and it leaks the one it does. Please keep the earlier ownership: create a tag, and remove that same tag on unmount.

A question about the comment above, not a request. I read `embed.js`, and its mount guard appears to sit in `sessionStorage` under a `mounting` key. Removing the script tag does not reset that key. If I read it correctly, an interrupted mount blocks every later mount for the rest of the tab session. What do you see? I may be misreading a minified file.

src/components/embeds/TikTokEmbed.tsx
```diff
@@ -19,18 +28,42 @@ export function TikTokEmbed({
...
         };
+
+        const observer = new IntersectionObserver(
+            (entries) => {
+                if (entries.some((e) => e.isIntersecting)) {
+                    load();
+                    observer.disconnect();
+                }
+            },
+            { rootMargin: "400% 0px" },
```
kouloumos, line 65: For the record, you are right here and the bot review is wrong. A vertical `rootMargin` percentage resolves against the root's height, not its width. At a 713px viewport, `entry.rootBounds.top` measures `-2852`, which is exactly 4 x 713.

src/components/embeds/__tests__/TikTokEmbed.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    it('loads embed.js as the reader approaches the video', () => {
+        renderEmbed();
+
+        approach();
+
+        expect(scripts()).toHaveLength(1);
+        expect(disconnect).toHaveBeenCalled();
+    });
+
+    it('loads embed.js once per document, whatever the number of embeds', () => {
```
kouloumos: This test asserts the de-duplication that the `remove()` call provides, so it goes with that change. `embed.js` already guards its own library by element id, so a second tag costs a cached request and nothing else.

src/components/embeds/__tests__/AgendaViewer.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { AgendaViewer } from '../AgendaViewer';
+
+// The lightbox and the localized Link are not under test here.
+jest.mock('../ZoomLightbox', () => ({ ZoomLightbox: () => null }));
+jest.mock('@/i18n/routing', () => ({
+    Link: ({ children }: { children: React.ReactNode }) => <a href="/">{children}</a>,
+}));
+
+describe('AgendaViewer', () => {
+    it('leaves the PDF to load as the reader approaches it', () => {
```
kouloumos, line 11: Not a blocker. This test asserts a static JSX attribute, and it needs two module mocks to do it. It can only fail if someone deletes the attribute, which the diff shows. I would drop the file, but I do not feel strongly.

src/app/[locale]/(other)/explain/ExplainReader.tsx
```diff
@@ -33,14 +57,68 @@ export function ExplainReader({
...
+            if (!(e.target instanceof Element)) return;
+            const link = e.target.closest<HTMLAnchorElement>('a[href^="#"]');
+            if (!link) return;
+            const id = fragmentId(link.getAttribute("href") ?? "");
+            // a link to the section already in the URL changes nothing, so no
+            // hashchange follows it — highlight here instead
+            if (id === fragmentId(window.location.href)) highlight(id);
+        };
+
+        if (arrival.current === undefined) arrival.current = fragmentId(window.location.href);
```
Comment (kouloumos) [inline src/app/[locale]/(other)/explain/ExplainReader.tsx]: For when this returns as its own PR. `arrival.current` is set once and never cleared, and line 107 runs on every pass of the effect. `sections` is the module constant `NAV_SECTIONS` today, so the effect runs once and nothing shows. If `sections` ever becomes locale-aware or built inline, the outline returns on every render, wherever the reader has scrolled. `arrival.current = null` after the first highlight closes it.

kouloumos (commented):
Thanks for this!

The sticky change is the right call. `position: sticky` is what this always wanted to be, and you checked that the workaround's premise is gone instead of assuming it. It is also the only part of the PR that makes the file smaller.

`77a63262` is worth more than the commit message claims. On production today, `/explain` sends five requests to TikTok before the reader scrolls at all: `embed.js`, two static bundles, an API call, and the full `embed/v2` player iframe. The video sits 38,475px down an 82,118px page. So the cost is not only the load. Every visitor's browser contacts TikTok, with an IP address and a referrer, whether or not the reader ever reaches the video. Your change limits that to the readers who get there.

I left the specifics inline. Two of them block the merge.

Last, please take `3e04d5f3` out of this PR. The arrival outline is outside what we scoped for the explain page, and it is not something we want now. This is a scope decision, not a quality one. The implementation is careful, and the tests document real browser behaviour.

One thing to weigh if you open it as its own issue. On desktop the table of contents already marks the section you arrive at. A cold load of `/explain#dimotiki-epitropi` sets `aria-current="true"` on that link and on its group heading, so the reader gets an orange marker with no new code. The gap is on mobile. There is no table of contents below `lg`, and the sticky subtitle bar cannot appear on arrival, because it waits for the heading to scroll above the top of the screen.

With the two inline fixes and `3e04d5f3` removed, this is ready to merge.

angelospk:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/3e04d5f35ffc7aa17ce4c31682adf840d0bf6328..b9f3e167c7e55c4e3c36e8dec1c6d4ee99913bf0) to drop the arrival outline and to fix the two blockers:

- Removed `3e04d5f3`. I keep it on a branch.
- A browser without `IntersectionObserver` now loads the script on mount. A test covers it.
- Each mount creates its own tag again, and removes that tag on unmount. The `remove()` call is gone.

You were right about the ownership, and the fix also removes the reason for the workaround. I read the first version of the commit again. It had `if (document.querySelector(...)) return;`, and it had dropped the `script.remove()` cleanup. The two together caused the bug I reported: the tag that leaked from the first visit made the guard return early on the second visit, so embed.js never ran again. The `remove()` call hid that cause instead of fixing it.

I checked this against production, which still has the ownership model. I drove Chrome to `/explain`, clicked through to `/about`, and went back, with no reload. I marked `window` first, and the mark survived. The player appeared after the return. So the bug does not reproduce on `main` on that path, and the ownership is enough on its own.

The same run measured the leaked tag. On `/about`, production holds 0 tags with this src. The preview held 1.

On `sessionStorage`: you read it correctly. The key is `__tt_embed__mounting`. The wrapper adds the `__tt_embed__` prefix to every key. `mount()` returns early while the value is `"true"`. It writes `"true"` before `checkLib()`, and it writes `"false"` only after `libHandle()` resolves. If `checkLib()` returns false, or if the page unloads in between, the value stays `"true"` for the rest of the tab session. Removal of the script tag does not reset it. The value was `"false"` in every measurement I took, on production and on the preview. This PR does not change the exposure, because `main` also re-runs the script on remount.

The test that asserted one tag per document is gone. It is an ownership test now: two embeds hold two tags, and an unmount of one removes only its own tag. The observer stub honours `disconnect()` now. Without that, an unmounted component still answered a later scroll.

I left `src/components/embeds/__tests__/AgendaViewer.test.tsx` in place. Tell me if you prefer it dropped, and I will remove it.

kouloumos (approved):
Both fixes are right. The observer path and the fallback path each create one tag, and each removes that same tag on unmount. The component no longer touches an element it did not create.

The new ownership test is better than the test it replaces. It renders two embeds and unmounts one, so it proves the property that matters. The old test only counted script tags.

I checked the preview. `/explain` loads with no request to TikTok. The blockquote and the caption stay in the server HTML. The player appears when the reader reaches the video.

One note on the remaining Greptile comment. It says the test casts to `any`. It does not. The file declares `class IntersectionObserverStub implements IntersectionObserver`, and `npx tsc --project tsconfig.jest.json --noEmit` passes in CI.

Thanks for measuring `rootMargin` in Chrome. Your numbers match mine.

src/components/embeds/TikTokEmbed.tsx
kouloumos: Fixed in `b9f3e167`. The fallback loads the script at mount and removes it on unmount, and a test covers both paths.

src/components/embeds/TikTokEmbed.tsx
kouloumos: Fixed in `b9f3e167`. Each mount creates its own tag and removes that same tag, so the component no longer touches an element it did not create. The `sessionStorage` question stays open. It does not block this PR.

src/components/embeds/TikTokEmbed.tsx
kouloumos: Settled. Your measurements match mine, and the bot withdrew the comment.

src/components/embeds/__tests__/TikTokEmbed.test.tsx
kouloumos: Done. The replacement test renders two embeds and unmounts one, so it checks the property that matters instead of a script count.

src/components/embeds/__tests__/AgendaViewer.test.tsx
kouloumos: Keeping the file. The point was optional, and the test costs nothing.

src/app/[locale]/(other)/explain/ExplainReader.tsx
Comment (kouloumos) [reply src/app/[locale]/(other)/explain/ExplainReader.tsx]: The commit is out of this PR, so this does not apply here any more. Keep it for the follow-up if you open one.

kouloumos:
Your reply on the `rootMargin` thread is the model. You measured it in Chrome rather than arguing from the rule, you gave the numbers, and you said plainly that you were leaving the code as it is. That answers a reviewer's three questions at once: what you found, how you know, and what you decided. It settled the disagreement in one pass, and the bot withdrew. Two notes on the rest.

You fixed both defects, and you fixed them well. You answered none of the six threads. So I read the whole diff again to work out what you changed, what you kept, and what you chose not to do. That last difference matters. A fix and a considered refusal look the same in a diff, and only a reply tells them apart. I also asked whether `embed.js` keeps its mount guard in `sessionStorage`. I still do not know if you checked. A reply of two lines costs you a minute, and it saves the reviewer a re-read.

I edited the title and the description. Both still promised the arrival outline. The description also still carried the demo video, a "Notes for review" section that explains only the outline's internals, and a "Testing" section counting ten tests in a file this branch no longer has. Greptile then reported a P1 saying the destination highlight was removed. The finding is wrong, and the description is what produced it. The cost is not only a bot. A person who opens this PR cold reads a promise the branch does not keep.

The description is a claim about the branch, so it changes when the branch changes. A force-push that drops a commit is the moment to re-read it.

The three-clause title is the root of it. One claim left, and the whole description became wrong at once. A PR that makes one claim stays accurate while it changes.

Both fixes landed clean, and the ownership test you wrote is better than the one it replaced. Keep the threads and the description in step with the code, and the next review is fast.

I replied on each thread and resolved them.

kouloumos:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/b9f3e167c7e55c4e3c36e8dec1c6d4ee99913bf0..4f0186b6879ce1cf3c001d7551889f3ab3d66cd0>) to rebase the branch onto `main` before the merge. The diff is unchanged.

kouloumos:
I pushed one commit to this branch, `56fcaad4`.

Andreas tested the pinned column on a laptop. It pins correctly, but it scrolls its own content on a short screen. At 1366x768 the list is 858px tall inside a 672px box, so 5 of the 21 entries sit below the fold of the column. The reader scrolls into a section whose entry is out of sight, and nothing brings it back.

The commit moves the column when the active section changes. It stops at the neighbour past the active entry, so the entry the reader wants does not sit flush against the edge.

Two measurements are worth keeping, because each one rejected the first thing I wrote.

`scrollIntoView({ block: "nearest" })` looks like the native answer here. It is wrong. It walks every scrollable ancestor, and on this sticky box it moved the page by 517px while it moved the column by 15px. A write to `scrollTop` moves the column alone: the same 15px, and the page stays where the reader left it.

A fixed margin fails as well. Entry heights are 28px, 31px and 51px, because 5 of the Greek titles wrap to two lines. A 40px margin shows a one-line neighbour and cuts a wrapped one in half. Reading the neighbour's rectangle adapts to whichever entry follows.

Both reasons sit in the code as comments, so neither gets simplified away later.

I also [force-pushed](<https://github.com/schemalabz/opencouncil/compare/046b10def946a01fc5cfca470b4d4bb99a51de94..56fcaad4876c2edd302c39944c4f6080651fe46c>) to fold the offset into that commit instead of stacking a second one.

PR #759, opened 2026-09-16
Register via QR code and record voiceprint consent

pkakelas:
## What this does

Superadmins can now download a PDF of QR codes for a city's council: one strip per council member, mayor and deputy mayor. We print it, cut the strips and leave one on each desk. The person scans their strip, signs in or signs up, and their account is linked to their person page. After that they can edit the page themselves.

The profile page also gets a checkbox for voiceprint consent. Only an account linked to a person sees it. We only store the answer for now. Nothing reads it yet.

Screenshot 2026-09-16 at 3 18 23 PM

Screenshot 2026-09-16 at 3 47 24 PM

## How it works

**Download.** Each row in `/admin/cities` has a QR button. It calls `GET /api/admin/cities/[cityId]/qr-strips` (superadmin only, never cached) and renders the PDF in the browser, like the offer and brochure PDFs. Each strip has the QR code, the name, the role, the last valid day and a warning that the code is personal. The texts use the city's language, because the councillors read them, not us.

**Codes.** A code is a signed token with the person id and an expiry. It is not stored anywhere. Codes are valid for 5 days, and every download creates new ones. Codes from an earlier download still work until they expire. The strip shows the day before the real expiry, so a code never stops working on a day the strip says it's valid.

**Claiming.** The QR code points to `/api/join/[token]`. If the person is signed out, they go through the normal magic-link sign-in and come back to the same link. Then we create an `Administers` row for the person and redirect to the profile, which shows what happened. A person can be claimed once. A second scan by another account is refused, and a race between two scans is handled by a serializable transaction. If the wrong account claims someone, fix it in `/admin/users`. Every successful claim posts to Discord.

The route is under `/api` on purpose. The token contains a dot, and the proxy matcher skips any path with a dot. A top-level `/join` would therefore skip the proxy, including basic auth on previews.

**Consent.** The consent is stored in its own table, `PersonVoicePrintConsent`, not as a column on `Person`. Our public people endpoints and pages return the whole `Person` row, so a column would have been public. A row means consent was given, with the time. Unticking deletes it. Only the account linked to the person can set it. A superadmin can't tick it for someone else.

## Worth a closer look

- **Signed tokens now have a kind.** The unsubscribe token and the claim token use the same secret and format. Without a kind, a claim code photographed from a strip would pass as an unsubscribe token with no user id. The unsubscribe route would then have unsubscribed every user, because Prisma drops an `undefined` filter. Each verifier now accepts only its own kind. Unsubscribe links in emails sent before this deploy have no kind, and they still work for their 30 days as long as they carry a user id.
- The relative redirect that `/qr` and the utterance route each built by hand now lives in `src/lib/utils/relativeRedirect.ts`, shared with the join route.

## Migration

`20260916080000_person_voiceprint_consent` creates one table. It changes no existing data.

## Testing

- Unit tests cover the tokens (kinds, expiry, tampering), the join route, claiming (including the retry), the strip data, the QR strips route, the consent rules and the profile form.
- Manually, against staging data: downloaded the PDF for Chania (45 strips, 7 pages) and checked that two downloads give different codes. Claimed a person with a real code, checked that the profile shows the result and that the person drops out of the next download, then removed the link.

<h2><a href="https://app.greptile.com/api/retrigger?id=64955767"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because the profile consent control still omits information needed for an informed voiceprint-consent decision.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Fcomponents%2Fprofile%2FUserInfoForm.tsx%3A262-266%0AThe%20profile%20now%20records%20voiceprint%20consent%20using%20only%20a%20short%20checkbox%20label.%20It%20no%20longer%20explains%20how%20the%20voiceprint%20will%20be%20created%20and%20used%20or%20that%20consent%20can%20be%20withdrawn%2C%20even%20though%20the%20join%20flow%20still%20provides%20that%20information.%20Someone%20granting%20consent%20from%20the%20profile%20therefore%20cannot%20make%20the%20same%20informed%20choice.%0A%0ANote%3A%20If%20this%20suggestion%20doesn't%20match%20your%20team's%20coding%20style%2C%20reply%20to%20this%20and%20let%20me%20know.%20I'll%20remember%20it%20for%20next%20time!%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=schemalabz%2Fopencouncil&pr=759&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Consent lacks essential context** <a href="https://github.com/schemalabz/opencouncil/pull/759#discussion_r4048721554">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
src/components/profile/UserInfoForm.tsx:262-266
The profile now records voiceprint consent using only a short checkbox label. It no longer explains how the voiceprint will be created and used or that consent can be withdrawn, even though the join flow still provides that information. Someone granting consent from the profile therefore cannot make the same informed choice.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

---

`````

</details>

<h3>Summary</h3>

This PR adds QR-based person claiming and records voiceprint consent periods.
- Generates localized council-member QR-strip PDFs through a superadmin-only, non-cacheable endpoint.
- Adds a magic-link-backed join flow with signed, expiring, identity-bound confirmation markers.
- Introduces claimed-person relationships and auditable voiceprint consent periods.
- Adds profile consent controls, localized copy, database migrations, and test coverage.

<sub>Reviews (15) · Last reviewed commit: ["fix(auth): bind the email grace mark to ..."](https://github.com/schemalabz/opencouncil/commit/cb4b7d700cdf02df5536429215fae96a68fcd6c7)</sub>

### CODE REVIEW

File: src/components/admin/cities/CouncilQrStripsButton.tsx [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+        download(async () => {
+            const response = await fetch(`/api/admin/cities/${encodeURIComponent(cityId)}/qr-strips`, { cache: "no-store" });
+            if (!response.ok) throw new Error(`QR strips request failed with ${response.status}`);
+            const strips: CouncilQrStrips = await response.json();
+            if (strips.people.length === 0) {
+                toast({ description: t("empty") });
+                return null;
+            }
+            // Lazy, as the other PDFs: @react-pdf/renderer stays out of the admin bundle.
+            const { CouncilQrStripsPdf } = await import("./council-qr-strips-pdf");
```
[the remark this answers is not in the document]
pkakelas: This follows the existing pattern for client-side PDFs (offer letter and brochure). council-qr-strips-pdf pulls in @react-pdf/renderer, which is about 500 KB gzipped. A static import would add that to every load of /admin/cities and would also undo the lazy import inside usePdfDownload. CLAUDE.md allows a dynamic import when it is necessary, and the comment above the import states the reason.

pkakelas:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/7125b4af6044be7946edec55b70a38e04344f5d7..aae9f32fc8daf1410b9ff4faa361db29efa53390) to remove the "Αναγνώριση φωνής" heading above the voiceprint consent box, along with its unused message in the four locales.

prisma/schema.prisma
```diff
@@ -280,6 +282,17 @@ model TaskStatus {
   @@index([councilMeetingId, cityId])
 }
 
+/// The person's own consent to a voiceprint, ticked on the profile by the
+/// account that administers them. A row is the consent; withdrawal deletes
+/// it. Its own table, not a column on Person: every public read of a person
+/// (the API, the pages) carries the whole row, and consent is not public.
+/// Recorded only for now: nothing reads it before generation or transcription.
+model PersonVoicePrintConsent {
```
kouloumos: **The problem.** A row is the consent and a withdrawal deletes it, so the table can hold
only the current answer. It cannot say who gave the consent, and after a withdrawal it
cannot say that a consent ever existed. Art. 7(1) asks us to demonstrate the consent, and
Art. 7(3) only helps if we can still show the consent was in force for the period we
processed in.

**A suggestion.** Store the period instead of the flag. `Role` already uses this pattern
here, and officer rotation depends on it.

```prisma
model VoicePrintConsent {
  id          String    @id @default(cuid())
  personId    String
  person      Person    @relation(fields: [personId], references: [id], onDelete: Cascade)
  userId      String?
  user        User?     @relation(fields: [userId], references: [id], onDelete: SetNull)
  givenAt     DateTime  @default(now())
  withdrawnAt DateTime?
  @@index([personId, givenAt])
}
```
```sql
CREATE UNIQUE INDEX "VoicePrintConsent_open_key"
  ON "VoicePrintConsent"("personId") WHERE "withdrawnAt" IS NULL;
ALTER TABLE "VoicePrintConsent" ADD CONSTRAINT "VoicePrintConsent_period_valid"
  CHECK ("withdrawnAt" IS NULL OR "withdrawnAt" > "givenAt");
```

The `onDelete` values are deliberately different. A deleted person removes the `VoicePrint`
rows, so no processing remains to justify. A deleted account must not erase the proof that
processing was lawful while it ran.

src/lib/db/personConsent.ts
```diff
@@ -0,0 +1,39 @@
...
+    const user = await getCurrentUser();
+    if (!user) throw new UnauthorizedError("Not signed in");
+    if (!user.administers.some((a) => a.personId === personId)) {
+        throw new ForbiddenError("Only the person's own account can give voiceprint consent");
+    }
+
+    if (consent) {
+        await prisma.personVoicePrintConsent.upsert({ where: { personId }, create: { personId }, update: {} });
+    } else {
+        await prisma.personVoicePrintConsent.deleteMany({ where: { personId } });
```
kouloumos: **The problem.** `deleteMany` removes the record of the consent along with the consent
itself. After a councillor withdraws, we cannot show that their consent covered the period
in which we generated their voiceprint.

**A suggestion.** With the period table above, a withdrawal closes the period:
```ts
await prisma.voicePrintConsent.updateMany({
    where: { personId, withdrawnAt: null },
    data: { withdrawnAt: new Date() },
});
```

src/lib/db/personConsent.ts
```diff
@@ -0,0 +1,39 @@
...
+    // A server action has no runtime types: a stale client once sent undefined.
+    if (typeof consent !== "boolean") throw new BadRequestError("consent must be a boolean");
+    const user = await getCurrentUser();
+    if (!user) throw new UnauthorizedError("Not signed in");
+    if (!user.administers.some((a) => a.personId === personId)) {
+        throw new ForbiddenError("Only the person's own account can give voiceprint consent");
+    }
+
+    if (consent) {
+        await prisma.personVoicePrintConsent.upsert({ where: { personId }, create: { personId }, update: {} });
```
kouloumos: **The problem.** `upsert` with an empty `update: {}` is not always a single statement. When
Prisma falls back to find-then-create, two concurrent ticks both find nothing, and the loser
gets `P2002`. Nothing catches it, so `UserInfoForm` shows the generic consent error for what
was a success.

**A suggestion.** The partial unique index above removes this case: a repeat tick becomes a
`create` that fails only when a period is already open, which is the right answer. If you
keep the current table instead, `createMany({ data: [{ personId }], skipDuplicates: true })`
is one `INSERT ... ON CONFLICT DO NOTHING` and matches the stated semantics.

src/components/profile/UserInfoForm.tsx
```diff
@@ -73,27 +90,48 @@ export function UserInfoForm({ user, isOnboarded }: UserInfoFormProps) {
...
         e.preventDefault();
         if (phoneSubmitBlocked) return;
-        await saveToApi({
+        const saved = await saveToApi({
             name: formData.name,
             phone: phoneValidity.isEmpty ? null : formData.phone,
             onboarded: true,
-        });
+        }, { refresh: persons.length === 0 });
+        if (saved && persons.length > 0) await saveConsents();
```
kouloumos: Reproduced on the preview: a consent withdrawal is silently discarded when an unrelated
field fails.

Steps: untick the consent box, enter a phone number that another account already holds,
then save. `saveToApi` returns `false`, so `saveConsents()` never runs. No error appears
for the consent. `consentEdits` is not cleared and `router.refresh()` is skipped, so the
box keeps showing unchecked. A fresh server render of the same page returns
`data-state="checked"`. The councillor believes they withdrew consent. They did not.

The root cause is that the consent shares a submit button with the account fields. Give
the consent its own section and its own save. That also answers Art. 7(2), which asks
for a consent request inside a wider declaration to be clearly distinguishable.

src/components/profile/UserInfoForm.tsx
```diff
@@ -48,16 +57,24 @@ export function UserInfoForm({ user, isOnboarded }: UserInfoFormProps) {
     const [serverPhoneError, setServerPhoneError] = useState<string | null>(null);
 
     const [formData, setFormData] = useState({
-        name: user.name || "",
+        // A councillor who signed up through their QR has no name yet: start
+        // from the name on their council record. Saved only with the form.
+        name: user.name || persons[0]?.name || "",
```
kouloumos: @greptile-apps raised this on 2026-09-16 and the thread is still open. It is real:
`currentUser()` loads `administers` with no `orderBy`, so `persons[0]` is whatever Prisma
returns. The population is narrow and the value is editable before it persists, so I read it as
a nit rather than a blocker.

One angle to add: the prefill keys off `persons`, not off the claim. A person admin that a
superadmin created, who never scanned a QR, also gets the council name pre-filled and the save
button pre-enabled on first load. `persons.length === 1 ? persons[0].name : ""` covers the case
this exists for.

src/lib/utils/roles.ts
```diff
@@ -414,6 +414,49 @@ export function isMayorRole(role: { isHead: boolean; cityId?: string | null; par
...
+/**
+ * Whether a city-level role is a deputy mayor, by its title: Αντιδήμαρχος and
+ * Αναπληρωτής Δήμαρχος, adjoint(e) au maire ("1er adjoint délégué …"),
+ * заменик градоначелника. The schema has no field for it; the title is all
+ * there is. An unknown title is not a deputy mayor.
+ */
+export function isDeputyMayorRole(role: TitledCityRole): boolean {
+  if (!role.name || !isCityLevelRole(role)) return false;
+  const title = normalizeRoleTitle(role.name);
+  return title.includes('αντιδημαρχ')
```
kouloumos: **These two predicates are redundant.** An Αντιδήμαρχος is appointed from among the δημοτικοί
σύμβουλοι, so a deputy mayor always holds a council seat as well. The same is true of a
δημοτικός σύμβουλος by definition. The `council` administrative-body branch above already
admits both.

The data agrees. I measured all four routes into `isCouncilMemberOrMayor` across every city on
the preview:

| | count |
|---|---|
| eligible for a strip | 337 |
| matched by `isMayor` | 10 |
| matched by a `council` body role | 329 |
| matched by `isDeputyMayorRole` | 82 |
| matched by `isCouncillorTitleRole` | 7 |
| **qualify only via the deputy-mayor title** | **0** |
| **qualify only via the councillor title** | **0** |

Both predicates fire, so the check is live. Neither admits one person the council body role does
not already admit.

The branch came from a Greptile comment speculating that a deputy mayor might hold only a
city-level role. That person does not exist.

**Why it is not free.** Title matching is what lets these through, and each one gets a code that
permanently links a personal account to a councillor's page:

| title | isDeputyMayorRole |
|---|---|
| `Γραμματέας Αντιδημάρχου` (secretary **to** the deputy mayor) | true |
| `Σύμβουλος Αντιδημάρχου` (adviser **to** the deputy mayor) | true |
| `Πρώην Αντιδήμαρχος` (**former** deputy mayor) | true |
| `Adjoint technique` (French staff grade) | true |
| `Adjointe administrative` | true |

`title.includes('αντιδημαρχ')` has no anchor, and `/^(\d+(er|re|e)\s+)?adjointe?\b/` matches
any French "adjoint" grade. The docstring says city staff must not be able to claim a page, and
these titles defeat that.

**A suggestion.** Remove `isDeputyMayorRole`, `isCouncillorTitleRole` and the private helpers they
need. `isCouncilMemberOrMayor` reduces to:

```ts
return isMayor(person) || person.roles.some((r) => r.administrativeBody?.type === "council");
```

That changes who gets a strip by zero people and removes the false-positive surface entirely.
Tightening the regexes instead would also work, but it keeps a matcher we do not need and invites
the next locale to be added as another pattern.

**One case this does not fix, and should not be fixed here.** Београд returns zero strips today.
Its two people carry a city-level `Одборник` role and no council body, so they fall through all
four branches — `isCouncillorTitleRole` has Greek and French patterns only. Removing the
predicates leaves that unchanged. The fix there is to record council-body roles for Serbian
cities, not to grow the title regexes, which is what the current design invites.

**One thing worth keeping.** `normalizeRoleTitle` exists because the data carries `Αντιδήµαρχος`
with a U+00B5 micro sign. `simplifyRoleName` in this same file is the pre-existing deputy matcher
— `CouncilMeetingDocx.tsx:114` uses it to shorten a portfolio title to `Αντιδήμαρχος` in the
minutes — and it still matches with a plain `indexOf` and a normal μ:

```
"Αντιδήμαρχος Οικονομικών"  simplifyRoleName -> "Αντιδήμαρχος"
"Αντιδήµαρχος Οικονομικών"  simplifyRoleName -> "Αντιδήµαρχος Οικονομικών"   <- not shortened
```

A councillor with the micro sign gets their full portfolio printed in the minutes. That is a
pre-existing bug, but this PR is what found the cause. NFKC folds µ to μ without changing length,
so `name.normalize('NFKC').indexOf('Αντιδήμαρχος')` fixes it and still gives a valid offset into
the original text. Worth carrying over even as the rest of the matcher goes.

I could not confirm the micro sign in data I can reach — zero instances across Chania, Athens,
Argos, Sparta and Zografou on the preview — so that part rests on the comment in the PR.

src/lib/auth/__tests__/personClaim.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    it('rejects an expired token', () => {
+        const now = Date.now();
+        const spy = jest.spyOn(Date, 'now').mockReturnValue(now);
+        const token = generatePersonClaimToken('person-1');
+        spy.mockReturnValue(now + 6 * 24 * 60 * 60 * 1000);
+        expect(verifyPersonClaimToken(token)).toBeNull();
+        spy.mockRestore();
+    });
+
+    it('is not interchangeable with an unsubscribe token, in either direction', async () => {
```
kouloumos: This test does not test the kind check. I deleted the `kind` line from `verifyPayload`
and all 23 tests in `personClaim` and `tokens` still passed. Each verifier's own shape
guard (`typeof data.userId`, `typeof data.personId`) rejects the other payload first.
Deleting the `timingSafeEqual` comparison also leaves all 8 `personClaim` tests green.
The two checks mask each other, and only removing both fails a test.

Assert on `verifyPayload` directly, where no wrapper guard intervenes:
```ts
expect(verifyPayload('unsubscribe', signPayload('person-claim', { exp }))).toBeNull();
```
`signedPayload.ts` has no test file of its own, and it is the primitive both tokens rest on.

src/lib/admin/councilQrStrips.ts
```diff
@@ -0,0 +1,99 @@
...
+ * query below asks for those.
+ */
+/**
+ * Who gets a strip: a council member, the mayor or a deputy mayor. A council
+ * seat is enough on its own; without one, the city-level title decides. Not
+ * every city-level role: a General Secretary is city staff, not elected, and
+ * must not be able to claim a page, so an unrecognised title gets no strip.
+ * The roles are the active ones already.
+ */
+function isCouncilMemberOrMayor(person: PersonWithRelations): boolean {
```
kouloumos, line 40: Two docblocks are stacked here. The first one ("The title under a name on a strip…")
describes `roleLabel`, which is 9 lines below. Move it there.

src/lib/discord.ts
```diff
@@ -188,6 +188,28 @@ export async function sendUserOnboardedAdminAlert(data: {
...
+export async function sendPersonClaimedAdminAlert(data: {
+    cityId: string;
+    personName: string;
+}): Promise<void> {
+    await sendAdminAlert({
+        title: `🪪 Person Claimed - ${data.cityId}`,
+        description: `A councillor linked their account through a claim QR. Review the link under ${env.NEXTAUTH_URL}/admin/users.`,
+        color: 0x1abc9c, // Teal
+        fields: [
+            { name: 'Municipality', value: data.cityId, inline: true },
```
kouloumos: `cityId` where the sibling alerts pass `cityName`.

src/app/[locale]/(other)/profile/page.tsx
```diff
@@ -17,9 +20,24 @@ export const metadata: Metadata = {
...
+        name: p.name,
+        voicePrintConsent: consented.has(p.id),
+    }));
+    // A success needs the link to exist: the status is a query parameter,
+    // and anyone can type one.
+    const claimKey = claimMessageKey(claim);
+    // A first claim needs no notice: the profile itself is the answer. A code
+    // scanned again says so. "Already yours" needs the link to exist, since
+    // the status is a query parameter and anyone can type one.
+    const claimShown = claimKey === "alreadyYours" ? persons.length > 0 : claimKey !== null && claimKey !== "linked";
```
Comment (kouloumos) [inline src/app/[locale]/(other)/profile/page.tsx]: A successful claim shows nothing. I checked all five statuses on the preview:
`already_linked`, `already_yours`, `invalid` and `not_found` each render a notice, and
`linked` renders none. The one outcome that changed state is the only silent one.

`?claim=linked` also stays in the address bar forever, because `DropClaimParam` only mounts
with a notice. Render it whenever `claim` is present.

src/lib/admin/__tests__/councilQrStrips.test.ts
```diff
@@ -0,0 +1,115 @@
...
+            person('committee-only', 'Γιώργος Παπαδάκης', [
+                role({ administrativeBodyId: 'body-committee', administrativeBody: { id: 'body-committee', name: 'Επιτροπή', type: 'committee' } }),
+            ]),
+        ]);
+        mockGetClaimedPersonIds.mockResolvedValue(new Set(['claimed']));
+
+        const strips = await getCouncilQrStrips('chania');
+
+        expect(strips!.cityName).toBe('Χανιά');
+        expect(strips!.people.map((p) => p.id).sort()).toEqual(['councillor-title-only', 'deputy', 'deputy-title-only', 'mayor', 'member']);
```
kouloumos: **The problem.** The sheet is meant to print in the same order the people page shows — the
docstring on `getCouncilQrStrips` says so, and it is what makes a printed stack usable when you
walk the chamber. I checked it against real Chania data: `sortPeople(filter(people), "council")`
gives exactly the same 45 names, in the same order, as the people page produces both in its
default view and on its council tab. So the property holds today.

Nothing protects it. `.map((p) => p.id).sort()` sorts the ids before comparing, so this assertion
cannot see order at all. Replacing `sortPeople` with `.reverse()` in `councilQrStrips.ts:85` fails
no test in the suite.

**A suggestion.** Drop the `.sort()` and assert the array as it comes, on a fixture with a mayor
and two members whose `electedOrder` differs. That pins the one property the sheet needs and the
current assertion is blind to.

kouloumos (commented):
The QR and claim half is good. I drove it end to end on the preview against real Chania
data: `linked`, `already_yours`, `already_linked` and the signed-out spent code all land
where they should, a live claim token is refused by the unsubscribe page, and the Athens
sheet renders as 7 pages of 46 strips with the codes drawn as vectors. The `kind` change
fixes a real hole. The serializable transaction is the right primitive. I have no concerns
about shipping that part.

I have one substantial point, and it is about the consent table.

### What the current shape records

`PersonVoicePrintConsent` is `(personId, consentedAt)`. A row is the consent. A withdrawal
deletes the row.

That shape records that a consent exists, and when. It does not record who gave it. It does
not record that a withdrawal happened.

### Why I think that is a problem

Article 9 consent has to be demonstrable. Art. 7(1) asks us to show that the data subject
consented. Art. 7(3) says a withdrawal does not affect the lawfulness of processing before
it — but to rely on that, we must still show the consent was in force during that period.
A deleted row shows nothing.

Three consequences follow. I reproduced each one on the preview:

- **No actor.** A superadmin gave themselves an `Administers` row over a councillor through
  `/admin/users`, then recorded that councillor's consent. The resulting row is identical to
  a self-given one. The PR description says a superadmin cannot tick it for someone else.
  Two API calls in the same admin area do exactly that.
- **A withdrawal erases the evidence.** After a withdrawal we cannot show that consent
  covered the period in which we generated a voiceprint.
- **Consent outlives the link that authorised it.** Consent given under account A, then A
  unlinked, then the person handed to account B: B's profile shows the box already ticked.
  B never consented.

None of these is a defect in the code you wrote. They all follow from the shape of the table.

### A suggestion

Store a period instead of a flag. `Role` already does this in our schema, and officer
rotation already relies on it — we close a role rather than delete it. The shape is inline
on the new model.

A withdrawal then becomes an `UPDATE`, never a `DELETE`. "Was consent in force on 14 March?"
becomes one predicate. A partial unique index keeps at most one open period per person, which
also removes the `upsert` race noted inline.

To keep this to one new table rather than two, the identity fact could live on `Administers`:
`claimedAt DateTime?`, set only by `claimPerson`, plus
`CREATE UNIQUE INDEX ... ON "Administers"("personId") WHERE "claimedAt" IS NOT NULL`. That
separates "this account is this person" from "this account may edit this person". The consent
check needs that distinction and the schema does not make it today. It also fixes a side
effect I measured: a delegate link silently removes a councillor from the strip sheet,
because `getClaimedPersonIds` cannot tell the two apart.

The timing argument is the one I would weigh most. There are zero consent rows today, so the
shape is free to change now. Later it becomes a data migration on Article 9 records, and we
cannot go back and establish who gave a consent we already collected.

### On sequencing

My suggestion is to ship the QR and claim flow now, and hold the consent box. Nothing reads
the flag yet, so holding it costs nothing. It also resolves the mismatch where the copy says
"Μπορείτε να το ανακαλέσετε ανά πάσα στιγμή" while `generateVoiceprint` never reads the value.

One more point, which I think has the same root cause. The consent control sits inside the
account form and shares its submit button. That causes the `UserInfoForm.tsx` bug inline, and
it is also what Art. 7(2) asks us to avoid: a consent request inside a declaration about other
matters has to be clearly distinguishable. The heading that the force-push removed was doing
some of that work. A separate section with its own save would fix the bug and the
distinguishability together.

Everything else is inline.

src/app/[locale]/(other)/profile/page.tsx
Comment (pkakelas) [reply src/app/[locale]/(other)/profile/page.tsx]: A successful claim lands on the profile, which is also what you see when you just open the page. That is the default outcome, so a banner seems too much for me. 

The second point is fixed in 53ed9f6d: `?claim=` is now dropped from the address bar even when no notice shows.

src/lib/utils/roles.ts
pkakelas: Τι να κάνεις, να κάτσεις να σκάσεις;

pkakelas:
I pushed a new join flow for the QR code ([compare](https://github.com/schemalabz/opencouncil/compare/8423280395c6a0f762a7bb857b9633777c9311d4..97d5ae6faa8ee1363c2fbb4f98970268116cf2ca)). Before, the QR opened sign-in and then the profile. Now it opens a flow in the style of the petition and notification signups: one question on each screen, one main button, plain words.

**The steps**

1. **"Είστε εσείς;"** The screen shows the name, the title and the city. "Δεν είμαι εγώ" opens a screen that asks the reader to give the paper to the correct person. That screen has a "back" button.
2. **"Πείτε μας το email σας".** The form catches a typo in the domain (for example `gmial.com`) and offers the fix. After the send, a screen tells the reader what to do in three numbered lines. The reader can resend after 30 seconds or change the email.
3. **Voiceprint consent.** Two answers with the same weight: "Ναι, συναινώ" and "Όχι τώρα". Both answers finish the flow. "Όχι τώρα" writes nothing.

A reader who is already signed in skips the email and sees two steps. The done screen does not repeat the consent answer.

**Layout**

- The flow is one column in the centre of the page, on a phone and on a desktop.
- On a phone, the buttons stay at the bottom of the screen, above the home indicator. On a desktop, they stay under the step.
- The other signup flows do not change.

**Codes that cannot continue**

- An expired or incorrect code shows "Αυτός ο κωδικός QR δεν ισχύει".
- When the person already has an account, a new scan shows "Αυτός ο κωδικός QR δεν ισχύει πλέον". This applies to all readers, the owner too. The owner gets a link to the profile. The other readers get the contact line, and a sign-in link when they are signed out.

**Other changes**

- The QR URL is now `/{cityId}/join?c=<code>`. The code is shorter, so the QR grid went from 69×69 to 41×41. Phones read it more easily. Codes printed before this change do not work.
- The claim completes the account: it sets the confirmed name if the account has none, and it marks the account as onboarded. No registration form shows on the profile after the flow.
- A reload, a second phone or the link in the email opens the correct step.
- If the code expires while the email is in transit, the link in the email still works for 24 hours. The server signs this link, so a person cannot add the grace to an expired code by hand.
- The consent write now checks the claim inside a serializable transaction.

**How to test**

Download the PDF from `/admin/cities`, scan a strip, and go through the steps signed out and signed in. Then scan the same strip again to see the "no longer valid" screen.

src/lib/personJoin/paths.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The path of the join flow for a code. The token rides in the query, not
+ * in the path: it holds a dot, and the proxy skips every dotted path, so a
+ * page under /join/<token> would lose locale routing on the other realms.
+ * No server imports: the flow's client components build this path too.
+ */
+export function personJoinPagePath(cityId: string, token: string): string {
```
kouloumos: **The problem.** The code now rides in the query of a normal page, so it reaches the analytics
scripts. I loaded a real QR URL on the preview and captured what Plausible sent:

```json
{"n":"pageview","v":36,
 "u":"https://pr-759.opencouncil.dev/chania/join?c=cm3glw2qp04jhckl3zbohiorx.tltpmq.BqXenqHGKINj88WN&step=2",
 "d":"opencouncil.gr","r":null}
```

PostHog does the same on production — `ConsentChip` captures `$current_url: window.location.href`
and `instrumentation-client.ts` sets no URL scrubbing. PostHog has no token on the preview, so
that half is from the code, not measured.

Two things go out, and they are worth separating:

- A live claim credential, to two processors, in their logs.
- The councillor's identity. The token's first segment is the plain `personId`, so every scan
  tells the processor which councillor scanned and when. This round dropped `utm_content=<personId>`
  from the QR on purpose. The URL capture puts it back.

The flow already measures itself, and at the granularity you chose:
`person_join_viewed`, `person_join_claimed`, `person_join_email_sent` and `person_join_consent`
all carry `city_id` and no person. So the automatic capture adds no number anyone reads.

This is not new in kind — the old flow put the code in `/sign-in?callbackUrl=/api/join/<token>`
for a signed-out scanner. It is new in reach: every scan now, and on the page the scan lands on.

**A suggestion.** Exclude the join path the way embeds are already excluded. `EMBED_PATH` gates
both `PlausibleAnalytics.tsx` and `instrumentation-client.ts`, so the mechanism exists. Stripping
`c` from the captured URL would also work and keeps the pageview.

src/lib/db/personConsent.ts
```diff
@@ -0,0 +1,86 @@
...
+ * closes every open period of the person and deletes nothing, so we can
+ * still show that consent was in force while a voiceprint was processed.
+ */
+export async function setVoicePrintConsent(personId: string, consent: boolean): Promise<void> {
+    // A server action has no runtime types: a stale client once sent undefined.
+    if (typeof consent !== "boolean") throw new BadRequestError("consent must be a boolean");
+    const user = await getCurrentUser();
+    if (!user) throw new UnauthorizedError("Not signed in");
+    const refused = () => new ForbiddenError("Only the account that claimed the person can give voiceprint consent");
+    if (!user.administers.some((a) => a.personId === personId && a.claimedAt)) throw refused();
```
kouloumos: **The problem.** Only an account with a claimed row can close a period. When the claimed row goes,
the open period stays open and nobody can close it — not the person, not a superadmin.

It takes three ordinary steps: scan and claim, tick consent, then a superadmin unlinks the account
(or the account is deleted). I did exactly that on the preview, and the row is still there:

```
person                 | given_by          | withdrawnAt | claimed rows for that person
Ευστρατία Βερυκάκη     | readonly@test.com | (open)      | 0
```

Four of the five open periods on the preview are healthy. That one cannot be closed by anyone.

It is harmless while `profile/page.tsx` is the only reader, because it filters by `userId`. It
stops being harmless when something asks "does this person consent?" without a user — which is
what the voiceprint gate will ask.

**A suggestion.** Let the write that removes a claim close the period, so the two stay symmetric.
`users.ts` already preserves `claimedAt` across the form's delete-and-recreate; it can close the
periods of a person whose claimed row is going away. `deleteUser` and `deleteCurrentUser` need the
same: today the `SetNull` leaves the period open with a null `userId`. A closed period keeps the
evidence; a deleted row would not.

src/lib/db/personConsent.ts
```diff
@@ -0,0 +1,86 @@
...
+                        where: { personId, withdrawnAt: null },
+                        data: { withdrawnAt: new Date() },
+                    });
+                    return;
+                }
+                // A period open under another account is stale: that account was
+                // the person once, and this one is now. Close it, so the open
+                // period is always the current account's own.
+                await tx.voicePrintConsent.updateMany({
+                    where: { personId, withdrawnAt: null, NOT: { userId: user.id } },
```
kouloumos: **The problem.** This closes a period that another account opened, and records it as
`withdrawnAt`. Nobody withdrew: the account stopped being the person. The column now carries two
facts that read the same — the data subject changed their mind, and the link changed.

For the lawfulness of past processing they are equivalent, so this is the weaker of the two. It
matters only if we are ever asked about withdrawal behaviour, where the table would overstate it.

**A suggestion.** Worth doing only if you take the one above. That fix adds a third meaning to the
same column — closed because the claim went — and at three it is worth naming the reason, either a
`closedReason` or a separate `supersededAt`. On its own I would leave it.

src/lib/personJoin/paths.ts
kouloumos: **The problem.** The code now rides in the query of a normal page, so it reaches the analytics scripts. I loaded a real QR URL on the preview and captured what Plausible sent:

```json
{"n":"pageview","v":36,"u":"https://pr-759.opencouncil.dev/chania/join?c=cm3glw2qp04jhckl3zbohiorx.tltpmq.BqXenqHGKINj88WN&step=2","d":"opencouncil.gr","r":null}
```

PostHog does the same on production — `ConsentChip` captures `$current_url: window.location.href`, and `instrumentation-client.ts` sets no URL scrubbing. PostHog has no token on the preview, so that half is from the code, not measured.

Two things go out, and they are worth separating. One is a live claim credential, to two processors, in their logs — it is valid for five days and it is the only thing standing between a reader and the councillor's page. The other is the councillor's identity: the token's first segment is the plain `personId`, so every scan tells the processor which councillor scanned and when. This round dropped `utm_content=<personId>` from the QR on purpose, and the URL capture puts it back.

The flow already measures itself, and at the granularity you chose. `person_join_viewed`, `person_join_claimed`, `person_join_email_sent` and `person_join_consent` all carry `city_id` and no person, so the automatic capture adds no number anyone reads.

This is not new in kind — the old flow put the code in `/sign-in?callbackUrl=/api/join/<token>` for a signed-out scanner. It is new in reach: every scan now, and on the page the scan lands on. The page already sets `referrer: no-referrer` and `noindex`, so the concern was clearly on your mind; this is the path that got missed.

**A suggestion.** Exclude the join path the way embeds are already excluded. `EMBED_PATH` gates both `PlausibleAnalytics.tsx` and `instrumentation-client.ts`, so the mechanism exists. Stripping `c` from the captured URL would also work and keeps the pageview.

kouloumos, line 7: **The problem.** The code now rides in the query of a normal page, so it reaches the analytics scripts. I loaded a real QR URL on the preview and captured what Plausible sent:

```json
{"n":"pageview","v":36,"u":"https://pr-759.opencouncil.dev/chania/join?c=cm3glw2qp04jhckl3zbohiorx.tltpmq.BqXenqHGKINj88WN&step=2","d":"opencouncil.gr","r":null}
```

PostHog does the same on production — `ConsentChip` captures `$current_url: window.location.href`, and `instrumentation-client.ts` sets no URL scrubbing. PostHog has no token on the preview, so that half is from the code, not measured.

Two things go out, and they are worth separating. One is a live claim credential, to two processors, in their logs — it is valid for five days and it is the only thing standing between a reader and the councillor's page. The other is the councillor's identity: the token's first segment is the plain `personId`, so every scan tells the processor which councillor scanned and when. This round dropped `utm_content=<personId>` from the QR on purpose, and the URL capture puts it back.

The flow already measures itself, and at the granularity you chose. `person_join_viewed`, `person_join_claimed`, `person_join_email_sent` and `person_join_consent` all carry `city_id` and no person, so the automatic capture adds no number anyone reads.

This is not new in kind — the old flow put the code in `/sign-in?callbackUrl=/api/join/<token>` for a signed-out scanner. It is new in reach: every scan now, and on the page the scan lands on. The page already sets `referrer: no-referrer` and `noindex`, so the concern was clearly on your mind; this is the path that got missed.

**A suggestion.** Exclude the join path the way embeds are already excluded. `EMBED_PATH` gates both `PlausibleAnalytics.tsx` and `instrumentation-client.ts`, so the mechanism exists. Stripping `c` from the captured URL would also work and keeps the pageview.

src/components/personJoin/PersonJoin.tsx
```diff
@@ -0,0 +1,208 @@
...
+            setBusy(false);
+        }
+        setFailures((n) => n + 1);
+        return false;
+    }
+
+    async function finish() {
+        if (!choice) return;
+        setConsentError(false);
+        if (choice === 'no') {
```
kouloumos, line 134: Nit. Declining writes nothing, which is right when there is no period to close — and on the normal path there never is, because the server sends `consented: true` straight to `done`. But `confirm()` also reaches this step on `already_yours`, and that branch does not read `stage.consented`.

I ran it on the preview: open the link in one tab and leave it on step 1, finish the flow on a second device including "Ναι, συναινώ", then come back to the first tab and press "Ναι, είμαι εγώ". It lands on the consent step, "Όχι τώρα" leads to "Όλα έτοιμα", and the period stays open — the reader said no and the record still says yes. It needs a tab opened before the claim and used after, so it is narrow, but that device switch is one the flow is built for.

**A suggestion.** Call `setVoicePrintConsent(person.id, false)` on decline too. It is an `updateMany` over open periods, so it costs one round trip and does nothing when there is nothing to close — and the correctness stops depending on the server stage and this branch agreeing.

kouloumos (commented):
One finding and one nit.

Ten of eleven landed, and the schema went further than I suggested — re-reading the claim inside the transaction, and preserving `claimedAt` across the admin form's delete-and-recreate, are both yours and both right. The compact token reads well: the MAC is checked before the expiry, the `person-claim:` prefix separates it from the confirmation mark, and 16 base64url characters are exactly 96 bits, so the signature malleability the old JSON format had is gone too. Moving the claim off a GET side effect and behind a signed confirmation answers the CSRF point properly.

I also went through the join module's server side: the truncated MAC is bijective with no alternate encodings, the `person-claim:`/`join-confirm:` domains cannot collide, `verifyJoinConfirmation` refuses a mark minted after the token expired, `callbackUrl` is un-influenceable from the browser, and the `!current && !(own && inFlow)` guard in `getJoinStage` holds across every combination — a hand-typed `?step=2` unlocks nothing for anyone but the account that already holds the claim. `getJoinPerson` exposes nothing that is not already public.

Greptile's open point about the confirmation mark not binding the recipient is a fair one, and I have not repeated it here.

pkakelas:
I pushed f8d777a6ca58f174d60a9d012367732d5a1d4817 to move the voiceprint consent into the personal details form on the profile:

- The profile shows only the checkbox label. The long description and the separate save button are gone.
- The checkbox is above "Αποθήκευση προσωπικών στοιχείων", and that button saves the consent too.
- The consent still saves through its own action, so a phone number that the server refuses does not block a withdrawal.

pkakelas:
I pushed cb4b7d700cdf02df5536429215fae96a68fcd6c7 to bind the mark of the sign-in email to the address of that email:

- The server signs the mark with the code, the send time and the address that the email goes to.
- `/api/join` accepts the mark only for the account with that address. A copied link claims nothing in a browser that is signed in as another account.
- A scan carries no mark, so it still does not read the session.

PR #770, opened 2026-09-17
fix(signup): stop losing readers who already have an account

christosporios:
## The problem, measured

PostHog, last three days on production:

| flow | code | failures | people |
|---|---|---|---|
| petition | `email_exists` | 20 | 18 |
| notification signup | `email_exists` | 16 | 15 |

**33 people** filled a whole form and were refused at the last press because their email already has an account — most of them do not know they have one, from an older signup or a petition. Of the 15 in the signup flow, **8 completed afterwards**; 9 asked for a sign-in link. So roughly a third of the funnel is lost here, and it is the largest leak in it.

Two things made the detour expensive:

1. The answer sent them somewhere else to type the same email again.
2. The link opened in a new tab — a new page load — so every location and topic they had picked was gone when they came back.

## The fix

1. **The link goes out from the server, at that moment.** The save action sends the magic link itself and answers `email_exists_link_sent`. The message reads as a notice — blue, a mail icon, `role="status"` — not a red alert, and the action bar stays as it was rather than shaking. A send that fails falls back to the old `email_exists` answer with the sign-in link in the alert, so the message never promises an email nobody sent.
2. **The form survives the round trip.** Both flows keep their answers in `localStorage` for a day, under a key per flow and municipality, and put them back on mount. The reader returns signed in to a filled form and presses the button once.

What a draft may put back is the flow's own decision (`draft.apply`), and it is deliberately narrow:

- the step comes from the URL, never the draft;
- the account fields belong to the session once there is one;
- **the WhatsApp tick is never restored** — Notis owns that answer, and a stale tick put back over his could resubscribe a reader who said ΣΤΟΠ.

## Verification

`tsc` (app + tests), eslint, `npm test` (225 suites, 2526 tests) — including new tests for the draft's key, its one-day expiry, its handling of an unreadable entry, and for the action bar staying calm when the link has been sent. The integration suite did not run locally (Docker is unresponsive on this machine); CI runs it.

Worth watching after deploy: `notification_signup_failed` / `petition_failed` with code `email_exists_link_sent` against completions — the recovery rate should climb from the ~50% measured above.


christosporios:
Pushed one more commit: `fix(signup): land the sign-in link back on the form`.

I had missed that `sendMagicLink` passed no destination, so the link landed on the home page — leaving the reader to find their way back to a form they had already filled, which is the detour this PR exists to remove. Both flows now send the page they are on (`returnTo`), and the link returns them to it, signed in, with the draft in place. The path goes through `safeRedirectPath`, like the sign-in page's own `callbackUrl`: a same-origin relative path or the profile, never another origin.

christosporios:
Pushed two commits for the review.

**Fixed**

- `8b005044` — the draft no longer speaks over the server. A reader editing a preference they already saved gets **no draft at all**, so an abandoned session can't replace their saved places and topics (and can't be written back by a save). The restore is state rather than a ref, so the write effect can't run in the same commit and put the pre-restore form over the draft it just read. Nothing is kept until the reader's first edit, so a visitor who types nothing leaves nothing behind. The effects depend on the draft's key rather than the object each flow builds inline — they were running on every render, serialising the whole form on every keystroke. A stored value that is not an object is refused. The manual sign-in link stays under both answers, so a slow or filtered email is never a dead end. New render test covers the restore, the no-clobber, the empty visitor, the first edit, and the `existing` case.
- `da66f03c` — the link-sent answer gets its own event (`notification_signup_link_sent`, `petition_link_sent`) instead of being counted as a failure, so the number this change is measured by can actually move.

**Not fixed, deliberately** — each is a decision rather than a defect:

- **The action mails any address, unthrottled.** Before this PR it only revealed that an address existed; now a loop against the Server Action sends a stranger a sign-in link each time. `/sign-in` has the same property, so this is not new to the product, but it is newly reachable without a page. Worth a rate limit on the sending path, as its own change.
- **`returnTo` resolves against `NEXTAUTH_URL`.** One deployment serves every realm domain, so a reader on `.fr`/`.rs` gets a link to the `.gr` host, where their city 404s. Greece is unaffected. The fix belongs with the realm-aware sign-in work, not here.
- **Restored `Topic` objects carry string dates under a `Date` type.** Nothing reads those fields today; storing ids and re-hydrating from the server-provided list would fix it and shrink the payload.
- **Name, email and phone sit in `localStorage` for a day.** That is what makes the inbox round trip survive; a shorter life or dropping the identity fields once the link is sent are the alternatives.
- `readStored`/`writeStored`/`removeStored` still live in `src/hooks/`; they belong in `src/lib/storage.ts`.

PR #776, opened 2026-09-18
fix(notis): the decision log, and a two-minute poller

christosporios:
Two changes to the same problem: Notis told a reader about a message it never sent, and new readers waited too long for the message it does send.

## 1. The decision log

A reader in Χαλάνδρι got this on 18 September:

> «Update στο έργο οδοστρώσεων 4 εκατ. ευρώ **που σου είχα πει**: …»

She had received one message in her life — the transition template on 2 September.

The wake of 11 September decided `silence` and sent nothing. Its rationale reads:

> "One item stands out: a €4M road resurfacing tender … **Sent one short message about it**; everything else stays silent as procedural."

One turn, `finish_wake` alone, no `send_message` call anywhere in the trace. The model wrote down the thing it had decided to do instead of the thing it did. Every later wake reads that log, so a week later the next wake believed it and wrote to her as if continuing a thread.

**The rationale carries judgement, not actions.** `system.md` asked for it in the voice of «της έστειλα…», which makes narrating a send the rationale's natural register whether or not one happened. It now asks why the event was, or was not, worth the reader's attention. `finish_wake`'s own schema says the same, because the model reads that too.

**The decision log says what the reader got.** A silence renders as `→ silence (nothing reached the reader)`, and the block opens with one line saying it holds reasoning, not a record — the conversation above is the record. A rationale that claims a send is now contradicted on the line above it.

**Both renderers.** The compaction pass folds aged-out wakes into `memory`, which never ages out and which the prompt renders with no caveat. It had its own copy of the renderer, so a false claim survived the fix and became permanent a few weeks later. It now uses the same `decisionOutcome` and carries the same line — which is what the block above it already did for messages, for the same reason.

**The prompt no longer contradicts itself.** Two passages still told the agent its decision log remembers what it told the reader; a promise belongs in `record_commitment`, which is the only row that holds one. A third forbade a send claim only when `send_message` was never called, which read as a licence to write «της έστειλα» after a real send.

This does not stop the model from dropping a `send_message` call. It stops a dropped call from becoming a memory of a message. The detector is the next step and is not here: a wake that sends nothing while its rationale claims otherwise is exactly checkable, and `runWake.ts` already carries the nudge machinery — it is gated on `hasUserMessage`, so it never runs on a proactive wake.

## 2. Two-minute poller

Enrollment happens in the poller and nowhere else, so its interval is what a new reader waits between finishing the signup and their first message. Five minutes was up to five minutes of silence under a screen that says «Ο Νότης θα σου γράψει σύντομα».

The interval is now a named constant, because every per-tick ceiling is a rate in disguise. `MAX_ENROLLMENTS_PER_TICK` releases eight readers a tick, so at two minutes it releases two and a half times the readers per hour that it did at five. That is the intent: faster, and still paced, rather than spending a whole launch cohort against the WhatsApp limit in one burst.

An earlier revision of this PR enrolled the reader inline, from the signup request, instead. It bought about a minute over this and cost a second enrollment path — which is where a review found an ordering bug, a split state between the two databases, an unpaced burst and a slower page load. One path is worth the minute.

### Two things to watch on the first day

- **Tick duration.** A tick that runs longer than the interval is skipped rather than queued, so the real cadence becomes the tick duration if a tick ever exceeds two minutes. The editorial passes are the long pole. The poller's own status row records each run.
- **The daily ceiling.** Eight per tick is a higher daily number than it was. It should stay under the number's messaging limit in WhatsApp Manager.

The tick also counts an intro only when the send reached the handset. A rail can suppress a message between the gate and Bird, and counting the attempt told the operator a reader was greeted when they were not.

## Verification

`tsc` clean in both apps. Notis: 36 suites / 407 tests. Main app: 226 suites / 2538 tests.

Three new cases in `prompt.test.ts`: a past silence renders the outcome and the block's warning, a send renders unchanged, an empty log says nothing about itself. The fake database now refuses a second subscription for one `userId`, the way Postgres does — nothing was testing the constraint the enrollment ceremony rests on.

Two rationale rows on that reader's subscription still carry the false claim, and one of them is the newest, so it is still in her next prompt. That is an operator edit, not a code change.


christosporios:
Pushed `ee37bf85`: the enrollment kick, as discussed. The signup now asks Notis to enroll the reader instead of leaving them to the next tick, so the intro arrives while the completion screen is still open.

The ceremony moved out of the poller's loop into `enrollment.ts` unchanged — both callers share it, and the poller's counters and alerts are as they were. A paused deployment, the quiet hours, an unreachable number and the +1 marketing hold all still refuse, and every refusal answers `next: "poller"`.

I also made the fake database refuse a second subscription for one `userId`. Nothing was testing that constraint, and the whole design rests on it: two callers can both pass the existence pre-read, and the loser must roll back rather than send a second intro.

The title and description now cover both changes.

christosporios:
Pushed `fb70a4d9`, the corrections from a review of this branch.

**The kick did not fire for two of its three callers.** `enrollNow` filtered the fan-out lookup on `notifyByPhone`, and `setNotisEnabled` writes that column *after* Notis answers — so Notis read the view while it still said `false` and declined every profile switch. The targeted lookup drops the filter. The poller's sweep keeps it, because the sweep reads the column to find the readers who want Νότης; this call already names one, and the request itself is the consent.

**The decision log has a second renderer.** `compaction.ts` folds aged-out wakes into `memory`, which never ages out, and it rendered a bare `→ silence` with the model's prose underneath. A rationale claiming a send would have become a permanent memory of a message — the bug this PR exists to fix, surviving it. It now uses `decisionOutcome` and carries the same warning as the live prompt, which is what the sibling block for messages already did.

Four smaller ones:

- The ceremony hands its send back to the caller instead of taking a `detachDelivery` flag. The poller awaits it and counts what the handset got rather than that a send was attempted, so a suppressed intro no longer reports as sent; the request path lets it run on.
- Nothing in the request path throws any more. A database blink returned 500, the main app read that as «ο Νότης δεν απαντά» and returned before writing the reader's consent — losing the request on both sides.
- The prompt contradicted itself: two passages still told the agent its decision log remembers what it told the reader, and one licensed a send claim after a real send.
- `enrollTarget` declines an empty row set rather than reading `rows[0]`.

The fakes now apply the view's predicates, so a lookup that stops matching them fails a test. Three cases were missing and are covered: enrollment before the consent column is written, a reader with no number, and an intro shell with no Bird project id.

36 suites / 416 tests on Notis, 226 / 2538 on the main app, `tsc` clean in both.

Open questions are in the thread below — I have not decided any of them here.

christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/fb70a4d979fef90eddd3c0d837ef72396c11128b..20e3929d274a31848ffc6eab00be0c1cad1fb1f5) to drop the inline enrollment and lower the poller interval instead.

The inline path bought about a minute over a two-minute tick, and it cost a second way for a reader to be enrolled. A review of it found an ordering bug that made it dead for the profile switch, a split state between the two databases when the main app's four-second timeout fired, no pacing budget at all on the request path, and a heavier import graph on a route every profile page load hits. None of those exist with one path.

What stays from that work: the compaction renderer, which is the half of the decision-log fix that matters most (it is the path that writes permanent memory), the prompt's self-contradictions, the fake database's unique-`userId` constraint, and the tick counting an intro only when the send reached the handset.

The title and description now describe the two changes that remain.

christosporios:
Pushed `17a6d992`, the corrections from a review of this branch.

**The interval had two definitions.** The admin panel kept its own `POLLER_INTERVAL_MS` at five minutes and counted down to a tick three minutes after the real one, so the operator's only poller heartbeat could never reach its overdue state. One dependency-free module now holds the interval and its label; the runtime and the panel both read it, and the two «κάθε 5′» labels follow it.

**An intro delivered over SMS counts as sent.** A WhatsApp template that fails terminally falls through to SMS with the same body, and the outcome keeps `status: "failed"` for the WhatsApp attempt. Reading the status alone reported every reader without WhatsApp as ungreeted — the inversion of what that counter is for.

**The decision log had two renderers and two row mappings**, and they had already drifted: compaction dropped the caveats that mark an entry as *not* a finished decision, so a wake cut at the token ceiling folded into `memory` as a whole one. One `decisionLine` and one `toDecisionEntry` now serve the prompt, the fold and the queue.

**Three prompt corrections.** `resolve_commitment` still instructed the model to write what it had done into the rationale. The new caveat tarred the shell's own `(σύστημα)` rows — which exist to be believed — with the model's unchecked prose. And the ban on describing actions left no way to write the commonest good reason for silence, that the reader has already been told; the rule now separates citing the conversation (the record) from citing yourself (a claim). `compaction.md` also told the summariser to extract what had been covered without saying which of its two sources can prove it.

Plus the fake database's upsert path, a real test for the compaction renderer, a send-decision assertion that could not fail, and the enrollment ceiling's arithmetic restored — 3,360/day against 1,344 before.

`tsc` clean, 36 suites / 408 tests.

Four things I did not change are in my summary to you rather than here; the two that touch this PR's own claims are the `→ send` half of the decision log, which the shell still cannot stand behind, and the tick duration, which can exceed the new interval.

PR #777, opened 2026-09-18
fix(notis): a reply card that reads, and a reader-level rate

christosporios:
The «Ποσοστό απάντησης» card printed `2 , 5%` over a line broken into three fragments, under a chip reading `↓ 48%`. Four faults in one tile.

## The number: `2 , 5%`

`tabular-nums` gives every glyph a digit-width cell, and in this font the comma takes one too — so the el-GR decimal separator rendered with a gap on each side. The string itself was always `"2,5%"`; only the metrics were wrong.

The headline drops `tabular-nums`. `fmtPct(…, fixed)` already pins the decimal place, which is what keeps the width steady as the value moves.

## The chip: `↓ 48%`

A rate does not move by a percentage of itself. 4,8% becoming 2,5% is **2,3 points**. Reported as a relative change it reads `↓ 48%` — a difference of five replies, described as a collapse. `DeltaChip` now takes the metric's unit and reports a rate in points (`↓ 2,3 μον.`). Counts are unchanged.

## The axis

A rate was drawn against a fixed 0–100 domain, so bucket heights could be compared. A reply rate lives between 2% and 5%: the line sat flat on the floor of the card, which is most of why it looked broken. It now scales to the data and stays anchored at zero, so the distance from zero is still honest.

## The line

News goes out on the days councils meet. A per-bucket rate is therefore a handful of sends against a handful of replies — most buckets have no denominator at all and plot null, and the ones that do swing between 0% and 100% on a single reply. That is what broke the line into fragments; the null-handling was working as designed on data that cannot support it.

The card now plots the rate **so far** at each bucket. One continuous line, no fragments, and its last point is the number printed above it. Buckets before the first send still plot null: there is no rate yet, and a zero would be one nobody earned. The tooltip carries the running counts, so a reader can see how thin the rate is.

## Note on what the card can answer

At 5 replies in 201 sends, the period figure is the honest signal and the shape is nearly flat. This makes the tile truthful and readable; it does not make a 24-hour reply rate a sensitive measure of engagement at this volume. A count of distinct readers who have ever replied would move less and say more, and is not in this PR.

## Verification

`tsc` clean, 36 suites / 411 tests. Four new cases in `metrics.test.ts` cover the running rate (carried forward across an empty bucket, no rate before the first send, last point equals the period figure) and the points-versus-percent rule.


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/364829432fea29929899095f3efe09fe2fae2fef..0269bbcb83d1d27495368be0aaed705d70f6f80b) to fix three regressions the first version introduced, and to put the chip's decision somewhere it can be tested.

**An absent baseline read as a rise.** `pointsChange` never returns null, so the «νέο» branch was unreachable for a rate — and the call site was collapsing "no rate" into `0`. A previous period with no news sends printed a green «↑ 2,5 μον.» against a baseline that does not exist, where `main` printed «νέο». The rate now reaches the chip as `null` and stays `null`.

**A 0% rate drew through the middle of the card.** `domain={[0, "dataMax"]}` degenerates to `[0, 0]` when every value is zero, and d3 maps a zero-width domain to the midpoint of the range — `scaleLinear().domain([0,0]).range([60,0])(0)` is 30. A window with sends and no replies rendered a headline of «0,0%» over a line halfway up the card. A rate now gets a ceiling rounded up to the next of 5 / 10 / 25 / 50 / 100, which is never degenerate and holds still while the rate wanders inside it.

**The tooltip asserted activity on quiet days.** The old guard was per-bucket; the new one only suppressed the hint before the first send in the window, so a Sunday that sent nothing read «Σαβ 17/8 2,5% 5/201». The label now says «έως» and the hint says «συνολικά».

Two more from the review:

- **The chip's decision moved to `deltaFor`**, beside the other metric rules. This jest project runs `.ts` only, so a decision living in the `.tsx` cannot be tested at all — which is why the bug above shipped. `DeliveryPanel` calls the same function, so the two rates on that screen no longer disagree about what a move is (0,5 points against 0,1) or about an absent baseline.
- **No rate is drawn until the denominator can carry one.** A cumulative rate is wildest where its denominator is smallest: a first bucket of 1/1 opens at 100%, sets the scale from a single reply, and squashes the settled rate back onto the floor — the failure this PR set out to remove.

36 suites / 413 tests, `tsc` clean. Nine new cases cover the null handling, the points-versus-percent rule, the threshold, the invert colouring and the denominator floor.

Two design questions I did not decide are in the thread below.

christosporios:
Pushed `4275eca8`: the card now measures readers, not messages.

**Why.** «Ποσοστό απάντησης» was replies per news message — 5 from 201, 2,5%. That number cannot tell five replies from one enthusiast apart from five replies from five people, and those are opposite answers to the question the card is asked. Its denominator was narrow too: agenda and summary wakes only, a reply only if it landed inside 24 hours, and only before that reader's next wake of any kind. Most of what a reader says never reached it.

**What it shows now.** «Αναγνώστες που απαντούν» — the share of readers who wrote to Νότης at least once in the period. Each reader counts once, whatever they sent and whenever. The detail line carries the counts: «12 από 102 αναγνώστες έγραψαν στον Νότη — ο καθένας μετράει μία φορά».

The denominator is every reader still subscribed, and it is the same number for both periods and for every bucket of the chart. So the delta moves only when the number of readers writing moves, not when the audience grows. The trade is that a window long enough for the audience to have changed measures its older buckets against today's readers, which reads low; that is in a comment at the call site.

**It also settles the chart.** The old rate's denominator only existed on the days councils met, which is why it needed a running total to draw a continuous line — and why the review found a running total damps recent movement. This rate has a denominator every day, so each bucket carries its own and the line shows when something changed. `cumulativeReplyRates` and its denominator floor are gone with the card they served.

Everything from the earlier review still applies and stays: the chip's `deltaFor` decision, the null-versus-zero handling, the rate ceiling on the axis, and tabular figures on integers only.

**One thing left behind, deliberately.** The message-level reply rate is still computed in `metrics.ts` and nothing renders it. Its query is a correlated subquery that runs three times per panel load. I did not remove it, because it is the only thing that can answer whether a *particular* message drew a reply — but it is dead weight today and it should either come back as a card or go.

36 suites / 411 tests, `tsc` clean.

PR #778, opened 2026-09-18
fix(notis): an empty agenda must not 400, and must not be lost

christosporios:
## The alert

Firing on production every tick since the agenda task completed:

```
notis poller: editorial pass failed for chalandri/sep30_2026
(cmu6viq3e00rlohpv89tnkawb): 400 {"type":"error","error":{"type":
"invalid_request_error","message":"output_config.format.schema: Invalid
schema: Enum must be a non-empty array"}}
```

## The cause

`briefSchema` pins `subjectId` to an enum of the meeting's subject ids:

```ts
subjectId: { type: "string", enum: subjectIds },
```

A meeting with no subjects makes that enum `[]`, which the API refuses outright. The failure is deterministic and the poller writes no processed row on a failed pass, so the same rejected request went out on every tick for as long as the row stayed inside `EVENT_LOOKBACK_MS` — seven days, 288 times a day at the production interval, each one a Discord alert.

`chalandri/sep30_2026` is an **Ειδική Συνεδρίαση Λογοδοσίας** on 30/09, not yet held. It has an agenda PDF and no numbered items, so the agenda task extracted no subjects and emitted its event anyway. Confirmed against the archive: `"subjects": []`. This is a normal shape for an accountability session, not a broken meeting, and it will happen again.

No money was spent on the loop: a 400 at request validation bills nothing. The cost was the alert noise and one MCP round trip per tick.

## The fix

**The pass asks nothing.** With no subjects there is nothing to score, so it returns an empty brief without calling the model.

**The tick consumes the event.** An empty brief gives a wake nothing to rank, and waking a whole municipality about a meeting the agent cannot describe is worse than silence. The event is recorded as consumed — the same path `stale` and `late-agenda` already take — so it stops re-surfacing. The tick counts these in `emptyBriefConsumed`, so the panel shows them rather than hiding them.

## A note on the fixture

The poller's shared `BRIEF` fixture carried `subjects: []`, so every fan-out test in that file was running a brief production never produces. Giving it a subject is what made the new behaviour visible; the empty case is now its own test.

## Verification

`tsc` clean, 36 suites / 417 tests. Two new cases: the pass makes no model request for a subjectless meeting, and the tick consumes such an event without enqueuing a wake.

## Before this merges

The alert keeps firing on production until this ships or the row ages out on 24/09. `POST /api/admin/poll {"seedOnly": true}` would silence it immediately, but it consumes **every** pending meeting event, so any real meeting waiting to fan out would be swallowed silently — check the panel's event backlog before reaching for it.


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/972d37354441f0a1784c641d6df044ba66b39ef6..8c95909bd4666c0dfe388381eabee32eb69930db). A review of the first version found that it fixed the alert storm by introducing a worse failure, so the fix now works at the root instead.

**The first version was dangerous.** `handleTaskUpdate` sets a task's status to `succeeded` *before* `processResult` writes its subjects, and `notis_meeting_events` filters on exactly that status — so a real meeting looks empty for a few seconds after its event appears. Consuming there turned a transient into permanent silence for a whole city, with no alarm and no undo: the dedup identity is `(city, meeting, type)`, so a re-run of the same task is filtered out before `classifyEvent` ever sees it, and nothing in the service deletes a processed row. The 400 was accidentally self-healing in that window. My fix removed the self-healing.

**The enum is the defect.** `subjectId: { type: "string", enum: subjectIds }` is the only reason this schema is hand-built, and it buys nothing: the response is matched by id afterwards, a subject the model omits already reads as unscored with its own copy, and an id the model invents lands in the map unread. It is a plain string now, so no meeting can make the schema invalid — and the class goes with it, not just this instance. Duplicate ids from a bad payload, or an enum past a size cap on a 200-item κοινότητα agenda, would have reproduced the same deterministic 400 and the same seven-day storm.

**The poller now reads the difference.** Inside a five-minute settle window an empty meeting is left alone, and the next tick fans it out normally. Past it, the event is consumed — an unconsumed event holds one of four per-tick slots for seven days and starves the feed — but an alarm goes out first, carrying the remedy, because consuming cannot be undone. That matches what `future-summary` twenty lines above already does for the other "a person should fix this" disposition.

`headline` is a sentence rather than `""`, because the admin ledger falls back with `??` and an empty string passes it.

**Still open, and not in this PR:** the retry loop itself. Any deterministic failure of the editorial pass still repeats for the whole lookback window with an un-deduped alert. The `max_tokens` failure is the same shape and worse — it throws *after* the call, so each retry is a fully billed completion whose usage is discarded, invisible in every spend surface.

36 suites / 419 tests, `tsc` clean. Four new cases: the schema carries no enum (mutation-checked), the pass makes no model request for a subjectless meeting, a settled agenda is consumed with an alarm, and a freshly completed task is left alone.

PR #779, opened 2026-09-18
fix(signup): stop the picker landing readers on the wrong municipality

christosporios:
A reader reported that they asked for Δήμος Αγίας Παρασκευής, but the welcome
email signed them in to `/heraklion/petition?step=2`, and their account showed
a request for Ηράκλειο.

PostHog shows what happened. On 2026-09-18 the reader searched the picker on
`/petition`, tapped the first row — «Α Αγία Παρασκευή 10+ δημότες το ζήτησαν
ήδη» — and landed on step 2. They pressed Back 1.3 seconds later. Four seconds
after that they tapped the first row again and landed on Ηράκλειο. Both
autocapture chains carry `li nth-child="1"`: the same row, a different
municipality. They filled in the form and submitted, and nothing on the step
told them which municipality they were asking for.

This is not a single reader. Over the last 7 days, 15 of 184
`petition_city_picked` events were a different municipality re-picked within
60 seconds.

## What changes

**The picker keeps its search in the URL.** `MunicipalityPicker` held the
search in React state alone. Tapping a municipality and pressing Back remounts
the picker, the search comes back empty, and the list falls back to its default
order — where Ηράκλειο sits first at 25+. The search now goes to `?q=` with
`replaceState`, and the picker reads the parameter back after mount. The
notifications page now seeds the picker from the same parameter; the petition
page already did.

**Step 2 names the municipality where the reader is looking.** The picker links
straight to `?step=2` of both flows, which skips step 1 — the only step whose
title carried the name. Inside the step every string read the same for every
municipality: «Ποιος το ζητάει;», «Ποια είναι η σχέση σου με τον δήμο;», «Μένω
στον δήμο.». The name was on the page, in the header breadcrumb, but it is
small, grey, far from the form and offers no way to change the municipality.
The reader in the report read past it and filled in the whole form. Step 2 now
carries the same card as step 1 — seal, name, status, and «Αλλαγή» back to the
picker. `PetitionCityCard` and `SignupCityCard` hold that card for each flow,
so the two steps cannot drift apart.

## Scope note

The report came from the petition. The notifications signup has the same hole,
reached through the same picker and the same `?step=2` link, so both flows are
fixed here. Say the word if you would rather split the notifications half into
its own PR.

## Not in this PR

The welcome email's landing page comes from the `Referer` header, not from an
intended destination. In `savePetition` the new-user branch calls
`sendMagicLink(email)` with no `returnTo`, so Auth.js falls back to
`headers.get("Referer") ?? "/"` — the page the Server Action ran from. The
existing-account branch 24 lines above does pass `returnTo`. The link was
therefore "correct": it returned the reader to the form they submitted, which
was the wrong municipality. It is also fragile, because a stripped Referer
sends a new reader to `/`. That is a separate change.

The reporter now holds two petitions, Ηράκλειο and Αγία Παρασκευή. The
Ηράκλειο one still needs to be deleted by hand.

## Testing

- `MunicipalityPicker.test.tsx` gains the regression: search, unmount, remount
  with no server query, and the filtered list is still there. It fails on the
  old component and passes on the new one.
- The suite also resets the URL between tests, because the picker now writes
  to it.
- `npx jest`: 2421 passed. Two suites under `subjectImages` fail to load on
  `main` as well, so they are untouched by this branch.
- `npx tsc --project tsconfig.jest.json --noEmit` and `eslint` are clean.
- Verified on this PR's preview. On `/petition`, typing «Παρασκ» put `?q=Παρασκ`
  in the URL; tapping Αγία Παρασκευή landed on step 2, which shows «Δήμος Αγίας
  Παρασκευής» and «Αλλαγή»; Back restored both the search box and the filtered
  list. `/chalandri/notifications?step=2` shows «Δήμος Χαλανδρίου» and «Επίσημη
  υποστήριξη». Checked at 1440px and at 390px.


christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/366316bd9d96b1d1923aacf31ba29f3954d817a3..f1347fbe771bd8fa763b7439f8bb2abbba0c5d23>) to apply the findings of a review of this branch:

**The URL write (commit 1).** Next patches `history.replaceState` and skips its own router sync when the state it receives already carries the routing state, which `window.history.state` always does — so the router never learned about `?q=`, and any later router-state commit could erase it. The hook now passes `null`. The write is also debounced by 300ms, because a browser limits history calls and WebKit throws past about 100 in 30 seconds; a row flushes the pending write before it navigates. An absent parameter now clears the box, so a search the reader deleted is no longer restored. Whitespace alone writes nothing. Both pages read the parameter through `firstSearchParam`, because Next hands a page `string | string[]` and `?q=a&q=b` crashed the server render. The hook moved to `src/hooks/useQueryParamState.ts`.

**The card (commit 2).** The notifications flow has three steps and saves on step 3; the card went only on step 2, so the whole reported path still reproduced there on a phone. Every step of both flows now carries it. «Αλλαγή» carries the reader's search back to the picker — it previously dropped it and landed them on the default order, which is the original bug. It now asks before discarding answers (drafts are per municipality, and absent for a reader who is updating) and stops while a submit is in flight. A later step now says when the reader has already signed. `autoFocus` on the free-text relation no longer scrolls the card off a phone screen. The card wrappers carry tests.

**Telemetry (commit 3).** Keeping the search in the URL fed the reader's own words to posthog-js, which copies `$current_url` onto every event, against a durable profile. `sanitize_properties` now strips the search parameters. This also covers the list search, which has carried `?search=` for longer.

Three commits rather than two fixups, so each message describes the final state. `npx tsc` is clean and `npx jest` is at 2438 passing; the two failing `subjectImages` suites fail on `main` as well.

PR #781, opened 2026-09-18
fix(notis): divide the reply rate by the readers written to

christosporios:
## The problem

«ΑΝΑΓΝΩΣΤΕΣ ΠΟΥ ΑΠΑΝΤΟΥΝ» divided by everyone still subscribed:

> **10,1%** — 52 από 513 αναγνώστες έγραψαν στον Νότη

Νότης is quiet by design. Three wakes in four end in silence, so most of those 513 had nothing to reply to in the window. Dividing by all of them measures how often he writes, not how well.

The size of the error is easy to bound: however many messages went out, they reached at most that many distinct people. Against the ~300 readers he actually wrote to, the same 52 repliers is **17%**, not 10%. The card was understating engagement by about half.

## The change

The denominator is the readers who received at least one message in the period — a response rate.

```
52 από 300 αναγνώστες που έλαβαν μήνυμα απάντησαν — ο καθένας μετράει μία φορά
```

The chart follows: each bucket divides by the readers written to in *that* bucket. A bucket he wrote nothing in has no rate at all rather than a zero — nobody failed to answer a message that was never sent — so the line breaks there. The fixed denominator goes with it, and so does the distortion where a long window measured its older buckets against an audience that has since grown.

## The one edge, left uncorrected

A reader who answers this week a message from last week lands in the numerator without being in this window's denominator. At any volume this service has seen that cannot push the rate over 100%, and intersecting the two sets would drop genuinely engaged readers for the accident of when they were last written to. It is documented at the function.

## Shape

`recipients` mirrors the `repliers` plumbing exactly — the same `groupBy` on `subscriptionId` and the same `COUNT(DISTINCT "subscriptionId")` bucket query, filtered to `outbound` instead of `inbound`, in both the period stats and the series.

## Verification

`tsc` clean, 36 suites / 425 tests. The rate's tests now pin the thing that changed: the same numerator against the two denominators, and no rate at all when he wrote to nobody.


christosporios:
I [force-pushed](https://github.com/schemalabz/opencouncil/compare/a53eb8fd7dbfe1ebbb1053cd2933e2d3a137a95e..3855a65a08ea2a66192904098b8bb0831fb3455c) after a review of this branch. One finding was the same bug this PR fixes, one level down.

**`recipients` counted sends that never reached anyone.** Filtering on `direction = 'outbound'` alone counts rows that are `suppressed`, `failed` or still `pending`. A reader whose only message that week was held by a rail had nothing to reply to, and was inflating the denominator — exactly the distortion this PR exists to remove. The filter now uses the same status list as `REACHED_STATUSES` in `queue.ts`, rather than inventing a third definition: `failRate` in this very file already subtracts suppressed rows for the same reason, forty lines down.

**The rate is clamped at 100%.** The docstring defended the cross-window leak — a reader answering this week a message from last week — on the grounds that it cannot outrun a denominator in the hundreds. True for a period, false for a chart bucket one minute wide, where two repliers against one recipient plots 200%. That value also misses every entry in `RATE_CEILINGS`, so the axis falls back to 100 and every honest bucket is flattened onto the floor.

**A period with no rate now reports no change, not «νέο».** `deltaFor` returned `new` when the *current* value was null, which is the inverse of what «νέο» means. That branch was effectively unreachable before, because the old denominator was every subscriber and never zero; with this denominator, a quiet hour is ordinary.

**Both series come from one query.** They were byte-identical apart from `'inbound'`/`'outbound'`, and the rate divides one by the other — so the two halves have to agree about what counts. One `GROUP BY 1, 2` keeps the delivery filter in one place and saves a round trip.

**And the docstring now names the coupling in both directions.** A per-recipient rate falls when Νότης writes to *more* people, because a wider send reaches more readers who were never going to answer. A recess week of 30 recipients and 9 repliers reads higher than a busy week of 400 and 60, though seven times as many readers wrote back. Worth reading beside «ΕΛΗΦΘΗΣΑΝ» rather than alone.

36 suites / 427 tests, `tsc` clean.

**Still thin:** `replierRateSeries` and the SQL have no test of their own, and there is no integration test for `periodStats` or `bucketedSeries` anywhere — so the new query runs for the first time in production. Worth a fixture through the series function at some point.

PR #796, opened 2026-09-22
refactor(ui): extract the orange call to action into one component

christosporios:
## What changed

Six surfaces hand-rolled the orange call to action. Each copy repeated the same class list and the same sliding-arrow markup. This PR replaces all six with one component, `src/components/ui/cta-button.tsx`.

## Why

`src/app/globals.css` underlines every anchor on hover:

```css
a:hover { text-decoration: underline; }
```

Every copy that renders an anchor must therefore turn that off with `hover:no-underline`, and each author had to remember it independently. The copy in `JoinScreens.tsx` did not have the guard, so its label underlined on hover inside the filled pill. `CtaButton` carries the guard, so no caller can forget it.

## Two notes for the reviewer

**The component has two variants, not one.** Four surfaces are the filled orange pill (`solid`). The two municipality pickers are bare orange *text* links: no background, no rounding, no white text. They share only the `group/cta` sliding arrow and the underline guard with the pill. One variant for both would have changed how the pickers look, so they get `variant="text"`.

**One hover treatment changed on purpose.** The two city cards faded the whole pill with `transition-opacity hover:opacity-90`, which also fades the white label. The other two solid surfaces faded only the background with `hover:bg-[hsl(var(--orange-deep))]/90`. The shared component uses the background version, so the label stays pure white. This is a small hover difference on those two cards. Tell me if you prefer the old behaviour per surface.

The solid variant builds on the existing `Button`, so the two city cards also get a keyboard focus ring that they did not have.

## Sizes

Heights differ per surface and stay as they are. `size="sm"` is `h-10`, for a call to action in a card. `size="md"` is `h-12 px-5 text-[15px]`, for one in a flow. Width and maximum width stay at the call site, because they are layout, not button size.

## Call sites

| File | Element | Variant |
| --- | --- | --- |
| `CityNotificationCard.tsx` | `TrackedLink` | `solid`, `sm` |
| `CityPetitionCard.tsx` | `TrackedLink` | `solid`, `sm` |
| `JoinScreens.tsx` | `Link` | `solid`, `md`, no arrow |
| `SignupChrome.tsx` (`SignupFooter`) | `button` | `solid`, `md` |
| `MunicipalityPicker.tsx` | `Link` | `text` |
| `notifications/page.tsx` | `Link` | `text` |

Every destination, label, and analytics event (`event` and `eventProps`) stays exactly as it was. The footer keeps its red flash state: the caller overrides the background through `className`, and `cn` uses `tailwind-merge`, so the override wins.

## Verification

- `grep -rn "group-hover/cta:translate-x-0.5" src` matches only the new component.
- `npm test` passes: 260 suites, 2926 tests.
- `npx tsc --noEmit` passes.
- `npm run lint` passes.

Not verified in a browser. The six surfaces need a visual check before merge.

<h2><a href="https://app.greptile.com/api/retrigger?id=67365432"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with no outstanding correctness, security, or repository-rule issues identified.

<h3>Summary</h3>

This PR centralizes six orange calls to action in a reusable `CtaButton` component while preserving link destinations, analytics events, action behavior, and per-surface sizing.
- Supports solid and text variants, link and button rendering, optional tracking, and optional arrows.
- Applies consistent hover, focus, disabled, and no-underline behavior.
- Moves and rewords the notification channel note, updating all locale catalogs and its OG-image usage.
- The previously reported inherited `Button` opacity and shadow behavior is fully fixed by implementing the CTA styles without the default `Button` variant.

<sub>Reviews (3) · Last reviewed commit: ["fix(signup): move the channels line next..."](https://github.com/schemalabz/opencouncil/commit/76326db6c6d05a1f1ea05246de8ebd1db5ee5670)</sub>


christosporios:
I [force-pushed](<https://github.com/schemalabz/opencouncil/compare/a5db011c19ae45caeb613136b4f114e674933ac1..e981f56029ab5c41fbf7a4f415f9b99940df0d8b>) to rebase onto `main`, which had moved 17 commits ahead. The rebase is not clean, so the diff changed:

- #795 added a `PRIMARY_BUTTON` constant and a seventh call to action in `JoinScreens.tsx` (`NotificationsInvite`). The constant is gone and both surfaces now use `CtaButton`. `QUIET_BUTTON` stays, because it is a ghost button.
- On the done screen the primary action is now conditional: with the notifications on offer the link to the person's page turns quiet, so it renders as a ghost `Button`, not as a `CtaButton`.

I also stopped building `CtaButton` on `Button`. `Button` applies its `default` variant unless you name another one, and that variant brings its own orange background, `shadow-md` and `hover:opacity-90`. Overriding all three through `className` is fragile, so the component now states the focus ring and the disabled state itself. One effect: the pill is flat everywhere. The two surfaces that went through `Button` had `shadow-md`; the two city cards never did.

christosporios:
Pushed [76326db6c](<https://github.com/schemalabz/opencouncil/commit/76326db6c>), which moves the channels line on the municipality picker.

The line about WhatsApp, SMS and email sat last, under the link to the petition, so it read as a condition of the petition. It now sits directly under the picker, and the petition link stays last.

The text was a fragment that leaned on the lead above it. It is now a sentence that stands on its own, because `/api/og` uses the same string as the note on the share card, where it sits under the lead. The key is `pickerChannels` in all four locales, because the line is no longer a footnote.

